PR #212, opened 2014-12-05
TriggerBuilder.UsingJobData(JobDataMap newJobDataMap) should ovewrite existing

RyanMelenaNoesis:
Update TriggerBuilder.UsingJobData(JobDataMap newJobDataMap) behavior to match that of TriggerBuilder.UsingJobData(string key, T value) overloads.  When duplicate key is found use value from newJobDataMap.


lahma:
Thanks for the contribution! A good fix.

kwaclaw:
Somehow this fix did not make it into current master (as of two years later!).

PR #230, opened 2015-02-01
Fix errors when using OleDb ? parameter placeholders

NathanBaulch:


lahma:
Thank you for the contribution, appreciated!

PR #238, opened 2015-02-24
Fix typos in SchedulerMetaData.cs

ubikuity:


lahma:
Thanks for the contribution!

PR #239, opened 2015-03-03
add mysql 6.9.5 provider support

Jetsly:
add mysql 6.9.5 provider support


lahma:
Thanks for the contribution!

PR #243, opened 2015-03-12
Fixed more typos and xmldoc issues

NathanBaulch:


src/Quartz/Impl/Calendar/DailyCalendar.cs
```diff
@@ -105,6 +105,8 @@ public class DailyCalendar : BaseCalendar
         ///     </li>  
         /// </ul>
         /// </summary>
+        /// <param name="rangeStartingTime">The range starting time in millis.</param>
```
lahma: if I remember right this will give compilation exception?

NathanBaulch: i'm pretty sure this is the normal way to document parameters. `<param>` is used to describe parameters and `<paramref>` is used inside `<summary>` blocks to reference parameters by name.

lahma: You don't get compile error after this?

NathanBaulch: No, what error are you seeing?

lahma:
My mistake, checked and compiles fine. Thanks for the contribution!

PR #245, opened 2015-03-17
More extensible QuartzScheduler plus minor DirectSchedulerFactory initialization fix

NathanBaulch:
I'm doing some advanced work with running the scheduler (`QuartzScheduler`) and worker (`JobRunShell`) in different OS processes. I really need a default public constructor on `QuartzScheduler` and all its notify methods to be public virtual. I hope you'll accept these very low risk tweaks.

This pull request also includes a minor bug fix in `DirectSchedulerFactory` to ensure that the `InstanceId` and `InstanceName` properties are set before `IThreadPool.Initialize` is called.

### DIFF
--- a/src/Quartz/Core/QuartzScheduler.cs
+++ b/src/Quartz/Core/QuartzScheduler.cs
@@ -1603,7 +1603,7 @@ namespace Quartz.Core
             }
         }
 
-        protected internal void NotifyJobStoreJobVetoed(IOperableTrigger trigger, IJobDetail detail, SchedulerInstruction instCode)
+        public virtual void NotifyJobStoreJobVetoed(IOperableTrigger trigger, IJobDetail detail, SchedulerInstruction instCode)
         {
             resources.JobStore.TriggeredJobComplete(trigger, detail, instCode);
         }
@@ -1614,7 +1614,7 @@ namespace Quartz.Core
         /// <param name="trigger">The trigger.</param>
         /// <param name="detail">The detail.</param>
         /// <param name="instCode">The instruction code.</param>
-        protected internal virtual void NotifyJobStoreJobComplete(IOperableTrigger trigger, IJobDetail detail, SchedulerInstruction instCode)
+        public virtual void NotifyJobStoreJobComplete(IOperableTrigger trigger, IJobDetail detail, SchedulerInstruction instCode)
         {
             resources.JobStore.TriggeredJobComplete(trigger, detail, instCode);
         }
@@ -2145,7 +2145,7 @@ namespace Quartz.Core
             }
         }
 
-        public void NotifySchedulerListenersInStandbyMode()
+        public virtual void NotifySchedulerListenersInStandbyMode()
         {
             // notify all scheduler listeners
             foreach (ISchedulerListener listener in BuildSchedulerListenerList())
--- a/src/Quartz/Core/QuartzScheduler.cs
+++ b/src/Quartz/Core/QuartzScheduler.cs
@@ -308,6 +308,15 @@ namespace Quartz.Core
         }
 
         /// <summary>
+        /// Create a <see cref="QuartzScheduler" />.
+        /// </summary>
+        /// <seealso cref="QuartzSchedulerResources" />
+        public QuartzScheduler()
+        {
+            log = LogManager.GetLogger(GetType());
+        }
+
+        /// <summary>
         /// Create a <see cref="QuartzScheduler" /> with the given configuration
         /// properties.
         /// </summary>
--- a/src/Quartz/Impl/DirectSchedulerFactory.cs
+++ b/src/Quartz/Impl/DirectSchedulerFactory.cs
@@ -317,14 +317,13 @@ namespace Quartz.Impl
 
             // Fire everything u
             // ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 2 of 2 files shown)


lahma:
Having the default constructor will lead to possibility of having the scheduler in wrong state (fields not initialized). I wonder if it would suffice to make the methods protected virtual the methods are something that sub-classing should allow to modify).

Could you elaborate your use case? With reflection you could call protected constructor and methods too. Seems very specific that you're trying to achieve.

NathanBaulch:
You're right, it is very specific, I was just hoping you wouldn't mind adding a default constructor and a few more virtual methods because it allows me to reuse the handy `JobRunShell.Run` method in my heavily customized scheduler server. I'd be totally fine with a protected parameterless constructor if you prefer as I'm already subclassing `QuartzScheduler` in my code.

Here's a very basic working example of how `JobRunShell` can be used with a custom `QuartzScheduler` that ignores job store notifications:

```csharp
var qs = new MyQuartzScheduler();
var tfb = new TriggerFiredBundle(
    new JobDetailImpl("test", typeof (TestJob)),
    new SimpleTriggerImpl(),
    null, false, null, null, null, null);
var jrs = new JobRunShell(new StdScheduler(qs), tfb);
jrs.Initialize(qs);
jrs.Run();

public class MyQuartzScheduler : QuartzScheduler
{
    public override void NotifyJobStoreJobComplete(
        IOperableTrigger trigger, IJobDetail detail, SchedulerInstruction instCode) {}
    public override void NotifyJobStoreJobVetoed(
        IOperableTrigger trigger, IJobDetail detail, SchedulerInstruction instCode) {}
}
```

It doesn't matter that the `resources`, `schedThread`, `jobMgr`, `errLogger` and `signaler` fields aren't initialized because they're never used in this scenario. Prior to patching Quartz I was using `FormatterServices.GetUninitializedObject` to create an uninitialized `QuartzScheduler` and reflection to set the fields I need, but that requires quite a lot of ugly reflection code.

lahma:
Would https://github.com/quartznet/quartznet/commit/8b7a40b7e3f63276cea83c00dabf76091a501850 fit your requirements?

NathanBaulch:
Yep, that should do the trick! Thanks.

lahma:
Cool, now in master!

NathanBaulch:
Great, out of curiosity when are you planning the next official release?

lahma:
I think the bits and pieces begin to be there to cut out a release. Hopefully this week.

lahma:
And the 2.3.2 release is out containing this change, hope it helps!

NathanBaulch:
Out of curiosity, why are the two protected internal `NotifyJobStore*` methods more sensitive than the other 25 public `Notify*` methods in `QuartzScheduler`? It's not critical but it would be nice if I could call these without using reflection, which I need to do once my child process finishes executing a job so the DB is updated.

NathanBaulch:
Oops, I spoke too soon. I can't use reflection because I need to call the `NotifyJobStore*` methods across a remoting boundary (between my child worker process and the scheduler in my parent process). So I'm forced to use a patched version once again. Any chance you'd consider making these methods public like the other `Notify*` methods?

lahma:
protected internal allows both Quartz internals call them and you to override them, so they are a bit more open than the others - you shouldn't need reflection. I usually opt to create public methods that call the protected versions if public interface is really needed - like in unit tests.

Something tells me that there might be another plugin point that you are after. Like easier to change QuartzScheduler implementation etc. You can already call Quartz via remoting if you export - so cross app domain call would come from there.

NathanBaulch:
I'm doing a very different kind of remoting here. The normal export feature allows an external party to remotely manage the scheduler. My remote process is actually executing the job and needs to make sure all the relevant `Notify*` events are forwarded back to the central `QuartzScheduler` instance. Subclassing isn't feasible because the `QuartzScheduler` constructor is called deep within the very large `StdSchedulerFactory.Instantiate` method (there's no `IQuartzSchedulerFactory` that I can easily replace).

So replacing those two words "internal protected" with "public" (like all the others) means I don't have to figure out an alternative mechanism to pass these important job store IPC messages.

Perhaps some sample code would help illustrate?

lahma:
Maybe even higher level view of the problem that the code is trying to address. It would be great to understand the original requirements and how Quartz falls short with them. What parts need to run inside a different process and why etc.

lahma:
Maybe even higher level view of the problem that the code is trying to address. It would be great to understand the original requirements and how Quartz falls short with them. What parts need to run inside a different process and why etc.

NathanBaulch:
Sure, our server needs to survive fatal `OutOfMemoryException` and `StackOverflowException` when jobs (which can be written by third parties) behave badly. Plus our jobs depend on a native 32-bit component so we're forced to run in x86 mode, meaning all threads share the same ~1.7GB of memory.
I've fixed all these problems by creating an `IThreadPool` implementation that uses processes instead of threads. It marshals the central server's `QuartzScheduler` instance to each child process during initialization so that my custom `JobRunShell` can send notifications back to the parent process.

lahma:
OK, thanks for clarifying. This might be something to support natively too, need to think about this for a moment.

I guess `StackOverflowException` shouldn't be a problem, that you can handle with listeners and cancel retries if needed but `OutOfMemoryException ` is a different beast.

You are still running non-clustered single-server setup that only spawns multiple processes or do you have clustered setup?

NathanBaulch:
`StackOverflowException` is not catchable/recoverable. I have a mixture of clustered and non-clustered instances that all use my custom process pool.

PR #260, opened 2015-04-24
QuartzScheduler inheritance and BlockForAvailableThreads fixes

NathanBaulch:
Pretty please can you make the two `NotifyJobStore*` methods in `QuartzScheduler` public? I really need this, plus it makes them consistent with all the other virtual methods in this class.

I've also fixed the shutdown issue that I described in the groups [here](https://groups.google.com/forum/#!topic/quartznet/E1x2HSoGJ4U). I realise that the `BlockForAvailableThreads` name doesn't make as much sense if it only blocks for 500ms, but this very small change easily fixes the shutdown issue because `QuartzSchedulerThread` already handles zero return values correctly on [line 281](https://github.com/quartznet/quartznet/blob/133c99e7527f6e1301e5ba866d11f9dd056ff0b4/src/Quartz/Core/QuartzSchedulerThread.cs#L281).

### DIFF
--- a/src/Quartz/Core/QuartzScheduler.cs
+++ b/src/Quartz/Core/QuartzScheduler.cs
@@ -1608,7 +1608,7 @@ namespace Quartz.Core
             }
         }
 
-        protected internal virtual void NotifyJobStoreJobVetoed(IOperableTrigger trigger, IJobDetail detail, SchedulerInstruction instCode)
+        public virtual void NotifyJobStoreJobVetoed(IOperableTrigger trigger, IJobDetail detail, SchedulerInstruction instCode)
         {
             resources.JobStore.TriggeredJobComplete(trigger, detail, instCode);
         }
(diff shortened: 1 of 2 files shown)


NathanBaulch:
I hate to beg but I'm getting desperate at this point. The strong name key used to sign official NuGet releases isn't published so I can't use a binding redirect to point existing assemblies at my patched  Quartz.dll which has a different public key token. This means all my customers have to recompile their job assemblies!

lahma:
Sorry for the delay, I've been travelling and have had some matters that have limited my time with these issues. I'll try to go through these PRs soon. Thanks for hanging in there!

lahma:
I think that the change to public is fine but the other change is not. Changing BlockForAvailableThreads to work this way will break things. The case you describe seems a bit artificial, but the real issue probably is that you have long blocking jobs.

Best practice is to have jobs not to block this way (similar to Thread.Sleep). They should check whether run can proceed and if not then reschedule themselves at later time. I'd say it's an alarming condition to have entire thread pool exhausted by such jobs that never finish?

NathanBaulch:
The change to public is the most important thing in this pull request, so feel free to cherry-pick that commit.

I'm not exactly sure what you mean by "long blocking jobs", but we have customers that run data import jobs that can take over an hour. If they happen to shrink the thread pool to 3 workers for example and schedule 3 simultaneous imports then the scheduler can't be shutdown for at least an hour, even if the jobs support interruption. There shouldn't need to be a free worker in order to send each busy worker an interruption signal.

I'm curious, how does my `BlockForAvailableThreads` tweak break things? It just shifts the retry loop from `SimpleThreadPool` up to `QuartzSchedulerThread`. I've been running this patch for a while now without any problems.

NathanBaulch:
Do you want me to create a new pull request with just the `NotifyJobStore*` commit? It's a very simple change that I really need in the next release.

lahma:
I've cherry-picked the commit, it's now in master.

lahma:
I'm closing this as obsolete, please open new PRs if you are still missing something, thanks.

PR #264, opened 2015-05-08
Cleanup

NathanBaulch:
More cleanup, inspired by diamondo25's fix.


lahma:
Thanks, surprised how many problems there were. Good catches.

PR #270, opened 2015-06-03
Store null JobData in JobDetails if it's empty

### LINKED ISSUE #269: Use NULL to store JobDataMap when the map is empty
In some of our scenarios we have a lot of Jobs and we don't use JobDataMap at all. In that case, the column `JOB_DATA` is filled with an array of bytes of the binary serialization of an empty `JobDataMap` (or `NameValueCollection` in case of `useProperties` to be `true`).

This adds irrelevant data and calls unnecessary serialization/deserialization logic that could be avoided by simply storing NULL and returning a new `JobDataMap` in the case that the blob in the DB is NULL.

mcliment:
Mimicked Trigger insert behavior. Null reading is already safe.

Fixes #269


lahma:
Thanks for the contribution!

PR #309, opened 2015-10-25
AbstractTrigger.StartTimeUtc setter loses UTC offset information

AlexeyVlg:
When it invokes startTimeUtc = new DateTime(value.Year, value.Month, value.Day, value.Hour, value.Minute, value.Second), it creates time value in current TimeZone, but not in UTC.


lahma:
Good catch! Thank you for the contribution!

GSPP:
The new code might leave fractional milliseconds in place. Is that intentional?

Issue #306, opened 2015-10-23
IDirectoryScanListener listen to multiple directories


cknightdevelopment:
Right now it looks like the `IDirectoryScanListener` is only able to listen to one directory at a time. I think it would be nice to have it be able to listen to multiple directories. Would this be a welcome contribution to the project right now? I would be happy to work on this in a fork and submit a pull request.

cknightdevelopment:
I just wanted to circle back and see if this might be a welcome contribution to the project before I start working on it. What are your thoughts? Thank you.

lahma:
Sorry for the delay. You can actually have as many job+listener pairs as you want. I would recommend having job per directory so get the benefits of concurrency.

Does having multiple listeners (one per directory job) sound feasible?

cknightdevelopment:
What I am doing today is I am scanning multiple different directories, each with their own job+listener pair. What I would like to have is one job+listener pair (scanner) that can scan multiple directories. My use case is that I have several directories with the same kind of files in them that need to be processed in the same way. I do not want to have to make a separate job+listener pair for each one (can be up to a hundred directories). In this scenario I would like the `IDirectoryScanListener` to be able to be passed a collection of directories to scan. Does this make sense? Is this use case currently covered by the library? Thanks!

lahma:
OK that sounds pretty reasonable considering your use case. Probably the easiest change would be to have both DIRECTORY_NAME and DIRECTORY_NAMES supported by the job and then aggregate results for all directories and then notify single plugin.

If done against quartznet-3 branch could use Parallel etc for better performance (release time frame uncertain then though).

cknightdevelopment:
I will plan on making this enhancement on a fork of the `quartznet-3` branch, and work to make use of the new features available in the framework. Thank you for the direction on how to go about implementing this feature. I will work on this and submit a pull request when I think it is ready for review. Thanks!

cknightdevelopment:
I am trying to use the build file and had a few issues with it trying to use variables that were not assigned. I got those out of the way for now, but it looks like NAnt is trying to use .NET 4.0 and csc.exe for that, which is causing it to throw errors when it hits new C# 6 language features. How are you working around this? Is there a way to point NAnt to the newer compiler?

cknightdevelopment:
I was able to just build the Quartz project in Visual Studio, after I had NAnt create assembly info classes for me. That should be good enough for now.

PR #311, opened 2015-11-14
Directory scanner updates

cknightdevelopment:
### Overview

This pull request is in regards to issue https://github.com/quartznet/quartznet/issues/306 that I submitted. The goal was to allow one instance of a `DirectoryScanJob` to be able to monitor more than one directory. Obviously this needed to be implemented in a way that will still allow a `DirectoryScanJob` to monitor one folder in the same way that it does today (no breaking changes). 

### High Level Changes

* Added two new keys that will be looked for on the merged `JobDataMap`. 
   * `DIRECTORY_NAMES` - Value is a semi-colon (;) separated string of directory paths to scan
   * `DIRECTORY_PROVIDER_NAME` - Value is the name of the explicit `IDirectoryProvider` the users wants to use (see below for more details on `IDirectoryProvider`)
* Moved much of the logic for getting the data in and out of the merged `JobDataMap` that the `DirectoryScanJob` into a new internal class called `DirectoryScanJobModel`
* Added a new interface called `IDirectoryProvider` whose implementations are used to get the directory paths to be scanned. 
   * Created a default internal implementation called `DefaultDirectoryProvider` that will use the existing `DIRECTORY_NAME` and the new `DIRECTORY_NAMES` merged `JobDataMap` entries
   * If a user assigns a `DIRECTORY_PROVIDER_NAME` key in the merged `JobDataMap`, then the implementation they specified will be used instead of the default one. Users add their explicit implementation to the `SchedulerContext` in the same way as implementations of `IDirectoryScanListener` are added today. [It is retrieved from the merged `JobDataMap`](https://github.com/quartznet/quartznet/compare/quartznet-3...cknightdevelopment:directory-scanner-updates?expand=1#diff-1bf77f883391b28a1b30a0e809bde451R97) in the same way as well.
* Added comments to the code (following the patterns I saw in the existing code) pertaining to the changes I made

### Comments

I tested both the new functionality as well as for backwards compatibility, and all seems to be working correctly and performing well. I believe that the addition of the `DirectoryScanJobModel` nicely abstracts the implementation details of getting data in and out of the merged `JobDataMap`. Additionally, it will make code changes easier if we want to add more options via this mechanism later. The addition of the option to allow users to assign an explicit implementation of `IDirectoryProvider` to retrieve the directory paths to scan adds nice flexibility as to the source where of this data comes from. Also, if users want to change the directories to scan over time (perhaps they are querying a database to get directory paths), their implementation can grab those changes just tell us which ones they want to scan in a more dynamic/customization fashion.

### In Closing

Please let me know if you have questions concerning these commits or if you would like to discuss anything else. I welcome whatever feedback you may have. Thanks!

### DIFF
--- a/src/Quartz/DefaultDirectoryProvider.cs
+++ b/src/Quartz/DefaultDirectoryProvider.cs
@@ -1,0 +1,36 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Quartz.Job;
+
+namespace Quartz
+{
+    internal class DefaultDirectoryProvider : IDirectoryProvider
+    {
+        public IEnumerable<string> GetDirectoriesToScan(JobDataMap mergedJobDataMap)
+        {
+            List<string> directoriesToScan = new List<string>();
+            string dirName = mergedJobDataMap.GetString(DirectoryScanJob.DirectoryName);
+            string dirNames = mergedJobDataMap.GetString(DirectoryScanJob.DirectoryNames);
+
+            if (dirName == null && dirNames == null)
+            {
+                throw new JobExecutionException($"The parameter '{DirectoryScanJob.DirectoryName}' or '{DirectoryScanJob.DirectoryNames}' " +
+                                                "is required and was not found in merged JobDataMap");
+            }
+
+            if (dirName != null)
+            {
+                directoriesToScan.Add(dirName);
+            }
+            else
+            {
+                directoriesToScan.AddRange(
+                    dirNames.Split(new[] { ";" }, StringSplitOptions.RemoveEmptyEntries)
+                        .Distinct()); // just in case their are duplicates
+            }
+
+            return directoriesToScan;
+        }
+    }
+}
--- a/src/Quartz/DirectoryScanJobModel.cs
+++ b/src/Quartz/DirectoryScanJobModel.cs
@@ -2,6 +2,8 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using Quartz.Job;
+using Quartz.Simpl;
+using Quartz.Spi;
 
 namespace Quartz
 {
@@ -10,17 +12,22 @@ namespace Quartz
     /// </summary>
     internal class DirectoryScanJobModel
     {
-        internal IReadOnlyCollection<string> DirectoriesToScan { get; private set; }
+        /// <summary>
+        /// We only want this type of object to be instantiated by inspecting the data 
+        /// of a IJobExecutionContext <see cref="IJobExecutionContext"/>. Use the 
+        /// GetInstance() <see cref="GetInstance"/> method to create an instance of this
+        /// object type
+        /// </summary>
+        private DirectoryScanJobModel()
+        {
+        }
+
+        internal IReadOnlyList<string> DirectoriesToScan { get; private set; }
         internal IDirectoryScanListener DirectoryScanListener { get; private set; }
         internal DateTime LastModTime { get; private set; }
         internal DateTime MaxAgeDate => DateTime.Now - this.MinUpdateAge;
         private TimeSpan MinUpdateAge { get; set; }
         private JobDataMap JobDetailJobDataMap { get; set; }
-
-
-        private DirectoryScanJobModel()
-        {
-        }
 
         /// <summary>
         /// Creates an instance of DirectoryScanJobModel by inspecting the provided IJobExecutionContext <see cref="IJobExecutionContext"/>
@@ -51,6 +58,7 @@ namespace Quartz
                     : TimeSpan.FromSeconds(5), // default of 5 seconds
                 JobDetailJobDataMap = context.JobDetail.JobDataMap,
                 DirectoriesToScan = GetDirectoriesToScan(schedCtxt, mergedJobDataMap)
+                    .Distinct().ToList()
             };
 
             return model;
--- a/src/Quartz/DirectoryScanJobModel.cs
+++ b/src/Quartz/DirectoryScanJobModel.cs
@@ -1,0 +1,120 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Quartz.Job;
+
+namespace Quartz
+{
+    /// <summary>
+    /// Internal model to hold settings used by <see cref="DirectoryScanJob"/>
+    /// </summary>
+    internal class DirectoryScanJobModel
+    {
+        internal IReadOnlyCollection<string> DirectoriesToScan { get; private set; }
+        internal IDirectoryScanListener DirectoryScanListener { get; private set; }
+        internal DateTime LastModTime { get; private set; }
+        internal DateTime MaxAgeDate => DateTime.Now - this.MinUpdateAge;
+        private TimeSpan MinUpdateAge { get; set; }
+        private JobDataMap JobDetailJobDataMap { get; set; }
+
+
+        private DirectoryScanJobModel()
+        {
+        }
+
+        /// <summary>
+        /// Creates an instance of DirectoryScanJobModel by inspecting the provided IJobExecutionContext <see cref="IJobExecutionContext"/>
+        /// </summary>
+        /// <param name="context">Content of the job execution <see cref="IJobExecutionContext"/></param>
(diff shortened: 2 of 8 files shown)


src/Quartz/DefaultDirectoryProvider.cs
```diff
@@ -0,0 +1,36 @@
...
+{
+    internal class DefaultDirectoryProvider : IDirectoryProvider
+    {
+        public IEnumerable<string> GetDirectoriesToScan(JobDataMap mergedJobDataMap)
+        {
+            List<string> directoriesToScan = new List<string>();
+            string dirName = mergedJobDataMap.GetString(DirectoryScanJob.DirectoryName);
+            string dirNames = mergedJobDataMap.GetString(DirectoryScanJob.DirectoryNames);
+
+            if (dirName == null && dirNames == null)
```
lahma: If user supplies both we are not handling it nor are we giving an error?

src/Quartz/DirectoryScanJobModel.cs
```diff
@@ -0,0 +1,120 @@
...
+    {
+        internal IReadOnlyCollection<string> DirectoriesToScan { get; private set; }
+        internal IDirectoryScanListener DirectoryScanListener { get; private set; }
+        internal DateTime LastModTime { get; private set; }
+        internal DateTime MaxAgeDate => DateTime.Now - this.MinUpdateAge;
+        private TimeSpan MinUpdateAge { get; set; }
+        private JobDataMap JobDetailJobDataMap { get; set; }
+
+
+        private DirectoryScanJobModel()
```
lahma: Constructors should be up top after fields, before properties.

src/Quartz/IDirectoryProvider.cs
```diff
@@ -0,0 +1,20 @@
+using System.Collections.Generic;
+using Quartz.Job;
+
+namespace Quartz
```
lahma: This probably belongs to namespace Quartz.Spi

src/Quartz/Job/DirectoryScanJob.cs
```diff
@@ -57,95 +66,52 @@ public DirectoryScanJob()
...
-            if (listener == null)
-            {
-                throw new JobExecutionException("DirectoryScanListener named '" +
-                                                listenerName + "' not found in SchedulerContext");
-            }
+            DirectoryScanJobModel model = DirectoryScanJobModel.GetInstance(context);
 
-            DateTime lastDate = DateTime.MinValue;
-            if (mergedJobDataMap.ContainsKey(LastModifiedTime))
+            List<FileInfo> updatedFiles = new List<FileInfo>();
```
lahma: List is not a thread-safe collection, should either use a concurrent collection or explicit lock object.

src/Quartz/Job/DirectoryScanJob.cs
```diff
@@ -57,95 +66,52 @@ public DirectoryScanJob()
...
+                }
+            });
 
             if (updatedFiles.Any())
             {
                 // notify call back...
-                log.Info("Directory '" + dirName + "' contents updated, notifying listener.");
-                listener.FilesUpdatedOrAdded(updatedFiles);
+                updatedFiles.Select(x => x.DirectoryName)
+                    .Distinct().ToList()
```
lahma: Maybe just a foreach here for clarity, should already be distinct before this as they are provided to model as is (logged differently otherwise)?

src/Quartz/DirectoryScanJobModel.cs
```diff
@@ -0,0 +1,120 @@
...
+using Quartz.Job;
+
+namespace Quartz
+{
+    /// <summary>
+    /// Internal model to hold settings used by <see cref="DirectoryScanJob"/>
+    /// </summary>
+    internal class DirectoryScanJobModel
+    {
+        internal IReadOnlyCollection<string> DirectoriesToScan { get; private set; }
```
lahma: I usually prefer IReadOnlyList if it has true list semantics.

src/Quartz/Job/DirectoryScanJob.cs
```diff
@@ -57,95 +66,52 @@ public DirectoryScanJob()
...
+
+                DateTime latestWriteTimeFromFiles = updatedFiles.Select(x => x.LastWriteTime).Max();
+                model.UpdateLastModifiedDate(latestWriteTimeFromFiles);
             }
             else if (log.IsDebugEnabled())
             {
-                log.Debug("Directory '" + dirName + "' contents unchanged.");
+                model.DirectoriesToScan.ToList().ForEach(dir =>
+                {
+                    log.Debug("Directory '" + dir + "' contents unchanged.");
```
lahma: Looks little more like model's concern?

lahma:
Nice job! Could you please have a look at my review comments if there is something that you would like to see to before we merge this?

src/Quartz/DefaultDirectoryProvider.cs
```diff
@@ -0,0 +1,36 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Quartz.Job;
+
+namespace Quartz
```
lahma: This probably belongs to namespace Quartz.Simpl

src/Quartz/DefaultDirectoryProvider.cs
cknightdevelopment: Right now if the user supplies neither `DIRECTORY_NAME` nor `DIRECTORY_NAMES`, [then a `JobExecutionException` is thrown](https://github.com/cknightdevelopment/quartznet/blob/directory-scanner-updates/src/Quartz/DefaultDirectoryProvider.cs#L18). If both are supplied, then currently only `DIRECTORY_NAME` is used. Perhaps, we should just use both in that case (removing duplicates)? Or would you rather throw an exception?

cknightdevelopment:
Thank you for these comments @lahma! I will make these edits later today and commit the changes for review. I did have a question for you regarding your comment on when they supply both or neither of `DIRECTORY_NAME` nor `DIRECTORY_NAMES`, but I commented directly on that item.

src/Quartz/Job/DirectoryScanJob.cs
cknightdevelopment: The `DirectoryScanJobModel` currently does not have an `ILog` member, and it also is unaware of whether or not the directories have files in them or not. I thought of the model class as just being a way to get the data that the `DirectoryScanJob` needs in and out of the merged `JobDataMap`. The usage of that data (e.g. checking to see if the directories has files or not) is the job of the `DirectoryScanJob`. That is why I had the job logging if the directories did not contain any changed files, because it is aware of that fact. What do you think?

cknightdevelopment:
Hey @lahma, I made the updates from your comments in this last commit. Thanks.

cknightdevelopment:
Hello @lahma. Were there any other edits you were wanting to see for this pull request? Thanks.

lahma:
Sorry for the delay, life has kept a bit busy. Still some thoughts:

* DirectoryScanJobModel probably should not be in root namespace as it's purely for job use
* should we keep IDirectoryProvider internal as I'd rather not expose such a specialized interface for end users (yet), multiple implementations seem unlikely for now 

What do you think?

cknightdevelopment:
I agree that we should move the `DirectoryScanJobModel` out of the root namespace, I can make that change. 

However, I think there are some good reasons to make `IDirectoryProvider` public. This adds nice flexibility as to the source of where the directory list comes from, and does not require it to be static. I actually have a use case where the directories I need to scan change over time (I am querying a database to get directory paths that are updated/added from a UI). I would need an implementation that can grab those changes and tell Quartz which directories to scan in a dynamic/customization fashion over time. What are your thoughts?

cknightdevelopment:
I have made the change discussed.

lahma:
Thanks I have merged this now. I think I'll have another pass later on to make the directory source provider concept a bit more generic. Thanks for you contribution!

cknightdevelopment:
Terrific! I am very glad that I could contribute to such a wonderful project. We all appreciate your work on quartz!

Issue #291, opened 2015-08-29
SQL Server database design is ..... less than optimal ......


marcscheuner:
SQL Server database structure leaves a lot to be desired.....

Have a look at just this table:

```
CREATE TABLE [dbo].[QRTZ_CALENDARS]
(
    [SCHED_NAME] [NVARCHAR] (100) NOT NULL ,
    [CALENDAR_NAME] [NVARCHAR] (200) NOT NULL ,
    [CALENDAR] [IMAGE] NOT NULL
 )

ALTER TABLE [dbo].[QRTZ_CALENDARS] WITH NOCHECK 
ADD CONSTRAINT [PK_QRTZ_CALENDARS] 
PRIMARY KEY CLUSTERED ([SCHED_NAME], [CALENDAR_NAME]) 
```

Two main points I would fix right away:
- do not use the "Image" datatype anymore - it's been deprecated since SQL Server 2005 and will be removed soon. Use "VARBINARY(MAX)" instead which offers all the same goodness
- Having the clustering key on the two rather large NVARCHAR columns is bad for two reasons:
  (1) it's really large (potentially up to 604 bytes!! Darn close to the 900 byte index entry limit)
  (2) it's of varying length (twice!) 
  
  Both points together this for a horribly bad clustering key. I would recommend using a "INT IDENTITY" or "BIGINT IDENTITY" surrogate key instead which would be _significantly_ more efficient 

And also: ALL UPPERCASE is really not necessary anymore and makes it just sooo much harder to read the table and column names. Can't we use normal Camel casing (initial uppercase, rest of the word is lowercase) for identifiers, and ALL UPPERCASE only for T-SQL keywords (TABLE, INT, etc.)?

lahma:
Yes this is a known feature/problem. The database is modeled quite directly to be compatible with Java  version which makes the code also follow the same logic. 

It would be quite an undertaking to change the all scripts, create migration scripts for *all* supported databases and rewrite the DB access layer. The solution should also map nicely between databases. This is something that I have not yet found the dire need to do.

But a pull request is always welcome. There's a lot to migrate and test + document if this needs to be done.

lahma:
Varbinary is now in master, bigger primary key change would need as noted, quite a lot of work.

PR #312, opened 2015-11-30
Change SQL database IMAGE types to VARBINARY

kchenery:
Decided I would have a go at fixing up the use of the deprecated IMAGE datatype to its replacement VARBINARY(MAX) datatype. As reported in issue #291.

In its simplest form the tables can be changed and no code modifications take place. However; the calls to SQL still result in parameters of the IMAGE datatype.  Changing it to VARBINARY results in it using a size of VARBINARY(8000). Quite possibly fine for most payloads but to ensure VARBINARY(MAX) is used a size of -1 must be specified to the parameter.  This resulted in a new overload of the AdoUtil.AddCommandParameter method.

Have tried to keep everything else as is.


lukeryannetnz:
+1

lahma:
I've tweaked a bit and merged to master, so will be in v3 releases. Thank you for doing the investigative work.

PR #316, opened 2015-12-21
Resolve build error

chilin:
Upgrade odp.net.managed from 121.1.1 to 121.1.2

### DIFF
--- a/src/Quartz.Tests.Integration/Quartz.Tests.Integration.2010.csproj
+++ b/src/Quartz.Tests.Integration/Quartz.Tests.Integration.2010.csproj
@@ -93,8 +93,9 @@
       <SpecificVersion>False</SpecificVersion>
       <HintPath>..\..\packages\NUnit.2.6.3\lib\nunit.framework.dll</HintPath>
     </Reference>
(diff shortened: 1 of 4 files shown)


mj1856:
I'm seeing the same thing.  121.1.1 has been removed from Nuget so this is essential in order to build.  Please merge.  Thanks.

mj1856:
Also it looks like there's a reference in [Quartz.build](https://github.com/quartznet/quartznet/blob/d54171ba05b1876b2434f2c005689bfcbbe941b0/Quartz.build#L595) that needs to be updated.

Not sure if it matters or not.

lahma:
Should probably replace the whole package with this official one I guess? 

https://www.nuget.org/packages/Oracle.ManagedDataAccess/

Now the PR addresses only version upgrade instead of package change.

lahma:
There seems to be a another pull request that addresses the issue with correct package https://github.com/quartznet/quartznet/pull/328

danielmarbach:
See also https://github.com/quartznet/quartznet/pull/338

lahma:
I took in the  #338 with required alterations as it has the complete fix, including build file changes.

PR #317, opened 2015-12-23
Handle DST better

mj1856:
This PR addresses the challenges of scheduling a daily event when considering daylight saving time transitions.

First, it's worth pointing out that the Quartz.net documentation's description of how DST is handled is wrong.  It states multiple times ([in the FAQ here](http://www.quartz-scheduler.net/documentation/faq.html#daylight-saving-time-and-triggers), [in the CronTrigger Tutorial here](http://www.quartz-scheduler.net/documentation/quartz-2.x/tutorial/crontrigger.html#notes), [in the Best Practices here](http://www.quartz-scheduler.net/documentation/best-practices.html#daylight-savings-time), and probably in the API docs somewhere) that an occurrence in the spring-forward gap could be _skipped_ and an occurrence in the fall-back overlap could execute _twice_.  **This is not actually the current implemented behavior**.

The current implemented behavior is actually:
- In the spring, when a local time doesn't exist, run at the same moment in time as if the transition did not occur.
- In the fall, when two occurrences exist with the same local time, run at the _standard time_ occurrence.

This behavior is ultimately coming from `TimeZoneInfo.GetUtcOffset`, which always returns the standard time offset when in doubt.

While the spring behavior is fine, unfortunately the fall behavior is usually undesired in scheduling systems.  When there are two occurrences of the same local time, a daily event should be expected to run at the _first_ occurrence encountered, since time progresses in a forward direction.  The first occurrence corresponds to the _daylight_ time, not the _standard_ time.

This PR addresses this by adding an overload of `Quartz.Util.TimeZoneUtil.GetUtcOffset` that accepts a `DateTime`, and has the desired implementation.  Then throughout the code, anywhere that `TimeZoneInfo.GetUtcOffset` was used, this new method is used instead.

Unit tests are included, and all existing unit tests pass.

The documentation should also be updated to reflect the actual behavior.  Thanks.

### DIFF
--- a/src/Quartz.Tests.Unit/DaylightSavingTimeTests.cs
+++ b/src/Quartz.Tests.Unit/DaylightSavingTimeTests.cs
@@ -26,6 +26,21 @@ namespace Quartz.Tests.Unit
     [TestFixture]
     public class DaylightSavingTimeTest
     {
+        private Func<DateTimeOffset> OriginalUtcNow;
+
+        [OneTimeSetUp]
+        public void Init()
+        {
+            OriginalUtcNow = SystemTime.UtcNow;
+            SystemTime.UtcNow = () => new DateTimeOffset(2016, 1, 1, 0, 0, 0, TimeSpan.Zero);
+        }
+
+        [OneTimeTearDown]
+        public void Dispose()
+        {
+            SystemTime.UtcNow = OriginalUtcNow;
+        }
+
         [Test]
         public void ShouldHandleDstSpringForwardTransition()
         {
--- a/src/Quartz.Tests.Unit/DaylightSavingTimeTests.cs
+++ b/src/Quartz.Tests.Unit/DaylightSavingTimeTests.cs
@@ -1,0 +1,107 @@
+﻿#region License
+/* 
+ * All content copyright Terracotta, Inc., unless otherwise indicated. All rights reserved. 
+ * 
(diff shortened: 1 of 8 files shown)


drusellers:
:clap:

mj1856:
You know, it's been almost a year, and I forgot all about this.

I will rebase it to the current master shortly.  Is there any other reason this wasn't merged previously?

lahma:
I guess this PR flew under the radar for me somehow, so I'm to blame. The master now is ongoing work for 3.0 (now in alpha). I guess this would be quite safe to be included in 2.x maintenance branch. I can backport this unless you are willing to throw yet another pull request for the 2.x branch.

lahma:
Excellent PR, thank you for taking the time to prepare it with test cases included. Now part of 3.0 deliverables.

drusellers:
SWEET! Looking forward to 3.0

mj1856:
Thanks.  I'll leave the backporting to you.  I don't have an active project that uses Quartz.net at the moment, just I recommend it a lot. ;)

mj1856:
Also, reminder - you'll want to update the doc locations I pointed out originally, and any others that describe DST behavior.  With this change, the rules are fixed to:

- Pick the earlier (daylight) occurrence when ambiguous.  (1:30 PDT, not 1:30 PST)
- Skip by the gap duration (usually 1h) when invalid. (2:30 PST => 3:30 PDT)

PR #328, opened 2016-03-02
Minor fixes to tests and adding missing AssemblyInfo.cs

mishrsud:
Added missing AssemblyInfo.cs from Sourceforge to Github
Updated the package odp.net.managed (doesn't seem to be available) to Oracle.ManagedDataAccess
HolidayCalendarTest.cs: Fixed tests to work with daylight saving
DailyTimeIntervalTriggerImplTest.cs: Fixed tests to work with daylight saving

### DIFF
--- a/src/Quartz.Tests.Integration/Quartz.Tests.Integration.2010.csproj
+++ b/src/Quartz.Tests.Integration/Quartz.Tests.Integration.2010.csproj
@@ -93,8 +93,9 @@
       <SpecificVersion>False</SpecificVersion>
       <HintPath>..\..\packages\NUnit.2.6.3\lib\nunit.framework.dll</HintPath>
     </Reference>
-    <Reference Include="Oracle.ManagedDataAccess">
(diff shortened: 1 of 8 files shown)


lahma:
Seems that assembly info change is no longer there which is correct, it should not be part of of version control as it's generated. Maybe should not have the app.config changes as they only add clutter?

mishrsud:
Hi @lahma sorry, I'm confused - When I tried building from source code without the AssemblyInfo.cs in src, it did not generate that file. Is there a build script that I should be using? Also, the source on Sourceforge has that file.
Have you had a chance to look at the fixes to the tests? I can probably do a pull request for those?
The nuget update was also required as I couldn't build some projects without that.

danielmarbach:
See also https://github.com/quartznet/quartznet/pull/338

lahma:
@mishrsud could you please rebase against master and include only the test related fixed, I've pulled the Oracle library fix already. You can get the assembly info generated by running the Build.cmd.

mishrsud:
Sure thing! Thanks for your patience!

lahma:
Closing due to inactivity. We can reopen if needed.

PR #329, opened 2016-03-03
Include inner exception

kk3399:
This could happen when there isn't enough memory - OutOfMemoryException


lahma:
Thanks for the contribution!

PR #330, opened 2016-03-11
Use `Task.Run` inside `QuartzSchedulerThread`

danielmarbach:
`Task.Factory.StartNew` with async delegate returns `Task<Task>`, you have to call `.Unwrap()`. For majority of cases `Task.Run` is the safer option.


lahma:
Looking better, thanks for the contribution!

PR #333, opened 2016-03-15
fixes #332 Trigger loop encountered using DailyTimeIntervalTrigger across DST start boundary

### LINKED ISSUE #332: Trigger loop encountered using DailyTimeIntervalTrigger across DST start boundary
I encountered an issue that just occurred on the last Daylight Savings Time (DST) change on March 12, 2016.  I'm using a DailyTimeIntervalTrigger that was set to fire before, during and after the DST change. The triggers are set to run in various time zones (according to location). It appears that about an **hour** after the DST change, the trigger keeps firing without pause (like it is in an infinite loop).

A simplified version of the code I'm using for the trigger is as follows:

``` c#
TimeZoneInfo tz = TimeZoneInfo.Local;
ITrigger trigger = TriggerBuilder.Create()
    .WithIdentity("MyJob")
    .WithDailyTimeIntervalSchedule(
        x => x.WithIntervalInMinutes(5)
            .StartingDailyAt(TimeOfDay.HourAndMinuteOfDay(0, 0)) 
            .EndingDailyAt(TimeOfDay.HourAndMinuteOfDay(23, 59))
            .OnDaysOfTheWeek(new DayOfWeek[] { DayOfWeek.Sunday, DayOfWeek.Monday, DayOfWeek.Tuesday, DayOfWeek.Wednesday, DayOfWeek.Thursday, DayOfWeek.Friday, DayOfWeek.Saturday })
            .InTimeZone(tz)
).Build();
```

I've Attached a sample project [QuartzTest.zip](https://github.com/quartznet/quartznet/files/173131/QuartzTest.zip)
If you set your clock to ~30 minutes before the DST start, and run it, you find that the program enters a tight loop of triggers an hour after DST changes. 

I put in some debug code in the test sample that prints out the result of ComputeFireTimes. I noticed that there was repetition after the DST start (I'm in Pacific Time, so DST start is at 10 AM UTC). I'm pretty sure that the repetition in the computation is the part of the problem. I haven't been able to investigate further yet. I'm testing with the 2.3.3 stable.

> 3/13/2016 9:35:00 AM +00:00
> 3/13/2016 9:40:00 AM +00:00
> 3/13/2016 9:45:00 AM +00:00
> 3/13/2016 9:50:00 AM +00:00
> 3/13/2016 9:55:00 AM +00:00
> 3/13/2016 10:00:00 AM +00:00
> 3/13/2016 10:05:00 AM +00:00
> 3/13/2016 10:10:00 AM +00:00
> 3/13/2016 10:15:00 AM +00:00
> 3/13/2016 10:20:00 AM +00:00
> 3/13/2016 10:25:00 AM +00:00
> 3/13/2016 10:30:00 AM +00:00
> 3/13/2016 10:35:00 AM +00:00
> 3/13/2016 10:40:00 AM +00:00
> 3/13/2016 10:45:00 AM +00:00
> 3/13/2016 10:50:00 AM +00:00
> 3/13/2016 10:55:00 AM +00:00
> 3/13/2016 10:00:00 AM +00:00
> 3/13/2016 10:05:00 AM +00:00
> 3/13/2016 10:10:00 AM +00:00
> 3/13/2016 10:15:00 AM +00:00
> 3/13/2016 10:20:00 AM +00:00
> 3/13/2016 10:25:00 AM +00:00
> 3/13/2016 10:30:00 AM +00:00
> 3/13/2016 10:35:00 AM +00:00
> 3/13/2016 10:40:00 AM +00:00
> 3/13/2016 10:45:00 AM +00:00
> 3/13/2016 10:50:00 AM +00:00
> 3/13/2016 10:55:00 AM +00:00
> 3/13/2016 10:00:00 AM +00:00
> 3/13/2016 10:05:00 AM +00:00
> 3/13/2016 10:10:00 AM +00:00
> 3/13/2016 10:15:00 AM +00:00
> 3/13/2016 10:20:00 AM +00:00
> 3/13/2016 10:25:00 AM +00:00
> 3/13/2016 10:30:00 AM +00:00
> 3/13/2016 10:35:00 AM +00:00
> 3/13/2016 10:40:00 AM +00:00
> 3/13/2016 10:45:00 AM +00:00
> 3/13/2016 10:50:00 AM +00:00
> 3/13/2016 10:55:00 AM +00:00
> 3/13/2016 10:00:00 AM +00:00
> 3/13/2016 10:05:00 AM +00:00
> 3/13/2016 10:10:00 AM +00:00
> 3/13/2016 10:15:00 AM +00:00
> 3/13/2016 10:20:00 AM +00:00
> 3/13/2016 10:25:00 AM +00:00
> 3/13/2016 10:30:00 AM +00:00
> 3/13/2016 10:35:00 AM +00:00
> 3/13/2016 10:40:00 AM +00:00

dandufeu:
I've found that the changes here correct this issue for me.

### DIFF
--- a/src/Quartz/Impl/Triggers/DailyTimeIntervalTriggerImpl.cs
+++ b/src/Quartz/Impl/Triggers/DailyTimeIntervalTriggerImpl.cs
@@ -678,7 +678,7 @@ namespace Quartz.Impl.Triggers
             long secondsAfterStart = (long)(fireTime.Value - startTimeUtc).TotalSeconds;
             long repeatLong = RepeatInterval;
 
-            DateTimeOffset sTime = fireTimeStartDate;
+            DateTimeOffset sTime = fireTimeStartDate.ToUniversalTime();
             IntervalUnit repeatUnit = RepeatIntervalUnit;
             if (repeatUnit == IntervalUnit.Second)
             {
@@ -689,7 +689,7 @@ namespace Quartz.Impl.Triggers
                 }
 
                 sTime = sTime.AddSeconds(RepeatInterval * (int)jumpCount);
-                fireTime = sTime;
+                fireTime = TimeZoneUtil.ConvertTime(sTime, TimeZone);
             }
             else if (repeatUnit == IntervalUnit.Minute)
             {
@@ -699,7 +699,7 @@ namespace Quartz.Impl.Triggers
                     jumpCount++;
                 }
                 sTime = sTime.AddMinutes(RepeatInterval * (int)jumpCount);
-                fireTime = sTime;
+                fireTime = TimeZoneUtil.ConvertTime(sTime, TimeZone);
             }
             else if (repeatUnit == IntervalUnit.Hour)
             {
@@ -709,7 +709,7 @@ namespace Quartz.Impl.Triggers
                     jumpCount++;
                 }
                 sTime = sTime.AddHours(RepeatInterval * (int)jumpCount);
-                fireTime = sTime;
+                fireTime = TimeZoneUtil.ConvertTime(sTime, TimeZone);
             }
 
             // g. Ensure this new fireTime is within the day, or else we need to advance to next day.
@@ -726,11 +726,7 @@ namespace Quartz.Impl.Triggers
                 return null;
             }
 
-            // apply proper offset
-            var d = fireTime.Value;
-            d = new DateTimeOffset(d.Year, d.Month, d.Day, d.Hour, d.Minute, d.Second, TimeZone.GetUtcOffset(d.DateTime));
-
-            return d.ToUniversalTime();
+            return fireTime.Value.ToUniversalTime();
         }
 
         private bool IsSameDay(DateTimeOffset d1, DateTimeOffset d2)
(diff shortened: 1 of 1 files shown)


lahma:
Thanks for this, and sorry it sat open for so long. Closing it, because every part of it has since landed or been bettered elsewhere — but the loop you reported in #332 is real again from a different cause, and that is now fixed in #3190 (3.x) and #3191 (main).

Where the three changes in this PR ended up:

| change | status |
|---|---|
| `sTime.ToUniversalTime()`, `TimeZoneUtil.ConvertTime(sTime, TimeZone)`, and dropping the trailing "apply proper offset" block | Applied in 7d0c637855 back in 2017, present on both `3.x` and `main` today |
| `fireTime.Value.ToUniversalTime() < fireTimeStartDate.ToUniversalTime()` | No-op — `DateTimeOffset`'s relational operators already compare instants, not wall-clock values |
| the `IsAmbiguousTime` guard, to capture the first of the two ambiguous hours | Superseded by `TimeZoneUtil.GetUtcOffset(DateTime, TimeZoneInfo)`, which returns `GetAmbiguousTimeOffsets(dt).Max()` — the daylight, and therefore first, occurrence — for every ambiguous time rather than only at this one call site. Covered by `TestDstFallBackWithAmbiguousStartTimeOfDay` |

So there was nothing left to rebase onto the current code.

On the loop itself: it came back, twice over. The fall-back fix in #2917 / #2921 introduced a date correction that applied to every interval size in either direction, and on a spring-forward day — a 23 hour local day — it moved the fire time *backwards*, wedging every trigger with an interval of an hour or less in every DST time zone. Separately, and predating all of this, resetting the time of day to `startTimeOfDay` inherited a stale UTC offset, which in time zones that move the clock at midnight made `GetFireTimeAfter` return the very time it was passed. Both are fixed in #3190 / #3191, with regression tests over 3024 time zone / interval / start-time combinations.

The 2017 test that was supposed to guard this, `TestDayLightSaving2`, never set `StartTimeUtc`, so it was asserting against a clamped "now" and could not fail. That is fixed too.

Issue #331, opened 2016-03-11
Quartz 3 and async


danielmarbach:
I'm only starting to dig into the inner workings of Quartz.NET but I already have doubts about the internal design when it comes to scheduling and async. All the things I mention here are not at all a personal critique! Please take no offense. I'm just trying to be helpful and highlight potential problems

Here are a few random thoughts:
- `JobRunShell` is a mixture of blocking and async calls. It should be async only
- I get the intent that the library would like to be backwards compatible. That's why one introduced `IJob`. Since scheduling, persisting, executing jobs, the business logic inside the jobs will likely to touch IO-bound operations I'd suggest to go with a major breaking change and only provide the async version of the job. It is trivial to implement a base class for jobs to remove the crutch of returning a completed task. This removes the unnecessary complexity in the library of handling the sync and the async version and the potential bugs that come with managing two almost distinct code bases.
- The abstractions for the thread pool and the workers do no longer hold up in the async world. The interfaces are built for blocking interactions. I propose to diverge from the java version and introduce new better Task-based abstractions to simplify the code drastically. For example the limiting of concurrency can easily be implemented with `SemaphoreSlim` or if more fanciness is needed with custom task schedulers
- `QuartySchedulerThread`is a weird mixture of async code interacting with the above-mentioned thread pool abstraction and some fancy blocking and non-trivial `Monitor`interaction. `Monitor.Wait`, `ThreadAbortedExceptions`, Pulse etc. belongs to the thread world and should not really be mixed with async code

I'm aware that especially my view on sync vs. async APIs is really provoking. 

If you are interested I'd be willing to help out and see if we can pull quartz .net towards a more but truly async design.
## Plan of Attack
- [x] Make it compile again #335 #338 
- [x] Remove `IAsyncJob` and make `IJob` async #336
- [x] Remove `Async` suffix from methods, see my [reasoning for NServiceBus](https://groups.google.com/forum/#!topic/particularsoftware/xrSJpgjowJQ). I think the same applies here as well.
- [x] Remove interruptable job and replace by cancellation token #339
- [x] Remove blocking calls in `JobRunShell` #345 
- [ ] SimpleThreadPool should become ClrThreadPool (merge Clr to be Simple) informing users about possible no longer supported properties when tried to be set, should use QueuedTaskScheduler to support thread counts and priorities
- TODO

roryprimrose:
Please please please accept this offer from @danielmarbach, he knows his async

lahma:
[...]

PR #336, opened 2016-03-22
Make IJob async only

danielmarbach:
Connects to #331 

Removes `IAsyncJob` and makes `IJob` async only

I synchronous version of a job is super simple by doing:

```
class MySyncJob : IJob {
   public Task Execute(...) {
      return Task.FromResult(0); // or Task.CompletedTask for 4.6
  }
}
```

When #335 and #338 ar merged I will rebase and force push


danielmarbach:
Still catching up with the tests and their structure. Please bear with me

lahma:
Sure, no worries. I'm trying to get the project.json structure to work without csproj references, will inform when pushed.

lahma:
This looks good, after rebase probably ready to be pulled in. The IJob/IAsyncJob was truly a crutch and it was probably a bad decision to support both when starting the conversion.

danielmarbach:
Was on vacation. Will rebase next week

danielmarbach:
Rebased and force pushed

danielmarbach:
Build seems to fail with

> Specify which project or solution file to use because the folder contains more than one project or solution file.

lahma:
Thank you for tweaking the commit, don't mind about the appveyor messages for now as I'm going to remake the build scripts anyway. Thanks for this change!

PR #339, opened 2016-03-24
Interruptable job

danielmarbach:
Connects to #331 

This gets rid of `IInterruptableJob` in favour of CancellationToken. The CancellationToken is available on the job execution context and can be queried for cancellation and/or passed into the async methods supporting cancellation. For separation of concerns, I introduced a `ICancellableJobExecutionContext` which allows canceling a job.

There is one TODO. I'd argue we could remove the try/catch. But since the member `cancel` is virtual should we leave it? 

https://github.com/quartznet/quartznet/pull/339/files#diff-3670816342511c486ea84cb94c3ef967R491

Currently based on top of #336, will rebase when those are merged

### DIFF
--- a/src/Quartz.Examples/example11/SimpleJob.cs
+++ b/src/Quartz.Examples/example11/SimpleJob.cs
@@ -53,7 +53,7 @@ namespace Quartz.Examples.Example11
             // wait for a period of time
             long delayTime = context.JobDetail.JobDataMap.GetLong(DelayTime);
 
-            await Task.Delay(new TimeSpan(10000 * delayTime));
+            await Task.Delay(new TimeSpan(10000 * delayTime), context.CancellationToken);
 
             log.InfoFormat("Finished Executing job: {0} at {1}", jobKey, DateTime.Now.ToString("r"));
         }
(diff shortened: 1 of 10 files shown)


lahma:
I'm having some mixed feelings about this. If possible we should keep public API changes to minimum for now (like keeping IReadOnlyList<T>, maybe even IJobExecutionContext). Callbacks like JobWasExecutedAsync(ICancellableJobExecutionContext context) do not quite make sense as there's no point of trying to cancel the job execution. We could offer CancellationToken via IJobExecutionContext for now and allow usage when needed.

I'm not saying that public API changes cannot be done, just would like to keep them reasonable while the internals are changed.

danielmarbach:
Fair enough. I'll back track on a few API changes then and rebase with a smaller API surface

danielmarbach:
Note: Like I said I will backtrack on a few changes in this PR and make this more focused. In wanted to make a small note. Since we are talking about a major version it might make sense to adopt semver or semverish versioning and make a few APIs internal while still moving the necessary public APIs into other directions. If you need someone to brainstorm about it I have some experience

danielmarbach:
Rebased, backtracked on a few changes. The caveat is now that I have to use `OfType` to select the cancellable jobs. But since @lahma wants to keep the API changes minimal I think this is a good tradeoff.

danielmarbach:
@lahma please see also https://github.com/quartznet/quartznet/pull/339#issuecomment-201346049 Would be glad to help out

danielmarbach:
Same build error as in #336

danielmarbach:
Rebased and force pushed

lahma:
Looking good, thank you!

PR #347, opened 2016-04-04
Removed dbFailureRetryInterval since it is no longer used

danielmarbach:
This caught me badly on the quartz 3 branch. Master already has the `dbFailureRetryInterval` lingering around although no longer used. This then introduced to [a recursive call on the current 3.0 branch](https://github.com/quartznet/quartznet/blob/quartznet-3/src/Quartz/Impl/DirectSchedulerFactory.cs#L233) which leads to a StackOverflowException.


danielmarbach:
Still don't know how you guys handle breaking changes so I went ahead and just did it

lahma:
Thank you! Marked for 2.6 so at least some indication about changes to be aware of.

danielmarbach:
You marked it as 2.4 and not 2.6

lahma:
Yes, sorry I meant 2.4 which is next from the current 2.3.3.

PR #348, opened 2016-04-04
Removed dbFailureRetryInterval since it is no longer used

danielmarbach:
Connects to #331 and #347


lahma:
This seems to be a duplicate of 347? Removal of using NUnit.Framework at least looks wrong. Builds should now roll if you rebase against quartznet-3 branch. I made minimal patches there too, separate work going on creating new setup with FAKE.

danielmarbach:
Not a duplicate. One is against master. Will check regarding NUnit

danielmarbach:
Rebased, updated and pushed. Don't know how the nunit stuff got removed. Magic fingers ;)

lahma:
Thank you!

PR #353, opened 2016-04-20
add limit for mysql (slow when has large misfire job)

vinhhrv:
add limit for mysql (slow when has large misfire job)

### DIFF
--- a/src/Quartz/Impl/AdoJobStore/MySQLDelegate.cs
+++ b/src/Quartz/Impl/AdoJobStore/MySQLDelegate.cs
@@ -34,5 +34,10 @@ namespace Quartz.Impl.AdoJobStore
         {
             return SqlSelectNextTriggerToAcquire + " LIMIT " + maxCount;
         }
+
+        protected override string GetSelectNextMisfiredTriggersInStateToAcquireSql(int maxCount)
+        {
+            return SqlSelectHasMisfiredTriggersInState + " LIMIT " + maxCount;
+        }
     }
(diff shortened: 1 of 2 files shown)


src/Quartz/Impl/AdoJobStore/MySQLDelegate.cs
```diff
@@ -34,5 +34,10 @@ protected override string GetSelectNextTriggerToAcquireSql(int maxCount)
         {
             return SqlSelectNextTriggerToAcquire + " LIMIT " + maxCount;
         }
+
+        protected override string GetSelectNextMisfiredTriggersInStateToAcquireSql(int maxCount)
+        {
+            return SqlSelectHasMisfiredTriggersInState + " LIMIT " + maxCount;
```
lahma: HasMisfiredTriggersInState has API promise "The most misfired triggers to return, negative for all", how does MySQL limit behave with -1?

vinhhrv: Sori lahma,

I miss this case,  i found this line
int maxMisfiresToHandleAtATime = (recovering) ? -1 : MaxMisfiresToHandleAtATime;
20 or -1

I just commited.

Tks.

lahma:
Looking good, thanks for the contribution!

vinhhrv:
Hi lahma,

Can you bump to new version for this issue ?,
I have waited for a long time

Thank you

PR #359, opened 2016-04-27
Initial NET Core targeting support

lahma:
Changes against v3 branch that enable use against upcoming NET Core.


src/Quartz.Tests.Integration/Quartz.Tests.Integration.csproj
```diff
@@ -70,6 +70,26 @@
     </Reference>
     <Reference Include="System" />
     <Reference Include="System.Core" />
+    <Reference Include="FakeItEasy, Version=1.25.2.0, Culture=neutral, PublicKeyToken=eff28e2146d5fd2c, processorArchitecture=MSIL">
```
lahma: Do we still need hard references like these? I was so pleased about the project.json resolving issue by being the glue between csproj and packages.

src/Quartz/App_Packages/LibLog.4.2/LibLog.cs
```diff
@@ -492,7 +492,7 @@ static ILog For<T>()
             return GetLogger(typeof(T));
         }
 
-#if !LIBLOG_PORTABLE
+#if STACKTRACE && !LIBLOG_PORTABLE
```
lahma: LibLog project itself might benefit from these changes

src/Quartz/Collection/TreeSet.cs
```diff
@@ -20,26 +20,37 @@
...
 namespace Quartz.Collection
 {
     /// <summary>
     /// Only for backwards compatibility with serialization!
     /// </summary>
     /// <typeparam name="T"></typeparam>
+#if BINARY_SERIALIZATION
     [Serializable]
+#endif // BINARY_SERIALIZATION
+    [DataContract]  // TODO (NetCore Port): Confirm that data contract serialization works as expected here
```
lahma: DataContract attributes should not be needed with Json.NET? These pollute the code base a bit.

src/Quartz/CronExpression.cs
```diff
@@ -269,84 +278,127 @@ public class CronExpression : ICloneable, IDeserializationCallback
...
+        // Serializing TimeZones is tricky in .NET Core. This helper will ensure that we get the same timezone on a given platform,
+        // but there's not yet a good method of serializing/deserializing timezones cross-platform since Windows timezone IDs don't
+        // match IANA tz IDs (https://en.wikipedia.org/wiki/List_of_tz_database_time_zones). This feature is coming, but depending
+        // on timelines, it may be worth doign the mapping here.
+        // More info: https://github.com/dotnet/corefx/issues/7757
+        [DataMember]
+        private string timeZoneInfoId
+        {
+            get { return timeZone?.Id; }
+            set { timeZone = (value == null ? null : TimeZoneInfo.FindSystemTimeZoneById(value)); }
```
lahma: We have TimeZoneUtil.FindTimeZoneById that has some of the mappings already

src/Quartz/CronExpression.cs
```diff
@@ -208,12 +208,21 @@ namespace Quartz
     /// <author>Contributions from Mads Henderson</author>
     /// <author>Refactoring from CronTrigger to CronExpression by Aaron Craven</author>
     /// <author>Marko Lahma (.NET)</author>
+#if BINARY_SERIALIZATION
     [Serializable]
-    public class CronExpression : ICloneable, IDeserializationCallback
+#endif // BINARY_SERIALIZATION
+    [DataContract]
+    public class CronExpression : object
+#if ICLONEABLE
```
lahma: ICloneable might be something that we could just drop. We can still offer the Clone method that could return the actual type and state clearly whether it's a deep or a shallow copy.

src/Quartz/Impl/AdoJobStore/Common/dbproviders.netstandard15.properties
```diff
@@ -0,0 +1,19 @@
+# Database provider configuration data
+# Core information taken from Spring Framework .NET - All credits to their great work!
+
+# SQL SERVER
+quartz.dbprovider.SqlServer-20.productName=Microsoft SQL Server Core, provider V4.1.0.0 in framework .NET Core 1
```
lahma: is it dangerous to indicate that this is the same fellow from .NET 2.0 era or are they so equal that can be called SqlServer-20?

src/Quartz/Impl/AdoJobStore/JobStoreSupport.cs
```diff
@@ -3485,7 +3485,11 @@ private static void CheckNotZombied(ConnectionAndTransactionHolder cth)
 
             if (cth.Transaction != null && cth.Transaction.Connection == null)
             {
-                throw new DataException("Transaction not connected, or was disconnected");
+                // TODO (NetCore Port): I think keeping the exception type consistent between desktop and Core versions of Quartz
+                //                      is more important than keeping the desktop one consistent between the previous version and this one.
+                //                      It's not an obvious call, though, so may be worth discussing.
+                // throw new DataException("Transaction not connected, or was disconnected");
+                throw new InvalidOperationException("Transaction not connected, or was disconnected");
```
lahma: this is so rare case that we can probably just change this, things are pretty broken at this stage when this exception is thrown

src/Quartz/Impl/AdoJobStore/LockException.cs
```diff
@@ -22,15 +22,18 @@
...
+    /// resource lock.
+    /// </summary>
+    /// <seealso cref="ISemaphore" />
+    /// <author>James House</author>
+    /// <author>Marko Lahma (.NET)</author>
+#if BINARY_SERIALIZATION
+    [Serializable]
+#endif // BINARY_SERIALIZATION
+    // TODO (NetCore Port): Is this actually persisted anywhere or is it only serializable for remoting?
+    public class LockException : JobPersistenceException
```
lahma: We can probably make Exceptions non-serializable as Remoting is going away

src/Quartz/Impl/AdoJobStore/SimpleSemaphore.cs
```diff
@@ -41,7 +43,11 @@ namespace Quartz.Impl.AdoJobStore
     /// <author>Marko Lahma (.NET)</author>
     public class SimpleSemaphore : ISemaphore
     {
+#if REMOTING
```
lahma: here we could just have the non-remoting code path

src/Quartz/Impl/AdoJobStore/UpdateRowLockSemaphore.cs
```diff
@@ -129,7 +129,11 @@ public UpdateLockRowSemaphore(IDbProvider provider)
 
                 if (await cmd.ExecuteNonQueryAsync().ConfigureAwait(false) != 1)
                 {
-                    throw new DataException(
+                    // TODO (NetCore Port): I think keeping the exception type consistent between desktop and Core versions of Quartz
+                    //                      is more important than keeping the desktop one consistent between the previous version and this one.
+                    //                      It's not an obvious call, though, so may be worth discussing.
+                    // throw new DataException(
+                    throw new InvalidOperationException(    // Maybe SqlException instead?
```
lahma: is SqlException now DB-agnostic, before it was directly related to MS SQL Server exception?

src/Quartz/Impl/Matchers/StringOperator.cs
```diff
@@ -20,13 +20,17 @@
...
 namespace Quartz.Impl.Matchers
 {
     /// <summary>
     /// Operators available for comparing string values.
     /// </summary>
+#if BINARY_SERIALIZATION
     [Serializable]
+#endif // BINARY_SERIALIZATION
+    [DataContract]
     public abstract class StringOperator : IEquatable<StringOperator>
```
lahma: as operators are really not persisted, we could probably drop serialization support altogether now

src/Quartz/Impl/StdSchedulerFactory.cs
```diff
@@ -508,11 +516,14 @@ protected virtual void ValidateConfiguration()
 
             Type tpType = loadHelper.LoadType(cfg.GetStringProperty(PropertyThreadPoolType)) ?? typeof(SimpleThreadPool);
 
+#if WINDOWS_THREADPOOL  // TODO (NetCore Port) - We can't use the ClrThreadPool on .NET Core because cross-plat threadpool implementations don't expose all the APIs we need
+                        // Use the SimpleThreadPool there for now and in the future create a new async-friendly threadpool that will work cross-plat.
             if (tpType == typeof(SimpleThreadPool))
```
lahma: Yes, idea is to ditch the ClrThreadPool as soon as possible, but requires rewriting as you said. I probably need to learn more about the ".NET Core because cross-plat threadpool implementations don't expose all the APIs we need".

src/Quartz/Impl/AdoJobStore/Common/ConfigurationBasedDbMetadataFactory.cs
```diff
@@ -38,7 +40,11 @@ public ConfigurationBasedDbMetadataFactory(string sectionName, string providerNa
         /// <returns>The properties parser</returns>
         protected virtual PropertiesParser GetPropertiesParser()
         {
-            NameValueCollection settings = (NameValueCollection)ConfigurationManager.GetSection(sectionName) ?? new NameValueCollection();
+            NameValueCollection settings =
+#if CONFIGURATION
+                (NameValueCollection)ConfigurationManager.GetSection(sectionName) ??
```
lahma: So there's no old configuration infrastructure on core at all?

src/Quartz/Job/SendMailJob.cs
```diff
@@ -80,6 +82,7 @@ public virtual Task Execute(IJobExecutionContext context)
         {
             JobDataMap data = context.MergedJobDataMap;
 
+#if MAIL
```
lahma: We can hide the entire class behind the pre-processor directive. Users can be guided to have a look at the implementation and and write own Core version with mail library they like the most.

src/Quartz/NetCoreStubs/TypeFilterLevel.cs
```diff
@@ -0,0 +1,29 @@
...
+
+#if REMOTING
+[assembly: System.Runtime.CompilerServices.TypeForwardedTo(typeof(System.Runtime.Serialization.Formatters.TypeFilterLevel))]
+#else // REMOTING
+namespace System.Runtime.Serialization.Formatters
+{
+    //
+    // Summary:
+    //     Specifies the level of automatic deserialization for .NET Framework remoting.
+    public enum TypeFilterLevel
```
lahma: Hopefully won't be needed at all as remoting support will be removed.

src/Quartz/Simpl/DefaultObjectSerializer.cs
```diff
@@ -12,6 +16,35 @@ namespace Quartz.Simpl
     /// <author>Marko Lahma</author>
     public class DefaultObjectSerializer : IObjectSerializer
```
lahma: would separate JsonObjectSerializer and BinaryObjectSerializer now be the correct separation? We could make it mandatory for user to define in configuration.

mjrousos: Yes, I think that makes sense.

src/Quartz.Tests.Integration/Quartz.Tests.Integration.csproj
mjrousos: Ah, good catch. We don't need these. I think VS made these changes automatically because I was changing the oracle dependency through VS's interface (instead of directly in the project.json, as would have been better). I've reverted these changes (in the Quartz.Tests.Unit.csproj file, as well).

src/Quartz/App_Packages/LibLog.4.2/LibLog.cs
mjrousos: I was actually able to revert these changes because LibLog portable works directly on .NET Core now.

src/Quartz/Collection/TreeSet.cs
mjrousos: We can probably remove many of these. They're here for two reasons - one of which is a good reason and one of which is not :)

The good reason: By default, Json.NET will serialize all public fields/properties (even if no DataContract attribute is present), but it won't serialize private members. If a DataContract attribute is applied, it will instead serialize all members attributed with DataMember (whether public or private) and nothing else. So, in some cases, using data contract annotation members is important to make sure we're serializing the same members that would have been serialized previously. I suspect that we have (many) more attributes than are really necessary just for that, though (this comment is on one such attribute which could be removed).

The not very good reason they're here: When I started working on this port, I was using DataContractSerialization instead of Json.Net and DCS requires many more data contract attributes. Because of that, I attributed pretty much everything. In the end, I switched to Json.Net because I found DCS is not very good at handling serialization of types that aren't known at compile time (if i included custom types in a JobDataMap it caused problems, for example). However, after switching to Json.Net the attributes were still left behind.

Now that we're not using DCS, many of these probably aren't necessary anymore. I've created mjrousos/quartznet#1 (and assigned it to myself) to track this cleanup

src/Quartz/CronExpression.cs
mjrousos: If we don't need to implement ICloneable on desktop .NET then dropping it would make the source a little cleaner.

src/Quartz/CronExpression.cs
mjrousos: Ah, I didn't see that. Cool. I'll call that method instead of TZI.FindSystemTimeZoneById directly and if we need better mapping, we can make changes in TimeZoneUtil.

src/Quartz/Impl/AdoJobStore/Common/ConfigurationBasedDbMetadataFactory.cs
mjrousos: No, unfortunately not. The rationale is that different app models will have different config models that make sense for them (web apps use one model, store apps use another, etc.). So, we've implemented new config models in Framework layers that are targeted at specific app models (ASP.NET Core 1 has a new config infrastructure, for example), but there's nothing common for libraries to use. The guidance for libraries is to allow apps to pass configuration data in programatically and make the apps responsible for reading the actual config files.

In Quartz.Net's case, the quartz.config files still work since we read them directly, but there's no support for reading from an app.config like there was before.

src/Quartz/Impl/AdoJobStore/Common/dbproviders.netstandard15.properties
mjrousos: Good point. They are very similar, but we should probably use a different identifier here. I'll change it to SqlServer-41 or something like that.

src/Quartz/Impl/AdoJobStore/JobStoreSupport.cs
mjrousos: Ok, I'll just use InvalidOperationException and remove the note about DataException.

src/Quartz/Impl/AdoJobStore/LockException.cs
mjrousos: Yes, that makes sense. I'll remove the serializable attribute, etc.

src/Quartz/Impl/AdoJobStore/SimpleSemaphore.cs
mjrousos: Yes, that will be simpler.

src/Quartz/Impl/AdoJobStore/UpdateRowLockSemaphore.cs
mjrousos: Good point. This should probably be 'DbException' (which is the more common base class SqlException derives from.

mjrousos: Ah, I just realized DbException is an abstract class. We're not meant to throw it directly. Perhaps throwing InvalidOperationException here is good enough.

src/Quartz/Job/SendMailJob.cs
mjrousos: Sounds good.

src/Quartz/NetCoreStubs/TypeFilterLevel.cs
mjrousos: Yes, we should be able to remove this once remoting support is gone. For the moment I needed it because RemotingSchedulerExporter has a public member (TypeFilterLevel) of this type and I didn't want to change public surface area until you'd given input on the sorts of changes you were comfortable with.

src/Quartz/Impl/Matchers/StringOperator.cs
mjrousos: Sounds good. I've removed serialization attributes here.

src/Quartz/Impl/StdSchedulerFactory.cs
mjrousos: @ericeil, I'm helping to port the Quartz.Net scheduling library to Core and one of the obstacles is ThreadPool changes. The Quartz.Net library allows users to schedule tasks and also control the level of concurrency in the task runners that process those schedules and kick things off.

One of the ways this was done previously was by queuing tasks in the ThreadPool when their schedules triggered while using GetMaxThreads/SetMaxThreads/GetAvailableThreads to make sure that the number of tasks running at any particular time was capped at the limit set by the user.

So, two questions here -

1. Can you help us understand briefly what the rationale was for removing those APIs? Was it more a matter of what was practical to implement cross-plat or was it more that those APIs just aren't recommended for use anymore?
2. How would you suggest we make a library like this work in .NET Core? It seems like it should be possible to create a task-based solution that begins tasks at their trigger times while keeping a count of executing tasks so that we limit concurrency. All that could be done without using the CLR's thread pool. Do you think that's the best approach?

Thanks for letting us pick your brain!

ericeil: There is some discussion of the Min/MaxThreads issue here: https://github.com/dotnet/cli/issues/889#issuecomment-173331862.  

The simplest way to limit concurrency in .NET Core is probably to use SemaphoreSlim.  For example:

```csharp
    class ConcurrencyLimiter
    {
        private readonly SemaphoreSlim _sem;

        public ConcurrencyLimiter(int maxConcurrencyLevel)
        {
            _sem = new SemaphoreSlim(maxConcurrencyLevel);
        }

        public async Task Schedule(Func<Task> asyncOperation)
        {
            await _sem.WaitAsync();
            try
            {
                await asyncOperation();
            }
            finally
            {
                _sem.Release();
            }
        }
    }
```

mjrousos: Thanks, @ericeil. That makes sense. @lahma, I can take a look into making a task scheduler that works on .NET Core tomorrow. I'll keep you posted on that investigation.

lahma: I'd like to just point out the current situation, we had SimpleThreadPool that used to handle fixed size worker pool: https://github.com/quartznet/quartznet/blob/quartznet-3/src/Quartz/Simpl/SimpleThreadPool.cs . This was the guarantee to have predictability over .NET own thread pool. Fixed size, fixed availability consumed only by Quartz workers.

Initial thought (before .NET Core portability talk) was to try to utilize TPL's example QueuedTaskScheduler (Masstransit for one uses this now in async world): https://github.com/quartznet/quartznet/blob/quartznet-3/src/Quartz/Util/QueuedTaskScheduler.cs

The current ClrThreadPool was just initial torso there as a TODO.

Thank you both for helping to find a solution. Your help is much appreciated as I'm very new to .NET Core limitations and possibilities.

mjrousos: The QueuedTaskScheduler looks useful. We may be able to simplify it some, though, because I think Quartz.Net doesn't need a priority queue. The old threadpools accepted a priority, but that was just the priority for all Quartz tasks relative to other items that might be in the thread pool. The QueuedTaskScheduler's priority queue is for assigning different priorities to tasks coming from Quartz.Net, which I think isn't something the scheduler interfaces support currently.

I'm looking more carefully through QueuedTaskScheduler now and will likely have some more thoughts on it shortly.

mjrousos: Quick update - I've got an updated IThreadPool that seems to work well in Core but I want to go over the code a little more carefully before adding it to the PR. I'll clean it up early next week.

lahma: Sounds promising. Thread priorities are a bit of an oddity in a way. They do share the same priority in pool (the original implementation) but jobs themselves were able to boost priority when starting (setting current thread to higher priority). Thread pool resets worker's priority back to pool's default after job completes.

Quartz also has notion of trigger priority which determines the order how triggers are queued for execution when multiple triggers are scheduled to run in same retrieved batch. This affects execution order but not the resources allowed for the job running.

In my opinion, boosting thread/task priority can be left as an enhancement for later. In async/await world there might not be many who run own actual Thread instances beside Quartz's threads inside same process/appdomain - this is were the prioritization would have an effect. Per task priority of course is an interesting ability but yet again can be different between core and desktop.

mjrousos:
I've added some more commits to this PR which add support for task-based thread pools. I added an abstract base (TaskSchedulingThreadPool) and two implementations of it - one which schedules with TaskScheduler.Default and one that schedules with the QueuedTaskScheduler already in the Quartz.Net project.

There's no API for changing thread priority with these thread pools, but other functionality should be working.

src/Quartz/Simpl/TaskSchedulingThreadPool.cs
```diff
@@ -0,0 +1,284 @@
...
+        /// <param name="runnable">The action to be executed</param>
+        /// <returns>True if the task was successfully scheduled, false otherwise</returns>
+        public bool RunInThread(Action runnable)
+        {
+            if (runnable == null || !isInitialized || shutdownCancellation.IsCancellationRequested) return false;
+
+            // Acquire the semaphore (return false if shutdown occurs while waiting)
+            try
+            {
+                concurrencySemaphore.Wait(shutdownCancellation.Token);
```
ericeil: Does IThreadPool require that RunInThread block the current thread until a thread pool thread is available?  Or does it just require that the scheduler limit the actual concurrency?  If the latter, I'd do this wait asynchronously, to avoid blocking the current thread...

src/Quartz/Simpl/TaskSchedulingThreadPool.cs
```diff
@@ -0,0 +1,284 @@
...
+
+            return 0;
+        }
+
+        /// <summary>
+        /// Schedules an action to run (using the task scheduler) as soon as concurrency rules allow it
+        /// </summary>
+        /// <param name="runnable">The action to be executed</param>
+        /// <returns>True if the task was successfully scheduled, false otherwise</returns>
+        public bool RunInThread(Action runnable)
```
lahma: We no longer need this Action-based version. Func<Task> should suffice now.

mjrousos: Ok, I'll go ahead and remove this overload.

src/Quartz/Simpl/TaskSchedulingThreadPool.cs
mjrousos: That's a good question. Old IThreadPool implementations blocked the thread. I'm not sure if the schedulers that use these classes depend on that or not. I'll take a look at the schedulers to see if I can figure it out. Do you know, @lahma?

Would it make sense to make RunInThread async so that schedulers don't have to wait for the tasks to start or do the schedulers depend on this blocking?

lahma: I can't come up a reason to block from the top of my head. As long as concurrency is limited (max n threads run at a time) we should be covered. QuartzSchedulerThread pushes jobs to run and should be happy if queued item gets safely into queue and will be executed. JobRunShell that is the execution target will take care of notifications after ran has been completed.

mjrousos: Got it. I'll leave this as it is temporarily, but as part of the async-ification of Quartz.Net we should try making this whole method async. That way the scheduler won't block waiting for a thread to be available and the code will get a bit simpler since we can just await the task and not have to bother with RemoveTaskFromRunningList-related-cleanup in a separate method called by ContinueWith.

PR #360, opened 2016-05-03
Add initialize step for named scheduler

tarasMelnichuk:


lahma:
Thanks for this, @tarasMelnichuk — and sorry it sat here this long. The problem you described is real and was reported again nine years later as #2786: `GetScheduler(name)` returns `null` unless `GetScheduler()` happens to have been called first, even when the name asked for is the very scheduler the factory is configured to produce.

The behaviour you proposed now ships on both branches:

- **3.x**: the DI factory was fixed in #2845; #3188 does the same for the property-configured `StdSchedulerFactory`, which is the case in this PR. It also initializes the factory before reading the configured name (the patch here would have hit a `NullReferenceException` on an uninitialized factory) and compares the name case-insensitively, matching how `SchedulerRepository` indexes names.
- **4.0 / main**: `DefaultSchedulerFactory` already creates the scheduler on a named lookup; #3189 aligns its name comparison with the repository and adds the regression tests for the property-configured path.

In both, only the factory's **own** name creates — any other name stays a pure lookup, so probing for a scheduler somebody else owns has no side effects.

Closing this one, as the branch it targets no longer exists in this form and the change has been carried forward in #3188 and #3189. Credit is noted there.

PR #361, opened 2016-05-03
Fixup serialization attributes

mjrousos:
This removes all the unnecessary DataContractSerialization attributes.

It adds in a few needed Json.NET attributes needed, instead. Also, adds a custom Json.Net contract resolver to prevent serializing read-only properties (since they aren't needed for deserialization and only bloat the serialized json).


lahma:
Thank you! Already much cleaner. As you noted comments it would be good shift towards get/private set properties when possible, but that can wait too as we also need to keep track how much public API changes are made.

PR #365, opened 2016-05-07
Async cleanup

danielmarbach:


src/Quartz/Core/QuartzScheduler.cs
```diff
@@ -1255,6 +1255,7 @@ public virtual Task<IJobDetail> GetJobDetail(JobKey jobKey)
             return resources.JobStore.RetrieveJob(jobKey);
         }
 
+#pragma warning disable AsyncFixer01 // Unnecessary async/await usage
```
lahma: Nice, learned about a new extension this way!

src/Quartz/Core/QuartzScheduler.cs
```diff
@@ -1265,6 +1266,7 @@ public virtual Task<IJobDetail> GetJobDetail(JobKey jobKey)
 
             return await resources.JobStore.RetrieveTrigger(triggerKey).ConfigureAwait(false);
```
lahma: however don't see why the await is necessary here

danielmarbach: Yeah sorry I should have linked it. It is pretty handy. Especially of you do not know the drill yet.

danielmarbach: It is necessary because of how `Task<T>` is designed. It doesn't support the necessary variance to be able to treat `Task<IOperableSomething>` as the defined returned type of the method.

PR #367, opened 2016-05-17
[SPIKE] Task based API spike for job worker pool

danielmarbach:
* Switched to .NET 4.6.1 to have support for `TaskCreationOptions.RunContinuationsAsynchronously` (see bug in 4.6 https://github.com/dotnet/coreclr/issues/2021)
* Implemented `AsyncManualResetEvent` for pause support
* Introduced really dumb version `IJobWorkerPool`

### DIFF
--- a/src/Quartz.Examples/App.config
+++ b/src/Quartz.Examples/App.config
@@ -1,24 +1,23 @@
 <?xml version="1.0"?>
-
 <configuration>
 
   <configSections>
-    <section name="log4net" type="log4net.Config.Log4NetConfigurationSectionHandler, log4net" />
+    <section name="log4net" type="log4net.Config.Log4NetConfigurationSectionHandler, log4net"/>
   </configSections>
 
   <log4net>
     <appender name="ConsoleAppender" type="log4net.Appender.ColoredConsoleAppender">
       <layout type="log4net.Layout.PatternLayout">
-        <conversionPattern value="%d [%t] %-5p %logger - %m%n" />
+        <conversionPattern value="%d [%t] %-5p %logger - %m%n"/>
       </layout>
     </appender>
     <root>
-      <level value="INFO" />
-      <appender-ref ref="ConsoleAppender" />
+      <level value="INFO"/>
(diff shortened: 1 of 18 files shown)


danielmarbach:
@lahma this is far from complete, but as you can see by using async primitives we can get rid of some locking and field state toggling. We can also get rid of Monitor.Wait and Pulse and switch completely block free code. 

Furthermore we could also float the cancellation token into the job worker pool. We could even think about floating it over the worker pool into the `JobRunShell.Run` method by changing the signature of `Run` on the worker pool to something like `Run(Func<CancellationToken, Task> runnable, CancellationToken token = default(CancellationToken))`

lahma:
Based on the discussion of the bug it seems that even 4.6.1 does not fix it?

> unfortunately this was discovered too late for 4.6.1 but it will be in the next update. But it will be on coreclr before that.

Target frameworks might be a good candidate for a Gitter chat, tried to decipher alternatives last night and no fun.

lahma:
Thanks, looks promising!

danielmarbach:
@lahma why did you merge this??? Far from done and I haven't implemented the `Set` and `Reset` of the manual reset event yet. So this is super buggy right now. I suggest you revert

danielmarbach:
The idea of this PR was more to explore alternatives and drive the implementation to a point where we can say: That works and then reimplement the necessary changes on a clean slate

lahma:
I have reverted it now. Merged the PR as you didn't state that it shouldn't be pulled in.

danielmarbach:
@lahma 

> this is far from complete
https://github.com/quartznet/quartznet/pull/367#issuecomment-219866867

Sorry I thought it was clear. In other OSS projects we use conventions like [SPIKE] or [WIP] to indicate that a PR should not be merged and is either a thought experiment or work in progress.

danielmarbach:
Can you reopen this PR? It seems I can't without sending in a new one

lahma:
Sorry, there seems to be no option to reopen a PR that has been once merged.

PR #368, opened 2016-05-17
Test updates

mjrousos:
A handful of test updates (and a couple small product fixes) to account for recent serialization and async changes.


lahma:
Thanks, looking good!

Issue #352, opened 2016-04-16
Trigger GetNextFireTimeUtc not updated / wrong after first fire


KristjanLaane:
When scheduling with the following trigger, the first next time is correct:

```
 DailyTimeIntervalScheduleBuilder.Create()
                        .WithMisfireHandlingInstructionFireAndProceed()
                        .StartingDailyAt(Quartz.TimeOfDay.HourAndMinuteOfDay(hour, minutes))
                        .OnEveryDay()
                        .WithIntervalInHours(24)
                        .WithRepeatCount(9999)
                        .InTimeZone(timeZone)

```
But after the trigger has fired for the first time, the next time is still the same, not 24 hours later as it should be in the case. 

These reports suggest other people have had the same issue unresolved since 2010:
https://groups.google.com/forum/?fromgroups#!searchin/quartznet/firetime/quartznet/0ssZGCBEnlA/yzZZEPyTxcMJ
https://groups.google.com/forum/?fromgroups#!searchin/quartznet/firetime/quartznet/elHG3O4npj0/yVl6P_Swx2cJ

KristjanLaane:
How do you handle bugs? is this likely to get into milestone 3 https://github.com/quartznet/quartznet/milestones/3.0
or are bug prioritized on a separate track?
thanks.

lahma:
Sorry for the delay (well it's been only 6 years or so), I'll schedule this is for 2.4 for checkup, but to be honest my limited time has been mostly wrapped around 3.0 issues. A pull request with test case(s) is always welcome.

KristjanLaane:
As you pointed out in the pull request comments, GetNextFireTimeUtc provides the correct value if a temporary trigger is used:
var tempTrigger = scheduler.GetTrigger(trigger.Key);
As opposed to getting the wrong value if using a trigger reference that was used in scheduling in the first place. 

This is a problem from the point of view of the consumer of the Quartz library as the developer is required to possess conditional knowledge of the internal workings of the library and know that if using the very same property in one case the value is correct, but in other cases it is not. I can attest it has already lead to problems in production and frustration from users :( :) And I'm not the only one as the links in the original post indicate. In some ways an "unreliable" property is a highly undesirable thing to expose, because of the hidden dangers it brings with it. 

To avoid such confusion, perhaps the NextFireTime property should not be a property of ITrigger, but instead it should be accessible through IScheduler with something like GetNextFireTimeUtc(triggerKey)?

[...]

PR #374, opened 2016-06-01
Failing test for NextFireTimeUtc

KristjanLaane:
NextFireTimeUtc does not work correctly after a trigger has fired once, as detailed in this issue https://github.com/quartznet/quartznet/issues/352
The failure of the test added here proves it.

### DIFF
--- a/Quartz.build
+++ b/Quartz.build
@@ -389,6 +389,7 @@
         <include name="System.dll" />
         <include name="System.Data.dll" />
         <include name="System.Transactions.dll" />
+        <include name="System.Windows.Forms.dll" />
         <include name="${common-logging-dll}" />
         <include name="${common-logging-core-dll}" />
         <include name="${nunit-dll}" />
--- a/src/Quartz.Tests.Integration/Quartz.Tests.Integration.2010.csproj
+++ b/src/Quartz.Tests.Integration/Quartz.Tests.Integration.2010.csproj
@@ -105,6 +105,7 @@
       <RequiredTargetFramework>3.5</RequiredTargetFramework>
     </Reference>
     <Reference Include="System.Data" />
+    <Reference Include="System.Windows.Forms" />
   </ItemGroup>
   <ItemGroup>
     <Compile Include="..\AssemblyInfo.cs">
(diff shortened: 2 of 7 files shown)


lahma:
Thanks for the repro code. I'll try to dig deeper hopefully this weekend.

lahma:
The test case TestNextFireTimeUtc fails because you try to observe internal scheduler state via trigger's copy. Test case works as expected when you retrieve the trigger before checking fire times:

    trigger = scheduler.GetTrigger(trigger.Key);
    var firstNextFireTimeUtc = trigger.GetNextFireTimeUtc().Value;
    // cannot use Assert.True(firstCorrectNextFireTimeUtc.Equals(firstNextFireTimeUtc)) because the milliseconds differ
    var correctAndActualDifferenceFirst = (firstCorrectNextFireTimeUtc - firstNextFireTimeUtc).Duration();
    Assert.LessOrEqual(correctAndActualDifferenceFirst, TimeSpan.FromSeconds(2), "Next Fire Time Should Have Been {0}, but Was Wrongly {1}", firstCorrectNextFireTimeUtc, firstNextFireTimeUtc);
    Console.WriteLine("First Trigger Time Prediction Correct");
    // wait for first fire to pass
    System.Threading.Thread.Sleep(TimeSpan.FromSeconds(10));
    Console.WriteLine("Checking When Trigger Should Fire The Second Time");

    trigger = scheduler.GetTrigger(trigger.Key);
    var secondNextFireTimeUtc = trigger.GetNextFireTimeUtc().Value;
    var correctAndActualDifferenceSecond = (secondCorrectNextFireTimeUtc - secondNextFireTimeUtc).Duration();
    Assert.LessOrEqual(correctAndActualDifferenceSecond, TimeSpan.FromSeconds(2), "Next Fire Time Should Have Been {0}, but Was Wrongly {1}", secondCorrectNextFireTimeUtc, secondNextFireTimeUtc);
    Console.WriteLine("Second Trigger Time Prediction Also Correct");

lahma:
I'm closing this one as the assumption was wrong about how to monitor state of scheduler.

Issue #375, opened 2016-06-03
JobExecutionContext ScheduledFireTimeUtc Wrong so Misfire Detection Also Not Working


KristjanLaane:
Hi again!

I found another issue in production that is rather impactful but was hard to discover, which stems from ScheduledFireTimeUtc being provided as UTC now, when it should have been earlier in the past due to a misfire e.g. because the device was in sleep during the scheduled fire time or when the system clock is pushed into the future. As a consequence the code wrongly never sees any misfires right now, because misfires are detected based on ScheduledFireTimeUtc.

I have made a pull request with the test TestScheduledFireTimeUtc that reproduces this problem. Note that it is a semi-manual test requiring you to manually either push the operating system clock forward or put the device to sleep at the right time, in order to cause a misfire. 

As a start for the detective work of why ScheduledFireTimeUtc is wrong, it looks like [it is set from RAMJobStore as trigger.GetPreviousFireTimeUtc()](https://github.com/quartznet/quartznet/blob/b03aa020168a63ab062cb703804e845a07c2003f/src/Quartz/Simpl/RAMJobStore.cs#L1640). This may mean that PreviousFireTimeUtc is not working correctly as well, or it may mean that ScheduledFireTimeUtc should not be set as trigger.GetPreviousFireTimeUtc() in some circumstances, not sure? The other place that ScheduledFireTimeUtc is set [is in the AdoJobStore namespace](https://github.com/quartznet/quartznet/blob/b03aa020168a63ab062cb703804e845a07c2003f/src/Quartz/Impl/AdoJobStore/JobStoreSupport.cs#L2738).

Thanks for your expert work!

PR #376, opened 2016-06-03
Failing manual test for ScheduledFireTimeUtc

KristjanLaane:
ScheduledFireTimeUtc likes to stay in the present even though it should sometimes reflect the past, as detailed in this issue https://github.com/quartznet/quartznet/issues/375


lahma:
Root cause was fixed.

PR #406, opened 2016-09-19
fix for select SqlSelectJobForTrigger

00raq00:
now it's using primary key


lahma:
Wow that was a good catch, changing index scan to index seek. Thank you!

PR #407, opened 2016-09-20
fix for select SqlSelectJobForTrigger to 2.x

00raq00:
now it's using primary key

same fix as to master 3.0 alpha


lahma:
Thank you!

00raq00:
Could you publish new nuget package :) ?

lahma:
I'll try to publish new version soon. May I ask how did you originally face this issue/find out about the problem? Was this a stressing performance problem or something more minor that came up in database server logs?

00raq00:
We are using quartz in our wcf which under high system load threw a lot of timeouts. In SQL Profiler we saw that it took lots milliseconds and executing plan wasn't fine, there wasn't index seek but index scan.

Issue #413, opened 2016-10-12
A potential bug in QuartzNet on Oracle


chichaoatabc:
Hi,
        I found a potential bug in QuartzNet2.3.3 on Oracle database.
        The problem is that the next tigger can't be acquired normally, after the oracle executes automatic SQL tuning task.
        When the scheduler acquires next trigger from QRTZ_TRIGGERS on Oracle, Quartz invokes "GetSelectNextTriggerToAcquireSql()" method in "OracleDelegate.cs".The SQL string it returns is 
"SELECT trigger_name,trigger_group,next_fire_time,priority FROM QRTZ_TRIGGERS WHERE rownum<=1 AND SCHED_NAME=’ServerScheduler’ AND TRIGGER_STATE=’waiting’ AND NEXT_FIRE_TIME<=’636094216783726571’ AND (MISFIRE_INSTR=-1 OR (MISFIRE_INSTR <> -1 AND NEXT_FIRE_TIME >=’63609421618372657’)) ORDER BY NEXT_FIRE_TIME ASC,PRIORITY DESC" 
        It acquires the first waiting trigger from NEXT_FIRE_TIME ascending sort set,this trigger has the minimum NEXT_FIRE_TIME.The expected resultset is as follow.
![default](https://cloud.githubusercontent.com/assets/20850340/19295458/4aeeed08-9067-11e6-85d9-cc044d0ef379.jpg)

```
    But sometimes the set is...
```

![default](https://cloud.githubusercontent.com/assets/20850340/19295499/79b5a744-9067-11e6-81ab-7e1d331229cd.jpg)

```
   So the scvTrigger_4 will always be acquired, other triggers will never be checked.
```

ANALYSIS: 
       Oracle may execute automatic SQL tuning task at 22:00 every day.This task includes DB index statistics update.After tuning the index ,the result of the SQL string changed.The rownum will indicate the order of record inserted,not the order of NEXT_FIRE_TIME,as the index NEXT_FIRE_TIME will never be used in this SQL. Also the Oracle Technical Support suggest us "rownum" and NEXT_FIRE_TIME index shouldn't be used together in this SQL.

Solution:
       Subquery is needed,the method should be as follow.
public class OracleDelegate: StdAdoDelegate
{
    protecte over string GetSelectNextTriggerToAcquireSql(int maxCount)
    {
        return "select \* from ("+SqlSelectNextTriggerToAcquire+") where rownum<=" +maxCount;
    }
    ......
}

[...]

PR #414, opened 2016-10-12
Update OracleDelegate.cs

chichaoatabc:
Oracle may execute automatic SQL tuning task at 22:00 every day.This task includes DB index statistics update.After tuning the index ,the result of the SQL string changed.The rownum will indicate the order of record inserted,not the order of NEXT_FIRE_TIME,as the index NEXT_FIRE_TIME will never be used in this SQL. 
https://github.com/quartznet/quartznet/issues/413


lahma:
Thanks, I'll do minor extra tweaks but looks good!

PR #419, opened 2016-11-03
System.TimeZoneNotFoundException fix

shamork:
"Couldn't retrieve trigger: Exception of type 'System.TimeZoneNotFoundException' was thrown."

added timezone alias to fix this


lahma:
Thanks for your contribution!

PR #426, opened 2016-11-22
Create WorkerThread in virtual method

NathanBaulch:
Bit of a long story (which I can share if interested), but I really need to intercept the creation of `WorkerThread` instances in `SimpleThreadPool`. Hopefully an uncontroversial change.


lahma:
Looks good, thank you.

Issue #464, opened 2017-03-31
RemotingSchedulerExporter can't create remoting channel on Mono


mao29:
I'm using version 2.5 of Quartz.Net. I'm trying to run Quartz.Server.2010 on Linux (Ubuntu Mate) with Mono (4.6.1). I didn't change any code in Quartz.Server project and try to run it from MonoDevelop. I get this output in console:
<details>
<summary>Console output</summary>
Loaded assembly: /mnt/Work/Quartz.NET-2.4.1/build/4.0/Debug/Quartz.Server/Quartz.Server.exe
Loaded assembly: /mnt/Work/Quartz.NET-2.4.1/build/4.0/Debug/Quartz.Server/Topshelf.dll [External]
Loaded assembly: /usr/lib/mono/gac/System.Core/4.0.0.0__b77a5c561934e089/System.Core.dll [External]
Loaded assembly: /usr/lib/mono/gac/System/4.0.0.0__b77a5c561934e089/System.dll [External]
Loaded assembly: /usr/lib/mono/gac/System.ServiceProcess/4.0.0.0__b03f5f7f11d50a3a/System.ServiceProcess.dll [External]
Loaded assembly: /mnt/Work/Quartz.NET-2.4.1/build/4.0/Debug/Quartz.Server/Common.Logging.Core.dll [External]
Loaded assembly: /mnt/Work/Quartz.NET-2.4.1/build/4.0/Debug/Quartz.Server/Quartz.dll
Loaded assembly: /usr/lib/mono/gac/System.Configuration/4.0.0.0__b03f5f7f11d50a3a/System.Configuration.dll [External]
Loaded assembly: /usr/lib/mono/gac/System.Xml/4.0.0.0__b77a5c561934e089/System.Xml.dll [External]
Loaded assembly: /usr/lib/mono/gac/Mono.Security/4.0.0.0__0738eb9f132ed756/Mono.Security.dll [External]
Topshelf.HostFactory Information: 0 : Configuration Result:
[Success] Name QuartzServer
[Success] DisplayName Quartz Server
[Success] Description Quartz Job Scheduling Server
[Success] ServiceName QuartzServer
Configuration Result:
[Success] Name QuartzServer
[Success] DisplayName Quartz Server
[Success] Description Quartz Job Scheduling Server
[Success] ServiceName QuartzServer
[...]

PR #465, opened 2017-04-07
Fix for issue #464 RemotingSchedulerExporter can't create remoting channel on Mono

mao29:
Fix for issue #464


lahma:
This seems like a breaking change as this removes possibility to provide the three parameters that are valid through Mono's validation. Maybe we need to create a cleaned Hashtable from CreateConfiguration() output that contains all valid parameters that are present in configuration ("includeVersions", "strictBinding", "typeFilterLevel")?

mao29:
Thank you for response. I can see your point. I've made separate method to extract only valid parameters for BinaryServerFormatterSinkProvider.

lahma:
Thank you!

Issue #477, opened 2017-05-02
Quartz.Web in VS2017


br0dieman9:
Hi,

I saw issue #447 What is the status of Quartz.Web in v3 was posted back in February.

I was wondering if you'd made any progress with the web project since that issue was opened.

I've tried downloading a zip file from github, I can build the Quartz.sln in VS2017, but the Quartz.Web project will not compile. Build produces this error:
![image](https://cloud.githubusercontent.com/assets/6833000/25631824/8fa32b86-2f3f-11e7-9398-922687b80bc9.png)

Running nuget restore causes more errors:
![image](https://cloud.githubusercontent.com/assets/6833000/25631869/b859eede-2f3f-11e7-8d0d-61334d3657d0.png)

I've also tried to use the build.cmd file to build the project (after cloning the github repo), which compiles successfully. Running dotnet run after the build.cmd errors out also:
![image](https://cloud.githubusercontent.com/assets/6833000/25634087/5a3650e2-2f47-11e7-8ad6-ca43e071cc52.png)

Please let me know if I am missing a step somewhere.

lahma:
Closing as this is not currently in active development.

PR #497, opened 2017-07-11
Fix for #477:  converted Quartz.Web project to asp.net core 2.0

smartquant:
Hi Marko,
I have converted the Quartz.Web project to asp.net core 2.0. 

- Project compiles and it has been added to the Quartz.sln file
- Swagger works on localhost:5000/swagger
- SignalR is commented out for it is not included yet in the official asp.net core release
- Index.html is hosted by default, but it does not show anything but the title. I am not a web expert so I have not touched any of the web files in /App

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -36,6 +36,36 @@
             "type": "coreclr",
             "request": "attach",
             "processName": "<example>"
+        },
+        {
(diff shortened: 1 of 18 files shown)


lahma:
Thanks, much appreciated.

PR #512, opened 2017-08-18
add mysql 6.9.9 provider support

Wahitler:
my sql
image

mysql.data 6.9.9 is stable version in nuget ,so add mysql 6.9.9 provider support.
please update quartz.net in nuget.
thanks

### DIFF
--- a/src/Quartz/Impl/AdoJobStore/Common/dbproviders.properties
+++ b/src/Quartz/Impl/AdoJobStore/Common/dbproviders.properties
@@ -199,6 +199,25 @@ quartz.dbprovider.MySql-695.useParameterNamePrefixInParameterCollection=true
 quartz.dbprovider.MySql-695.bindByName=true
 quartz.dbprovider.MySql-695.dbBinaryTypeName=Blob
 
+
+# Connector/NET 6.9 (.NET 2.0/4.0) -->
+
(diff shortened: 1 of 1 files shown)


lahma:
I'm trying to steer away from creating version specific providers. Have your tried to use provider with name "MySql" instead of proposed "MySql-69". It binds to version that it can find.

Wahitler:
I tried and this is my quartz.config. i am working on quartz.net 2.6
image

and this is my falut.
image

I notice the lastest code haven't  specific providers,but tag v2.6 not.

qq 20170818184218

lahma:
Ah I see, it's only in v3 (master). Does binding redirect help?

Wahitler:
That means,If I work on 2.x,I shoud Redirect Assembly Versions?
and I only set provider as 'mysql'  when  I work on 3.x?

lahma:
Ideally if you can just do the binding redirect you won't need to wait for next servicing release for 2.6, it's not actively worked on at the moment. Probably will have a new version of MySQL by then... And yest, for 3.x should work with just the 'MySQL'. 

The 3 alpha should be pretty solid, I'm keeping the alpha sticker to allow further API development if needed.

Wahitler:
thanks,I tried on 2.6, that's ok through binding redirect
thank you for all /laugh

PR #552, opened 2017-11-23
Implement better random number generator

dbraaten42:
Implement Random number generator based on RNGCryptoServiceProvider which provides better randomness then System.Random.


lahma:

PR #580, opened 2018-01-26
Fixes serialization issues. Properties with backing fields are needed…

Sicos1977:
… when using [NonSerialized] attributes


lahma:
Thanks for handling this!

Sicos1977:
Your welcome... a small question. How do I setup trace logging in QuartzNet?

lahma:
You need to delegate to some logging framework via liblog and configure it. Serilog and log4net should be quite easy to set up. If you really don't want a dependency on some framework you could [write your own adapter](https://www.quartz-scheduler.net/documentation/quartz-3.x/quick-start.html#adding-logging).

PR #583, opened 2018-01-30
2 fixes for QuartzNet

Sicos1977:
- Prevent faulty JobListeners from influencing QuartzNet logic.
- Do not always trust ReSharper to use properties with expression bodies.

See the TestRemoting -> Test method, I call the StdSchedulerFactory overthere twice and the second call would raise an exception that the remoteinstance scheduler already existed.

### DIFF
--- a/src/Quartz/Impl/StdSchedulerFactory.cs
+++ b/src/Quartz/Impl/StdSchedulerFactory.cs
@@ -138,7 +138,11 @@ namespace Quartz.Impl
 
         private static readonly ILog log = LogProvider.GetLogger(typeof(StdSchedulerFactory));
 
-        private string SchedulerName => cfg.GetStringProperty(PropertySchedulerInstanceName, "QuartzScheduler");
+        private string SchedulerName
+        {
+            // ReSharper disable once ArrangeAccessorOwnerBody
+            get { return cfg.GetStringProperty(PropertySchedulerInstanceName, "QuartzScheduler"); }
+        }
 
         private ILog Log => log;
 
--- a/src/Quartz/Listener/BroadcastJobListener.cs
+++ b/src/Quartz/Listener/BroadcastJobListener.cs
@@ -25,6 +25,8 @@ using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 
+using Quartz.Logging;
+
 namespace Quartz.Listener
 {
     /// <summary>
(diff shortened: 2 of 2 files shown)


src/Quartz/Impl/StdSchedulerFactory.cs
```diff
@@ -138,7 +138,11 @@ public class StdSchedulerFactory : ISchedulerFactory
 
         private static readonly ILog log = LogProvider.GetLogger(typeof(StdSchedulerFactory));
 
-        private string SchedulerName => cfg.GetStringProperty(PropertySchedulerInstanceName, "QuartzScheduler");
+        private string SchedulerName
```
lahma: isn't this basically the very same code, expression bodied members translate to get/set?

src/Quartz/Listener/BroadcastJobListener.cs
```diff
@@ -99,22 +103,66 @@ public bool RemoveListener(string listenerName)
...
-            return Task.WhenAll(listeners.Select(l => l.JobToBeExecuted(context, cancellationToken)));
+            return Task.WhenAll(listeners.Select(l =>
+            {
+                try
+                {
+                    return l.JobToBeExecuted(context, cancellationToken);
+                }
+                catch (Exception e)
+                {
+                    if (log.IsDebugEnabled())
```
lahma: these should probably be warnings so that the problem raises some attention

src/Quartz/Listener/BroadcastJobListener.cs
```diff
@@ -99,22 +103,66 @@ public bool RemoveListener(string listenerName)
...
+                    {
+                        log.Debug($"JobListener {l.Name} - JobToBeExecuted method raised an exception: {e.Message}");
+                    }
+                    return Task.FromResult(0);
```
lahma: There's TaskUtil.CompletedTask

src/Quartz/Impl/StdSchedulerFactory.cs
Sicos1977: No it isn't, I thought the same as you but it looks like that the expression body way generates a backing field with a null check. Just try it out and you will see that the expression body works different then the property with only a getter.

Sicos1977:
I was wondering if you are going todo anything with my pull request? Do I still need to change something? Denying it is also an option, you wont hurt my feelings.

lahma:
Sure, sorry to keep you waiting. I usually handle these when I have free time which usually coindices with weekends. I do appreciate that you take the time to fix and help with outstanding issues.

lahma:
Thanks for fixing these!

PR #599, opened 2018-02-28
fixes #598 tables_oracle.sql should use NUMBER(19) for long properties

### LINKED ISSUE #598: tables_oracle.sql should use NUMBER(19) for long properties
### Expected behavior

Class attributes of type `long` should be mapped to the Oracle database using `NUMBER(19)` columns (see [Java Type to SQL Type Mappings](https://docs.oracle.com/cd/E19501-01/819-3659/gcmaz/) for reference).

### Actual behavior

The `long1` and `long2` fields from `SimplePropertiesTriggerProperties` are mapped to `NUMBER(13)` columns, which may be too short.

danilopiazza:


lahma:
Thank you!

PR #631, opened 2018-05-30
Make JobInterrupted method virtual in class SchedulerListenerSupport

mario-d-s:
It seems `JobInterrupted` had the wrong signature in the `SchedulerListenerSupport` class, it was impossible to execute logic on job interruption.

This very simple commit marks the method `virtual` so inheritors could override it.

### DIFF
--- a/src/Quartz/Listener/SchedulerListenerSupport.cs
+++ b/src/Quartz/Listener/SchedulerListenerSupport.cs
@@ -97,7 +97,7 @@ namespace Quartz.Listener
             return TaskUtil.CompletedTask;
         }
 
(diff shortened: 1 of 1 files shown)


lahma:
Indeed, thanks for the contribution!

mario-d-s:
@lahma Thanks for merging so quickly 😃 

Can you estimate when 3.0.6 will be released?

mario-d-s:
Of course, I understand. Copying the class was what I was planning to do, but thanks for the suggestion.

If I could chip in with Quartz development, please let me know what the priorities are, I'd be happy to help.

PR #681, opened 2018-09-20
Added transient codes from EF into new JobStore.

puddlewitt:
I have been receiving the following error when using Azure PaaS SqlDatabase

`System.InvalidOperationException: Transaction not connected, or was disconnected
at Quartz.Impl.AdoJobStore.ConnectionAndTransactionHolder.CheckNotZombied()
at Quartz.Impl.AdoJobStore.ConnectionAndTransactionHolder.Rollback(Boolean transientError)`

It appears that after this event `trigger_state` is left as `BLOCKED` and recovery does not occur.

- Single Instance Azure SqlDatabase (no georep)
- Single Instance Quartz Scheduler (not clustered)
- Quartz scheduler hosted in a WebJob (single instance)

After chatting on Gitter we think the database instance is being taken down by Azure, but Quartz is not identifying the errors codes as transient. I have copied the transient errors codes from EF and merged those with the msdn documentation. I have created a new `SqlServerJobStoreTX` so that the previous implementation can be easily reverted in a deployment.


lahma:
I think these should be added directly to JobStoreTX as it original code handling came from the code used by the transient fault handling application block, I guess the codes have been refined since. Adding extra job store just for this is too much I feel.

Could you please just add them to the current job store?

lahma:

Issue #692, opened 2018-10-10
Introduce new config option `ClusterCheckinMisfireThreshold`


footcha:
### Expected behavior

Clustered check-in can be missed due to ```GC.Collect()``` taking extraordinary long time - the cluster instance is "frozen". When garbage collector finishes its work then a scheduler instance is "unfrozen" and continues its regular work. However, there is a chance that in the meantime another cluster instance can detect the "frozen" instance as failed and clears it. After garbage collecting period, when the first instance is unfreezed and it detects "This scheduler instance (xxx) is still active but was recovered by another instance in the cluster. This may cause inconsistent behavior."
As a consequence, jobs marked with ```DisallowConcurrentExecutionAttribute``` do run in parallel among cluster instances.

It cannot be simply fixed by a greater value in ```quartz.jobStore.clusterCheckinInterval```. I suggest a new configuration option ```quartz.jobStore.clusterCheckinMisfireThreshold```. The sum of clusterCheckinInterval and clusterCheckinMisfireThreshold then will be the "maxNotResponding" interval for a scheduler instance - if it exceedes then other scheduler instances can consider this scheduler instance as failed.

lahma:
Interesting analysis, thanks for digging into this. I would love to get a PR showing the concept and ideally keeping the old behavior as intact as reasonable (maybe opt-in new behavior?).

Another thing is what might be the root cause for such long pause - how long pause are we talking about? Is Quartz generating the the GC stress or your jobs?

footcha:
> Another thing is what might be the root cause for such long pause - how long pause are we talking about? Is Quartz generating the the GC stress or your jobs?

I believe that image https://pasteboard.co/HHLs93x.png will help to understand my proposal. If there is a delay in regular clustered chechin (due to garbage collector, network latency, etc.) then there should be additional protection period (clusterCheckinMisfireThreshold) that discourages other scheduler instances from detecting this scheduler instance as failed.
In practice we have raher long ```GC.Collect()``` period (can be 150sec in some cases). Quartz is not to blame - Quartz is just a victim of such an extraordinary behaviour.

lahma:
OK thanks for the clarification. The 150sec is a really long pause indeed. Thanks for bringing this into attention and also providing a solution. Will be part of the next release!

adansklevanskis:
Hi lahma,
Can you say when this change will be published as a new version of Nuget package? 
Thanks!

lahma:
[...]

PR #693, opened 2018-10-10
Introduced a config parameter `ClusterCheckinMisfireThreshold`.

footcha:
This PR solves issue #692.

### DIFF
--- a/src/Quartz/Impl/AdoJobStore/JobStoreSupport.cs
+++ b/src/Quartz/Impl/AdoJobStore/JobStoreSupport.cs
@@ -75,6 +75,7 @@ namespace Quartz.Impl.AdoJobStore
             RetryableActionErrorLogThreshold = 4;
             DoubleCheckLockMisfireHandler = true;
             ClusterCheckinInterval = TimeSpan.FromMilliseconds(7500);
+            ClusterCheckinMisfireThreshold = TimeSpan.FromMilliseconds(7500);
             MaxMisfiresToHandleAtATime = 20;
             DbRetryInterval = TimeSpan.FromSeconds(15);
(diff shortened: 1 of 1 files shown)


footcha:
Yes, it indeed is. Thank you for your super fast response. Can you say when this change will be published as a new Nuget package? We at our company are quite in trouble without this change. Creating a workaround would cause a dirty mess in our codebase. I would really appreciate if you were able to to publish it soon but I do not want to stress you anyway :)

footcha:
I fully understand. Thank you for your proposals - I will discuss then with the rest of a team.
Can you please estimate when there will be the next regular release?

lahma:
There's nothing set in stone. Usually fast when there's critical bug or "enough" fixes have accumulated. The 3.0.7 was released just last weekend (unfortunate for you) so I would think that there's no need for next release for some weeks or even months.

The time between releases might indicate code maturity or my laziness:

https://www.nuget.org/packages/Quartz

lahma:
Great to hear you have a solution on company side. I'll try to find time this evening to generate a solution for running numbers, the current versioning is bad as you've noted.

lahma:
@footcha can you try out the MyGet feed here:

https://www.myget.org/F/quartznet/api/v3/index.json

It should have master builds from now on, but as it's brand new can't make too much promises how it'll behave in short term if changes are needed.

footcha:
It is working nicely. Thank you! Just be careful with storing API key for MyGet in source control. The key should sit only in build server, not in source control since it is available to public. I know that MyGet contains preview version only in case of Quartz, but still...

lahma:
No worries, the key in appveyor.yml is actually encrypted as the real key is a guid. Great to hear it's working for you.

PR #708, opened 2018-11-27
fixes#411 - Added ability to detect deleted files and notify the listener

senthilkumarmohan:
Hi,

I have added the ability to detect the deleted files to dir scanner job. Added FilesDeleted method to the listener interface. I have implemented this by saving the current file list in the JobDataMap and comparing the list during the  next iteration. Please review and share your feedback - happy to implement any changes.

Thanks,
Senthil

### DIFF
--- a/src/Quartz.Jobs/Job/DirectoryScanJob.cs
+++ b/src/Quartz.Jobs/Job/DirectoryScanJob.cs
@@ -118,6 +118,8 @@ namespace Quartz.Job
                 {
                     model.DirectoryScanListener.FilesDeleted(deletedFiles);
                 }
+
+                //Update current file list
                 model.UpdateFileList(allFiles);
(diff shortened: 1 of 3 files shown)


senthilkumarmohan:
Hi @lahma, could you please share your feedback on this PR?

lahma:
Thanks for taking the time to prepare this change and sorry for the late reply. The change looks good, only thing is that it changes the API so probably cannot be included in a minor release.

senthilkumarmohan:
Okay. Thanks for your response - leave it with you to decide on the version this can be merged into.

lahma:
As this only affects Quartz.Jobs package it's lighter procedure in my book. I'll probably brand next release as 3.1 so it will show some things have changed. Thanks again!

PR #719, opened 2019-01-02
Update RAMJobStore.AcquireNextTriggers() to reduce allocations

drieseng:
* Update `RAMJobStore.AcquireNextTriggers()` to reduce allocations when the store has no triggers.
* Use return value of `SortedSet.Add()` to determine whether a trigger was already acquired for a job that does not allow concurrent execution.
* Use hand-rolled `FirstOrDefault()` extension method for **SortedSet** to eliminate type checks.

I only have a benchmark result for the last change (as `RAMJobStore.AcquireNextTriggers()` is hard to benchmark as it changes state):

|                    Method |     Mean |     Error |    StdDev | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
|-------------------------- |---------:|----------:|----------:|------------:|------------:|------------:|--------------------:|
|       FirstOrDefault_Linq | 216.3 ns | 0.5124 ns | 0.4793 ns |      0.1216 |           - |           - |               192 B |
| FirstOrDefault_HandRolled | 162.4 ns | 3.8830 ns | 4.7686 ns |      0.0913 |           - |           - |               144 B |

### DIFF
--- a/src/Quartz/Simpl/RAMJobStore.cs
+++ b/src/Quartz/Simpl/RAMJobStore.cs
@@ -23,7 +23,6 @@ using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Globalization;
-using System.Linq;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


lahma:
Thank you!

PR #720, opened 2019-01-02
Reduced usage of ConcurrentDictionary in RAMJobStore

### LINKED ISSUE #715: RAMJobStore: ConcurrentDictionary solely used within lock
Upon first look, it appears as if the **ConcurrentDictionary** instance fields in **RAMJobStore** are only used within a synchronized block. As such, there's no reason why **ConcurrentDictionary** should be preferred over a **Dictionary**.

Since **ConcurrentDictionary** incurs quite some overhead and we do not have to worry about concurrency, we could just use a regular **Dictionary** instead.

Would you accept such a PR, or do you plan to remove the instance wide lock?

drieseng:
* Reduced usage of **ConcurrentDictionary** in **RAMJobStore** since we always operate in a synchronized block.
* Use return value of `Dictionary.TryGetValue(TKey key, out TValue value)` and `List<T>.Remove(T value)`.

Fixes #715.

**Note:**
I could not change **triggersByKey** to a regular **Dictionary**, as there's no replacement for `ConcurrentDictionary.TryRemove(TKey key, out TValue value)`. The `Dictionary.Remove(TKey, out TValue value)` method is only available in **.NET Core** and **.NET Standard 2.1**.

### DIFF
--- a/src/Quartz/Simpl/RAMJobStore.cs
+++ b/src/Quartz/Simpl/RAMJobStore.cs
@@ -52,12 +52,12 @@ namespace Quartz.Simpl
     {
         private readonly object lockObject = new object();
 
-        private readonly ConcurrentDictionary<JobKey, JobWrapper> jobsByKey = new ConcurrentDictionary<JobKey, JobWrapper>();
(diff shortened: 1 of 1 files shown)


lahma:
Thanks again!

PR #725, opened 2019-01-04
Minor optimizations for TaskSchedulingThreadPool

drieseng:
In `TaskSchedulingThreadPool.RemoveTaskFromRunningList(Task completedTask)`, do not first check if the completed task is null and in the list of running tasks. The task should never be null and always be in the list of running tasks, and, even if this does happen, invoking `runningTasks.Remove(completedTask)` with a null task is not a problem.

In `TaskSchedulingThreadPool.Shutdown(bool)` use the actual number of threads that we wait on in the log message. This also ensures we always access **runningTasks** in a synchronized block.

Remove an extra allocation in `TaskSchedulingThreadPool.Shutdown(bool)`.


lahma:
Thank you!

PR #727, opened 2019-01-04
Remove unused functionality from QueuedTaskScheduler

drieseng:
Removes unused functionality from **QueuedTaskScheduler** hereby reducing complexity and improving performance.


lahma:
This was the original implementation from Microsoft folks but as you noted, maybe a little overkill for the usage. Thank you!

PR #731, opened 2019-02-04
GetTimeBefore() and GetFinalFireTime() throw NotImplementedException

PressXtoChris:
It seems strange that public functions that are just TODOs would return null instead of throwing an exception, as this can be confusing for users who would expect the functionality to exist. Is there a particular reason this has been left as null?


lahma:
Yes, makes more sense. Thank you for the contribution!

lahma:
This was reverted as in 3.1 caused too many surprises for people using the `ITrigger.FinalFireTimeUtc` interface. Proper way to handle is probably to actually implement the functionality.

PR #745, opened 2019-03-28
Raise exception in case of duplicated trigger in xml

siyamandayubi:
Currently, in case of multiple triggers with the same name in xml files, Quartz ignores the first ones
which misleading us why the first triggers didn't fire. In such case, the xml config
is invalid and qaurtz should raises exception, so that we will know what is wrong

### DIFF
--- a/src/Quartz/Xml/XMLSchedulingDataProcessor.cs
+++ b/src/Quartz/Xml/XMLSchedulingDataProcessor.cs
@@ -24,6 +24,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using System.Xml;
 using System.Xml.Schema;
+using System.Linq;
 using System.Xml.Serialization;
 
 using Quartz.Impl.Matchers;
@@ -31,6 +32,7 @@ using Quartz.Logging;
 using Quartz.Spi;
 using Quartz.Util;
 using Quartz.Xml.JobSchedulingData20;
+using Quartz.Impl.AdoJobStore;
 
 namespace Quartz.Xml
 {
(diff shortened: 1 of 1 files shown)


src/Quartz/Xml/XMLSchedulingDataProcessor.cs
```diff
@@ -405,6 +407,13 @@ protected virtual void ProcessInternal(string xml)
                 }
             }
 
+            // check for triggers with the same name
+            var duplicatedTriggers = triggerEntries.GroupBy(c => c.Item.name).Where(c => c.Count() > 1);
```
lahma: I think the grouping should be done using name and group? Trigger's identity is based on both name and group.

lahma (changes requested):

lahma:
@siyamandayubi — thank you for this, and I am sorry it sat so long. The bug is still real on `main`.

Two triggers sharing a name and group in one file both land in `loadedTriggers`. `ScheduleJobs`
schedules the first, and when it reaches the second, `GetTrigger` returns the one it just wrote
(`XmlSchedulingDataProcessor.cs:769`). `OverwriteExistingData` defaults to true (line 93), so it
reschedules over it at Debug level. The `ReportDuplicateTrigger` warning at line 892 fires only when
the duplicate names a *different* job, so the same-job case is silent. Last definition wins. The JSON
loader inherits the same `ScheduleJobs`, so it has it too.

We are taking this, with credit to you. Two changes from your patch: the check keys on name **and**
group, as I asked back in 2019 — a trigger's identity is both — and it goes in
`AddTriggerToSchedule`, the one funnel both loaders pass through, throwing
`SchedulingDataValidationException`. `ignore-duplicates` and `overwrite-existing-data` are about the
file versus the scheduler; neither can express "declared twice in one document", so that stays an
error regardless of the directives.

Closing this in favour of the reworked change, tracked as #3416; you will be on it.

lahma:
Landed on `main` as #3464 (`81463df41`), with you as co-author on the commit. A trigger or job key declared twice in one scheduling-data file now throws `SchedulingDataValidationException` naming the key, on both the XML and the JSON loaders, and the directives do not suppress it. It ships in 4.0.0-alpha.3. Thank you again.

siyamand-ayubi:
@lahma Thanks for the update! Although I’m no longer working on the related project, it’s still great to see the fix merged. Much appreciated! 🙂

PR #746, opened 2019-03-28
remove assemblyinfo refs from projects

siyamandayubi:


lahma:
The reference is required as it contains dynamic information about build version etc. The file is generated by build.cmd for example.

Issue #758, opened 2019-05-17
Handle misfired triggers using batch commands


nicsur:
Misfired triggers are currently handled one-by-one. This generates a lot of database traffic when there is a large number of triggers. In turn, the recovery time is long and the misfire handler cannot handle a larger number of triggers during a run due to lock contentions.

This improvement will change the misfire handling to use batch database requests (for databases that allow batches). Using batches significantly reduces the roundtrips to the database allowing for faster recovery and larger number of triggers in one misfire handler run.

PR #759, opened 2019-05-17
Handling misfires using batch commands

nicsur:
Handling of misfired triggers using database batch commands. The goal is to minimize database round-trips to reduce handling time when there is a large number of misfires.
 - New boolean setting: quartz.jobStore.batchMisfireHandling to control feature.
 - When batchMisfireHandling is set, the misfire handler will process misfired triggers in batches of size maxMisfiresToHandleAtATime
- To support batch processing of triggers, multi-object methods have been added. They mirror the exiting single-object methods.
- The single-object methods are kept as they are for backward-compatibility.
- Not all supported stores can handle batches and this is controlled in code by IDriverDelegate.SupportsBatching override.
- Unit-tests and tests have been added to cover the changes.

#758

### DIFF
--- a/src/Quartz.Tests.Integration/MisfireBatchRecoveryTest.cs
+++ b/src/Quartz.Tests.Integration/MisfireBatchRecoveryTest.cs
@@ -1,0 +1,382 @@
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Collections.Specialized;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using System.Threading;
+using System.Threading.Tasks;
+using NUnit.Framework;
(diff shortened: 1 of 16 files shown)


lahma:
Sorry for keeping you hanging. Could you please rebase against master and I hopefully have soon time to go through this idea. Thanks for understanding.

lahma:
Thanks for this, and sorry it sat for so long.

Misfire batching has now landed on `main`: the write path in #2994, and the batched read plus `DbBatch` write in #3149. Between them they cover what this PR set out to do, so I'm closing it here.

The implementation ended up differing enough from this diff (the SQL, the delegate shape and the misfire path have all changed since 2019) that it was written fresh rather than rebased, but the idea and the case for it were yours — credited in #3149.

PR #765, opened 2019-06-05
Update TimeZoneUtil.cs

a407121393:
i think the "Asia/Beijing" is not exist


lahma:
Thanks for the fix! I'd recommend using the [TimeZoneConverter plugin](https://www.nuget.org/packages/Quartz.Plugins.TimeZoneConverter/) as it has far better compatibility, also between Windows/Linux.

PR #788, opened 2019-08-20
Add line about property injecting

matgrioni:
Small extension on what is currently in documentation about property injecting, to describe issue related to using property injection and a custom job factory.


lahma:
Thank you for taking the time to improve the documentation! It's always the weak link in OSS projects so this is highly appreciated 👍

PR #790, opened 2019-09-05
Feature/every nth week

nyakimov:
Ability to create cron expressions that execute every nth week on a specific weekday/s


nyakimov:
Please review and let me know what you think about the proposed addition.

lahma:
Looks good, thanks for taking the time to create this nice feature!

nyakimov:
I'm glad you like it. Thanks for merging it in!

Issue #817, opened 2020-01-15
Sql Server runs with very high CPU and generates massive numbers of query plans


IDDesigns:
We are running:

Sql Server 2017 with a single database containing only contents created by the Quartz 3.0.7 sql server script and Quartz 3.0.7.

Running against this database we have about 500 load balanced pairs (so around 250 unique sched_names) running from other servers.

The SQL server is constantly running with very high CPU and the query plan cache is regularly flushed due to so many plans being created, all of which is generated by Quartz traffic.
The load on the scheduler servers running the jobs is what I would expect from the jobs we are running.

Upon further investigation it would appear that the load on the sql server is, in part, due to the fact that the sched_name is hardcoded into the queries so query plans cannot be shared across multiple schedulers.

To help verify that this will help things I have forked the source and created the following branch: https://github.com/IDDesigns/quartznet/tree/features/parameteriseSchedName
which (hopefully) turns all the usages of sched_name into parameters. 
(I'm currently in the fairly early stages of testing this)

Is there any reason why this change wouldn't work?
If not, is this change something that would be considered for inclusion?

Many thanks

Ian

lahma:
This sounds like exactly what is needed. I have some preliminary work related to fixing locking code using regular varchars that causes index scans, but it shouldn't collide with this work.

A pull request would be very welcome 👍

lahma:
Thank you very much, great work!

PR #818, opened 2020-01-15
SchedName added to queries as sql paramter

IDDesigns:
Running multiple schedulers against a single database results in high cpu due to multiple query plans being created,
by adding the scheduler name as a sql parameter, sql server is able to reuse the plans across multiple schedulers.

The SchedulerNameLiteral property available on a number of classes is no longer required by the core Quartz code, but has been left to maintain backward compatibility.

#817

### DIFF
--- a/src/Quartz/Impl/AdoJobStore/AdoJobStoreUtil.cs
+++ b/src/Quartz/Impl/AdoJobStore/AdoJobStoreUtil.cs
@@ -17,6 +17,7 @@
  */
 #endregion
 
+using System;
 using System.Globalization;
 
 namespace Quartz.Impl.AdoJobStore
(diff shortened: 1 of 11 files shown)


src/Quartz/Impl/AdoJobStore/AdoJobStoreUtil.cs
```diff
@@ -28,17 +29,30 @@ namespace Quartz.Impl.AdoJobStore
...
+        /// <summary>
+        /// Replace the table prefix in a query by replacing any occurrences of
+        /// "{0}" with the table prefix.
+        /// </summary>
+        /// <param name="query">The unsubstituted query</param>
+        /// <param name="tablePrefix">The table prefix</param>
+        /// <param name="schedNameLiteral">the scheduler name - no longer required, scheduler name is now a sql parameter</param>
+        /// <returns>The query, with proper table prefix substituted</returns>
+        [Obsolete("Use overload with schedNameLiteral, scheduler name is now a sql parameter")]
+        public static string ReplaceTablePrefix(string query, string tablePrefix, string schedNameLiteral)
```
IDDesigns: I opted to maintain backward compatibility for any public or protected methods - they should all be marked as obsolete.

src/Quartz/Impl/AdoJobStore/SimplePropertiesTriggerPersistenceDelegateSupport.cs
```diff
@@ -110,17 +113,21 @@ public void Initialize(string tablePrefix, string schedName, IDbAccessor dbAcces
 
         protected string TablePrefix { get; private set; }
 
+        [Obsolete("Scheduler name is now added to queries as a parameter")]
         protected string SchedNameLiteral { get; private set; }
 
+        protected string SchedName { get; private set; }
```
IDDesigns: Anywhere the SchedNameLiteral property existed, I've added a SchedName so that anyone deriving from the class is able to get the property easily.

src/Quartz/Impl/AdoJobStore/StdAdoConstants.cs
```diff
@@ -33,317 +35,318 @@ public class StdAdoConstants : AdoConstants
         public const string TablePrefixSubst = "{0}";
 
         // table prefix substitution string
+        [Obsolete("Scheduler name injection has been replaced with sql parameter @schedulerName")]
         public const string SchedulerNameSubst = "{1}";
```
IDDesigns: I've replaced anywhere this was originally used with @schedulerName, and left this for an external uses. 
I'm hoping this will avoid breaking changes to external code.

PR #824, opened 2020-01-30
Remove indexes covering duplicate columns

lahma:
I'd love to have second eyes on this if either of you @IDDesigns or @jvavasseur (or both!) have the time and energy. I think you both have played your fare share with SQL Server and probably have the environment to test performance too.

When I checked some indexes I found out that the SQL Server ones don't make much sense when group is always and sometimes also the name (group+name) are part of the query. Having columns indexes the both ways just add works to the server and gives no benefit.

Thoughts?


IDDesigns:
I'll take a look - I should also be able to pull some stats about which indexes are used in case any of the are not.

jvavasseur:
Hello 


I had a quick look at the SQL Server scripts. The fist thing I have to say is that every Clustered Index are not ideal. 

They are not :
Narrow = small data type would be good. However they are all using composites key with NVARCHAR(120+). This lead to large key, i.e. the pointer used in non clustered indexed. It uses more space on both disk and memory cache, less data are stored and retrieved by queries.
Unique = They don’t seem to be unique. This will lead to a Uniquifier (4 bytes) id added to keys and more storage and memory cache being used
Static = they seems to be static (they don’t change)
Ever Increasing = They aren’t ever increasing. This means that new Clustered keys are added in between old one. This lead to fragmentation, data page creation and page movement
Fixed sise = They all use datatype with variable length (NVARCHAR)

The size of the clustered index key could partially explain why non clustered index may be poorly used and perform poorly: 
They are all between 160-420 characters long => 160*2 to 420*2 bytes + 4 bytes uiniquifier
This is really close to the maximum key length (900 bytes) and around 5-10% of a whole data page (8kb) 
This pointer is replicated in all non clustered index rows at the leaf level
The SQL Server Engine may think that reading and searching these huge index key is not the best option.

Many table use the same PK and Clustered index with [SCHED_NAME], [TRIGGER_NAME], [TRIGGER_GROUP]. 
This could be centralised in such a table : CREATE Table X (Id int or bigint PK Clustered, [SCHED_NAME] NVARCHAR(120), [TRIGGER_NAME] nvarchar(150), [TRIGGER_GROUP] nvarchar(150)
And then This Id could be referenced and indexed in most tables which are currently using theses 3 columns as PK. 

This is just theory and I haven’t look at real life index usage yet. I will try to do it this week using our product but this will be really dependent on our own product and its usage.

I haven’t mention it before when we first chatted because it may not be easy to change and update the whole schema without breaking changes. 
One way to overcome (and hide) this could be using views with structures and name similar to the actual tables and add new underneath tables with new names and a better structure.

Regards
Julien

lahma:
Thank you @jvavasseur for your insights! I appreciate you taking the time to investigate and bring forth the issues. The composite key is from day 0 and aligns with the Java version, so more than 10 years of legacy 🙂 

It might also be hard to create special views as that might vary from database to database - there are quite many supported at the moment. I'd see it as a possible improvement for 4.x or another bigger release to migrate the database to use numeric or GUID identifiers. That is just a little undertaking...

I'm hoping to see if we have any quick wins from tweaking the index setup for the current (non-optimal) table structure.

IDDesigns:
From what I can see a large amount of time spent when inserting or updating is for updating the indexes, and then the select queries all appear to use the primary key index.

I'm seeing what happens if I disable the indexes on some of the tables (expect the primary one), it may be it's possible to not bother with most of the indexes.

jvavasseur:
Well this is probably a side effect of what I said yesterday. Each update to an index must (re)write the 400-800 bytes key + indexed data.

jvavasseur:
Note that numeric and GUID (unique identifier), while being quite short (narrow) are not recommended because they are random and not ever increasing 
There are ways with views because they can be updated/have update and can involve triggers as well. I am only talking about SQL Server though.

IDDesigns:
I've disable all bar the primary indexes and it looks like both the CPU and IO have dropped, and the queries are running a little faster. I'll monitor for a couple of days to makes sure.

Using a lookup table sounds like it should work - you could use a GUID in the way that nHibernate uses it so that it is sortable and increases : https://nhibernate.info/blog/2009/05/21/using-the-guid-comb-identifier-strategy.html

jvavasseur:
With SQL Server, Int (4b) or Bigint (8b) are usually used. A Identity column can be added with Id Int Identity(0, 1) in order to avoid Id generation and overlap.

IDDesigns:
I've been running with all the indexes disabled (bar the primary clustered ones) for a couple of days and there is a visible drop in CPU usage - not massive, but significant enough.
So removing all the indexes is a good improvement over the existing release.

lahma:
@IDDesigns thank you for profiling and providing the information. Kind of makes me sad that the non-clustered indexes OOTB don't give any value... 

I wonder whether having the https://github.com/quartznet/quartznet/pull/818 makes any difference? The latest release can still cause index scans because the value is passed as varchar instead of nvarchar, the PR fixes that to utilize the indexes better. But if query optimizer always select clustered indexes anyway, I guess the other indexes don't kick in.

shatl:
> Note that numeric and GUID (unique identifier), while being quite short (narrow) are not recommended because they are random and not ever increasing There are ways with views because they can be updated/have update and can involve triggers as well. I am only talking about SQL Server though.
May be too late, but there is an excellent package https://github.com/phatboyg/NewId to generate sequential ids.

database/tables/tables_sqlServer.sql
```diff
@@ -346,23 +346,43 @@ ALTER TABLE [dbo].[QRTZ_TRIGGERS] ADD
...
+DROP INDEX IF EXISTS [IDX_QRTZ_FT_G_T] ON [dbo].[QRTZ_FIRED_TRIGGERS];
+GO
+
+
+CREATE INDEX [IDX_QRTZ_T_G_J]                 ON [dbo].[QRTZ_TRIGGERS](SCHED_NAME, JOB_GROUP, JOB_NAME);
+CREATE INDEX [IDX_QRTZ_T_C]                   ON [dbo].[QRTZ_TRIGGERS](SCHED_NAME, CALENDAR_NAME);
+
+CREATE INDEX [IDX_QRTZ_T_N_G_STATE]           ON [dbo].[QRTZ_TRIGGERS](SCHED_NAME, TRIGGER_GROUP, TRIGGER_STATE);
+CREATE INDEX [IDX_QRTZ_T_STATE]               ON [dbo].[QRTZ_TRIGGERS](SCHED_NAME, TRIGGER_STATE);
+CREATE INDEX [IDX_QRTZ_T_N_STATE]             ON [dbo].[QRTZ_TRIGGERS](SCHED_NAME, TRIGGER_NAME, TRIGGER_GROUP, TRIGGER_STATE);
```
shatl, line 379: If we define IDX_QRTZ_T_N_STATE as  `CREATE INDEX [IDX_QRTZ_T_N_STATE] ON [dbo].[QRTZ_TRIGGERS](SCHED_NAME, TRIGGER_STATE, TRIGGER_GROUP, TRIGGER_NAME);`

it will cover `IDX_QRTZ_T_N_G_STATE`, `IDX_QRTZ_T_STATE` as well

database/tables/tables_sqlServer.sql
```diff
@@ -346,23 +346,43 @@ ALTER TABLE [dbo].[QRTZ_TRIGGERS] ADD
...
+CREATE INDEX [IDX_QRTZ_T_NEXT_FIRE_TIME]      ON [dbo].[QRTZ_TRIGGERS](SCHED_NAME, NEXT_FIRE_TIME);
+CREATE INDEX [IDX_QRTZ_T_NFT_ST]              ON [dbo].[QRTZ_TRIGGERS](SCHED_NAME, TRIGGER_STATE, NEXT_FIRE_TIME);
+CREATE INDEX [IDX_QRTZ_T_NFT_ST_MISFIRE]      ON [dbo].[QRTZ_TRIGGERS](SCHED_NAME, MISFIRE_INSTR, NEXT_FIRE_TIME, TRIGGER_STATE);
+CREATE INDEX [IDX_QRTZ_T_NFT_ST_MISFIRE_GRP]  ON [dbo].[QRTZ_TRIGGERS](SCHED_NAME, MISFIRE_INSTR, NEXT_FIRE_TIME, TRIGGER_GROUP, TRIGGER_STATE);
```
shatl, line 383: If we define IDX_QRTZ_T_NFT_ST_MISFIRE_GRP as `CREATE INDEX [IDX_QRTZ_T_NFT_ST_MISFIRE_GRP]  ON [dbo].[QRTZ_TRIGGERS](SCHED_NAME, NEXT_FIRE_TIME, TRIGGER_STATE, MISFIRE_INSTR, TRIGGER_GROUP);`

it will cover `IDX_QRTZ_T_NEXT_FIRE_TIME`, `IDX_QRTZ_T_NFT_ST` and `IDX_QRTZ_T_NFT_ST_MISFIRE`

P.S. I haven't done any performance measurements on it, but it looks like number of indexes can be reduced even further

database/tables/tables_sqlServer.sql
lahma: Thanks for the input. I'm a bit hesitant to do this before it's proven efficient on the field. I'm sure reiterate this a lit later.

Issue #768, opened 2019-06-18
Support mysql-net/MySqlConnector


ghd258:
Support mysql-net/MySqlConnector
https://github.com/mysql-net/MySqlConnector

ghd258:
@lahma   doing？

lahma:
A PR would be appreciated, otherwise this will wait until I have some time spare.

RavindraV07:
I have MySQL 5.7.25 .and using Dot net core, Please let me know which connector I need to install from Nuget Manager?  

I am getting this error

"Could not load file or assembly 'MySql.Data, Culture=neutral, PublicKeyToken=null'. The system cannot find the file specified."

bgrainger:
The advantage of making this change is that `StdAdoDelegate` [uses `ExecuteNonQueryAsync`](https://github.com/quartznet/quartznet/blob/287f9865296a7f7f7db522d54e5e716d953d25fd/src/Quartz/Impl/AdoJobStore/StdAdoDelegate.cs#L201), `ExecuteReaderAsync`, etc.

It's a long-standing bug in MySql.Data ([bug 70111](https://bugs.mysql.com/bug.php?id=70111)) that async I/O is not implemented correctly; thus this method will actually run synchronously. (See also [here](https://stackoverflow.com/a/42305108/) and [here](https://stackoverflow.com/questions/30551459/mysql-c-sharp-async-methods-doesnt-work) on Stack Overflow.)

https://github.com/mysql-net/MySqlConnector is an OSS replacement for MySql.Data that supports true asynchronous I/O.

lahma:
@bgrainger kindly provided PR this, will be part of 3.1.

PR #825, opened 2020-02-03
Switch MySQL library to MySqlConnector

bgrainger:
#768

I wasn't sure if the existing library should be changed in-place (to give users good async behaviour by default) or if a new named provider should be added instead (less of a breaking change?).


lahma:
Thank you for working on this! I think we should keep the old version on the side if we want to get this to 3.x release train, otherwise would be quite breaking change as runtime type loading would fail. Can you make this a new dbprovider with the name MySqlConnector?

bgrainger:
I added `MySqlConnector` as a new provider (leaving `MySql` unchanged), and switched the integration tests to use it.

lahma:
Thank you for this 👍 For a release like 4.x we could consider dropping built-in support for old version if needed.

Issue #831, opened 2020-03-08
SchedulerStarted before StartPlugins leads to System.TimeZoneNotFoundException with 


klinki:
### Version: 3.0.7
 

OS: Windows 10 (I use Windows for development, Linux for production usage).

### Expected behavior
`Quartz.Plugins.TimeZoneConverter` should be used when installed and should be respected in `SchedulerStarted` method.  `System.TimeZoneNotFoundException` should not be thrown when starting scheduler.

### Actual behavior
`Quartz.Plugins.TimeZoneConverter` is not properly initialized so it is not used and `SchedulerStarted` method throws `System.TimeZoneNotFoundException`.

I believe in `QuartzScheduler.Start`
```
await StartPlugins(cancellationToken).ConfigureAwait(false);
```
should be run before 
```
await resources.JobStore.SchedulerStarted(cancellationToken).ConfigureAwait(false);
```
to allow initialization of Quartz.Plugins.TimeZoneConverter.

### Steps to reproduce

Scheduler configuration:
```json
  "quartz": {
    "quartz.serializer.type": "json",
    "quartz.threadPool.threadCount": "2",
    "quartz.jobStore.type": "Quartz.Impl.AdoJobStore.JobStoreTX, Quartz",
    "quartz.jobStore.useProperties": "true",
    "quartz.jobStore.tablePrefix": "__QRTZ_",
    "quartz.jobStore.dataSource": "mysql",
    "quartz.dataSource.mysql.provider": "MySql",
    "quartz.dataSource.mysql.connectionString": "....",
    "quartz.plugin.timezoneConverter.type": "Quartz.Plugin.TimeZoneConverter.TimeZoneConverterPlugin, Quartz.Plugins.TimeZoneConverter"
  },
```

```csharp
// scheduler and job configuration, SystemTime prefereably set to fixed point
```

Full exception:
```
2020-03-08 10:55:00.060 +01:00 [ERR] Failure occurred during job recovery: Couldn't retrieve trigger: Could not find time zone with id Europe/Budapest, consider using Quartz.Plugins.TimeZoneConverter for resolving more time zones ids
2020-03-08 10:55:00.060 +01:00 [ERR] Error while processing job
Quartz.JobPersistenceException: Couldn't retrieve trigger: Could not find time zone with id Europe/Budapest, consider using Quartz.Plugins.TimeZoneConverter for resolving more time zones ids
[...]

PR #832, opened 2020-03-10
Set TimeZoneUtil.CustomResolver in TimeZoneConverterPlugin.Initialize method

klinki:
This fixes issue #831 with System.TimeZoneNotFoundException thrown even when Quartz.Plugins.TimeZoneConverter is used.


lahma:
I did a bit more analysis and switching the order might affect how XmlSchedulingPlugin sees things. It would be initialized before job store is initialized which currently might not be a problem but there could potentially be plugins that expect job store to be ready.

I was thinking whether it would be better here to to use TimeZoneConverterPlugin.Initialize the time zone handling and forget the current start/stop that would allow returning to older version. It's hard to come up with a use case where one would want to stop the time zone handling as part of scheduler stopping. What do you think?

klinki:
That sounds good to me. You are right, there is very low probability of someone needing to restore old timezone handling after stopping scheduler.

klinki:
I updated the PR and description.

lahma:
Thank you for helping out! 👍

PR #835, opened 2020-03-30
Use TEXT type instead of fixed-length VARCHAR

mausch:
for Postgres tables.

It's quite annoying to get errors about exceeding the limit when e.g. some job class name is very long.
I don't see any reason not to represent them as unlimited TEXT... this data type as been in Postgres since very old versions (7.1 from 2001 has it!) and [Postgres docs say "There is no performance difference"](https://www.postgresql.org/docs/10/datatype-character.html)

### DIFF
--- a/database/tables/tables_postgres.sql
+++ b/database/tables/tables_postgres.sql
@@ -13,11 +13,11 @@ DROP TABLE IF EXISTS qrtz_calendars;
 
 CREATE TABLE qrtz_job_details
   (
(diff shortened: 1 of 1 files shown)


lahma:
Thank you for taking the time to do a PR, it indeed seems that there's no reason to have such limit.

mausch:
Thank _you_ for doing an amazing job building and maintaining Quartz.NET for so many years!!

Issue #837, opened 2020-03-31
Quartz.Serialization.Json doesn't handle Guid deserialization


olicooper:
Jobs persisted in a Mysql/SqlServer data store using the `Quartz.Serialization.Json` plugin store `System.Guid` differently than when using the RamJobStore. This means we have to parse the Guid from the string.

### Expected behavior

Persisted object is deserialized back to a System.Guid

### Actual behavior

Guid is deserialized to a System.String

##### Setup the test data
```csharp
public class SimpleTestClass
{
    public Guid Guid { get; set; }
    public SimpleTestClass()
    {
        Guid = Guid.NewGuid();
    }
}

var data = new JobDataMap();
data.Add("guid", Guid.NewGuid());
data.Add("class", new SimpleTestClass());
data.Add("int", 2345);

TriggerBuilder.Create().UsingJobData(data).WithIdentity("TEST").StartNow().Build();
```
##### Log the results when executing the job
```csharp
Logger.LogInformation($"typeof 'guid' == {context.Trigger.JobDataMap.Get("guid").GetType().Name}"); // outputs: typeof 'guid' == String
Logger.LogInformation($"typeof 'class' == {context.Trigger.JobDataMap.Get("class").GetType().Name}"); // outputs: typeof 'class' == SimpleTestClass
Logger.LogInformation($"typeof 'int' == {context.Trigger.JobDataMap.Get("int").GetType().Name}"); // outputs: typeof 'int' == Int64
Logger.LogInformation($"typeof 'Guid' in 'class' == {(context.Trigger.JobDataMap.Get("class") as Testing).Guid.GetType().Name}"); // outputs: typeof 'Guid' in 'class' == Guid
```

PR #838, opened 2020-03-31
Add handling of Guid's in the JobDataMap

olicooper:
Resolves: https://github.com/quartznet/quartznet/issues/837

This will need testing prior to release.

### DIFF
--- a/src/Quartz/Util/StringKeyDirtyFlagMap.cs
+++ b/src/Quartz/Util/StringKeyDirtyFlagMap.cs
@@ -204,15 +204,6 @@ namespace Quartz.Util
         }
 
         /// <summary>
-        /// Adds the given <see cref="Guid" /> value to the <see cref="IJob" />'s
-        /// data map.
-        /// </summary>
-        public virtual void Put(string key, Guid? value)
-        {
-            base.Put(key, value);
-        }
-
-        /// <summary>
         /// Retrieve the identified <see cref="int" /> value from the <see cref="JobDataMap" />.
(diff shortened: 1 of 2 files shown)


olicooper:
I'm unsure if the nullable Guid should be included as it doesn't quite fit with the other methods you have. Feel free to remove if it doesn't make sense.

src/Quartz/Util/StringKeyDirtyFlagMap.cs
```diff
@@ -347,5 +356,39 @@ public virtual TimeSpan GetTimeSpan(string key)
...
+        /// </summary>
+        public virtual Guid GetGuid(string key)
+        {
+            object obj = this[key];
+
+            try
+            {
+                return (Guid)obj;
+            }
+            catch (Exception)
```
9Rune5, line 371: This try..catch adds 7 lines of code just to change "System.InvalidCastException: Specified cast is not valid." into "System.InvalidCastException: Identified object is not a Guid.".

That seems a bit overkill.

olicooper: Thanks for pointing this out, I have removed the try/catch blocks so it throws the original exception.

olicooper:
I have compiled a version of this to use in a live project and I can confirm that using `JobDataMap.GetGuidValue("myGuid")` works as expected when using the JobStoreTX (MySqlConnector) provider.

lahma:
Thank you for this too, great work! 👍

PR #843, opened 2020-04-17
Fix Deserialization of DateTimeOffset when using JsonObjectSerializer

### LINKED ISSUE #834: DateTimeOffset stored as ISO 8601 string in JobDataMap is discarded when executing the job
### Version: 3.0.7
 
- dotnetcore 3.1.200

### Expected behavior

When storing a `DateTimeOffset` (ISO 8601) string in the `JobDataMap` and having it serialised into a ADO Store. I expect to be able to retrieve the `DateTimeOffset` from the job context when the job executes.

The following config is set (as well as the ADO store details):
- `quartz.jobStore.useProperties = true`
- `quartz.serializer.type = json` 

### Actual behavior

I am unable to get the `DateTimeOffset` (ISO 8601) string from the `MergedJobDataMap` associated with the job execution context. I need to prefix the string value with something else to be able to retrieve it. It is stored in the database, but when pulling it into the executing job it's not there.

My shot in the dark guess is that the deserialiser is picking up that the string is a `DateTimeOffset` 
 and is thus deserialising it into that type whereas JobDataMap is expecting all values to be strings (due to the `useProperties` config) and quietly discards anything which is not.

### Steps to reproduce

1. Set the `quartz.jobStore.useProperties = true` value in config
2. Set the `quartz.serializer.type = json` value in config
3. Create a new job/trigger with at least one `DateTimeOffset` stored as a string using the ISO 8601 format.
```
			.WithIdentity("Example", group)				
                         .UsingJobData("executed_at", DateTimeOffset.Now.ToString("O"))
                         .Build();

var jobTrigger = TriggerBuilder.Create()
				.WithIdentity("ExampleTrigger", group)
				.StartNow()
				.Build();
```
4. Now schedule the job. When it gets executed the `IJobExecutionContext` will not have the `executed_at` property defined in the `MergedJobDataMap`.

If I store the DateTimeOffset like this, then I can retrieve it as a string and parse it myself:
```
 .UsingJobData("executed_at", $"{nameof(DateTimeOffset)}:{DateTimeOffset.Now:O}")
```

### LINKED ISSUE #840: DateTimeOffset is incorrectly retrieved as DateTime from database
### Expected behavior

`JobDataMap.GetDateTimeOffsetValue("MyDateTimeOffset")` should return DateTimeOffset.

### Actual behavior

When calling `JobDataMap.GetDateTimeOffsetValue("MyDateTimeOffset")` I get an 'invalid cast from DateTime' exception.

### Steps to reproduce

1. Create a project that uses a persisted job store (I'm using MySql)
1. Create a job that contains a DateTimeOffset in the JobDataMap
1. When the job executes, try to extract the DateTimeOffset with `JobDataMap.GetDateTimeOffsetValue("MyDateTimeOffset")`

```csharp
var jobData = new JobDataMap();
jobData.Put("MyDateTimeOffset", DateTimeOffset.Now());
```

olicooper:
fixes #840
fixes #834

This fix instructs Newtonsoft to Serialize and Deserialize date objects as a DateTimeOffset. Because a DateTimeOffset can be cast in to a DateTime object, this fix shouldn't effect any DateTime serialization.

Here is an example that demonstrates the original issue: https://dotnetfiddle.net/5Q1gUZ

### DIFF
--- a/src/Quartz.Serialization.Json/Simpl/JsonObjectSerializer.cs
+++ b/src/Quartz.Serialization.Json/Simpl/JsonObjectSerializer.cs
@@ -46,7 +46,8 @@ namespace Quartz.Simpl
                 {
                     IgnoreSerializableInterface = true
                 },
-                NullValueHandling = NullValueHandling.Ignore
+                NullValueHandling = NullValueHandling.Ignore,
+                DateParseHandling = DateParseHandling.DateTimeOffset
             };
         }
 


lahma:
Awesome! Thank you for studying this and creating the fix.

PR #862, opened 2020-06-19
eliminate the chance that the factory code will run multiple times

Coder3333:
I changed dbMetadataLookup from ConcurrentDictionary<string, DbMetadata> to ConcurrentDictionary<string, Lazy<DbMetadata>> and changed GetDbMetadata to use GetOrAdd.  What this does is prevent dbMetadataFactory.GetDbMetadata from getting called multiple times if the same provider is requested at the same time on multiple threads at initialization.  It won't prevent any exceptions from the previous version of code, but it will eliminate the possibility that potentially slow code might execute more than once.


lahma:
Thanks for this version but I think the previous version is "good enough". This adds quite a lot complexity and the metadata building isn't that expensive operation, I think thread-safety is the main goal.

PR #879, opened 2020-07-13
Reuse lock requestor id for subsequent calls in ExecuteInNonManagedTXLock

### LINKED ISSUE #794: Deadlock when db not available.
Link to code where deadlock happened

https://github.com/quartznet/quartznet/blob/6025454ba072bd8e178ed0e96c44fcf3c4acb509/src/Quartz/Impl/AdoJobStore/JobStoreSupport.cs#L3806

lahma:
fixes #794

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -9,8 +9,12 @@
 * NEW FEATURES
 
 	* DI configuration now supports adding schdeuler, job and trigger listeners (#877)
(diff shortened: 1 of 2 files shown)


pharris8:
Everyone, thank you so much for fixing this.  It has been causing our implementation to die every few weeks.

PR #898, opened 2020-07-20
Update timezoneconverter-integration.md

Aenima4six2:
Update docs to use correct plugin namespace.

Using the current namespace from the documentation yields the following startup exception.
```
Using custom data access locking (synchronization): Quartz.Impl.AdoJobStore.UpdateLockRowSemaphore
Unhandled exception. Quartz.SchedulerException: SchedulerPlugin of type 'Quartz.Plugin.TimeZoneConverter, Quartz.Plugins.TimeZoneConverter' could not be instantiated.
```


lahma:
Thank you! Quick turnaround as the documentation was some 6 hours old 😅 Unfortunate copy-paste error.

Issue #911, opened 2020-08-02
Introduce separate Quartz.Extensions.Hosting


lahma:
See discussion https://github.com/quartznet/quartznet/discussions/910 , it makes sense to have separate package to allow "pure-hosting" without ASP.NET Core dependency. This new package will be dependency for current AspNetCore package.

lahma:
I've added info to change log and also [a teaser documentation](https://www.quartz-scheduler.net/documentation/quartz-3.x/packages/hosted-services-integration.html#installation) for the unpublished 3.2 release, also tweaked build file to produce the package as part of CI flow.

PR #913, opened 2020-08-02
Add Quartz.Extensions.Hosting project

andrewlock:
As discussed in https://github.com/quartznet/quartznet/discussions/910:

- Introduces new _Quartz.Extensions.Hosting_ project, that only depends on _Microsoft.Extensions.Hosting.Abstractions_ (and _Quartz_ and _Quartz.Extensions.DependencyInjection_)
- Re-platforms _Quartz.AspNetCore_ on top of _Quartz.Extensions.Hosting_ project. 
- Adds an example worker service project
 
Obviously open to discussion/suggestions/changes 🙂 For example, I had to split the `QuartzHealthCheck` in two due to the way the project split, but there are other ways we could achieve a similar result.

#911


src/Quartz.Extensions.Hosting/IQuartzHostedServiceListener.cs
```diff
@@ -0,0 +1,8 @@
+namespace Quartz
+{
+    public interface IQuartzHostedServiceListener
```
lahma, line 3: I think this interface might get troublesome as the error count is kind of ambiguous, but I don't have better ideas at the moment from the top of my head...

lahma:
This is great work! Thanks for taking the extra kilometer and also building example project, we can't have too many samples 👍 

Only one comment about the public interface which might get problematic / require re-thinking at some point, but I don't think it's in this PR's scope. 

I'm willing to merge as-is unless you have anything to add?

andrewlock:
Thanks @lahma, yeah that interface bothered me too, but in my initial testing it seemed necessary. After a bit more testing I've realised it's not required, and we can just rely on the existing `ISchedulerListener` support for the health check. That makes everything cleaner and removes the need for the public interface entirely 🙂

lahma:
Nice, looks clean now. A bit thank you once again for helping out!

andrewlock:
No problem at all, thanks for reviewing and merging so quickly 🙂

lahma:
And this package is now live on NuGet with version 3.2.0! 🚀

PR #917, opened 2020-08-06
Quartz.Net 3.1.0 introduces breaking changes in configuring Quartz.Impl.AdoJobStore.JobStoreSupport

footcha:
While upgrading from Quartz 3.0.7 to 3.1.0 I experienced some issues with a clustered scheduler. It seems that there is a hidden breaking change due to nice refactoring in #818. I tried to update a release notes page with some valuable comments on how to overcome this issue.


lahma:
Ah, really good catch. Thanks for updating the docs, should be online in a few.

PR #923, opened 2020-08-09
Use 2.1 as the minimum version for the .NET Platform Extensions

kevinchalet:
As discussed in https://github.com/openiddict/openiddict-core/issues/361, this PR updates the various .csproj to use 2.1 as the minimum version for the .NET Platform Extensions, but without forcing a specific range to allow users to explicitly reference newer versions.

/cc @lahma

### DIFF
--- a/src/Quartz.AspNetCore/Quartz.AspNetCore.csproj
+++ b/src/Quartz.AspNetCore/Quartz.AspNetCore.csproj
@@ -1,25 +1,34 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
-    <PropertyGroup>
-        <Description>Quartz.NET ASP.NET Core integration; $(Description)</Description>
-        <TargetFramework>netstandard2.0</TargetFramework>
-        <RootNamespace>Quartz</RootNamespace>
-        <GenerateDocumentationFile>True</GenerateDocumentationFile>
-        <Authors>Lewis Zou, Marko Lahma, Quartz.NET</Authors>
-    </PropertyGroup>
-    <ItemGroup>
-        <Compile Include="..\AssemblyInfo.cs" />
-    </ItemGroup>
+  <PropertyGroup>
+    <Description>Quartz.NET ASP.NET Core integration; $(Description)</Description>
+    <TargetFrameworks>netcoreapp2.2;netstandard2.0</TargetFrameworks>
+    <RootNamespace>Quartz</RootNamespace>
+    <GenerateDocumentationFile>True</GenerateDocumentationFile>
+    <Authors>Lewis Zou, Marko Lahma, Quartz.NET</Authors>
+  </PropertyGroup>
 
-    <ItemGroup>
-        <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="3.1.6" />
-        <PackageReference Include="Microsoft.AspNetCore.Diagnostics.HealthChecks" Version="2.2.0" />
-    </ItemGroup>
+  <ItemGroup>
+    <Compile Include="..\AssemblyInfo.cs" />
+  </ItemGroup>
 
-    <ItemGroup>
-      <ProjectReference Include="..\Quartz.Extensions.DependencyInjection\Quartz.Extensions.DependencyInjection.csproj" />
-      <ProjectReference Include="..\Quartz.Extensions.Hosting\Quartz.Extensions.Hosting.csproj" />
-      <ProjectReference Include="..\Quartz\Quartz.csproj" />
-    </ItemGroup>
+  <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' ">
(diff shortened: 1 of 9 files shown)


src/Quartz.AspNetCore/Quartz.AspNetCore.csproj
```diff
@@ -1,25 +1,34 @@
...
+  <ItemGroup>
+    <Compile Include="..\AssemblyInfo.cs" />
+  </ItemGroup>
 
-    <ItemGroup>
-      <ProjectReference Include="..\Quartz.Extensions.DependencyInjection\Quartz.Extensions.DependencyInjection.csproj" />
-      <ProjectReference Include="..\Quartz.Extensions.Hosting\Quartz.Extensions.Hosting.csproj" />
-      <ProjectReference Include="..\Quartz\Quartz.csproj" />
-    </ItemGroup>
+  <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' ">
```
kevinchalet, line 15: @lahma multi-targeting is a good option to add support features that are only present in newer versions of the .NET Platform Extensions while still supporting older ones. Not sure if it's something you're interested in doing so let me know if you don't like this approach and I'll revert these changes.

src/Quartz.AspNetCore/Quartz.AspNetCore.csproj
```diff
@@ -1,25 +1,34 @@
...
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="2.1.1" />
+  </ItemGroup>
+
+  <ItemGroup Condition=" '$(TargetFramework)' == 'netcoreapp2.2' ">
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="2.2.0" />
+    <PackageReference Include="Microsoft.AspNetCore.Diagnostics.HealthChecks" Version="2.2.0" />
```
kevinchalet, line 21: FYI, health checks are not tied to ASP.NET Core and you could have a `Quartz.Extensions.HealthChecks` package referencing `Microsoft.Extensions.Diagnostics.HealthChecks`. `Microsoft.AspNetCore.Diagnostics.HealthChecks` only contains the ASP.NET Core middleware responsible of handling the health requests.

src/Quartz.Tests.Integration/Quartz.Tests.Integration.csproj
```diff
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp3.1;$(TargetFullFrameworkVersion)</TargetFrameworks>
+    <TargetFrameworks>netcoreapp2.1;netcoreapp3.1;$(TargetFullFrameworkVersion)</TargetFrameworks>
```
kevinchalet, line 3: While convenient and useful, testing against .NET Core 2.1 is not mandatory. Let me know if you want me to revert this change.

src/Quartz.AspNetCore/Quartz.AspNetCore.csproj
lahma: I think this looks fine 👍

src/Quartz.AspNetCore/Quartz.AspNetCore.csproj
lahma: I'd like to keep AspNetCore project as an umbrella for now, otherwise there will be a lot of small dlls with very specific functionality (though it's more correct).

src/Quartz.Tests.Integration/Quartz.Tests.Integration.csproj
lahma: I think this is fine for now.

lahma:
Looks good to me, should we merge?

src/Quartz.AspNetCore/Quartz.AspNetCore.csproj
kevinchalet: Makes sense 😃

kevinchalet:
@lahma assuming you're happy with the changes, yes, please! 😃

lahma:
Thank you 🙏🏻

kevinchalet:
Thanks for merging it 😃 

FYI, the build failed with an error in the `TestCronExpressionWithExtraWhiteSpace` test: https://ci.appveyor.com/project/lahma/quartznet-6fcn8/builds/34562433.

Since the test has a `test failed before because of improper trimming` comment, I suspect it has already happened in the past?

lahma:
That test case seems to be a bit unstable... But went through after couple retries, packages are now on dev feed!

kevinchalet:
Thanks! I updated https://github.com/openiddict/openiddict-core/pull/1052 to use the latest bits and it works like a charm! 👏

PR #928, opened 2020-08-11
Fix - Parsing boolean db value

L3Z4:
DbValue for boolean is "0" or "1" and cannot be parsed by bool.Parse. It throws an exception.
Falling back on TryParse then check string equality on error.

### DIFF
--- a/src/Quartz/Impl/AdoJobStore/StdAdoDelegate.cs
+++ b/src/Quartz/Impl/AdoJobStore/StdAdoDelegate.cs
@@ -651,7 +651,7 @@ namespace Quartz.Impl.AdoJobStore
         {
             if (columnValue != null && columnValue != DBNull.Value)
             {
-                return Convert.ToBoolean(columnValue);
+                return !bool.TryParse(columnValue as string, out var success) ? columnValue.ToString() == "1" : success;
             }
 
             throw new ArgumentException("Value must be non-null.");


lahma:
@L3Z4 what database are you using? Might it be that you haven't chosen a proper ADO.NET delegate like `OracleDelegate`?

L3Z4:
I'm using SQL Server, which represents boolean values as 0/1. In dotnet
Boolean.Parse expect a True/false string to be successfully parsed.
Maybe I can create a dedicated overlaod, but this code is conservative and
will not impact previously working db.

Le mer. 12 août 2020 à 07:28, Marko Lahma <notifications@github.com> a
écrit :

lahma:
I'm a bit confused, have you configured `SqlServerDelegate`? This is what you really want to use to get proper optimization for trigger retrieval. There should be no need for this modification if you are using the standard SQL scripts.

L3Z4:
I configured SqlServerDelegate with the provided extension method. Sql
tables has been created using the provided SQL scripts, found in this
repository.

Let me do a double check to be absolutely sure

Le mer. 12 août 2020 à 07:48, Marko Lahma <notifications@github.com> a
écrit :

L3Z4:
Ok my bad, I guess I was relying on previous version of Sql service script. I have just re-created the database and it seems to works. 
Should I delete this merge request ?

lahma:
Great that you got it sorted. I'd rather have the error on having odd schema and guide people to use proper delegate when needed. I don't see that we should create a workaround here. Thanks for reaching out though 👍

PR #930, opened 2020-08-11
Register an ITypeLoadHelper by default in services.AddQuartz()

### LINKED ISSUE #924: Should services.AddQuartz() register a default ITypeLoadHelper service?
Currently, if you don't explicitly call `UseSimpleTypeLoader()` when using `services.AddQuartz()`, you get an exception from DI:

```
System.InvalidOperationException : 'No service for type 'Quartz.Spi.ITypeLoadHelper' has been registered.'
```

Assuming the default `SimpleTypeLoadHelper` implementation is appropriate for most use cases, maybe it should be directly registered by `services.AddQuartz()`?

/cc @lahma

andrewlock:
The `SimpleTypeLoadHelper` is used by default in the `StdSchedulerFactory` if no other `ITypeLoadHelper` is registered:

https://github.com/quartznet/quartznet/blob/9193fac3587b9a020bf861d2404d463d67e05b07/src/Quartz/Impl/StdSchedulerFactory.cs#L416

But if you don't explicitly call `UseSimpleTypeLoader()`, the required type isn't registered in DI and you get an `InvalidOperationException` on startup.

This changes registers the required service, as long as no other `ITypeLoadHelper` is added inside the call to `AddQuartz()`.

Unfortunately we can't call UseSimpleTypeLoader(), as that would overwrite the configuration for any other configured type loader (while leaving the service registration intact):

https://github.com/quartznet/quartznet/blob/89b414f142010086d057c70085d4fe3c65b39f26/src/Quartz.Extensions.DependencyInjection/ServiceCollectionQuartzConfigurator.cs#L76-L80

Fixes #924

### DIFF
--- a/src/Quartz.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
+++ b/src/Quartz.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
@@ -8,6 +8,7 @@ using Microsoft.Extensions.Logging;
 
 using Quartz.Logging;
 using Quartz.Simpl;
+using Quartz.Spi;
 
 namespace Quartz
 {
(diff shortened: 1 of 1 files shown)


lahma:
Thank you for fixing this! 👍

PR #963, opened 2020-09-14
Index Drop/Create modified to Drop_Existing = ON

WadeTheFade:
FROM:
https://www.tech-recipes.com/rx/60642/creating-index-drop_existing-sql-server/

If we are re-creating a clustered index on a table, which already has a non-clustered index, using DROP_EXISTING = ON in SQL Server while re-creating a clustered index has the following benefits:

• Non-clustered indexes are not re-built again. If we are re-creating a clustered index with DROP_EXISTING = ON in SQL Server, it knows the clustering key has not been changed in the Non-clustered index. This saves a significant amount of time. Every Non-clustered index has a clustering key which refers to the Clustered index. If a clustered index is rebuilt, then SQL Server has to modify the Non-Clustered index as well. However, using DROP_EXISTING = ON is beneficial because we do not need to rebuild the Non-clustered indexes again.

• If we use the old method to drop and re-create a clustered index, it causes all non-clustered index on that table to be rebuilt again to match the clustering key. This can be saved using DROP_EXISTING = ON with clustered index.


lahma:
Thanks for investigating and creating the PR. I see quite a lot of added complexity. Out of curiosity, what kind of row counts are you running when you decided to find a faster way to rebuild? And more importantly, which versions of SQL Server support this feature?

WadeTheFade:
I was going to leave two areas, one being commented out for supported SQL versions. But I was trying to figure out when DROP_Existing was introduced, but it looks like it goes all the way back to SQL 2000. So I just left it in.

The main reason I made these changes as it our CI/CD pipeline runs AliaSQL for versioning. Our INDEX scripts are in the EVERYTIME folder that gets run anytime there is a change in the file. This needs the existing check because `DROP_EXISTING=ON` has a weird requirement that it needs to already exist. The way its written allows you to deploy to any environment regardless of if the index existed before script or not.

We don't want to just run drops and creates manually in production. These changes will be in a branched and committed.

database/tables/tables_sqlServer.sql
```diff
@@ -347,42 +347,203 @@ ALTER TABLE [dbo].[QRTZ_TRIGGERS] ADD
...
+--CREATE INDEX [IDX_QRTZ_T_STATE]               ON [dbo].[QRTZ_TRIGGERS](SCHED_NAME, TRIGGER_STATE);
+--CREATE INDEX [IDX_QRTZ_T_N_STATE]             ON [dbo].[QRTZ_TRIGGERS](SCHED_NAME, TRIGGER_NAME, TRIGGER_GROUP, TRIGGER_STATE);
+---CREATE INDEX [IDX_QRTZ_T_NEXT_FIRE_TIME]      ON [dbo].[QRTZ_TRIGGERS](SCHED_NAME, NEXT_FIRE_TIME);
+--CREATE INDEX [IDX_QRTZ_T_NFT_ST]              ON [dbo].[QRTZ_TRIGGERS](SCHED_NAME, TRIGGER_STATE, NEXT_FIRE_TIME);
+--CREATE INDEX [IDX_QRTZ_T_NFT_ST_MISFIRE]      ON [dbo].[QRTZ_TRIGGERS](SCHED_NAME, MISFIRE_INSTR, NEXT_FIRE_TIME, TRIGGER_STATE);
+--CREATE INDEX [IDX_QRTZ_T_NFT_ST_MISFIRE_GRP]  ON [dbo].[QRTZ_TRIGGERS](SCHED_NAME, MISFIRE_INSTR, NEXT_FIRE_TIME, TRIGGER_GROUP, TRIGGER_STATE);
+--CREATE INDEX [IDX_QRTZ_FT_INST_JOB_REQ_RCVRY] ON [dbo].[QRTZ_FIRED_TRIGGERS](SCHED_NAME, INSTANCE_NAME, REQUESTS_RECOVERY);
+--CREATE INDEX [IDX_QRTZ_FT_G_J]                ON [dbo].[QRTZ_FIRED_TRIGGERS](SCHED_NAME, JOB_GROUP, JOB_NAME);
+--CREATE INDEX [IDX_QRTZ_FT_G_T]                ON [dbo].[QRTZ_FIRED_TRIGGERS](SCHED_NAME, TRIGGER_GROUP, TRIGGER_NAME);
+--GO
```
WadeTheFade, line 547: Oops this can be deleted. I've split out the original block of drops, to be "ones we are discarding" and "ones we can rebuild our index" in one of two ways.

lahma:
@WadeTheFade — thank you for this, and apologies for the very long silence.

The file you patched has changed shape since 2020. `tables/tables_sqlServer.sql` is now a
pristine-init script: everything destructive sits inside `IF @DropDb = 1` (line 12), and the indexes
are created once at the end, after the tables have just been made. There is no longer a
drop-then-recreate of an index that survives, so `DROP_EXISTING = ON` has nothing to save. Every
clustered index in the schema is a `PRIMARY KEY CLUSTERED` constraint that no script re-creates,
which is the case the option really pays off in.

The re-runnable index management you actually needed does exist now — it just moved. It is
`database/migrations/3.20/index_alignment_sqlServer.sql` and section 5 of
`migrations/4.0/schema_30_to_40_upgrade_sqlServer.sql`, both `IF NOT EXISTS … CREATE INDEX` and
`IF EXISTS … DROP INDEX`, safe to re-run from a deployment pipeline. Those files are generated from
`build/Build.DatabaseMigrations.Scripts.cs`, so a hand edit is no longer the way in.

Closing as superseded. If your pipeline still hits a rebuild cost against those migrations, open an
issue with the numbers and we will look again.

PR #978, opened 2020-09-30
Improve options pattern for DI

### LINKED ISSUE #955: Allow custom configuration via Options pattern
"I'm always frustrated when" I cannot access `IServiceProvider` in the `configure` action of `AddQuartz`.

`AddQuartz` should have an overload like `AddQuartz(this IServiceCollection serviceCollection, Action<IServiceProvider, IServiceCollectionQuartzConfigurator>? configure = null)` similar to eg. [`AddContext`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.entityframeworkservicecollectionextensions.adddbcontext?view=efcore-3.1#Microsoft_Extensions_DependencyInjection_EntityFrameworkServiceCollectionExtensions_AddDbContext__1_Microsoft_Extensions_DependencyInjection_IServiceCollection_System_Action_System_IServiceProvider_Microsoft_EntityFrameworkCore_DbContextOptionsBuilder__Microsoft_Extensions_DependencyInjection_ServiceLifetime_Microsoft_Extensions_DependencyInjection_ServiceLifetime_).

Alternatively Quartz could expose an options interface, and I could configure with eg. [`IConfigureOptions<QuartzOptions>`](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/options?view=aspnetcore-3.1#use-di-services-to-configure-options). For example [`AddCors`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.corsservicecollectionextensions.addcors?view=aspnetcore-3.1) does this with [`CorsOptions`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.cors.infrastructure.corsoptions?view=aspnetcore-3.1). One can use `IConfigureOptions<CorsOptions>` to configure cors and access `IServiceProvider`.

As a workaround I can manually bootstrap the dependencies I'd like to access in `AddQuartz`.

**Additional context**
My use cases are similar to Andrew Lock's described in [this blog post](https://andrewlock.net/avoiding-startup-service-injection-in-asp-net-core-3/). Hope you find some inspiration there.

Many thanks!

EDIT: To be clear, this is a minor improvement on the configuration interface as [feedback was requested](https://github.com/quartznet/quartznet/issues/877#issuecomment-658351482).

lahma:
fixes #955


src/Quartz.Extensions.DependencyInjection/MicrosoftDependencyInjectionJobFactory.cs
```diff
@@ -10,19 +11,19 @@ namespace Quartz
     internal sealed class MicrosoftDependencyInjectionJobFactory : PropertySettingJobFactory
     {
         private readonly IServiceProvider serviceProvider;
-        private readonly JobFactoryOptions options;
+        private readonly QuartzOptions options;
```
kevinchalet, line 14: If `MicrosoftDependencyInjectionJobFactory` is long-lived, maybe it would be better to have a reference to `IOptions<QuartzOptions>` here and call `options.Value` in `InstantiateJob`. (which would enable options reload support).

src/Quartz.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
```diff
@@ -31,91 +42,75 @@ public static class ServiceCollectionExtensions
...
-                    options.Set(key, builder.Properties[key]);
+                    options[key] = schedulerBuilder.Properties[key];
                 }
-
-                return new ServiceCollectionSchedulerFactory(serviceProvider, options);
             });
-            return services;
-        }
+            
+            services.AddSingleton<ContainerConfigurationProcessor>();
```
kevinchalet, line 64: Should you use `TryAdd` here to avoid re-registering these days if `AddQuartz()` is called multiple times? It would be great if `AddQuartz()` was fully idempotent.

src/Quartz.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
```diff
@@ -20,6 +20,17 @@ public static class ServiceCollectionExtensions
...
+        {
+            return AddQuartz(services, new NameValueCollection(), configure);
+        }
+
+        /// <summary>
+        /// Configures Quartz services to underlying service collection. This API maybe change!
+        /// </summary>
+        public static IServiceCollection AddQuartz(
+            this IServiceCollection services,
+            NameValueCollection properties,
```
kevinchalet, line 32: Using `NameValueCollection` in `IServiceCollection` is not very .NET Core/`Microsoft.Extensions.Options`-ish. If you're not against introducing breaking changes (I see a few ones are already introduced in other parts), maybe you could move this thing to a dedicated method on `IServiceCollectionQuartzConfigurator`? Something like `options.ImportProperties(new NameValueCollection())` that would call `services.Configure()` under the hood?

src/Quartz.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
```diff
@@ -20,6 +20,17 @@ public static class ServiceCollectionExtensions
         public static IServiceCollection AddQuartz(
```
kevinchalet, line 20: Now that Quartz.NET depends on `Microsoft.Extensions.Options`, I think it would be better to call `services.AddOptions()` to ensure the options services are automatically imported.

kevinchalet (commented):
Very promising stuff here 😄

src/Quartz.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
```diff
@@ -31,91 +42,75 @@ public static class ServiceCollectionExtensions
...
+                    options[key] = schedulerBuilder.Properties[key];
                 }
-
-                return new ServiceCollectionSchedulerFactory(serviceProvider, options);
             });
-            return services;
-        }
+            
+            services.AddSingleton<ContainerConfigurationProcessor>();
+            services.AddSingleton<IPostConfigureOptions<QuartzOptions>, QuartzConfiguration>();
```
kevinchalet, line 65: `QuartzConfiguration``currently does nothing so it likely doesn't matter, but if you're interested in making `AddQuartz()` idempotent, you'll probably want to use `TryAddEnumerable` for this service.

src/Quartz.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
lahma: Changed to use TryAdd

src/Quartz.Extensions.DependencyInjection/MicrosoftDependencyInjectionJobFactory.cs
lahma: Of course, changed.

src/Quartz.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
lahma: Added.

src/Quartz.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
lahma: Sounds reasonable, but I think that requires some changes... I'd rather leave that to later time as there's some scheduling (grin) pressure here to get a release out 😉 

For Quartz it's idiomatic to use `NameValueCollection` but that's a fair point that it's not something you'd see on public API.

lahma:
@kevinchalet I've addressed all but one of your comments. A big thank your for having a look (which I really was hoping for but not wanting to put pressure for). Hopefully latest commit takes this closer to merge-ready.

kevinchalet:
> A big thank your for having a look (which I really was hoping for but not wanting to put pressure for).

My pleasure :clap:

I'll give the new bits a try ASAP.

PR #994, opened 2020-10-16
Allow concurrent access to XML file with job and trigger definitions

### LINKED ISSUE #993: Process XML file with jobs without blocking it for reads
### Overview

XMLSchedulingDataProcessor [opens](https://github.com/quartznet/quartznet/blob/ca3c9cd311a8c0dc57fbb5967cce94af9c66a01e/src/Quartz/Xml/XMLSchedulingDataProcessor.cs#L176) the quartz_jobs.xml file with [no sharing](https://docs.microsoft.com/en-us/dotnet/api/system.io.file.open#System_IO_File_Open_System_String_System_IO_FileMode_System_IO_FileAccess_). If there is another process accessing the same file concurrently, such operation could result in the "System.IO.IOException: The process cannot access the file 'C:\some_path\quartz_jobs.xml' because it is being used by another process" error.

### Use case

To validate scheduled jobs agains some business rules, it is necessary to open quartz_jobs.xml in the read-only mode, parse the content, execute validation rules, report discovered violations.

### Proposed solution

XMLSchedulingDataProcessor should [open](https://docs.microsoft.com/en-us/dotnet/api/system.io.file.open?view=netcore-3.1#System_IO_File_Open_System_String_System_IO_FileMode_System_IO_FileAccess_System_IO_FileShare_) quartz_jobs.xml file with the [FileShare.Read](https://docs.microsoft.com/en-us/dotnet/api/system.io.fileshare) access.

### Alternatives

A external "validator" process can re-try the open file operation.

Anyway, taking into account that XMLSchedulingDataProcessor requires only read-only access to the XML file, and a small effort required to implement the change, it would be good to improve the code.

### Additional context

XMLSchedulingDataProcessorPlugin also [access](https://github.com/quartznet/quartznet/blob/37dc4454a3337167d073cb2ed69658c7679d9f83/src/Quartz.Plugins/Plugin/Xml/XMLSchedulingDataProcessorPlugin.cs#L359) the XML file using correct [FileShare.ReadWrite](https://docs.microsoft.com/en-us/dotnet/api/system.io.fileshare) option.

safor:
Changes the way how XMLSchedulingDataProcessor opens the XML file with job and trigger definitions to allow concurrent read-only access from other processes.

fixes #993

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -10,10 +10,7 @@ This is a maintenance release containing mostly bud fixes.
 
   * Make QuartzOptions Triggers and JobDetails public (#981)
   * Fix configuration system injection for dictionary/quartz.jobStore.misfireThreshold in DI (#983)
-
-* NEW FEATURE
-
-  * Allow concurrent access to XML file with job and trigger definitions (#993)
+  * XMLSchedulingDataProcessor can cause IOException due to file locking (#993)
 
 ## Release 3.2.0, Oct 1 2020
 
--- a/changelog.md
+++ b/changelog.md
@@ -11,6 +11,10 @@ This is a maintenance release containing mostly bud fixes.
   * Make QuartzOptions Triggers and JobDetails public (#981)
   * Fix configuration system injection for dictionary/quartz.jobStore.misfireThreshold in DI (#983)
 
+* NEW FEATURE
+
+  * Allow concurrent access to XML file with job and trigger definitions (#993)
+
 ## Release 3.2.0, Oct 1 2020
 
 This is a release that focuses on restructuring some packages that warrants for a minor version number increment.
--- a/src/Quartz/Xml/XMLSchedulingDataProcessor.cs
+++ b/src/Quartz/Xml/XMLSchedulingDataProcessor.cs
@@ -173,7 +173,7 @@ namespace Quartz.Xml
 
             Log.InfoFormat("Parsing XML file: {0} with systemId: {1}", fileName, systemId);
 
-            using (var stream = File.Open(fileName, FileMode.Open, FileAccess.Read))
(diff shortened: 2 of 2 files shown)


safor:
I admit that opening this PR without an agreement violates the [contributing guidelines](https://github.com/quartznet/quartznet/blob/master/CONTRIBUTING.md). I registered the issue #993 and decided to open PR, because the change is very small and it could speed up the delivery of the future. Hope the owner and community won't find this change redundant.

changelog.md
```diff
@@ -11,6 +11,10 @@ This is a maintenance release containing mostly bud fixes.
   * Make QuartzOptions Triggers and JobDetails public (#981)
   * Fix configuration system injection for dictionary/quartz.jobStore.misfireThreshold in DI (#983)
 
+* NEW FEATURE
```
lahma, line 14: Can we move this to bug fix side as I think this rather a bug? "XMLSchedulingDataProcessor can throw IOException due to file locking" or something similar?

lahma:
Thank you @safor , don't sweat about the contributing guidelines. These PRs are more than welcome and I'm happy to receive them, just didn't get the chance to review the issue before. Just a small nit about maybe this belongs to bug fix side of release notes?

changelog.md
safor: It proves that it makes sence to discuss an issue before opening PR :)

I was not sure whether it is a feature to bug fix. Still I think it is not a bug in this code, but I agree that the change is better to specify under FIXES.

The changelog was undeted in f5e549a

lahma:
Excellent, thank you very much! Will be part of 3.2.1 which will hopefully be out this weekend.

PR #1003, opened 2020-10-20
Add UseZeroSizeThreadPool to configuration

sergeyshaykhullin:


sergeyshaykhullin:
Something weird with commits =/

sergeyshaykhullin:
I found a problem with MassTransit integration

It is autostarts scheduler inside MassTransit

```cs
// SchedulerBusObserver.PostStart
await _scheduler.Start().ConfigureAwait(false);
```

Any way to detenmine using scheduler that it is running on `ZeroSizeThreadPool` and starting is illegal?

Maybe for `ZeroSizeThreadPool` `Start` can do nothing?

lahma:
> Any way to detenmine using scheduler that it is running on `ZeroSizeThreadPool` and starting is illegal?
> Maybe for `ZeroSizeThreadPool` `Start`/... can do nothing?

In MassTransit's case I think the default behavior is to expect the scheduler to start, otherwise things won't work. Maybe a parameter (if really needed) to suppress the start.

> I think that it might be... More abstract thing... Like Worker and Master with preconfigured values

Yes the `ZeroSizeThreadPool` and not starting the scheduler might not be too obvious. The DI interface might offer some readable way to state this.

lahma (approved):

lahma:
Thank you for this!

lahma:
Version 3.2.3 of Quartz.NET has been released including this new feature, thanks again for the contribution!

PR #1006, opened 2020-10-22
Add Quartz.OpenTracing support 

PavelStefanov:
Hi everyone!

I was really happy about OpenTelemetry support but it's still in beta and most projects use OpenTracing instead.
I added OpenTracing support and it'd be great if you could help me to finish it correctly and publish this NuGet package.

### DIFF
--- a/src/Quartz.OpenTracing/Quartz.OpenTracing.csproj
+++ b/src/Quartz.OpenTracing/Quartz.OpenTracing.csproj
@@ -1,22 +1,23 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
-	<PropertyGroup>
-		<TargetFramework>netstandard2.0</TargetFramework>
-		<RootNamespace>Quartz.OpenTracing</RootNamespace>
-		<Title>Quartz.NET OpenTracing integration</Title>
-		<Description>Quartz.NET OpenTracing integration</Description>
-		<GenerateDocumentationFile>True</GenerateDocumentationFile>
-	</PropertyGroup>
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <RootNamespace>Quartz.OpenTracing</RootNamespace>
+    <Title>Quartz.NET OpenTracing integration</Title>
+    <Description>Quartz.NET OpenTracing integration</Description>
+    <Authors>Pavel Stefanov, Quartz.NET</Authors>
+    <GenerateDocumentationFile>True</GenerateDocumentationFile>
+  </PropertyGroup>
 
-	<ItemGroup>
-		<PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="2.1.1" />
-		<PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" Version="2.1.1" />
-		<PackageReference Include="Microsoft.Extensions.Options" Version="2.1.1" />
-		<PackageReference Include="OpenTracing" Version="0.12.1" />
-	</ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="2.1.1" />
+    <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" Version="2.1.1" />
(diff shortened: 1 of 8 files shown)


src/Quartz.OpenTracing/QuartzDiagnosticOptions.cs
```diff
@@ -0,0 +1,42 @@
+using System;
+using System.Collections.Generic;
+
+namespace Quartz.OpenTracing
+{
+    public class QuartzDiagnosticOptions
```
lahma, line 6: I think OpenTelemetry share the same concepts here. Should these options be unified to be quite the same between the integrations, thoughts?

lahma (commented):
Looks like a great addition. I don't have much to give feature wise as I'm unfamiliar with OpenTracing. It seems that the PR is correctly hooked to produce the NuGet package already.

src/Quartz.OpenTracing/Quartz.OpenTracing.csproj
```diff
@@ -0,0 +1,22 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+	<PropertyGroup>
+		<TargetFramework>netstandard2.0</TargetFramework>
+		<RootNamespace>Quartz.OpenTracing</RootNamespace>
+		<Title>Quartz.NET OpenTracing integration</Title>
+		<Description>Quartz.NET OpenTracing integration</Description>
+		<GenerateDocumentationFile>True</GenerateDocumentationFile>
```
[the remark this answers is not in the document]
PavelStefanov: Thanks!
There's awesome approach to maintain coding style it's [EditorConfig](https://editorconfig.org/). It's supported by VS, Rider and VS Code. I generated .editorconfig using IntelliCode. It analyses current project and creates relevant rules. If you don't mind. I hope this helps in the future.

But I've few questions e.g. `Quartz.OpenTelemetry.Instrumentation.csproj` has some special vars like `$(TargetFullFrameworkVersion)` and `$(Description)`, should I add the same things?

src/Quartz.OpenTracing/QuartzDiagnosticOptions.cs
PavelStefanov: Yes, it looks similar and indeed OpenTracing and OpenTelemetry have the same concept, but there're difference between implementations. 

- Looks like in OpenTelemetry implementation we try to subscribe to Veto event also but I think for tracing we interest only for Execute event cuz we need tracing some long operations not just logging some statements
- In OpenTracing impl we have only event name and payload from diagnostic and then we manually create span and fill all tags ourself therefore we've `ComponentName` property and `OperationNameResolver` func in the options and users can override these as they want. At the same time as I can see OpenTelemetry gets already filled Activity and can only add some tags, it can't change OperationName cuz it's already filled from DiagnosticListener.
- If we merge these options we'll must move it to `Quartz` project or will create some new project to share options. I don't think so it's good idea to move it to `Quartz`, it's nothing connecting to main project it's implementation details of tracing.
And then we have to support it between impls and I expect that after OpenTelemetry is released it'll be main approach for distributing tracing and it can develop faster.

But I'm not familiar with OpenTelemetry implementation so well and mb I don't know something.

PavelStefanov: I also can add example code in `Quartz.Examples.AspNetCore` project but it looks like so overloaded already.

And I suggest to create another issue for `docs`.

PavelStefanov:
I also can add example code in `Quartz.Examples.AspNetCore` project but it looks like so overloaded already.

And I suggest to create another issue for docs.

lahma:
Thank you for this!

 I'll add some documentation and update both tracing solutions to include a note that they use beta packages and API might still change a bit. All PRs to fine-tune are greatly welcome as you are the one who knows how this thing actually is expected work 🙂

lahma:
I've added some basic documentation with https://github.com/quartznet/quartznet/commit/3ce179f252559b8725d3fa0f56fd6448099be2d6 and will already go live and state that the unpublished version will be required.

lahma:
Version 3.2.3 of Quartz.NET has been released including this new feature, thanks again for the contribution!

PR #1007, opened 2020-10-23
Empty statement redundancy and expression always true code fix 

onpoc:


onpoc:
@hacktoberfest
Empty statement redundancy and expression always true code fix

lahma:
Unfortunately this breaks Quartz serialization logic.

PR #1026, opened 2020-11-10
JobBuilder.UsingJobData(string key, string value) should be JobBuilder.UsingJobData(string key, string? value)

### LINKED ISSUE #1025: JobBuilder.UsingJobData(string key, string value) should be JobBuilder.UsingJobData(string key, string? value)
Because the library make use of the nullability operators, the definition of following member says:
I accept a nullable string as value.

```
public class StringKeyDirtyFlagMap : DirtyFlagMap<string, object>
{
        /// <summary>
        /// Adds the given <see cref="string" /> value to the <see cref="IJob" />'s
        /// data map.
        /// </summary>
        public virtual void Put(string key, string? value)
        {
            base.Put(key, value!);
        }
}
```

Then this should flow to all other usages:

```
    public class JobBuilder : IJobConfigurator
    {
        /// <summary>
        /// Add the given key-value pair to the JobDetail's <see cref="JobDataMap" />.
        /// </summary>
        ///<returns>the updated JobBuilder</returns>
        /// <seealso cref="IJobDetail.JobDataMap" />
        public JobBuilder UsingJobData(string key, string value)
        {
            jobDataMap.Put(key, value);
            return this;
        }
    }
```

I encountered this in version 3.2.3, but probably introduced in #866 .

Because of this we get now following warnings (static analyzer):
_Possible 'null' assignment to non-nullable entity_

dvdwouwe:
JobBuilder.UsingJobData(string key, string value) should be JobBuilder.UsingJobData(string key, string? value)

fixes #1025

### DIFF
--- a/src/Quartz/JobBuilder.cs
+++ b/src/Quartz/JobBuilder.cs
@@ -277,7 +277,7 @@ namespace Quartz
         /// </summary>
         ///<returns>the updated JobBuilder</returns>
         /// <seealso cref="IJobDetail.JobDataMap" />
-        public JobBuilder UsingJobData(string key, string value)
+        public JobBuilder UsingJobData(string key, string? value)
         {
             jobDataMap.Put(key, value);
             return this;


dvdwouwe:
First time I create a PR, dunno how to link to the issue #1025

lahma:
Thank you, I just added `fixes #1025` to the description and will allow us to automatically resolve issues via PR.

PR #1040, opened 2020-11-19
Handled exceptions thrown while retrieving the misfired trigger

prajumt:
When the scheduler tries to recover all misfired triggers, if one of the triggers throws an exception due to some reason(Invalid CRON expression in my case), the remaining triggers are skipped due to the fact that the exception is not handled. So, we made this change to make the misfired trigger recovery process resilient to any kind if exception.

### DIFF
--- a/src/Quartz/Impl/AdoJobStore/JobStoreSupport.cs
+++ b/src/Quartz/Impl/AdoJobStore/JobStoreSupport.cs
@@ -1,4 +1,4 @@
-#region License
+﻿#region License
 
 /*
  * All content copyright Marko Lahma, unless otherwise indicated. All rights reserved.
@@ -751,7 +751,17 @@ namespace Quartz.Impl.AdoJobStore
 
             foreach (TriggerKey triggerKey in misfiredTriggers)
             {
-                var trig = await RetrieveTrigger(conn, triggerKey, cancellationToken).ConfigureAwait(false);
+                IOperableTrigger? trig;
+                try
+                {
+                    trig = await RetrieveTrigger(conn, triggerKey, cancellationToken).ConfigureAwait(false);
+                }
+                catch (Exception e)
+                {
+                    Log.ErrorException($"Error retrieving the misfired trigger: '{triggerKey}'", e);
+                    continue;
+                }
+
                 if (trig == null)
                 {
                     continue;
--- a/src/Quartz/Impl/AdoJobStore/JobStoreSupport.cs
+++ b/src/Quartz/Impl/AdoJobStore/JobStoreSupport.cs
@@ -767,7 +767,15 @@ namespace Quartz.Impl.AdoJobStore
                     continue;
                 }
 
-                await DoUpdateOfMisfiredTrigger(conn, trig, false, StateWaiting, recovering).ConfigureAwait(false);
+                try
+                {
+                    await DoUpdateOfMisfiredTrigger(conn, trig, false, StateWaiting, recovering).ConfigureAwait(false);
+                }
+                catch (Exception e)
+                {
+                    Log.ErrorException($"Error updating misfired trigger: '{trig.Key}'", e);
+                    continue;
+                }
 
                 DateTimeOffset? nextTime = trig.GetNextFireTimeUtc();
                 if (nextTime.HasValue && nextTime.Value < earliestNewTime)


lahma:
How did you get the invalid cron expression into the database in the first place? This will start swallowing some exceptions that might indicate that there's something more dire wrong the whole process.

mghafoori:
Hi Marko. Thanks for your reply. We have been using Quartz.Net since version one, and our best guess as to how the bad CRON ended up in DB is a possible bug in the older versions of CRON validations.  The issue that this caused us was that such a bad CRON would break the recovery routine and would not recover the rest of jobs. We thought by handling the error and logging it, at least we ensue that the rest of jobs get recovered.

prajumt:
This change(d33a851) is done in the similar context as the previous one(6dfde9c). Misfired triggers are recovered in a loop. For any reason(Invalid or missing job type in my case), if there is an exception while recovering a specific trigger, the rest of the triggers are not getting recovered as the loop is broken. I have made this change to log the exception and continue the loop so that the remaining triggers can also be recovered.

lahma (approved):

lahma:

I hope this is not a common case for Quartz.NET users, but like you've said this allows better behavior in case of errors - as long as logs are monitored and library users understand why possibly some triggers are not handled as expected.

sberkovitz:
I will share an excerpt of the internal RCA we wrote regarding this issue, it might provide more insight into how we arrived at this problem/solution.  I hope our assumptions are right!

> In Quartz, there are two primary mechanisms for firing triggers: the main scheduler thread and a misfire thread.  The main scheduler thread runs continuously with a short delay between queries.  Each time it runs, it wiill query the Quartz datastore (a SQL database in our case) for tasks that should run within a time window with a start time defined as the current time plus the thread delay time, and an end time defined as the current time plus the misfire threshold.
> 
> The misfire thread also runs continuously and with a short delay between queries. The purpose of the misfire thread is to identify tasks that did not run as originally scheduled and, depending on the misfire configuration, either run the task immediately or reschedule it for the main thread to pick up on a subsequent run.
> 
> Misfires can occur for several reasons such as the scheduler being down (i.e. maintenance, reboot, etc.) or under load when there are not enough server resources to execute all of the jobs at once.
> 
> While our task infrastructure is sufficiently large,  the scheduling of tasks has high concentrations of tasks executing on the hour and 15/30 minutes on top of the hour. This has led to large CPU and memory spikes aligning with these intervals.  During these spikes, which typically only last for a minute or two, there is a significantly higher likelihood of a misfire occuring due to the rapid consumption of available threads. The number of available threads is primarily limited by the amount of available memory, which was also being negatively impacted by a few tasks that were utilizing large amounts of memory.
> 
> Under normal operations, a misfire would be remedied by the Quartz misfire thread and the task would run again at its next scheduled interval. Our troubleshooting eventually uncovered that the misfire thread was crashing without any recovery, which prevented it from executing and rescheduling tasks as expected.
> 
> The underlying cause of the crash was a single bad CRON expression which had an hour component of 24.  In order to determine the next scheduled run time, the misfire thread calculates the next run time based on the CRON expression. Because 24 is not a valid hour component for a CRON expression, the parser throws an exception that is not handled, which in turn crashes the misfire thread. We are unsure of how exactly this job was scheduled with this value since we validate this field prior to scheduling. We do know that this CRON expression had existed prior to the Quartz upgrade, and that previous versions appeared unaffected by this value, based on the run history of the trigger.
> 
> Another contributing factor to this incident was the unpredictability or seeming randomness of the tasks impacted.  When Quartz queries the datastore for triggers (for both the main thread and misfire thread), it does so without a sort order specified. Instead, the underlying datastore is responsible for the ordering. In our case, this is by the primary key which is a composite key of three values, one of which is fixed, one of which is the tenant ID, and the other is a randomly generated GUID (128bit value). Because our mitigation tool deletes and re-creates tasks, each time we ran it, we would generate new GUIDs and this would alter the ordering of results.  It also meant that the misfire thread would crash unpredictably – if the first task it picked up had this expression, it would crash right away without recovering any jobs. However if the bad CRON expression ended up later in the list, the misfire thread would have run correctly for all tasks before it hit the bad data.
> 
> On June 17, 2020, we introduced a configuration change to Quartz to increase the trigger batch size from 1 to 5.  Quartz.NET is a port of the Java version of Quartz.  The default batch size in the Java version was 5 whereas the default in .NET was 1.  A larger batch size reduces the number of SQL queries executed and allows Quartz to process multiple triggers in a single loop of its main thread. Upon completing this change, we did observe a reduction in failure rates. We now understand that this change helped because it increases the chance that a trigger will be picked up before it is considered misfired. Under normal operations, Quartz has a delay between trigger scans of between 30 and 37 seconds (0-7 seconds jitter, randomly generated).  With a batch size of 1, the process of picking up 5 triggers is
> 1.	Pickup 1 trigger
> 2.	Wait 30-37 seconds
> 3.	Repeat #1 and #2 four more times
> 
> This results in, a total wait of between 150 and 185 seconds (2.5-3min).  With a batch size of 5, this wait time is eliminated which greatly increases the likelihood of a trigger being picked up prior to misfiring.
> 
>

lahma:
@sberkovitz Thank you for sharing this detailed analysis. A very enlightening and interesting read! 

> The default batch size in the Java version was 5 whereas the default in .NET was 1. 

I think .NET and Java share the same batch size of 1 though, which is defaulted in Java Quartz configuration.

https://github.com/quartz-scheduler/quartz/blob/d42fb7770f287afbf91f6629d90e7698761ad7d8/quartz-core/src/main/java/org/quartz/impl/StdSchedulerFactory.java#L693

PR #1054, opened 2020-12-14
Clarifying some wording, fixing a couple typos in the Best Practices document

b4ux1t3:
There were a few unneeded commas, parentheses and a couple typos that were making this section a little difficult to read. 

I also wanted to add a little blurb about some of the symptoms that unhandled exceptions in jobs can lead to in order to make it easier for someone running into those kinds of issues to track down this specific section of documentation.


lahma:
Thank you, proof-reading is always welcomed! I'm not a native speaker so I will make a lot of typos for original content at least.

PR #1063, opened 2021-01-05
Fix UseTimeZoneConverter when using IServiceCollectionQuartzConfigurator

olicooper:
I am not sure if this is the correct thing to do, but it is easier to simply show you the suggested changes for you to review.
I get this compile error message as follows:
> The type 'Quartz.IServiceCollectionQuartzConfigurator' cannot be used as type parameter 'T' in the generic type or method 'TimeZonePluginConfigurationExtensions.UseTimeZoneConverter<T>(T)'. There is no implicit reference conversion from 'Quartz.IServiceCollectionQuartzConfigurator' to 'Quartz.IPropertyConfigurationRoot'

`IServiceCollectionQuartzConfigurator` has the interface `IPropertyConfigurer` which itself has the interface `IPropertySetter`. `IPropertyConfigurationRoot` also has the `IPropertySetter` interface. By changing the interface requirements the problem should disappear (not tested).

### DIFF
--- a/src/Quartz.Plugins.TimeZoneConverter/TimeZonePluginConfigurationExtensions.cs
+++ b/src/Quartz.Plugins.TimeZoneConverter/TimeZonePluginConfigurationExtensions.cs
@@ -5,7 +5,7 @@ namespace Quartz
 {
     public static class TimeZonePluginConfigurationExtensions
     {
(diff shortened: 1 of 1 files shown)


olicooper:
This issue looks like it was caused by using an older version of the library (v3.1). I believe it is okay in v3.2.3

PR #1065, opened 2021-01-06
Updated JobRunShell to handle only scheduler-related cancellations

### LINKED ISSUE #1064: JobRunShell silently handles OperationCanceledException which is not correct in terms of job retry handling
**Describe the bug**

JobRunShell has an ability to cancel the execution via OperationCanceledException but try/catch logic handles any OperationCanceledException being thrown in user code which is silently handled rather than generating exception and report exception to JobListeners.

**Version used**

Quartz v 3.2.3

**To Reproduce**

```
    public class TestJob : IJob
    {
        public Task Execute(IJobExecutionContext context)
        {
            throw new OperationCanceledException();
        }
    }
```

Quartz treats this Job as cancelled which is not correct since for example HttpClient.Timeout throws exact exception which is an exception in code execution rather than signal to Quartz to cancel the Job.

**Expected behavior**

Based on the code provided inside Job.Execute method, this is a user code and any exception thrown there should not be handled by Quartz, so JobListeners could log them and trigger retries.

**Additional context**
Based on JobRunShell implementation, each Job.Execute is wrapped with try/catch statement and there is a catch block which handles the OperationCanceledException

```
  // Execute the job
  try
  {
  ...
      await job.Execute(jec).ConfigureAwait(false);
  ...
  }
  catch (OperationCanceledException)
  {
      endTime = SystemTime.UtcNow();
      log.InfoFormat($"Job {jobDetail.Key} was cancelled");
  }
```
To have graceful handling of OperationCanceledException, I`m suggesting to perform check for CancellationToken (OperationCanceledException.CancellationToken) inside exception to check whether token is Quartz related (for example Scheduler signaled to cancel job) or it is user code related exception which should not be handled

nazar-kuzo:
Updated JobRunShell to handle only scheduler-related cancellations.
Closes #1064

### DIFF
--- a/src/Quartz/Core/JobRunShell.cs
+++ b/src/Quartz/Core/JobRunShell.cs
@@ -197,6 +197,9 @@ namespace Quartz.Core
                         endTime = SystemTime.UtcNow();
                     }
                     catch (OperationCanceledException)
+                    
+                    // handle only scheduler-related cancellations
+                    when (cancellationToken.IsCancellationRequested)
                     {
                         endTime = SystemTime.UtcNow();
                         log.InfoFormat($"Job {jobDetail.Key} was cancelled");


nazar-kuzo:
@lahma, this filter explicitly checks for cancellation token passed from Scheduler to be cancelled rather than handle any similar exception. I might be missing something from there, so please take a look and let me know your thoughts.

lahma (approved):

lahma:
Looks legit, thank for bringing up the issue and fixing it!

PR #1086, opened 2021-01-24
Remove LibLog and use Microsoft logging

lahma:


justmara:
omg, waiting for this so much. that static LogManager brings some pain to tests :)
it was nice ... like 15 years ago.
looks like progress stalled on this?

lahma:
Replaced by #1480

PR #1090, opened 2021-01-26
Add  a regular AddJob Type

startewho:
When I want to dynicmic add a Job ,it ususally use the reflection.so  i add a regular AddJob method .

### DIFF
--- a/src/Quartz.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
+++ b/src/Quartz.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
@@ -47,15 +47,15 @@ namespace Quartz
 
                 return LogProvider.CurrentLogProvider as MicrosoftLoggingProvider;
             });
-            
+
             var schedulerBuilder = SchedulerBuilder.Create(properties);
             if (configure != null)
             {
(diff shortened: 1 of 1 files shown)


src/Quartz.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
```diff
@@ -113,23 +113,37 @@ public static class ServiceCollectionExtensions
...
+        /// <summary>
+        /// Add job to underlying service collection.jobType shoud be implement `IJob`
+        /// </summary>
+        public static IServiceCollectionQuartzConfigurator AddJob(
+           this IServiceCollectionQuartzConfigurator options,
+           Type jobType,
+           JobKey? jobKey = null,
+           Action<IJobConfigurator>? configure = null)
+        {
+            if (typeof(IJob).IsAssignableFrom(jobType))
```
lahma, line 127: I think there it's an error condition if the job doesn't implement `IJob`, should throw `ArgumentException` as pre-check, will make the diff smaller also (less nested).

lahma (changes requested):
Thanks for submitting the PR! Please have a look at my comment.

lahma (approved):

lahma:

PR #1091, opened 2021-01-26
Update crontriggers.md

ahmad-moussawi:
Wrap the * between code backticks to prevent the false italic style


lahma:
Thank you! Should be online soon when the docs build pipeline completes.

PR #1127, opened 2021-03-11
Use ActivatorUtilities in MicrosoftDependencyInjectionJobFactory

### LINKED ISSUE #1120: Disposable job is disposed twice when using UseMicrosoftDependencyInjectionScopedJobFactory
**Describe the bug**

Quartz job that implements `IDisposable` is disposed twice when using `DependencyInjectionJobFactory` with `CreateScope` enabled (e.g. via `UseMicrosoftDependencyInjectionScopedJobFactory` extension method).

This shouldn't be a big problem, because properly implemented `Dispose` can be called multiple times safely according to Framework Design Guidelines. But still, it's something that could be improved.

The reason the duplicate disposal happens is because the job is first disposed when disposing the DI scope and then explicitly by `ScopedJob`:
https://github.com/quartznet/quartznet/blob/5411524b7c05317b5b895d696e9cf17da5793321/src/Quartz.Extensions.DependencyInjection/Simpl/MicrosoftDependencyInjectionJobFactory.cs#L70-L74

**Version used**

3.2.4

**To Reproduce**

Prepare a test project:
```
dotnet new nunit --name QuartzTests
cd QuartzTests
dotnet add package Quartz.Extensions.DependencyInjection --version 3.2.4
dotnet add package Microsoft.Extensions.DependencyInjection
dotnet add package Microsoft.Extensions.Logging
```

Set UnitTest1.cs to:
```csharp
using System;
using System.Threading;
using System.Threading.Tasks;
using Microsoft.Extensions.DependencyInjection;
using Microsoft.Extensions.Options;
using NUnit.Framework;
using Quartz;

namespace QuartzTests
{
    public class Tests
    {
        [Test]
        public async Task Test1()
        {
            var services = new ServiceCollection();
            services.AddLogging();
            services.AddQuartz(cfg =>
            {
                cfg.UseMicrosoftDependencyInjectionScopedJobFactory();
                cfg.ScheduleJob<MyJob>(t => { });
            });
            var sp = services.BuildServiceProvider();
            var schedulerFactory = sp.GetRequiredService<ISchedulerFactory>();
            var scheduler = await schedulerFactory.GetScheduler();

            await scheduler.Start();
            await scheduler.Shutdown(waitForJobsToComplete: true);

            Assert.AreEqual(1, MyJob.ExecuteCount);
            // FAILS, ACTUAL = 2
            Assert.AreEqual(1, MyJob.DisposeCount);
        }

        class MyJob : IJob, IDisposable
        {
            static int _executeCount;
            static int _disposeCount;

            public Task Execute(IJobExecutionContext context)
            {
                Interlocked.Increment(ref _executeCount);
                return Task.CompletedTask;
            }

            public void Dispose()
            {
                Interlocked.Increment(ref _disposeCount);
            }

            public static int ExecuteCount => _executeCount;

            public static int DisposeCount => _disposeCount;
        }
    }
}
```

**Expected behavior**

Job should be disposed only once.

ArturDorochowicz:
This PR changes `MicrosoftDependencyInjectionJobFactory` to stop resolving jobs from the container and instead creates them with `ActivatorUtilities` (i.e. job dependencies are resolved from the container, the job is newed up).

Job types should no longer be registered in the container. `ScheduleJob` extension is changed to no longer register the job type.

This change fixes double disposal of `IDisposable` jobs (#1120). Prior to this change, `IDisposable` jobs were disposed twice: by the container as well as by the job factory (`ScopedJob`). And when there was no scope (`CreateScope=false`), the container kept all job instances until the container itself was disposed, creating a resource leak.
 
`AllowDefaultConstructor` option is now redundant and has been removed.

Fixes #1120

### DIFF
--- a/src/Quartz.Extensions.DependencyInjection/JobFactoryOptions.cs
+++ b/src/Quartz.Extensions.DependencyInjection/JobFactoryOptions.cs
@@ -1,7 +1,15 @@
+using System;
+
 namespace Quartz
 {
     public class JobFactoryOptions
     {
+        /// <summary>
+        /// When DI has not been configured with the job type, should the default no-arg public constructor be tried.
+        /// </summary>
+        [Obsolete("The value is ignored as all jobs are created with ActivatorUtilities.CreateInstance (see https://github.com/quartznet/quartznet/issues/1120).")]
+        public bool AllowDefaultConstructor { get; set; }
+
         /// <summary>
         /// Whether to use scopes when building job instances, enables injection of scoped services like
         /// Entity Framework's DbContext.
--- a/src/Quartz.Extensions.DependencyInjection/JobFactoryOptions.cs
+++ b/src/Quartz.Extensions.DependencyInjection/JobFactoryOptions.cs
@@ -3,11 +3,6 @@ namespace Quartz
     public class JobFactoryOptions
     {
         /// <summary>
-        /// When DI has not been configured with the job type, should the default no-arg public constructor be tried.
-        /// </summary>
-        public bool AllowDefaultConstructor { get; set; }
-
-        /// <summary>
         /// Whether to use scopes when building job instances, enables injection of scoped services like
         /// Entity Framework's DbContext.
         /// </summary>
--- a/src/Quartz.Extensions.DependencyInjection/Simpl/JobActivatorCache.cs
+++ b/src/Quartz.Extensions.DependencyInjection/Simpl/JobActivatorCache.cs
@@ -5,7 +5,7 @@ using Microsoft.Extensions.DependencyInjection;
 
 namespace Quartz.Simpl
 {
-    public class JobActivatorCache
+    internal class JobActivatorCache
     {
         private readonly ConcurrentDictionary<Type, ObjectFactory> activatorCache = new ConcurrentDictionary<Type, ObjectFactory>();
         private readonly Func<Type, ObjectFactory> createFactory = type => ActivatorUtilities.CreateFactory(type, Type.EmptyTypes);
--- a/src/Quartz.Extensions.DependencyInjection/Simpl/JobActivatorCache.cs
+++ b/src/Quartz.Extensions.DependencyInjection/Simpl/JobActivatorCache.cs
@@ -1,0 +1,29 @@
+using System;
+using System.Collections.Concurrent;
+
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Quartz.Simpl
+{
+    public class JobActivatorCache
+    {
+        private readonly ConcurrentDictionary<Type, ObjectFactory> activatorCache = new ConcurrentDictionary<Type, ObjectFactory>();
+        private readonly Func<Type, ObjectFactory> createFactory = type => ActivatorUtilities.CreateFactory(type, Type.EmptyTypes);
+
+        public IJob CreateInstance(IServiceProvider serviceProvider, Type jobType)
+        {
+            if (serviceProvider is null)
+            {
+                throw new ArgumentNullException(nameof(serviceProvider));
+            }
+
+            if (jobType is null)
+            {
+                throw new ArgumentNullException(nameof(jobType));
+            }
+
+            var factory = activatorCache.GetOrAdd(jobType, createFactory);
+            return (IJob)factory(serviceProvider, null);
+        }
+    }
+}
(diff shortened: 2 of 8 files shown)


lahma:
Thank you for doing the hard work and implementing the changes, I probably need some time go through these and test a bit. Great work!

src/Quartz.Extensions.DependencyInjection/JobFactoryOptions.cs
```diff
@@ -2,11 +2,6 @@ namespace Quartz
 {
     public class JobFactoryOptions
     {
-        /// <summary>
-        /// When DI has not been configured with the job type, should the default no-arg public constructor be tried.
-        /// </summary>
-        public bool AllowDefaultConstructor { get; set; }
```
lahma, line 8: I think for 3.3 release it would be nicer to mark this as obsolete and and comment that no longer used, maybe link to GH issue?

src/Quartz.Extensions.DependencyInjection/Simpl/JobActivatorCache.cs
```diff
@@ -0,0 +1,29 @@
+using System;
+using System.Collections.Concurrent;
+
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Quartz.Simpl
+{
+    public class JobActivatorCache
```
lahma, line 8: Can we make this internal? I'd rather hide everything we can and only open when request with good reasoning comes in. Makes it easier to move the cheese.

src/Quartz.Extensions.DependencyInjection/Simpl/MicrosoftDependencyInjectionJobFactory.cs
```diff
@@ -15,13 +15,16 @@ public class MicrosoftDependencyInjectionJobFactory : PropertySettingJobFactory
     {
         private readonly IServiceProvider serviceProvider;
         private readonly IOptions<QuartzOptions> options;
+        private readonly JobActivatorCache activatorCache;
 
         public MicrosoftDependencyInjectionJobFactory(
             IServiceProvider serviceProvider,
-            IOptions<QuartzOptions> options)
+            IOptions<QuartzOptions> options,
+            JobActivatorCache activatorCache)
```
lahma, line 23: OK this probably makes it harder to change JobActivatorCache internal

lahma (changes requested):
Looking good. I think we should just obsolete the property not to break things too badly for people. Though the MS DI integration already has public constructor change...

You showed some test cases in #1120 , could they be added to test suite?

src/Quartz.Extensions.DependencyInjection/Simpl/MicrosoftDependencyInjectionJobFactory.cs
ArturDorochowicz: Since the job factory is registered as singleton anyway, I can either fully inline JobActivatorCache into it or new up the cache in the job factory's ctor/field initializer and then make it internal. Thoughts?

ArturDorochowicz:
> You showed some test cases in #1120 , could they be added to test suite?

That was mostly a demonstration of the issue, dressed up as (integration?) tests. I wasn't even sure they were deterministic. Specifically, does this deterministically execute the scheduled job?
```csharp
await scheduler.Start();
await scheduler.Shutdown(waitForJobsToComplete: true);
```

If you want to include this anyway, please advise were should I put it, as there doesn't seem to be an existing test project that touches this area.

src/Quartz.Extensions.DependencyInjection/Simpl/MicrosoftDependencyInjectionJobFactory.cs
lahma: I think that's a good idea 👍🏻

lahma:
I can check the tests when I'll do some testing and docs later on, so no need for now.

src/Quartz.Extensions.DependencyInjection/Simpl/JobActivatorCache.cs
ArturDorochowicz: It's internal now.

src/Quartz.Extensions.DependencyInjection/JobFactoryOptions.cs
ArturDorochowicz: Added Obsolete attribute.

lahma (approved):

lahma:
Great work! Thanks for hanging in there and tweaking 🚀

ArturDorochowicz:
@lahma Thanks for merging. I've just realized that I missed one place where job is registered in the container - in AddJob. The xml doc comment should be changed too.

https://github.com/quartznet/quartznet/blob/1e82175be0e60fca3199269b9f7bb6384f60a59f/src/Quartz.Extensions.DependencyInjection/ServiceCollectionExtensions.cs#L111-L139

PR #1130, opened 2021-03-16
Remove redundant DI job registration from AddJob

ArturDorochowicz:
This removes redundant DI job registration which was missed in #1127 .


lahma (approved):

lahma:
Thank you!

PR #1159, opened 2021-04-09
Try resolving jobs from service provider before resorting to ActivatorUtilities

### LINKED ISSUE #1157: Can no longer register jobs via DI
See https://github.com/openiddict/openiddict-core/issues/1243#issuecomment-816019916 . Ideally we can support both discovery mechanisms.

lahma:
fixes #1157

### DIFF
--- a/src/Quartz.Extensions.DependencyInjection/ServiceCollectionQuartzConfigurator.cs
+++ b/src/Quartz.Extensions.DependencyInjection/ServiceCollectionQuartzConfigurator.cs
@@ -61,15 +61,10 @@ namespace Quartz
             UseJobFactory<MicrosoftDependencyInjectionJobFactory>(configure);
         }
 
+        [Obsolete("Jobs are always created with some")]
         public void UseMicrosoftDependencyInjectionScopedJobFactory(Action<JobFactoryOptions>? configure = null)
         {
-            void ScopeConfiguration(JobFactoryOptions options)
-            {
-                options.CreateScope = true;
-                configure?.Invoke(options);
-            }
-
-            UseJobFactory<MicrosoftDependencyInjectionJobFactory>(ScopeConfiguration);
+            UseJobFactory<MicrosoftDependencyInjectionJobFactory>(configure);
         }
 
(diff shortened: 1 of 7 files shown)


kevinchalet:
Should `(InnerJob as IDisposable)?.Dispose();` be behind a `if` to prevent double-disposal if the service comes from DI? (when it's created using `ActivatorUtilities`, it must be disposed of manually)

lahma:
> Should `(InnerJob as IDisposable)?.Dispose();` be behind a `if` to prevent double-disposal if the service comes from DI? (when it's created using `ActivatorUtilities`, it must be disposed of manually)

Good catch, added check for job's origin and only disposing if it came from ? ActivatorUtilities`.

ArturDorochowicz:
Looks good.

src/Quartz.Extensions.DependencyInjection/ServiceCollectionQuartzConfigurator.cs
```diff
@@ -61,15 +61,10 @@ public void UseMicrosoftDependencyInjectionJobFactory(Action<JobFactoryOptions>?
             UseJobFactory<MicrosoftDependencyInjectionJobFactory>(configure);
         }
 
+        [Obsolete("Jobs are always created with some")]
```
ArturDorochowicz, line 64: Did you mean `... with scope`?

lahma: Thanks, working in parallel with two bug fixes seems to freeze the brain. Pushed fix commit.

lahma:
I think we have the pieces together to patch the DI resolution support.. @kevinchalet and @ArturDorochowicz  are you happy with this so I can merge? Thank you both for helping out and spotting problems.

ArturDorochowicz:
@lahma I think it addresses all concerns. Ship it! :)
Thank you.

kevinchalet:
@lahma looks awesome, thanks for fixing it! 👏🏻

PR #1160, opened 2021-04-09
Simplify DNS name resolution in HostNameBasedIdGenerator

### LINKED ISSUE #1158: Can't get hostname on WSL2+docker host network
Hi, this is a rather obscure bug 🙂 

When running Quartz using the SimpleInstanceIdGenerator on WSL2 + docker + host network, it fails to get the current hostname and crashes when initializing the scheduler factory.

I've created a repro in https://github.com/mausch/QuartzNetHostAddressTest

The exception is "Cannot run without an instance id." thrown at https://github.com/quartznet/quartznet/blob/a22915a9abac1568accb93eb24b4cce5331c8249/src/Quartz/Impl/StdSchedulerFactory.cs#L930 .

BTW it seems that throw should carry the inner exception i.e. make it:
```
throw new InvalidOperationException("Cannot run without an instance id.", e);
```

The way I propose to fix this is to simplify https://github.com/quartznet/quartznet/blob/2fccbc1a17c3fc1bd6f7433f8aedc2c541a1fa46/src/Quartz/Simpl/HostNameBasedIdGenerator.cs#L61-L62

Not sure why it's doing all that DNS resolution just to get the hostname? Couldn't that be just a call to `Dns.GetHostName()` ? I looked through the history of that file but couldn't figure out why, maybe someone else knows!

Many thanks,
Mauricio

lahma:
fixes #1158

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -8,9 +8,12 @@ This release returns the possibility to resolve jobs from Microsoft DI container
 ActivatorUtilities is used to construct the type with constructor injection support. Now both AllowDefaultConstructor and CreateScope have
 been obsoleted as behavior is now either via DI construction or ActivatorUtilities and scope is always created to prevent resource leaks / double disposal.
 
+Also a problem with host name resolution under WSL2 scenario was fixed.
+
 * FIXES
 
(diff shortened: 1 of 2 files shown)


mausch:
:+1:

PR #1181, opened 2021-04-21
Restore System.Data.SqlClient support on .NET Core

devigo:
Added the ability to optionally use `System.Data.SqlClient` on .NET Core (https://github.com/quartznet/quartznet/discussions/950#discussioncomment-415731)

If we want to use `System.Data.SqlServer` in .NET Core we can specify it explicitly with this parameter
`quartz.dataSource.myDS.provider = SystemDataSqlClient`


lahma (approved):

lahma:
Looks good, thanks for also remembering to update the docs.

PR #1185, opened 2021-04-23
Use custom InstantiateType for all instantiations in StdSchedulerFactory

miegir:
Use InstantiateType that can be overloaded to allow custom instantiation of plugins, job listeners and trigger listeners. When used with Quartz.AspNetCore, it will allow to use dependency injection in plugins and other types.


miegir:
I have no idea whiy that test is failed. On my local machine, it and many ither tests failed even without changes.

lahma:
It's not the most rewarding thing trying to have CI system for a threaded / timing critical library 😄 

PR #1186, opened 2021-04-24
Update StdAdoConstants.cs

chrisedebo:
Seemed like these might have been missed at some point. Not sure what to do about SIMPROP_TRIGGERS


lahma:
Thank you!

PR #1189, opened 2021-04-30
Added a configuration point to the private Scope of MicrosoftDependencyInjectionJobFactory

arielmoraes:
This PR aims to enable a configuration point to the private `IServiceScope` created by the `MicrosoftDependencyInjectionJobFactory`

With that solution in place we could inherit the `MicrosoftDependencyInjectionJobFactory` and configure scoped types based on Job Data or another property from the ambient context.

### DIFF
--- a/src/Quartz.Extensions.DependencyInjection/Simpl/MicrosoftDependencyInjectionJobFactory.cs
+++ b/src/Quartz.Extensions.DependencyInjection/Simpl/MicrosoftDependencyInjectionJobFactory.cs
@@ -30,14 +30,16 @@ namespace Quartz.Simpl
             //  Generate a scope for the job, this allows the job to be registered
             //	using .AddScoped<T>() which means we can use scoped dependencies 
             //	e.g. database contexts
-            var scope = ConfigureScope(bundle, scheduler);
+            var scope = serviceProvider.CreateScope();
+            ConfigureScope(scope, bundle, scheduler);
             var (job, fromContainer) = CreateJob(bundle, scope.ServiceProvider);
             return new ScopedJob(scope, job, canDispose: !fromContainer);
         }
 
-        protected virtual IServiceScope ConfigureScope(TriggerFiredBundle bundle, IScheduler scheduler)
+        protected virtual void ConfigureScope(IServiceScope scope, TriggerFiredBundle bundle, IScheduler scheduler)
         {
-            return serviceProvider.CreateScope();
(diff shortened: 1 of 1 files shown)


src/Quartz.Extensions.DependencyInjection/Simpl/MicrosoftDependencyInjectionJobFactory.cs
```diff
@@ -30,11 +30,16 @@ protected override IJob InstantiateJob(TriggerFiredBundle bundle, IScheduler sch
             //  Generate a scope for the job, this allows the job to be registered
             //	using .AddScoped<T>() which means we can use scoped dependencies 
             //	e.g. database contexts
-            var scope = serviceProvider.CreateScope();
+            var scope = ConfigureScope(bundle, scheduler);
             var (job, fromContainer) = CreateJob(bundle, scope.ServiceProvider);
             return new ScopedJob(scope, job, canDispose: !fromContainer);
         }
 
+        protected virtual IServiceScope ConfigureScope(TriggerFiredBundle bundle, IScheduler scheduler)
```
lahma, line 38: I think it would be more understandable API if it's either `void ConfigureScope(IServiceScope scope)` which configures one already created or another option is  `IServiceScope CreateScope(IServiceProvider serviceProvider)` what do you think? CreateScope could be parameterless if services are not needed.

arielmoraes: The idea is to allow the configuration to be based on the job context. For example, I have a multi tenant application and I set the TenantId as a Job Data, so to configure the scope I need access to the context.

arielmoraes: Indeed the scope could be created and passed along with the context, that would avoid dispose problems for the caller.

lahma: So would you like to refine the solution still?

arielmoraes: I've just pushed the changes

lahma:
Great, looking logical and good now. Thank you for your contribution!

PR #1244, opened 2021-07-14
Update job-stores.md

davemcdermid:
Fixed casing for SqlServerDelegate to prevent type-load errors if people copy/paste the name


lahma:
Good catch, thank you!

Issue #1211, opened 2021-05-28
ScheduleJob shorthand: Job name should match trigger name by default


Timovzl:
**Describe the bug**

[ScheduleJob](https://www.quartz-scheduler.net/2020/10/01/quartznet-3-2-released/#schedulejob) is documented as a shorthand for `AddJob` plus `AddTrigger`, and it is wonderful.

Configuring the trigger is mandatory. Configuring the job is optional. It is, by far, the most readable to omit the job configuration - even the example in the docs omits it.

**However, if no job name is explicitly specified, a UUID is generated for it.** This seems like a poor default. Consider the example from the docs:

```cs
q.ScheduleJob<ExampleJob>(trigger => trigger
    .WithIdentity("Combined Configuration Trigger")
    .StartAt(DateBuilder.EvenSecondDate(DateTimeOffset.UtcNow.AddSeconds(7)))
    .WithDailyTimeIntervalSchedule(x => x.WithInterval(10, IntervalUnit.Second))
    .WithDescription("my awesome trigger configured for a job with single call")
);
```

The trigger name will be "Combined Configuration Trigger". What should the job name be? I would infer that, considering what the shorthand was designed for in the first place, **the job name should match the trigger name by default**.

**Version used**

3.3.2.

**To Reproduce**

Simply implement the [example](https://www.quartz-scheduler.net/2020/10/01/quartznet-3-2-released/#schedulejob) from the docs.

**Expected behavior**

The job name (in the backing store) should equal the trigger name.

**Additional Context**

The only workaround we currently have is significantly uglier than the example, and seems to somewhat cancel out the benefits of the shorthand:

```cs
q.ScheduleJob<ExampleJob>(trigger => trigger
    .WithIdentity("Combined Configuration Trigger")
    .StartAt(DateBuilder.EvenSecondDate(DateTimeOffset.UtcNow.AddSeconds(7)))
    .WithDailyTimeIntervalSchedule(x => x.WithInterval(10, IntervalUnit.Second))
    .WithDescription("my awesome trigger configured for a job with single call"),
    job => job.WithIdentity("Combined Configuration Trigger")
);
```

More importantly, it violates the DRY (Don't Repeat Yourself) principle. We could store the job name in a variable or constant, but that would require even more code.

lahma:
Thanks for the feedback, this is quite logical but also a breaking change, but again with GUID ids it doesn't make much of a breaking change until someone clashes with existing registration. Would you like to make a PR for this?

[...]

PR #1260, opened 2021-07-29
ScheduleJob(trigger => trigger.WithIdentity()): Set job identity to match

Timovzl:
Implements https://github.com/quartznet/quartznet/issues/1211.

`ScheduleJob(trigger => trigger.WithIdentity(...))`, without a job configurator, now sets the job identity to match the trigger's (instead of `"DEFAULT-[UUID]"`).


lahma (approved):
Great work, kudos for the test case.

Issue #1259, opened 2021-07-28
Feature Request: Built-in Recurrence Rule Trigger


tomXGames:
**Problem**
I am trying to scheduled tasks, but I only have iCalendar recurrence rules describing how they repeat. Quartz.NET currently doesn't support these rules. This is especially frustrating as there are lots of patterns that can't be expressed as CRON expression but can be expressed as recurrence rules. I have tried writing a converter that goes from recurrence rule to cron, but as stated above there are loads of cases where you just can't express the pattern as a CRON expression.

**Solution**
It would be nice to have a simple trigger like where you could just pass in a recurrence rule and it just works. There was a [pull request](https://github.com/quartz-scheduler/quartz/pull/153) for the regular Java Quartz Scheduler that wanted to add support for RRules, but is has been closed a while ago.

lahma:
So would this be something that you would like to champion and possibly implement? If third-party library is needed I guess a separate project/NuGet package would be in order.

tomXGames:
I could take a look at it and see if I could implement it, but I think it's possible. I will fork the repo and check it asap. I'll keep you updated!

tomXGames:
So I took a look at it and I think this shouldn't be too hard. I was wondering though I could use an existing RecurrenceRule class so I don't have to fight with the parsing. I could use [this ](https://github.com/EWSoftware/PDI/blob/master/Source/EWSPDI/Recurrence.cs) class by @EWSoftware. But if you prefer I implement this from scratch, I could also do that.

lahma:
Existing work is OK as it seems that it has permissive license, as long as headers maintained (I'm not a lawyer though) - probably should be internal classes on Quartz side if brought as source. Source files has the benefit of now having dependency which would require separate NuGet package.

EWSoftware:
My PDI project is licensed using Ms-PL so you can reuse the code here if you like.  Just keep the header comments intact with the license notice and link to the original project.  I don't have any problems with that.

lahma:
@EWSoftware thank you for the clarification, great to hear that this is something that we can do. Copyright headers and attributions will be intact if used. Great work by the way 👍🏻

tomXGames:
I took a look at the code base and I will implement it asap (probably in the course of the next week). Thanks to both for the input!

PR #1276, opened 2021-08-15
Add RecurrenceRule-Trigger 

tommasopeduzzi:
This has been discussed in this issue: #1259 
This is still very much a WIP and I am happy to make any changes.
List of changes:
- Added the Recurrence-class with all it's dependencies to Impl/PDI
- Added RecurrenceRuleTrigger Interface as well as Implementation as a sort of interface between the scheduler and the Recurrence class
- Added RecurrenceScheduleBuilder for easy creation of the trigger
- Added tests testing the RecurrenceScheduleBuilder.

### DIFF
--- a/src/Quartz.Tests.Unit/RecurrenceScheduleBuilderTest.cs
+++ b/src/Quartz.Tests.Unit/RecurrenceScheduleBuilderTest.cs
@@ -1,0 +1,43 @@
+﻿using System;
+using System.Collections.Generic;
+using NUnit.Framework;
+using Quartz.Impl.Triggers;
+
+namespace Quartz.Tests.Unit
+{
+    /// <summary>
+    /// Unit tests for <see cref="RecurrenceScheduleBuilder"/>
+    /// </summary>
+    /// <author>Tommaso Peduzzi</author>
+    [TestFixture]
+    public class RecurrenceScheduleBuilderTest
+    {
+        [Test]
+        public void TestWeekly()
+        {
+            var weekDays = new List<DayOfWeek>();
+            weekDays.Add(DayOfWeek.Monday);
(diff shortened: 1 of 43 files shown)


lahma:
@tommasopeduzzi sorry it's been such a long time that it has taken me to address this.

I had some concerns like hiding the implementation as internal (anything from the 3rd party library), which I tried to push to your branch, but as you've made the PR from main - I cannot do that. Ideally a PR would come from a separate branch to keep main "clean".

Second things I noticed that the PDI library expect local times and Quartz always uses UTC times as base. I'm not sure if the conversions are handled properly - the trigger doesn't have concept of current time zone like cron trigger does. What are your thoughts?

tommasopeduzzi:
Hello @lahma! No problem!

You're right about the git mess, I am going to make a new branch and cherrypick them over. Is that a valid solution to the problem?

I could probably change the Recurrence-Class in such a way, that it converts all the local times to UTC Time on assignment, that shouldn't be too hard and would be an easy fix to the problem... 

Writing a new Recurrence class would also be an option and shouldn't be too hard, based on the [specifications](https://icalendar.org/iCalendar-RFC-5545/3-8-5-3-recurrence-rule.html). Is that a better solution?

lahma:
I'm quite fine with any solution you find plausible for you to work with. Using the current libraries is fine, but just the concern about how "local time" behaves if it cannot be defined (parametrizable time zone). Generally triggers and time zones can get a bit tricky, so bugs might be ahead, but it's understandable for a new feature not to be 100% bullet proof in v1.

lahma:
Thanks for the PR, this was adapted as new feature with different implementation and was released.

PR #1286, opened 2021-08-28
fix for job type loading after version change (#1283)

### LINKED ISSUE #1283: SimpleTypeLoader fails on generic jobs after version change
**Describe the bug**

When using generic job classes, after version change it fails storing/retrieving triggers/jobs:
```csharp
Couldn't store trigger 'CreateProviderOrdersJob`1.db9ec53b-5eaf-409c-8765-c90379a3d89d' for 'CreateProviderOrdersJob`1.MyJobSubType' job: Couldn't retrieve job because a required type was not found: Could not load type 'Infrastructure.MyJob`1[[Abstractions.MyJobSubType, Abstractions, Version=1.0.226.0, Culture=neutral, PublicKeyToken=null]], Infrastructure'
```
This happens because FullName from generic includes version:
`Quartz.Tests.Unit.JobDetailTest+GenericJob`1[[Quartz.Tests.Unit.JobDetailTest+IJobSubType, Quartz.Tests.Unit, Version=3.3.3.0, Culture=neutral, PublicKeyToken=f6b8c98a402cc8a4]]`

This method is used both when storing jobs/riggers and loading back jobs from store. So if you register jobs with 1.0 version, when you deploy 1.1 you need to re-register job under same name but with new version in type string and then restart all the failed (because of job load error) triggers.

**Version used**

Checked on 3.3.3 version also

**To Reproduce**

```
        [Test]
        public void GenericJobTypeShouldBeLoadable()
        {
            var type = typeof(GenericJob<IJobSubType>);
            var typeString = type.AssemblyQualifiedNameWithoutVersion();
            var loadedType = new SimpleTypeLoadHelper().LoadType(typeString);

            Assert.That(typeString, Is.Not.Contains(", Version="));
            
            Assert.That(loadedType, Is.Not.Null);
            Assert.That(loadedType, Is.EqualTo(type));
        }

        public class GenericJob<T> : IJob
        {
            public Task Execute(IJobExecutionContext context) => Task.CompletedTask;
        }
        
        public interface IJobSubType { }
```

**Expected behavior**

After changing version of assembly containing job class, loading of jobs and storing of triggers should not fail.

**Additional context**

Fixed this behavior modifying ObjectExtensions like this:
```
public static class ObjectExtensions
    {
        private static readonly ConcurrentDictionary<Type, string> assemblyQualifiedNameCache = new ConcurrentDictionary<Type, string>();
        private static readonly Regex cleanup = new Regex(", (Version|Culture|PublicKeyToken)=[0-9.\\w]+", RegexOptions.Compiled);
        
        public static string AssemblyQualifiedNameWithoutVersion(this Type type)
            => assemblyQualifiedNameCache.GetOrAdd(type, x => GetTypeString(x) + ", " + x.GetTypeInfo().Assembly.GetName().Name);

        public static string? GetTypeString(Type type)
            => type.IsGenericType
                ? GenericTypeString(type.FullName)
                : type.FullName;
        
        public static string? GenericTypeString(string? name)
            => string.IsNullOrEmpty(name)
                ? null
                : cleanup.Replace(name, "");
    }
```
This simplifies type string representation, removing version/culture/token information but still allows Type.Load() to successfully discover and load that type.

justmara:
fixes #1283

### DIFF
--- a/src/Quartz/Util/ObjectExtensions.cs
+++ b/src/Quartz/Util/ObjectExtensions.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Concurrent;
 using System.Reflection;
+using System.Text.RegularExpressions;
 
 namespace Quartz.Util
 {
@@ -10,8 +11,19 @@ namespace Quartz.Util
     public static class ObjectExtensions
     {
         private static readonly ConcurrentDictionary<Type, string> assemblyQualifiedNameCache = new ConcurrentDictionary<Type, string>();
-
+        private static readonly Regex cleanup = new Regex(", (Version|Culture|PublicKeyToken)=[0-9.\\w]+", RegexOptions.Compiled);
+        
         public static string AssemblyQualifiedNameWithoutVersion(this Type type)
-            => assemblyQualifiedNameCache.GetOrAdd(type, x => x.FullName + ", " + x.GetTypeInfo().Assembly.GetName().Name);
+            => assemblyQualifiedNameCache.GetOrAdd(type, x => GetTypeString(x) + ", " + x.GetTypeInfo().Assembly.GetName().Name);
+
+        public static string? GetTypeString(Type type)
+            => type.IsGenericType
+                ? GenericTypeString(type.FullName)
+                : type.FullName;
+        
+        public static string? GenericTypeString(string? name)
+            => string.IsNullOrEmpty(name)
+                ? null
+                : cleanup.Replace(name, "");
     }
 }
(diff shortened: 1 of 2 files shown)


src/Quartz/Util/ObjectExtensions.cs
```diff
@@ -10,8 +11,19 @@ namespace Quartz.Util
...
+            => assemblyQualifiedNameCache.GetOrAdd(type, x => GetTypeString(x) + ", " + x.GetTypeInfo().Assembly.GetName().Name);
+
+        public static string? GetTypeString(Type type)
+            => type.IsGenericType
+                ? GenericTypeString(type.FullName)
+                : type.FullName;
+        
+        public static string? GenericTypeString(string? name)
+            => string.IsNullOrEmpty(name)
+                ? null
```
lahma, line 26: Should this return name instead of null?

lahma:
Thank you!

PR #1298, opened 2021-09-13
Doc typo

kolpav:
https://github.com/quartznet/quartznet/blob/29ba6a0fb6a1d9fb0ec36480a4cd34c3d99e7105/src/Quartz.Plugins/Plugin/History/LoggingTriggerHistoryPlugin.cs#L29-L31

https://github.com/quartznet/quartznet/blob/29ba6a0fb6a1d9fb0ec36480a4cd34c3d99e7105/src/Quartz.Plugins/Plugin/Management/ShutdownHookPlugin.cs#L28-L30


lahma:
Thank you!

PR #1314, opened 2021-09-27
Fix StartDelay delaying the start of other hosted services

### LINKED ISSUE #1313: StartDelay option delays web application start and not just the scheduler
**Describe the bug**

There's possibly an issue with `QuartzHostedServiceOptions.StartDelay` option which was recently introduced in 3.3.3 (via #1166).

If this value is set, all `IHostedService`s after `QuartzHostedService` are delayed. In a typical Asp.Net Core application this will include the web host itself. I don't think this is the desired result.
 
I looked at #1166 and saw that it ultimately is about awaiting `QuartzScheduler.StartDelayed`. I was surprised to learn that the implementation of `StartDelayed` is a `Task.Delay` followed by a `Start`. I was somehow expecting a regular start and then a delay in scheduling, if it makes sense.

I hope I can offer some critique of the current implementation of `QuartzScheduler.StartDelayed`. If it is the desired behavior then I think such api is redundant and leads to confusion. It's redundant, because developers are perfectly capable of adding a delay on their own before calling `Start`. As to confusion, I think that the comment in #1166: 
```csharp
// when we need to init another IHostedServices first
options.StartDelay = TimeSpan.FromSeconds(10);
```
shows that the author of that PR made the same assumption I would make and thought that `StartDelayed` would return normally (quickly) and only somehow delay the scheduling (though that's wrong thinking still as it's prone to races; the correct approach would be to simply register other hosted services before Quartz)     

**Version used**

3.3.3

**To Reproduce**

**Expected behavior**

`StartDelay` should not cause a delay in starting other hosted services.

ArturDorochowicz:
QuartzScheduler.StartDelay is changed not to wait on the background Task anymore.  Returning a task is preserved so that the API does not change.

Update TestStartDelayed test to take less time. This test is more non-deterministic now.

Fixes #1313

### DIFF
--- a/src/Quartz.Extensions.Hosting/QuartzHostedService.cs
+++ b/src/Quartz.Extensions.Hosting/QuartzHostedService.cs
@@ -6,10 +6,11 @@ using Microsoft.Extensions.Options;
 
 namespace Quartz
 {
-    internal class QuartzHostedService : BackgroundService
+    internal class QuartzHostedService : IHostedService
     {
         private readonly ISchedulerFactory schedulerFactory;
         private readonly IOptions<QuartzHostedServiceOptions> options;
+        private IScheduler scheduler = null!;
 
         public QuartzHostedService(
             ISchedulerFactory schedulerFactory,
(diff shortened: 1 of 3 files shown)


lahma (approved):
Great, looks good to my eye.

PR #1375, opened 2021-10-27
JobKey, TriggerKey and Key<T> are now immutable

drieseng:
Made **JobKey**, **TriggerKey** and **Key\<T>** immutable, and therefore **Key\<T>** no longer defines a default protected ctor. 

The implementation of `Equals(object? obj)` has simplified due to this change, but the positive effect on micro-benchmarks (that I added in this PR) is within noise range.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -16,6 +16,8 @@
 		* FullName
 
 	* Triggers can no longer be constructed with a **null** group name (#1359)
+	
+	* **JobKey**, **TriggerKey** and **Key\<T>** are now immutable. **Key\<T>** therefore no longer has a default protected ctor.
 
 ## Release 3.3.3, Aug 1 2021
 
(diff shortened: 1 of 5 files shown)


lahma:
@drieseng — you were right, and this is in. I am closing it as superseded rather than merging it,
because the branch it targets (`v4`) stopped moving in March 2022; 4.x development went to `main`.

On `main` today, `Key<T>` lives at `src/Quartz/Key.cs` and reads:

```csharp
private readonly string name;
private readonly string group;

public string Name => name;
public string Group => group;
```

No `protected Key()`, no setters, and `Equals` is the one-line comparison you wanted. `JobKey` and
`TriggerKey` are both `sealed`, with only the `(name)` and `(name, group)` constructors. It landed in
`8300b61de` (#3210) during the 4.0 API pass, on your reasoning: a key is a value and a dictionary key
in every job store, so it has no business being mutable.

Your open question from this thread got answered there too — serialization is handled by an
immutable-key contract resolver on the Newtonsoft side, with `KeySerializationTest` pinning that both
serializers rebuild keys through the constructor. And your `KeyBenchmark` from #1367 is still in the
tree.

Sorry it took the long way round, and thank you for the push.

PR #1387, opened 2021-11-04
Simplify TaskExtensions.IsCompletedSuccessfully(this Task t)

drieseng:
In `TaskExtensions.IsCompletedSuccessfully(this Task t)` we currently consider the task to be completed successfully if all of the following conditions are met:
* **Status** is **RanToCompletion**.
* **IsFaulted** is **false**.
* **IsCanceled** is **false**.

These last two conditions are not necessary as **Status** is only **RanToCompletion** if the task completed successfully.

I added unit tests for `TaskExtensions.IsCompletedSuccessfully(this Task t)`, but they may require a little tuning to get them to produce stable results.

### DIFF
--- a/src/Quartz.Tests.Unit/TaskExtensionsTest.cs
+++ b/src/Quartz.Tests.Unit/TaskExtensionsTest.cs
@@ -1,0 +1,107 @@
+using NUnit.Framework;
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Quartz.Tests.Unit
+{
+    [TestFixture]
+    public class TaskExtensionsTest
+    {
+        [Test]
+        public void IsCompletedSuccessfully_Created()
+        {
+            var task = new Task(() => Task.Delay(300));
+
+            Assert.False(task.IsCompletedSuccessfully());
+            Assert.AreEqual(TaskStatus.Created, task.Status);
+        }
+
+        [Test]
+        public void IsCompletedSuccessfully_WaitingToRun()
+        {
+            var task = new Task(() => Thread.Sleep(300));
+            task.Start();
+
+            Assert.False(task.IsCompletedSuccessfully());
+            Assert.AreEqual(TaskStatus.WaitingToRun, task.Status);
+        }
+
+        [Test]
+        public void IsCompletedSuccessfully_Running()
+        {
+            var task = new Task(() => Thread.Sleep(300));
+            task.Start();
+
+            task.Wait(50);
+
+            Assert.False(task.IsCompletedSuccessfully());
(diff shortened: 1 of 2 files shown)


lahma:
Should we just keep the old expression and short-circuit for the fast one on .NET Core?

drieseng:
Are you afraid it's not correct?
I was already convinced from the docs (**RanToCompletion** = "The task completed execution successfully."), but added unit tests just to be 100% sure.

lahma:
Mmm.. why would there be new `IsCompletedSuccessfully` if it's the same thing? Looking at the sources it checks a bit different flags.

drieseng:
These have an underlying flags enum (called **m_stateFlags**). In that enum you **can** have the **RanToCompletion** flag set in a combination with **Canceled** or **Faulted**. 

Which is why in **Status**, they have the following logic:

```cs
if ((sf & (int)TaskStateFlags.Faulted) != 0) rval = TaskStatus.Faulted;
else if ((sf & (int)TaskStateFlags.Canceled) != 0) rval = TaskStatus.Canceled;
else if ((sf & (int)TaskStateFlags.RanToCompletion) != 0) rval = TaskStatus.RanToCompletion;
...
```

You'll only get **RanToCompletion** as **Status** if neither the **Faulted** nor **Canceled** flags where set on the underlying (flags) enum.

The **IsCompletedSuccessfully** property was introduced for convenience, and it should be a little faster since they only perform one check to see if **RanToCompletion** is the only flag - of the **Canceled**, **Faulted** and **RanToCompletion** flags - that is set on the underlying flags enum:

```cs
public bool IsCompletedSuccessfully => (m_stateFlags & (int)TaskStateFlags.CompletedMask) == (int)TaskStateFlags.RanToCompletion;
```

lahma:
@drieseng — you were right about the semantics, and the thread is a good read: `Status` resolves
`Faulted` and `Canceled` ahead of `RanToCompletion` out of the same flags, so the two extra checks
never changed an answer. My hesitation at the time was misplaced.

The reason I am closing rather than merging is that the code is gone. `TaskExtensions` was deleted in
`aa5300658` (#2371, May 2024) — `src/Quartz/TaskExtensions.cs` no longer exists anywhere in the
repository. Everything targets `net10.0` now, so the three remaining call sites use the BCL property
directly: `JobRunShell.cs:450`, `QuartzScheduler.cs:2074` and `PropertySettingJobFactory.cs:103`. The
extension existed only to backfill the property on older targets, and there is nothing left to
backfill.

It also targets `v4`, which stopped moving in March 2022; 4.x lives on `main`.

Thank you for chasing this down to the `m_stateFlags` level instead of taking my word for it. That is
the right instinct, and it happened to be the correct call.

PR #1400, opened 2021-11-08
Add unit tests for SimpleTriggerImpl.GetFireTimeAfter(...)

drieseng:
Add unit tests for `SimpleTriggerImpl.GetFireTimeAfter(...)`.

### DIFF
--- a/src/Quartz.Tests.Unit/Impl/Triggers/SimpleTriggerImplTest.cs
+++ b/src/Quartz.Tests.Unit/Impl/Triggers/SimpleTriggerImplTest.cs
@@ -1,0 +1,379 @@
+using NUnit.Framework;
+using Quartz.Impl.Triggers;
+using System;
+
+namespace Quartz.Tests.Unit.Impl.Triggers
+{
+    [TestFixture]
+    public class SimpleTriggerImplTest
+    {
+        private Random _random;
+
+        public SimpleTriggerImplTest()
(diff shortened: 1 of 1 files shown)


src/Quartz.Tests.Unit/Impl/Triggers/SimpleTriggerImplTest.cs
```diff
@@ -0,0 +1,379 @@
+using NUnit.Framework;
+using Quartz.Impl.Triggers;
+using System;
+
+namespace Quartz.Tests.Unit.Impl.Triggers
+{
+    [TestFixture]
```
lahma, line 7: nit: NUnit doesn't require `[TestFixture]` anymore, `[Test]` is enough for discovery

src/Quartz.Tests.Unit/Impl/Triggers/SimpleTriggerImplTest.cs
```diff
@@ -0,0 +1,379 @@
...
+    public class SimpleTriggerImplTest
+    {
+        private Random _random;
+
+        public SimpleTriggerImplTest()
+        {
+            _random = new Random();
```
lahma, line 14: in a CI environment a seed for random might be good, but this way it will show instability though...

lahma (approved):

PR #1404, opened 2021-11-10
Avoid a few interface calls in JobRunShell

drieseng:
Avoids a few interface calls in **JobRunShell**.


drieseng:
@lahma I'm pretty sure the test failure are unrelated.

PR #1408, opened 2021-11-11
Verify scheduler configudation values, validate IdleWaitTime

### LINKED ISSUE #1394: QuartzScheduler: idleWaitTime that is zero (or less) is silently ignored
When **idleWaitTime** is **TimeSpan.Zero** (or less), it is silently ignored by **QuartzScheduler** and, as result, the default idle wait time of 30 seconds is used by **QuartzSchedulerThread**.

I'd propose to throw an **ArgumentException** either when **idleWaitTime** is **TimeSpan.Zero** or less, or - which is what I prefer - only reject **idleWaitTime** if it's negative.

Why reject an **idleWaitTime** that is **TimeSpan.Zero**, but allow one that represents a single tick?

drieseng:
We now verify whether valid values are specified for **IdleWaitTime**, **MaxBatchSize** and **BatchTimeoutWindow**, and no longer replace **TimeSpan.Zero** with a default of **30 seconds** if explicitly set.

This last modification is to allow for scenarios where someone wants a very low latency and precise scheduler.
We should not implicitly / silently change the configuration specified by a user.

<details>
<summary>Detailed changes</summary>
&nbsp;

**StdSchedulerFactory**
* Use default value of **30 seconds** for idle wait time if not explicitly set
* No longer reject value greater than **zero** and less than **1000 milliseconds** for idle wait time.

**DirectSchedulerFactory**
* Throw **ArgumentOutOfRangeException** if **idleWaitTime** is less than **TimeSpan.Zero**, **maxBatchSize** is less than **one** or **batchTimeWindow** is less than **TimeSpan.Zero**.
* Set **IdleWaitTime** property on **QuartzSchedulerResources** instead of passing value in ctor of **QuartzScheduler**.
* Document exceptions.

**ExceptionHelper**
* Add `ThrowArgumentOutOfRangeException(...)` method.

**QuartzScheduler**
* Make **resources** and **schedThread** fields internal for unit testing purposes.
* Remove **idleWaitTime** argument from ctor.

**QuartzSchedulerResources**
* Add **IdleWaitTime** property.
* Throw **ArgumentOutOfRangeException** if value for **BatchTimeWindow** is less than **TimeSpan.Zero**.
* Throw **ArgumentOutOfRangeException** if value for **MaxBatchSize** is less than **one**.

**QuartzSchedulerThread**
* Remove **IdleWaitTime** property, and instead use the **IdleWaitTime** on **QuartzSchedulerResources**.

Update benchmarks to explicitly specifiy 30 seconds for idle wait time, and update for **QuartzScheduler** ctor that no longer takes **idleWaitTime** argument.
Add unit tests.
</details>

Fixes #1394.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -13,10 +13,16 @@
       * JobName
       * JobGroup
       * FullName
-
     * Triggers can no longer be constructed with a **null** group name (#1359)
-
     * The *endUtc* argument of **SimpleTriggerImpl** is no longer nullable.
+    * If a value is explicitly specified for "IdleWaitTime", we will no longer silently ignore the value (and use 
+      a default value of 30 seconds instead) if it's less than or equal to **zero**.
+    * If you use **StdSchedulerFactory** to create a scheduler, we will no longer reject an **IdleWaitTime** that
+      is greater than **zero** but less than **1000 milliseconds**.
+    * An negative value for **IdleWaitTime** or **BatchTimeWindow** will no longer be accepted.
+    * For **MaxBatchSize**, a value less than or equal to **zero** will be rejected.
+    * The ctor for **QuartzScheduler** no longer takes an **idleWaitTime** argument. This value
+      is now obtained from a newly introduced **IdleWaitTime** property on **QuartzSchedulerResources**.
 
 ## Release 3.3.3, Aug 1 2021
 
--- a/src/Quartz.Benchmark/JobRunShellBenchmark.cs
+++ b/src/Quartz.Benchmark/JobRunShellBenchmark.cs
@@ -51,11 +51,12 @@ namespace Quartz.Benchmark
                 InstanceId = instanceId,
                 ThreadPool = new DefaultThreadPool { MaxConcurrency = threadCount },
                 JobStore = new NoOpJobStore(),
+                IdleWaitTime = TimeSpan.FromSeconds(30),
                 MaxBatchSize = threadCount,
                 BatchTimeWindow = TimeSpan.Zero
             };
 
-            return new QuartzScheduler(res, TimeSpan.Zero);
+            return new QuartzScheduler(res);
         }
 
         private TriggerFiredBundle CreateTriggerFiredBundle()
--- a/src/Quartz.Benchmark/QuartSchedulerBenchmark.cs
+++ b/src/Quartz.Benchmark/QuartSchedulerBenchmark.cs
@@ -395,11 +395,12 @@ namespace Quartz.Benchmark
                 InstanceId = instanceId,
                 ThreadPool = new DefaultThreadPool { MaxConcurrency = threadCount },
                 JobStore = new RAMJobStore(),
+                IdleWaitTime = TimeSpan.FromSeconds(30),
                 MaxBatchSize = threadCount,
                 BatchTimeWindow = TimeSpan.Zero
             };
 
-            return new QuartzScheduler(res, TimeSpan.Zero);
+            return new QuartzScheduler(res);
         }
 
         private JobExecutionContextImpl CreateJobExecutionContext(IScheduler scheduler, IOperableTrigger trigger)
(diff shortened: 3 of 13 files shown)


drieseng:
@lahma Locally, the `DirectSchedulerFactoryTest.TestPlugins()` unit tests sometimes/mostly enters a loop (within NUnit it seems). I'll have a closer look if this also affects the **AppVeyor** build.

drieseng:
@lahma I was able to resolve the issue (where running the unit tests would take several minutes) by making sure that schedulers that I created are effectively shut down.

There were a few causes for that issue:
* If you create a scheduler, but not start it, you shouldn't have to perform a shutdown. However, in the ctor of **QuartzScheduler** we initialize but also start a **QuartzSchedulerThread**.  This is not what I (and I think most people) would expect. I'll submit a PR to change this, but it's up to you to decide of course.
* If you do not shut down the **QuartzScheduler**, we will never await the task that we created in `QuartzSchedulerThread.Start()`.  This apparently comes back with a vengeance if you create several schedulers, and finally call `Shutdown()` for one of them. I have no explanation as to why it behaves this way, do you? We may consider implementing **IDisposable** on **QuartzSchedulerThread** and add a finalizer, to ensure the task is always awaited.

You can easily reproduce this issue with the following code fragment:
```cs
for (var i = 0; i < 100; i++)
{
    var schedulerName = i.ToString();
    var schedulerInstanceId = i.ToString();

    DirectSchedulerFactory.Instance.CreateScheduler(schedulerName, schedulerInstanceId, new DefaultThreadPool(), new RAMJobStore());

    var scheduler = await DirectSchedulerFactory.Instance.GetScheduler(schedulerName);
    // await scheduler.Shutdown();
}

DirectSchedulerFactory.Instance.CreateScheduler(new DefaultThreadPool(), new RAMJobStore());
var defaultScheduler = await DirectSchedulerFactory.Instance.GetScheduler();
await defaultScheduler.Start();
await defaultScheduler.Shutdown();
```

If you comment out the `await scheduler.Shutdown()` line in the for loop, it'll finish almost immediately.

I was able to fix this issue by changing `QuartzSchedulerThread.Start()` to create a long-running task.
I think this makes sense, even if it wouldn't fix this issue.
Is it ok if I create another PR for this?
I'll do some more investigation tomorrow to see if that still "leaks" a thread if we never shut down the **QuartzSchedulerThread**.

drieseng:
@lahma Do you think you'll have time to review this one?

lahma (approved):

lahma:
Thank you!

drieseng:
@lahma I was actually expecting you to have remarks on this one. I wonder if we should do this in two phases:
1) Reject the value **TimeSpan.Zero** for **IdleWaitTime**. This will avoid people going from "I used **TimeSpan.Zero** to get a 30 second **IdleWaitTime**" to effectively getting a zero **IdleWaitTime**.
2) Once we've released a version with the changes in this PR and item **1**, we can move back to allowing **TimeSpan.Zero** for **IdleWaitTime**.

What do you think?

lahma:
> @lahma I was actually expecting you to have remarks on this one. 

Sorry to fail your expectations.

> 1) Reject the value **TimeSpan.Zero** for **IdleWaitTime**. This will avoid people going from "I used **TimeSpan.Zero** to get a 30 second **IdleWaitTime**" to effectively getting a zero **IdleWaitTime**.

We are in the zone where people want to hurt themselves, not sure how much should be supported.

> 2) Once we've released a version with the changes in this PR and item **1**, we can move back to allowing **TimeSpan.Zero** for **IdleWaitTime**.

Again, this is as corner-case as it gets if I understand correctly. How much do we need to support?

drieseng:
> > @lahma I was actually expecting you to have remarks on this one.
> 
> Sorry to fail your expectations.

LOL

PR #1409, opened 2021-11-13
RAMJobStore: Remove allocation and simplify

drieseng:
Do not allocate a **List\<IOperableTrigger>** when no time triggers are available.
Remove unnessary null checks, and remove unnecessary lock.


lahma:
Looks good!

PR #1411, opened 2021-11-13
Use JobBuilder to create IJobDetail instances

drieseng:
We should give preference to using **JobBuilder** to create a **IJobDetail** instance over directly initializing a **JobDetailImpl**.

### DIFF
--- a/src/Quartz.Tests.Integration/ExceptionPolicy/ExceptionJobTest.cs
+++ b/src/Quartz.Tests.Integration/ExceptionPolicy/ExceptionJobTest.cs
@@ -32,8 +32,13 @@ namespace Quartz.Tests.Integration.ExceptionPolicy
             await sched.Start();
             string jobName = "ExceptionPolicyUnscheduleFiringTrigger";
             string jobGroup = "ExceptionPolicyUnscheduleFiringTriggerGroup";
-            JobDetailImpl myDesc = new JobDetailImpl(jobName, jobGroup, typeof(ExceptionJob));
-            myDesc.Durable = true;
+
+            var myDesc = JobBuilder.Create()
+                                   .OfType<ExceptionJob>()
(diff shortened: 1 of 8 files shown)


lahma (approved):
Nice cleanup!

src/Quartz.Tests.Integration/ExceptionPolicy/ExceptionJobTest.cs
```diff
@@ -32,8 +32,13 @@ public async Task ExceptionJobUnscheduleFiringTrigger()
             await sched.Start();
             string jobName = "ExceptionPolicyUnscheduleFiringTrigger";
             string jobGroup = "ExceptionPolicyUnscheduleFiringTriggerGroup";
-            JobDetailImpl myDesc = new JobDetailImpl(jobName, jobGroup, typeof(ExceptionJob));
-            myDesc.Durable = true;
+
+            var myDesc = JobBuilder.Create()
+                                   .OfType<ExceptionJob>()
+                                   .WithIdentity(new JobKey(jobName, jobGroup))
```
drieseng, line 38: Was wondering all the time "should I introduce an overload takes the name and group as **string** arguments?" ...
Now that you've merged I noticed that we already such an overload :(
Let me know if you want me to use that overload. It's a little less verbose for those cases where we don't have a **JobKey** (which is most of the time).

lahma: I was on the fence with nitpicking about using the overload, but in this context doesn't matter. 

Thanks for cleaning up the code!

drieseng: You should not hesitate to give feedback like that.
I have a tick skin, I can take a lot more than that :smile:

PR #1412, opened 2021-11-13
JobBuilder: Allow PersistJobDataAfterExecution and ConcurrentExecutionDisallowed to be set

drieseng:
Allow **PersistJobDataAfterExecution** and **ConcurrentExecutionDisallowed** to be explicitly set when building a IJobDetail.
When not set, we continue to use reflection to determine the values for these properties based on the presence (or absence) of **PersistJobDataAfterExecutionAttribute** and **DisallowConcurrentExecutionAttribute**.

If the job type is specified when the **JobDetailImpl** is constructed (using **JobBuilder**), then determining the value for these properties has now become a one-time cost.

Until this PR we payed for this each time these properties are accessed.
In case of **ConcurrentExecutionDisallowed** this is when:
* the trigger is acquired
* the trigger is fired
* the trigger(ed job) has completed

The creation and caching of **JobTypeInformation** for a given job type has now also been encapsulated in the *JobTypeInformation* class.

This PR solely exists to sollicit for feedback, and to have something more concrete to discuss https://github.com/quartznet/quartznet/discussions/1392.

### DIFF
--- a/src/Quartz/Impl/JobDetailImpl.cs
+++ b/src/Quartz/Impl/JobDetailImpl.cs
@@ -1,4 +1,4 @@
-﻿#region License
+#region License
 
 /*
  * All content copyright Marko Lahma, unless otherwise indicated. All rights reserved.
@@ -35,10 +35,38 @@ namespace Quartz.Impl
     /// </summary>
     /// <seealso cref="DisallowConcurrentExecutionAttribute"/>
     /// <seealso cref="PersistJobDataAfterExecutionAttribute"/>
-    internal class JobTypeInformation
+    internal sealed class JobTypeInformation
     {
-        internal bool ConcurrentExecutionDisallowed { get; set; }
-        internal bool PersistJobDataAfterExecution { get; set; }
+        private static readonly ConcurrentDictionary<Type, JobTypeInformation> jobTypeCache = new ConcurrentDictionary<Type, JobTypeInformation>();
+
+        public JobTypeInformation(bool concurrentExecutionDisallowed, bool persistJobDataAfterExecution)
+        {
+            ConcurrentExecutionDisallowed = concurrentExecutionDisallowed;
+            PersistJobDataAfterExecution = persistJobDataAfterExecution;
+        }
+
+        /// <summary>
+        /// Return information about JobType as an instance
+        /// </summary>
+        /// <param name="jobType">The type for which information will be searched</param>
+        /// <returns>
+        /// An <see cref="JobTypeInformation"/> object that describe specified type 
+        /// </returns>
+        public static JobTypeInformation GetOrCreate(Type jobType)
+        {
+            return jobTypeCache.GetOrAdd(jobType, Create(jobType));
+        }
+
+        private static JobTypeInformation Create(Type jobType)
+        {
+            var concurrentExecutionDisallowed = ObjectUtils.IsAnyInterfaceAttributePresent(jobType, typeof(DisallowConcurrentExecutionAttribute));
+            var persistJobDataAfterExecution = ObjectUtils.IsAnyInterfaceAttributePresent(jobType, typeof(PersistJobDataAfterExecutionAttribute));
+
+            return new JobTypeInformation(concurrentExecutionDisallowed, persistJobDataAfterExecution);
+        }
+
+        public bool ConcurrentExecutionDisallowed { get; }
+        public bool PersistJobDataAfterExecution { get; }
     }
 
     /// <summary>
@@ -67,13 +95,13 @@ namespace Quartz.Impl
     [Serializable]
     public class JobDetailImpl : IJobDetail
     {
-        private static readonly ConcurrentDictionary<Type, JobTypeInformation> jobTypeCache = new ConcurrentDictionary<Type, JobTypeInformation>();
-
         private string name = null!;
         private string group = SchedulerConstants.DefaultGroup;
         private string? description;
         private JobDataMap jobDataMap = null!;
         private Type jobType = null!;
+        private bool? disallowConcurrentExecution;
+        private bool? persistJobDataAfterExecution;
 
         [NonSerialized] // we have the key in string fields
         private JobKey key = null!;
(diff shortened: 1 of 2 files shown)


src/Quartz/Impl/JobDetailImpl.cs
```diff
@@ -35,10 +35,38 @@ namespace Quartz.Impl
     /// </summary>
     /// <seealso cref="DisallowConcurrentExecutionAttribute"/>
     /// <seealso cref="PersistJobDataAfterExecutionAttribute"/>
-    internal class JobTypeInformation
+    internal sealed class JobTypeInformation
```
drieseng, line 38: This type is now immutable.

src/Quartz/Impl/JobDetailImpl.cs
```diff
@@ -212,7 +276,7 @@ public JobKey Key
 
                 return key;
             }
-            set
+            internal set
```
drieseng, line 279: This one cannot be private since we use it in `ServiceCollectionExtensions.ScheduleJob<T>(this ServiceCollectionQuartzConfigurator options, Action<ITriggerConfigurator> trigger, Action<IJobConfigurator>? job = null)`.

I wonder if:
* we shouldn't throw (in `JobBuilder.Build()`if you attempt to build a **IJobDetail** without having assign a key?
* if makes sense to assign a job key that uses the name and group of the trigger in `ServiceCollectionExtensions.ScheduleJob<T>(this ServiceCollectionQuartzConfigurator options, Action<ITriggerConfigurator> trigger, Action<IJobConfigurator>? job = null)`

src/Quartz/Impl/JobDetailImpl.cs
```diff
@@ -236,7 +300,7 @@ public JobKey Key
         public string? Description
         {
             get => description;
-            set => description = value;
+            private set => description = value;
```
drieseng, line 303: Personally, I'd prefer to remove all setters (and make them - and the type itself - sealed).
But I know this could be considered a major breaking change.

src/Quartz/Impl/JobDetailImpl.cs
```diff
@@ -303,17 +367,48 @@ public virtual JobDataMap JobDataMap
...
+        /// <value>
+        /// <see langword="true"/> if job data should be re-stored when execution of the job completes; otherwise,
+        /// <see langword="false"/>.
+        /// </value>
+        /// <seealso cref="PersistJobDataAfterExecutionAttribute"/>
+        public virtual bool PersistJobDataAfterExecution
+        {
+            get
+            {
+                if (!persistJobDataAfterExecution.HasValue)
```
drieseng, line 384: If **JobType** cannot be set after "build", we can remove this logic.

src/Quartz/Impl/JobDetailImpl.cs
```diff
@@ -67,13 +95,13 @@ internal class JobTypeInformation
     [Serializable]
     public class JobDetailImpl : IJobDetail
     {
-        private static readonly ConcurrentDictionary<Type, JobTypeInformation> jobTypeCache = new ConcurrentDictionary<Type, JobTypeInformation>();
-
         private string name = null!;
         private string group = SchedulerConstants.DefaultGroup;
         private string? description;
         private JobDataMap jobDataMap = null!;
         private Type jobType = null!;
```
drieseng, line 102: We actually lie about this not being nullable. We have a use case in [`StdAdoDelegate.SelectJobForTrigger(...)`](https://github.com/quartznet/quartznet/blob/73d385526dec96b4c36e03a0ef75a4dfc5a32306/src/Quartz/Impl/AdoJobStore/StdAdoDelegate.Jobs.cs#L208) where we allow an  **IJobDetail** to be built without a job type. We invoke this method with **loadJobType** set to **false** from [`JobStoreSupport.RemoveTrigger(ConnectionAndTransactionHolder conn, TriggerKey triggerKey, IJobDetail? job, CancellationToken cancellationToken = default)`](https://github.com/quartznet/quartznet/blob/73d385526dec96b4c36e03a0ef75a4dfc5a32306/src/Quartz/Impl/AdoJobStore/JobStoreSupport.cs#L1295) if the **job** argument is **null**.

One way of dealing with this would be to:
1.  Always build a full **IJobDetail** instance (wih job type). This would be the easiest solution, but I'm pretty sure this is not what you want.
2. Introduce a specialized **IJobDetail**-like type that only represents a **JobType** and whether it's durable or not. We can then remove the **loadJobType** argument from `SelectJobForTrigger(ConnectionAndTransactionHolder conn,             TriggerKey triggerKey, ITypeLoadHelper loadHelper, bool loadJobType, CancellationToken cancellationToken = default)` and introduce another (similar) method that would return the new **IJobDetail**-like type.
3. Update **JobDetailImpl.JobType** (and other properties or methods in **JobDetailImpl** that use **JobType**) to throw an exception if **JobType** was not set.  There are a few problematic methods though:
     * JobDetailImpl.ToString()
    Here we use the **JobType** property, but if this can now throw we end up with a `.ToString()` override that can 
    throw. I'm sure we do not want this.
     * JobDetailImpl.Validate()
    He were use the **jobType** (and other) instance fields directly. This is not correct from an extensibility point-of- 
    view as we allow the corresponding properties to be overridden, but ignore them.

Personally, I prefer option **2**.
Option **3** is only an option if we stop marking some properties virtual.
There are lots of places where we mark properties or methods virtual, but - since we do not always invoke these properties or methods - it does not make much sense to do so.

drieseng:
@lahma Do you think you can find some time to discuss this PR?

drieseng:
@lahma Do you think you can find some time to review and discuss this PR?

drieseng:
@lahma, (how) do you want me to proceed with this one? I'd like to get this one in before I move on to other optimizations.
* Do I add the `DisallowConcurrentExecution(...)` and `PersistJobDataAfterExecution(...)` methods to **IJobConfigurator**.
* Do want to keep the setters on **JobDetailImpl** public?

Is there anything else you want me to change before you take this PR under consideration?

src/Quartz/Impl/JobDetailImpl.cs
lahma: There's probably a more larger concern where we would not want to rely on `Type`, when a web interface interacts and handles jobs it shouldn't be required it to have the DLL (concrete types) deployed. So a big design mistake in the early days was to require a *resolvable* type in many places, when it should have been enough to resolve it at store time.

src/Quartz/Impl/JobDetailImpl.cs
```diff
@@ -140,6 +168,42 @@ public JobDetailImpl(string name, string group, Type jobType, bool isDurable, bo
...
+        /// <param name="key">The key of the job.</param>
+        /// <param name="jobType">Type of the job.</param>
+        /// <param name="description">The description given to the <see cref="IJob" /> instance by its creator.</param>
+        /// <param name="isDurable">if set to <c>true</c>, job will be durable.</param>
+        /// <param name="requestsRecovery">if set to <c>true</c>, job will request recovery.</param>
+        /// <param name="jobDataMap">The data that is associated with the <see cref="IJob" />.</param>
+        /// <param name="disallowConcurrentExecution">Indicates whether or not concurrent exection of the job should be disallowed.</param>
+        /// <param name="persistJobDataAfterExecution">Indicates whether or not job data should re-stored when execution of the job completes.</param>
+        /// <exception cref="ArgumentNullException"><paramref name="key"/> is <see langword="null"/>.</exception>
+        internal JobDetailImpl(JobKey key,
```
lahma, line 184: nit: I prefer every arguments on new line when there are arguments on new lines, so this would have less indent too

src/Quartz/Impl/JobDetailImpl.cs
lahma: Generally having job/trigger with just a name is all fine, less friction, but defaulting to trigger's name is problematic - job can have N triggers.

src/Quartz/Impl/JobDetailImpl.cs
lahma: Well if this is v4 work... we can break as much as we want. We can always re-open sealed classes with good reasoning behind the request.

src/Quartz/JobBuilder.cs
```diff
@@ -67,9 +67,11 @@ public class JobBuilder : IJobConfigurator
     {
         private JobKey? key;
         private string? description;
-        private Type jobType = null!;
+        private Type? jobType;
         private bool durability;
         private bool shouldRecover;
+        private bool? _concurrentExecutionDisallowed;
```
lahma, line 73: currently we don't have `_` prefix deployed around the code base, I'm fine with using it, just need to make things align

lahma (approved):

lahma:
Thank you, sorry to keep you waiting for such a long time!

PR #1414, opened 2021-11-14
Remove use of GetTypeInfo()

drieseng:
Remove use of `Type.GetTypeInfo()` that dates back to early days of .NET Core.


drieseng:
This should be good to go now. Had to delay the launch a few times :smile:

lahma (approved):
Nice cleanup, oh the glory days of porting to .NET Standard with old SDKs.

PR #1419, opened 2021-11-16
Merge changes from main into v4

drieseng:


drieseng:
@lahma `SchedulerTest.ReschedulingTriggerShouldKeepOriginalNextFireTime()` test apparently occasionally fails on macOS too (it's already ignored for Linux).

But it's not that weird that it can fail if you look at the code:
```
IOperableTrigger trigger = (IOperableTrigger) TriggerBuilder.Create()
    .WithSimpleSchedule(x => x.WithIntervalInHours(1).RepeatForever())
    .ForJob(job)
    .StartNow()
    .Build();

trigger = (IOperableTrigger) await scheduler.GetTrigger(trigger.Key);
Assert.That(trigger.GetPreviousFireTimeUtc(), Is.EqualTo(null));
```

We're expecting the trigger not to fire between scheduling it and obtaining the trigger, but nothing prevents it from doing so.
I'm sure that means we're just too damn fast :smile:

drieseng:
@lahma Can you approve this one?

About the test failure (see me comment above):
Can we just schedule the trigger to start in a few seconds instead of "now"?
If you think that is ok, I'll submit a separate PR for this.

lahma:
> Can we just schedule the trigger to start in a few seconds instead of "now"?
> If you think that is ok, I'll submit a separate PR for this.

Should be fine.

PR #1420, opened 2021-11-16
Let exceptions bubble up that occur while cloning

### LINKED ISSUE #1415: Remove try/catch from Clone() methods
We always have a try/catch block in our `Clone()` methods to hide the underlying cause when cloning fails.

**For example:**
```cs
/// <summary>
/// Creates a new object that is a copy of the current instance.
/// </summary>
/// <returns>
/// A new object that is a copy of this instance.
/// </returns>
public virtual DirtyFlagMap<TKey, TValue> Clone()
{
    DirtyFlagMap<TKey, TValue> copy;
    try
    {
        copy = (DirtyFlagMap<TKey, TValue>) MemberwiseClone();
        copy.map = new Dictionary<TKey, TValue>(map);
    }
    catch (Exception)
    {
        throw new Exception("Not Cloneable.");
    }

    return copy;
}
```

This makes the generated IL/ASM bloated, and does not offer any added value as the caller can now only guess why cloning failed.

@lahma, is it ok to remove these try/catch blocks?

drieseng:
Do not catch exceptions that occur while cloning.
Fixes #1415.

### DIFF
--- a/src/Quartz/CronExpression.cs
+++ b/src/Quartz/CronExpression.cs
@@ -2209,17 +2209,8 @@ namespace Quartz
         /// </returns>
         public object Clone()
         {
-            CronExpression copy;
-            try
-            {
-                copy = new CronExpression(CronExpressionString);
-                copy.TimeZone = TimeZone;
-            }
-            catch (FormatException e)
-            {
-                // never happens since the source is valid...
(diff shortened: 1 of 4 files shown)


lahma (approved):

lahma:
Thank you.

PR #1424, opened 2021-11-17
StringOperator: improve dealing with null values

drieseng:
Update the **Equality** operator to return **true** in `Evaluate(string value, string compareTo)` when both **value** and **compareTo** are **null**. Previously, a **NullReferenceException** would be thrown.

For all other built-in operators, `Evaluate(string value, string compareTo)` now returns **false** instead of throwing a **NullReferenceException** or **ArgumentNullException** when **value** or **compareTo** (or both) are **null**.

Added unit tests.

### DIFF
--- a/src/Quartz.Tests.Unit/Impl/Matchers/StringOperatorTest.cs
+++ b/src/Quartz.Tests.Unit/Impl/Matchers/StringOperatorTest.cs
@@ -1,0 +1,297 @@
+using NUnit.Framework;
+using Quartz.Impl.Matchers;
+using System;
+using System.IO;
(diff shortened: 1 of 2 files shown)


drieseng:
@lahma I was in doubt whether or not it would be better to throw an **ArgumentNullException** if **compareTo** is **null**. This is what we do in **StringMatcher** anyway.

I've decided to update the PR to do just that. I'll mark it as draft until this is done.

drieseng:
@lahma Test failure is unrelated.

lahma (approved):
Looks good.

PR #1431, opened 2021-11-23
Improve performance, reduce allocations and reduce contention in TaskSchedulingThreadPool 

drieseng:
Improve performance, reduce allocations and reduce contention of **TaskSchedulingThreadPool** by:
* Setting continuation on the wrapper task insteaad of the unwrapped task.
* Using a **CountDownEvent** instead of a **List\<Task>** to keep track of the (number of) running tasks and wait until all tasks have completed.
* Caching the delegate that we use to mark the task complete.

The biggest impact on allocations and certainly contention/locking comes from the usage of **CountDownEvent**.

<details>
<summary>Benchmark results</summary>

|                                                           Method | Branch |    Mean  |    Error |   StdDev |  Gen 0 |  Gen 1 |  Gen 2 | Allocated |
|----------------------------------------------------------------- |--------|---------:|---------:|---------:|-------:|-------:|-------:|----------:|
| RunInThread_CompletedTask_MaxConcurrencyIsMaxValue_MultiThreaded | main   | 184.6 us | 10.96 us | 32.15 us | 0.0800 | 0.0300 | 0.0100 |     426 B |
| RunInThread_CompletedTask_MaxConcurrencyIsMaxValue_MultiThreaded | PR     | 575.1 ns | 11.37 ns | 14.38 ns | 0.0430 | 0.0170 | 0.0050 |     259 B |
</details>

Note that the time unit for **main** is microseconds, while the unit for the **PR** is nanoseconds.
For this benchmark, the new threadpool implementation is **320** times faster.

I had to temporarily reduce the number of tasks that are created in that benchmark from 1 million to 100.000 in order to get results for **main**. If not, it would take ages to complete.

### DIFF
--- a/src/Quartz/Simpl/TaskSchedulingThreadPool.cs
+++ b/src/Quartz/Simpl/TaskSchedulingThreadPool.cs
@@ -1,5 +1,4 @@
-﻿using System;
-using System.Collections.Generic;
+using System;
 using System.Threading;
 using System.Threading.Tasks;
 
@@ -20,10 +19,15 @@ namespace Quartz.Simpl
         // only to the thread pool functions themselves (such as scheduling tasks).
         private readonly CancellationTokenSource shutdownCancellation = new CancellationTokenSource();
 
-        // A list of running tasks (needed to optionally wait for executing tasks at shutdown)
-        private readonly List<Task> runningTasks = new List<Task>();
+        /// <summary>
+        /// Allows us to wait until no running tasks remain.
+        /// </summary>
+        private CountdownEvent runningTasksCountdown = null!;
 
-        private readonly object taskListLock = new object();
+        /// <summary>
+        /// Cached delegate to mark a given task as complete.
+        /// </summary>
+        private Action<Task> completeTask = null!;
 
         // The semaphore used to limit concurrency and integers representing maximim concurrent tasks
         private SemaphoreSlim concurrencySemaphore = null!;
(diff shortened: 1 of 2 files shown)


src/Quartz/Simpl/TaskSchedulingThreadPool.cs
```diff
@@ -185,25 +200,25 @@ public bool RunInThread(Func<Task> runnable)
                 return false;
             }
 
+            // Wrap the runnable in a Task to start it asynchronously
             var task = new Task<Task>(runnable);
 
-            // Unrap the task so that we can work with the underlying task
-            var unwrappedTask = task.Unwrap();
-            lock (taskListLock)
+            lock (runningTasksCountdown)
```
drieseng, line 206: @lahma Let me know if you prefer to use a specific lock object instead.

src/Quartz/Simpl/TaskSchedulingThreadPool.cs
```diff
@@ -212,44 +227,47 @@ public bool RunInThread(Func<Task> runnable)
...
             if (waitForJobsToComplete)
             {
-                Task[] tasksArray;
-                lock (taskListLock)
+                lock (runningTasksCountdown)
                 {
                     // Cancellation has been signaled, so no new tasks will begin once
                     // shutdown has acquired this lock
-                    tasksArray = runningTasks.ToArray();
+                    log.DebugFormat($"Waiting for {runningTasksCountdown.CurrentCount.ToString()} threads to complete.");
```
drieseng, line 258: This is a trap that a lot of people run into. String interpolation, just like `string.Format(...)` will box (struct) parameters.

drieseng:
This one is ready to go.

 I'll have to revisit some benchmarks where we are slower with new implementation because we put an enormous pressure on the .NET threadpool. These are benchmarks that I haven't even pushed yet, where I configured the threadpool to have a **MaxConcurrency** of **int.MaxValue**. Let's say that this is not your typical configuration :-)

drieseng:
@lahma Did you have time to review this one? Let me know if you have any questions or remarks.

lahma:
Thank you, I'm not entirely sure about all the ramifications, but I guess we'll see later..

Issue #1417, opened 2021-11-15
Review/discuss DirtyFlagMap


drieseng:
While scanning for opportunities to improve performance and/or reduce memory allocations, I stumbled upon **DirtyFlagMap\<TKey,TValue**.

I discovered the following issues (or at least things to discuss, clarify and document) while browsing the code:

- [x] 1. The `GetObjectData(SerializationInfo info, StreamingContext context)` method, which is part of infrastructure code for binary serialization, is current **public**. I propose to make this method **protected**.
- [ ] 2. The **WrappedMap** property is currently **public**. I propose to make this method **protected internal**, or even **internal**.

- [x] 3. The out argument of `TryGetValue(TKey key, out TValue value)` should be annotated with a **[MaybeNullWhen(false)]** attribute. This matches the corresponding method in **Dictionary\<TKey,TValue>**.
- [x] 4. I propose to remove the `TValue Get(TKey key)` method as we already have two other ways to obtain the value for a given key.
  If we keep this method, we should at least make the return value nullable.
- [ ] 5. The `this[TKey key]` indexer should throw a **KeyNotFoundException** if no entry exists for the specified key.
   This matches the implementation of the corresponding indexer in **Dictionary\<TKey,TValue>**, and matches how it's documented in **IDictionary\<TKey,TValue>**.
- [ ] 6. For the setter of that indexer, we should decide if we actually want to set the **Dirty** flag when there's already an entry with the specified key **and** with a value that equals the specified value.
  **Note:**
  Checking whether or not the value has changed will cost a little more, as we'd have to lookup the current value first and compare it with the new value.
- [ ] 7. I propose to make `Remove(KeyValuePair<TKey, TValue> item)` an explicit interface implementation.
- [ ] 8. We should also consider updating the `Remove(KeyValuePair<TKey, TValue> item)` to only remove the entry if the current value equals the value of the **KeyValuePair\<TKey, TValue>**. This is how it's implemented in **Dictionary\<TKey,TValue>**.
- [ ] 9. We should consider updating `Add(TKey, TValue)` to throw an **ArgumentException** if an entry exists with the specified key.
This is how it's implemented in **Dictionary\<TKey,TValue>**, and how it's documented in **IDictionary\<TKey,TValue>**.
- [ ] 10. We should consider updating `Add(KeyValuePair<TKey, TValue>)` to call `Add(TKey, TValue)` instead of `Put(TKey, TValue)`.
[...]

PR #1444, opened 2021-11-29
DirtyFlagMap: Update Put(TKey key, TValue val) to only set dirty flag if successful

drieseng:
Update `DirtyFlagMap.Put(TKey key, TValue val)`to only set the dirty flaw when we know that we managed to add or update an entry in the underlying dictionary.

Fixes first part of issue 14 in #1417


lahma:
Thank you.

PR #1445, opened 2021-11-29
DirtyFlagMap: Remove Get(TKey key)

drieseng:
Remove `Get(TKey key)`, and replace its usage with the indexer.
Document `TryGetValue(TKey key, out TValue value)` and add **MaybeNullWhen** attribute on out parameter.

Fixes items 3 and 4 of #1417.


lahma:
Thanks.

PR #1448, opened 2021-11-30
Revert change to output path to allow us to run benchmarks

drieseng:
Revert change to output path - that was done as part of the [migration](https://github.com/quartznet/quartznet/commit/6053359d8756e3dbbd0786272ddf17f9a4bbb28c#diff-5c1900dca99df01c3ddc3a884282164b04deee76e79b176e2d34cfa47f3b8907) to NUKE - to allow the assembly key file to be found by projects that are generated by BDN.


drieseng:
@lahma Is this safe to do, or do we need changes elsewhere?

lahma:
I think this will break the packaging logic as it was expecting to find things from sub-directories under src. The build output folder was making it hard to build the release zip with the binaries in it.

lahma:
This it the logic that gathers artifacts: 

https://github.com/quartznet/quartznet/blob/42f0ec89bb567d40bfc771f56724746fe61831f0/build/Build.cs#L160-L166 

With this change is also now writing to NUKE project direcotry (build).

drieseng:
I'll see if I can find another way to fix this.

Issue #1432, opened 2021-11-23
QuartzHostedService: Wait for ApplicationStarted


Timovzl:
Hosted services and (by extension) **Quartz jobs are started before the application has successfully started**. For example, if the application serves HTTP requests, the jobs are started before it is ready to respond to those requests - **even before the app can respond to health checks**.

This is particularly annoying for smart deployments. For example, one might use Kubernetes to spin up new application instances and consider them successfully started once they answer a health check positively. Only after this happens do we include the new instances in the load balancer and do we spin down the old application instances. However, if the new instances fail to become healthy in a timely manner, they are spun down again. **The issue is that such never-officially-included instances may have already started running jobs.**

**Solution**

The problem would be mostly **mitigated if the Quartz hosted service would wait for application startup to complete**:

```cs
await Task.Delay(Timeout.InfiniteTimeSpan, // Wait "indefinitely", until ApplicationStarted is triggered
    this.HostApplicationLifetime.ApplicationStarted) // IHostApplicationLifetime injected in ctor
    .ContinueWith( _ => {}, TaskContinuationOptions.OnlyOnCanceled) // Without an OperationCanceledException
    .ConfigureAwait(false));
```

This way, at least we know that the application has managed to start successfully. For example, any unrelated hosted services that cause startup failure will now automatically _precede_ our jobs, which is good.

Optionally, this behavior could be configured by a boolean, `QuartsHostedServiceOptions.AwaitApplicationStarted`. However, I see no issue with the delay, and I consider the timing to be more correct, so I believe the adjusted behavior should be the default.

Could such a default be considered a breaking change? Probably not: Currently, there are no guarantees that a failure in a Quartz job will occur _before_ startup completes (and thus would have the ability to prevent a successful startup). In actuality, they _might_ happen early and prevent startup, or they _might_ happen later and _not_ prevent it. Going from an unpredictable scenario to a predictable seems non-breaking to me, but please correct me if I am overlooking something.

**Alternatives**

There is `QuartzHostedServiceOptions.StartDelay`. However, arbitrary delays are bad; startup times may be unpredictable.

**Additional Context**

We might consider the need to solve this problem as a design flaw in `BackgroundService`.

Services that perform startup and/or shutdown logic should implement `IHostedService`. Services that implement _background logic_ (like Quartz jobs, arguably) should inherit from `BackgroundService`. This changes what methods the service needs to implement, in a way that makes the most sense for the situation.

[...]

PR #1449, opened 2021-12-01
QuartzHostedService now has jobs start after application startup

Timovzl:
Implements #1432.

QuartzHostedService now has jobs start after application startup completes, unless QuartzHostedServiceOptions.AwaitApplicationStarted=false.


lahma (approved):
Looks good to me. If @andrewlock does not have improvement ideas I will merge this in a few days. Thanks for this splendid contribution, kudos for the test coverage!

Timovzl:
@lahma I wanted to check with you that ISchedulerFactory.GetScheduler() does _not_ cause job exection to start. Can you confirm that? Since we’re still calling that immediately (which seems good for perhaps ensuring connectivity to the remote).

Timovzl:
I’m also now considering the removal of lines 36-42 of QuartzHostedService, since I just realized that the task never completes synchronously (unless cancellation _just_ occurred, in which case we do not care about the exception).

Seems like I followed BackgroundService’s pattern needlessly in this case.

Timovzl (commented):
TO DO: Remove QuartzHostedService lines 36-42, as the task does not complete synchronously?

lahma:
> I wanted to check with you that ISchedulerFactory.GetScheduler() does _not_ cause job exection to start. Can you confirm that? Since we’re still calling that immediately (which seems good for perhaps ensuring connectivity to the remote).

The scheduler starts when `Start()` is called, it can be "no-op" when it's configured with `ZeroSizeThreadPool`, but the `Start` should be the key here. There are some funky code paths which require `GetScheduler` to be Task, can't remember where the rabbit hole went... Generally it should always be non-async in real like (until proven otherwise).

> TO DO: Remove QuartzHostedService lines 36-42, as the task does not complete synchronously?

I'm not sure I quite follow.

Timovzl:
Those lines contain some code that await the task if it is already completed, so as to propagate any exceptions in that case.

However, since the Task starts with a delay, it practically never completes synchronously, so that code is redundant.

lahma:
> Those lines contain some code that await the task if it is already completed, so as to propagate any exceptions in that case.
> 
> However, since the Task starts with a delay, it practically never completes synchronously, so that code is redundant.

Sounds reasonable, I trust your judgement here.

Timovzl:
Actually, scratch that. Changed my mind. It’s more correct this way in the case where ApplicationStarted is already triggered (tests?) and it’s better to follow the standard pattern. Leaving it as-is.

src/Quartz.Extensions.Hosting/QuartzHostedService.cs
```diff
@@ -1,44 +1,94 @@
...
+                // Follow the pattern from BackgroundService.StartAsync: https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.Extensions.Hosting.Abstractions/src/BackgroundService.cs
+
+                var deferredSchedulerStartTask = AwaitStartupCompletionAndStartSchedulerAsync(cancellationToken);
+
+                // If the task completed synchronously, await it in order to bubble potential cancellation/failure to the caller
+                // Otherwise, return, allowing application startup to complete
+                if (deferredSchedulerStartTask.IsCompleted)
+                {
+                    await deferredSchedulerStartTask.ConfigureAwait(false);
+                }
```
andrewlock, line 42: I don't think this pattern is quite correct 🤔 If `!deferredSchedulerStartTask.IsCompleted`, I think you need to store that task in a field and await it in `StopAsync`, similar to [the way `BackgroundService` works](https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.Extensions.Hosting.Abstractions/src/BackgroundService.cs)?

Timovzl: I think you're right. I'll make a change.

(For the sake of completeness, note that this is quite niche. It is only for one scenario: when the application is stopped _while our deferred startup task has not yet completed_. It achieves only one thing: whilst we are cancelling, the await in `StopAsync` helps shutdown wait until our cancellation completes. For `BackgroundService` it is much more relevant, as its execute task is expected to keep running during the application lifetime.)

src/Quartz.Tests.Unit/QuartzHostedServiceTests.cs
```diff
@@ -0,0 +1,469 @@
...
+
+            using var startupCts = new CancellationTokenSource();
+
+            await quartzHostedService.StartAsync(startupCts.Token);
+
+            Assert.NotNull(schedulerFactory.LastCreatedScheduler);
+            Assert.AreEqual(shouldSchedulerBeStartedImmediately, schedulerFactory.LastCreatedScheduler.IsStarted);
+
+            appliationLifetime.SetStarted();
+            await Task.Delay(20); // Give the hosted service some time to respond to the ApplicationStarted token
```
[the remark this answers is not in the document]
Timovzl: Agreed.

Timovzl:
@andrewlock I have implemented your suggestions.

Timovzl:
@lahma I see that `TestNoConcurrentExecOnSameJob` has failed. I find it hard to imagine that this was caused by the change. Could it be that this is a sensitive test that can sometimes fail due to timing or concurrency issues? Could you rerun the check to see if it goes away?

lahma:
Don't worry about it, some tests might have timing issues, single failing agent doesn't rock the boat - I'm rerunning them though.

lahma:
Great work, thank you!

PR #1456, opened 2021-12-04
Do not sign assemblies produced by BenchmarkDotNet

drieseng:
BDN produces MSBuild projects to run benchmarks. These projects are created in subdirectories of the output path (eg. **src\Quartz.Benchmark\bin\release\net5.0\DefaultJob**), and as such our **Directory.Build.props** (which is **src**) is also applied for these projects.

In **Directory.Build.props** we enable signing using a key file with the relative path "../../quartz.net.snk".

When MSBuild now tries to sign the assembly produced by BDN, it resolves that relative path to "src\Quartz.Benchmark\bin\release\quartz.net.snk". This fails of course:

`CSC : error CS7027: Error signing output with public key from file '../../quartz.net.snk' -- File not found. [C:\development\quartznet\src\Quartz.Benchmark\bin\release\net5.0\DefaultJob\BenchmarkDotNet.Autogenerated.csproj]`

To resolve this issue, I've updated **Directory.Build.props** to not sign the assembly if name of the project that is being built contains "BenchmarkDotNet.Autogenerated".

I also updated BDN to version **0.13.1** as that fixes [this](https://github.com/dotnet/BenchmarkDotNet/issues/1452) issue.


lahma:
Should this go master as well?

drieseng:
@lahma Ideally, yes.

PR #1457, opened 2021-12-04
General improvements to scheduler benchmarks.

drieseng:
Improvements to scheduler benchmarks in order to get more accurate results.


drieseng:
@lahma Test failure is unrelated, and should be fixed by #1458.

PR #1458, opened 2021-12-05
*NOMERGE* Stabilize flaky tests playground

drieseng:
Stabilize the following tests:
* Quartz.Tests.Unit.JobExecutionAttributesInterfaceInheritanceTest.TestNoConcurrentExecOnSameJob(): add a small tolerance as **DateTime.(Utc)Now** does not have a high precision.
* Quartz.Tests.Unit.DisallowConcurrentExecutionJobTest.TestNoConcurrentExecOnSameJob(): add a small tolerance as **DateTime.(Utc)Now** does not have a high precision.
* Quartz.Tests.Unit.SchedulerTest.ReschedulingTriggerShouldKeepOriginalNextFireTime(): Ensure trigger does not immediately fire. Enable test on Linux.

Re-enable shutdown test in **Quartz.Tests.Unit.SchedulerTest**, and add test to verify that we do not wait for tasks to complete when **waitForJobsToComplete** is **false**.

### DIFF
--- a/build/Build.cs
+++ b/build/Build.cs
@@ -100,6 +100,7 @@ namespace System {
             DotNetTest(s => s
                 .EnableNoRestore()
                 .EnableNoBuild()
+                .AddLoggers("console;verbosity=normal")
                 .SetProjectFile(Solution.GetProject("Quartz.Tests.Unit"))
                 .SetConfiguration(Configuration)
                 .SetFramework(framework)
--- a/src/Quartz.Benchmark/JobExecutionContextImplBenchmark.cs
+++ b/src/Quartz.Benchmark/JobExecutionContextImplBenchmark.cs
@@ -103,7 +103,7 @@ namespace Quartz.Benchmark
                     BatchTimeWindow = TimeSpan.Zero
                 };
 
-            return new QuartzScheduler(res, TimeSpan.Zero);
+            return new QuartzScheduler(res);
         }
     }
 }
--- a/src/Quartz.Tests.Unit/DailyTimeIntervalScheduleBuilderTest.cs
+++ b/src/Quartz.Tests.Unit/DailyTimeIntervalScheduleBuilderTest.cs
@@ -1,4 +1,4 @@
-﻿#region License
+#region License
 
 /*
  * All content copyright Marko Lahma, unless otherwise indicated. All rights reserved.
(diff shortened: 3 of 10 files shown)


drieseng:
@lahma I've managed to stabilize some tests. I also re-enabled / updated one that was ignored, and add a test.

But....

The test that I added - **Quartz.Tests.Unit.SchedulerTest.TestShutdownWithoutWaitShouldNotBlockUntilAllTasksHaveCompleted** - is flaky. In that test I register a trigger that executes a given job every millisecond. The job actually sleeps for 200 ms to allow us to verify that a shutdown where **waitForJobsToComplete** is **false** does not wait until the task for executing this job completes.

I use a **Stopwatch** to track how much time it takes to execute the `IScheduler.Shutdown(bool waitForJobsToComplete, CancellationToken cancellationToken = default)` method. This should take very little time as we pass **false** for **waitForJobsToComplete**, but sometimes it takes more than 200 ms to complete. You'd think that there's a bug, and that we are effectively waiting until all tasks have completed (even though we pass **false** for **waitForJobsToComplete**).

I cannot reproduce this issue locally, so I added some troubleshooting code to **Quartz.NET** (that should not be merged, of course). I updated `QuartzScheduler.Shutdown(bool waitForJobsToComplete, CancellationToken cancellationToken = default)` to return a string containing the timings of just about each line in that method. When CI fails, it now outputs that information. But even when the **Stopwatch** in **Quartz.Tests.Unit.SchedulerTest.TestShutdownWithoutWaitShouldNotBlockUntilAllTasksHaveCompleted** indicates that shutdown took 200+ ms, the timings in that method indicate that it took less than a millisecond:

```
Before Standby = 0
After Standby = 0
After halt of scheduler thread = 0
After NotifySchedulerListenersShuttingdown = 0
After cancel jobs = 0
After shutdown of threadpool = 0
After shutdown of scheduler thread = 0
After unbind = 0
After shutdown of plugins = 0
After shutdown of jobstore = 0
After NotifySchedulerListenersShutdown = 0
After remove of scheduler instance = 0
Final 70270bf2-509c-4dc7-ad19-01a51a6e7268_$_NON_CLUSTERED = 0
```

Do you have any idea what may be causing this?

drieseng:
@lahma I expected this had to do with the fact that the test method was async, and the last two runs appear to confirm this.
I'll do some more pushes to force additional CI builds.

PR #1460, opened 2021-12-06
Stabilize tests

drieseng:
Attempt to stabilize more tests:
* `Quartz.Tests.Unit.DailyTimeIntervalScheduleBuilderTest.TestScheduleInMiddleOfDailyInterval()`: provide more meaningful information when asserts fail.
* `Quartz.Tests.Unit.JobExecutionAttributesInterfaceInheritanceTest.TestNoConcurrentExecOnSameJob()`: add a small tolerance as **DateTime.(Utc)Now** does not have a high precision.
* `Quartz.Tests.Unit.DisallowConcurrentExecutionJobTest.TestNoConcurrentExecOnSameJob()`: add a small tolerance as **DateTime.(Utc)Now** does not have a high precision.
* `Quartz.Tests.Unit.SchedulerTest.ReschedulingTriggerShouldKeepOriginalNextFireTime()`: Ensure trigger does not immediately fire. Enable test on Linux.

Re-enable shutdown test in **Quartz.Tests.Unit.SchedulerTest**, and add test to verify that we do not wait for tasks to complete when **waitForJobsToComplete** is **false**.
I had to make both tests synchronous to get them stable. I cannot explain this though.

Replaces #1458. I'm keeping that one open to play around with some async weirdness.

### DIFF
--- a/src/Quartz.Tests.Unit/DailyTimeIntervalScheduleBuilderTest.cs
+++ b/src/Quartz.Tests.Unit/DailyTimeIntervalScheduleBuilderTest.cs
@@ -1,4 +1,4 @@
-﻿#region License
+#region License
 
 /*
  * All content copyright Marko Lahma, unless otherwise indicated. All rights reserved.
@@ -42,6 +42,7 @@ namespace Quartz.Tests.Unit
     /// </summary>
     /// <author>Zemian Deng saltnlight5@gmail.com</author>
     /// <author>Nuno Maia (.NET)</author>
+    [TestFixture]
     public class DailyTimeIntervalScheduleBuilderTest
     {
         [Test]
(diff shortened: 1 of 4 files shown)


drieseng:
@lahma Can you force the CI build to run several times for this PR without having to commit any changes?

drieseng:
[This](https://docs.github.com/en/actions/managing-workflow-runs/re-running-workflows-and-jobs) page describes how to do this, but I don't think I have access to do so.

lahma:
Five consecutive re-runs were successful 🥳

drieseng:
@lahma If ok for you, you can merge these changes. I'd still like to know why I had to make the **TestShutdownWithoutWaitShouldNotBlockUntilAllTasksHaveCompleted** test in **Quartz.Tests.Unit.SchedulerTest** synchronous, but that should not block this PR.

lahma:
Thank you,  probably need to find time to dig into it.

PR #1463, opened 2021-12-10
Rename FloatValue into MyFloatValue

CodeDrop:
Name property same as JobData key "myFloatValue" used in previous samples to make "injection" work.


lahma:
Great work on docs, thank you!

PR #1464, opened 2021-12-10
Fix typo in "More About Jobs & Job Details"

CodeDrop:
[...] than -> then only one instance of "SalesReportForJoe" can execute at a given time.


lahma:
Thank you!

PR #1466, opened 2021-12-11
Remove unnecessary null check from TriggerWrapperComparator.GetHashCode()

drieseng:
Removes unnecessary null check from `TriggerWrapperComparator.GetHashCode()` as the **ttc** field is readonly and always initialized.


lahma:
Thank you.

PR #1470, opened 2021-12-12
Make internal classes sealed

drieseng:
Make all our internal classes sealed hereby (potentially) allowing more calls to be devirtualized.


lahma:
Thank you.

PR #1476, opened 2021-12-22
docs: (Triggers) Be explicit about trigger priority.

### LINKED ISSUE #1459: Priority documentation
Could you please document somewhere very clearly (other than in code) that the bigger number in Priority indicates higher priority? It's far from intuitive (for example, in Jira lower values indicate higher priority -  https://beam.apache.org/contribute/jira-priorities/ ).

jafin:
Resolves #1459


lahma:
Thank you!

PR #1479, opened 2021-12-26
Fix assertions to compare against LocalTime

jafin:
Fix assertions to compare against LocalTime to avoid error when run in positive UTC timezone environment.

When The Unit tests were run in UTC+10 timezone they would fail due to the comparison being off by one in UTC compare.


lahma:
Thank you!

Issue #988, opened 2020-10-11
Quartz.NET 4.0 Roadmap


lahma:
This is an umbrella issue gathering things that could be done in 4.0 timeframe.

- [ ] Refactor persistence API to be more document database friendly, https://github.com/quartznet/quartznet/issues/814 probably contributes to this
- [x] Use `ValueTask` instead of `Task`, RamJobStore doesn't need heavy machinery
- [x] Use `Microsoft.Extensions.Logging.Abstractions` instead of liblog
- [x] Add HTTP API for scheduler, maybe via Quartz.AspNetCore package, existing issue #511
- [x] Use the new TimeProvider abstraction #2079
- [x] Remove binary serialization support  #2324
- [x] Remove remoting support #2447
- [x] Remove full framework support #2226

Sicos1977:
A replacement for .net remoting like a rest api would also be a nice thing

lahma:
Thanks @Sicos1977 , added https://github.com/quartznet/quartznet/issues/511 to the list

Pinox:
An option to load persistent data from cloud provider like Azure Table Storage (No SQL) would be an awesome low cost solution for running quartz.  

Currently I have quartz running on seperate SQL DB and it's such a waste of money especially the fact the I'm not running many jobs on the DB.   Table storage where you only pay for the data that you use will make much more sense for many people as I think most Quartz db's are probably very small in size and with something like table storage it's cheap , accessible from anywhere with near infinite scaling.   A 1GB solution on Azure Table Storage will cost you $0.1265 per GB

https://azure.microsoft.com/en-gb/services/storage/tables/

lahma:
@Pinox what I've understood the row key is used for fast lookups in table storage. Quartz needs to check next fire time which is changing value for triggers and if you have thousands of triggers might become a problem. Azure table storage might not scale well in this case?

Pinox:
@lahma   I know there is a limit for transactions (20 000 per second) and entities (2000 per second.) per partition.  So yes if you do more than thousands per second it will not scale well on Azure Table Storage but you still have general NoSQL DB's as option for such scenario's.

I just think if Quartz was able to work on NoSQL then for smaller type workloads it will be a perfect fit for Azure Table Storage and if you really need massive record update per second then you can always throw it in bigger NoSQL DB like CosmoDB / MongoDB or any big DB vendor out there and then you are can use the most appropriate hardware for your task that is persistent and very cost effective.

[...]

PR #1480, opened 2021-12-26
Add microsoft logging and replace LibLog

jafin:
This is an extension of work started in #1086 by @lahma  Please peer review, and offer any improvements.  If later merged, suggest aborting #1086 as this would supersede that PR.

Feature supports #988 

+ Remove LibLog.
+ Added to Examples sample wireups for `Microsoft Logging` and `Serilog`. 
+ Added `SetLoggerFactory` to `QuartzConfigurator`, by default if not set, a NullLogger will be returned
+ Update all log instances to use `ILogger<T>`
+ Update log messages to MessageTemplates with template values as parameters.
+ Replaced `FailFastLoggerFactory` with `TestLoggerHelper` using MELT for log assertion.

The general approach, get logger via a service locator pattern, to inject logger in constructor appears it would need a much larger refactor, outside scope of this change.

:warning: NOTE: I have not run the integration tests, I've yet been unable to configure the required dependencies.

## TODO

Question:  `XMLSchedulingDataProcessor` takes logger in the constructor, which is different to everywhere else, perhaps this should be reverted to a service locator pattern and create a logger in the constructor for consistency?


lahma (changes requested):
Great work @jafin ! This is quite big PR so I'm afraid I need ask a couple of things.

* this requires rebase, can you do that? (might make it easier if you squash some commits first)
* can you separate documentation changes to a separate PR that I could review after this one?

Overall I'm eager to get this into v4 branch as this is very important groundwork. Sorry it took a while for me to go start reviewing this.

jafin:
@lahma 
- Doco has been rebased.
- Rebased on v4 branch.

Let me know if you need further changes.

src/Quartz.Examples/Program.cs
```diff
@@ -38,8 +38,20 @@ public static async Task Main()
         {
             try
             {
-                var logRepository = log4net.LogManager.GetRepository(Assembly.GetEntryAssembly());
-                log4net.Config.XmlConfigurator.Configure(logRepository, new System.IO.FileInfo("log4net.config"));
+                var logOption = AnsiConsole.Prompt(
```
lahma, line 41: Nice!

src/Quartz.Plugins/Plugin/History/LoggingTriggerHistoryPlugin.cs
```diff
@@ -213,7 +217,7 @@ public class LoggingTriggerHistoryPlugin : ISchedulerPlugin, ITriggerListener
         /// <summary>
         /// Logger instance to use. Defaults to common logging.
```
lahma, line 218: Seems outdated

lahma (approved):
Great work, thank you for finalizing this. Next step will be utilizing MS DI fully and injecting dependencies using it.

PR #1481, opened 2021-12-27
Update changelog - hyperlink issues

jafin:


lahma:
Thanks for going through the links but I think this might be a bit of a maintenance burden. Changelog mostly serves GitHub browsing and it's not published elsewhere except for the ZIP package.

jafin:
All good, i'll close.

PR #1490, opened 2022-01-12
Testcase for #1441

DirkieDurky:
As discussed in https://github.com/quartznet/quartznet/discussions/1441, here is a test case that properly breaks.
SystemTime.[UtcNow|Now] doesn't break it, we need to actually change the system time.

### DIFF
--- a/src/Quartz.Tests.Unit/SystemTimeChange.cs
+++ b/src/Quartz.Tests.Unit/SystemTimeChange.cs
@@ -1,0 +1,104 @@
+using System;
+using System.Runtime.InteropServices;
+using System.Threading.Tasks;
+using NUnit.Framework;
+using Quartz.Impl;
+
+namespace Quartz.Tests.Unit
+{
+    internal class SystemTimeChange
+    {
+        public static Boolean Ran;
+
+        [StructLayout(LayoutKind.Sequential)]
+        public struct SYSTEMTIME
+        {
+            public Int16 wYear;
+            public Int16 wMonth;
+            public Int16 wDayOfWeek;
+            public Int16 wDay;
+            public Int16 wHour;
+            public Int16 wMinute;
+            public Int16 wSecond;
+            public Int16 wMilliseconds;
+        }
+
+        [TestFixture]
+        public class Testcase : IJob
+        {
+            public Task Execute(IJobExecutionContext context)
+            {
+                Ran = true;
+                return Task.CompletedTask;
+            }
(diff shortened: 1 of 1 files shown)


lahma:
Thanks for this, and sorry it sat here for so long. Closing it now that the underlying problem is fixed — but the report was valid and the diagnosis in #1508 was spot on.

**Root cause**

The bug was a sleep duration computed by subtracting two *wall-clock* timestamps and then handed to a *monotonic* wait. When the system clock jumps backward by `D`, the computed duration inflates by `D`, and the thread then sleeps for `D` of **real** time — long after the clock has been restored. That is exactly the `Monitor.Wait(sigLock, timeUntilTrigger)` that @EbenZhang pinpointed in https://github.com/quartznet/quartznet/issues/1508#issuecomment-1057882900.

The same mistake appeared in three places:

| Where | Effect of a backward jump |
|---|---|
| `QuartzSchedulerThread` trigger wait | triggers stop firing |
| `MisfireHandler` | misfire recovery stops |
| `ClusterManager` | check-in stops; peer nodes declare this instance failed and recover its triggers |

**Fixes**

- #2929 capped the `QuartzSchedulerThread` wait at `idleWaitTime` — this is what @GieltjE and @DirkieDurky were hitting.
- The `MisfireHandler` and `ClusterManager` sleeps are now clamped to one `misfireHandlerFrequency` / `clusterCheckinInterval` respectively (3.x: #3147, 4.x: #3148).

**Why this particular test can't be merged**

`kernel32!SetSystemTime` is Windows-only, requires elevation, and mutating the machine clock inside CI isn't viable — a failed test run would leave the agent's clock wrong. Instead the clock is faked while monotonic timers keep running normally, which is precisely the condition that triggers the bug: `SystemTime.UtcNow` on 3.x, a custom `TimeProvider` on 4.x. See `SystemTimeClockJumpTest`, plus unit tests over the two sleep computations.

**What to expect now**

Recovery is bounded, not instantaneous. Quartz schedules against the wall clock, so moving the clock backward an hour genuinely does defer a trigger by an hour — that part is by design. What's fixed is that the scheduler now *notices* when the clock comes back:

- firing resumes within one `quartz.scheduler.idleWaitTime` (30 s default),
- with AdoJobStore, misfire handling within one `misfireHandlerFrequency` and cluster check-in within one `clusterCheckinInterval`.

Thanks again for the report and the reproduction — it's what made the failure mode concrete.

PR #1497, opened 2022-01-20
Make the health check not fail if there are transient failures

### LINKED ISSUE #1496: QuartzHealthCheck never recovers after detecting failure
**Describe the bug**

The error count can only be incremented but is never cleared if things start working.

**Version used**

3.3.3

**To Reproduce**

Start a quartz app with the DB down and then start the DB. The health check always reports an error.

**Expected behavior**

I'd expect the health check to start reporting success when the DB comes up

bencyoung-Fignum:
@lahma as discussed. I've not given this a full test, but it's in the direction I had in mind...

fixes #1496


bencyoung-Fignum:
https://github.com/quartznet/quartznet/issues/1496

lahma:
Have you tested and happy with the new behavior? Are the vscode task definition changes needed?

bencyoung-Fignum:
> 

The vscode changes just seemed to happen automatically in the latest version... I'll try and check it locally but it's not easy for me!

bencyoung-Fignum:
I've now tested this and it works, and is simpler than before

bencyoung-Fignum:
Test failure looks transient

lahma (approved):
I did some small tweaks and fixed the health check UI to show status so this could be verified. Looks good, thank you!

bencyoung-Fignum:
Thank you so much!

PR #1513, opened 2022-02-08
remove outdated Info about TriggerUtils

Strandinator:
I got a little confused when browsing the docs.
As far as I can tell TriggerUtils used to be able to create commonly used Triggers in a short way.
Currently it seems you are supposed to use the TriggerBuilder DSL.


lahma:
Thank you!

PR #1534, opened 2022-03-03
Update quick-start.md

marceloceccon:
In the first example of the code, it declares IScheduler sched = but later down the code it invokes scheduler.Start


lahma:
Thank you!

PR #1537, opened 2022-03-06
Make GetObjectData(SerializationInfo info, StreamingContext context) protected

drieseng:
Make GetObjectData(SerializationInfo info, StreamingContext context) protected.
Resolves item 1 of issue #1417.


drieseng:
@lahma I had hoped to introduce unit tests to verify that we can still deserialize version **0** of the **DirtyFlagMap** serialization "contract". In commit 750dc833e1b700fc6308b1695c1e7860324f3deb (dates back to yesterday or something, I'm sure you still recall) you introduced changes to allow **DirtyFlagMap** to deserialize that version.

If I try to deserialize that version, I get an exception because - in version **0** - **DirtyFlagMap** did not have generic type parameters (TKey, TValue). Did you introduce the changes in that commit to "unbreak" binary serialization or was it for **Json.NET**?

lahma:
Yeah l remember that like yesterday, that was 10 years ago, probably in a hangover, being young and still having the life ahead of me. Honestly I can't remember the specifics on this, it should mostly be about making sure fields were (de)serialized properly, I don't think I had anything related to type definition in mind back then... Maybe we don't need to sweat about it too much if it's been broken all along..

PR ready for merge?

drieseng:
@lahma Yes, it's ready. I actually wanted to make sure the "complexity" was necessary, and add a unit test to make sure we don't regress. I think it can be removed, but if you're not sure why you added it and if it fixed some scenario then we should keep it as is.

Perhaps I'll still add a unit test for it (without using the actual output of the V0 serialization), but I'll do that in a separate PR.

lahma:
Thank you!

PR #1548, opened 2022-03-20
Merge main into v4

jafin:
Sync merge `main` into `v4`


lahma:
I'm also considering the option of doing plain rebase for v4 before merging to main. V3.4 shouldn't be that far away and after that should branch that to 3.x maintenance.

jafin:
@lahma Ok sounds good, I'll cancel this PR.  Was just concerned about diverging to much from the inclusions since 3.3

PR #1559, opened 2022-03-25
Fix spelling.

jafin:
Fix minor spelling.


lahma (approved):
Thanks, especially for targeting v4 to keep v3 changes to mininum .

PR #1561, opened 2022-03-26
Use job and trigger listeners without configurations

VilleHakli:
Quartz will resolve job and trigger listeners from service provider only if there exists listener configuration class also in the service provider. This is added when IServiceCollectionQuartzConfigurator's add-methods are called. These add methods does not provide all the same overloads as service collection to add listeners and thus limits the possible ways to register listeners.

This PR changes the way Quartz resolves job and trigger listeners from service provider. All the listeners that are added to service collection will be used. If no configuration class is found, defaults are used.


lahma (approved):
Awesome, thank you!

PR #1564, opened 2022-03-27
fix (IntegrationTest): Ensure Scheduler removed in teardown, change test assertion

### LINKED ISSUE #1563: Integration Test JobTypeNotFoundShouldNotBlock does not cleanup scheduler  if fails and question on assertion.
**Describe the bug**

I am having a few issues running the integration test JobTypeNotFoundShouldNotBlock

* It appears to keep a scheduler thread running if the assertion fails.
* Assertion checks for TriggerState = 'Error'  Should it not be checking that TriggerState Is.Not.EqualTo(TriggerState.Blocked)?

**Version used**

main 618e7071588cc265bda94a76d96bc2da6652d7f8

**To Reproduce**

Run integration test `JobTypeNotFoundShouldNotBlock`

**Expected behavior**

Test pass/fail with cleanup.

**Additional context**

### Assertion
Change the test assertion be 
```c#
Assert.That(triggerState, Is.Not.EqualTo(TriggerState.Blocked));
```

instead of 
```c#
Assert.That(triggerState, Is.EqualTo(TriggerState.Error));
```

Given the test intent is `JobTypeNotFoundShouldNotBlock`
There is another motive here,  I actually can't get this test to pass it it's current form? If the assert was amended then it will pass.

### Test teardown.
If a test that uses the scheduler fails, I don't think cleanup is occurring as the test runner is hanging.
Suggest adding code to destroy running schedulers after test. AdoJobStoreSmokeTest:

```c#
[TearDown]
public async Task ShutdownSchedulers()
{
    var schedulers = await SchedulerRepository.Instance.LookupAll(CancellationToken.None);
    foreach (var scheduler in schedulers)
    {
        await scheduler.Shutdown(CancellationToken.None);
    }
}
```

can do PR will above changes if this makes sense.

jafin:
Ensure Scheduler is removed in teardown, change assertion to match test intent for test JobTypeNotFoundShouldNotBlock.

Closes #1563

### DIFF
--- a/src/Quartz.Tests.Integration/Impl/AdoJobStore/AdoJobStoreSmokeTest.cs
+++ b/src/Quartz.Tests.Integration/Impl/AdoJobStore/AdoJobStoreSmokeTest.cs
@@ -431,7 +431,7 @@ namespace Quartz.Tests.Integration.Impl.AdoJobStore
 
             manualResetEvent.Wait(TimeSpan.FromSeconds(20));
 
-            Assert.That(await scheduler.GetTriggerState(badTrigger.Key), Is.EqualTo(TriggerState.Error));
+            Assert.That(await scheduler.GetTriggerState(badTrigger.Key), Is.Not.EqualTo(TriggerState.Blocked));
         }
 
         private static async Task<IScheduler> CreateScheduler(NameValueCollection properties)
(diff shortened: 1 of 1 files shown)


lahma:
Thank you!

PR #1571, opened 2022-04-04
Pulled DisallowConcurrentExecution from JobBuilder class up to IJobCo…

saklig:
Pulled DisallowConcurrentExecution up to IJobConfigurator, so this property can be set when adding jobs from the ServiceCollectionExtensions.

Example:
services.AddQuartz(configurator =>
{
   configurator.AddJob(type, jobKey, jobConfig =>
   {
        jobConfig.DisallowConcurrentExecution();
   });
});

### DIFF
--- a/src/Quartz/IJobConfigurator.cs
+++ b/src/Quartz/IJobConfigurator.cs
@@ -163,5 +163,19 @@ namespace Quartz
         /// </remarks>
         /// <seealso cref="DisallowConcurrentExecutionAttribute"/>
         JobBuilder DisallowConcurrentExecution(bool concurrentExecutionDisallowed = true);
+
+        /// <summary>
+        /// Instructs the <see cref="IScheduler" /> whether or not job data should be re-stored when execution of the job completes.
+        /// </summary>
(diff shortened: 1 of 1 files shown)


lahma:
Thanks for the PR. This targets main so it will be part of v4 preview builds. If you wish it to land also on *possible* v3 patch release you should target 3.x branch.

drieseng:
@saklig, you may also want to expose **PersistJobDataAfterExecution**.

saklig:
> Thanks for the PR. This targets main so it will be part of v4 preview builds. If you wish it to land also on _possible_ v3 patch release you should target 3.x branch.

I would like to get these changes into 3.x as well, but the changes in this PR is dependent on the changes introduced in PR [1412](https://github.com/quartznet/quartznet/pull/1412), and judging only the conversation there, it seems like there might be some breaking changes for 3.x.

saklig:
> @saklig, you may also want to expose **PersistJobDataAfterExecution**.

Good point, I'll add that to the PR.

lahma (approved):

lahma:
Thank you!

PR #1575, opened 2022-04-07
Allow PersistJobDataAfterExecution and ConcurrentExecutionDisallowed to be explicitly in JobBuilder and pulled up to IJobConfigurator

saklig:
Cherry picked commits for exposing DisallowConcurrentExecution and PersistJobDataAfterExecution in IJobConfigurator, and dependent changes made to the JobBuilder and JobDetailImpl.
I see there is some changes to setters, from public to private, which might qualify as a breaking change in 3.x. In that case I will walk those changes back in a new commit to avoid breaking.

### DIFF
--- a/src/Quartz/IJobConfigurator.cs
+++ b/src/Quartz/IJobConfigurator.cs
@@ -163,5 +163,19 @@ namespace Quartz
         /// </remarks>
         /// <seealso cref="DisallowConcurrentExecutionAttribute"/>
         JobBuilder DisallowConcurrentExecution(bool concurrentExecutionDisallowed = true);
+
+        /// <summary>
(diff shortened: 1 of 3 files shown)


lahma:
@drieseng thoughts on having these on v3 as 3.5 release or similar? This is binary breaking change but I don't think we transfer job details on the wire.

drieseng:
@lahma I think this is safe to go into a 3.x release.

lahma:
Thanks for assessing this impact @drieseng !

lahma:
Thanks for the PR @saklig !

saklig:
@lahma Thanks for accepting it!

Issue #1453, opened 2021-12-02
Use injected repository service instead of static singleton


momvart:
**Is your feature request related to a problem? Please describe.**
I observed this problem when added quartz to my ASP.NET project and integration tests started to fail because of the error "Scheduler with the name xxx already exists". After investigations, I found out that this is because of a shared context between different instances of `WebApplicationFactory`. Reading the source codes led me to [this line](https://github.com/quartznet/quartznet/blob/42f0ec89bb567d40bfc771f56724746fe61831f0/src/Quartz/Impl/StdSchedulerFactory.cs#L1101) which shows a static repository is being used as a singleton. As the tests run in shared memory space, this static object is also shared between them.

Although the error could be solved by defining locks in custom hosted services, this behavior is breaking the isolation of tests and requires additional considerations in any case. For example, we cannot use separate databases for each test if they are going to run in parallel.

**Describe the solution you'd like**
Register the repository as a singleton service and inject it into the dependent classes.

Also relates to #399.

lahma:
I think this requires some breaking changes in order to work. Unless you can figure out a way to support in 3.x without breaking API and compatibility? Scheduling for v4 for now.

kolpav:
@lahma Is this problem still present on latest dev version? I added quartz into my integration tests using `WebApplicationFactory` with same problem described above its bummer quartz is not working well with it as the job scheduling is what I was looking the most to test. Is there some workaround for now? Knowing if this is still on your radar would be enough so I don't spend more time investigating well known problem. Other than that using quartz is a pleasure 🙏

lahma:
Probably is still present in the latest dev version. A lot of changes are needed to make Quartz use DI services instead of current configuration scheme (and still support the old one to a degree).

In tests you could manually try to remove scheduler from scheduler repository or assign unique name (GUID or something) for each created scheduler.

kolpav:
@lahma Exactly what I am trying right now, thank you for quick response.

rogierpennink:
> @lahma Exactly what I am trying right now, thank you for quick response.

I'm not sure if you ever got it to work. I ran into exactly the same issue: I'm trying to run multiple integration tests for jobs in parallel using `WebApplicationFactory`.

The solution that ended up working for me was instantiating a new `WebApplicationFactory` instance per test, along with an integration-test specific path for the quartz configuration in my Startup.cs file:

```csharp
services.AddQuartz(q =>
{
    q.UseInMemoryStore();
[...]

PR #1588, opened 2022-04-22
Make StdSchedulerFactory and ServiceCollectionSchedulerFactories threadsafe

### LINKED ISSUE #1587: StdSchedulerFactory and derived factories are not threadsafe
`GetScheduler(CancellationToken)` uses `SchedulerRepository.Instance` to check if the the scheduler already exists. If it does not, then then `Instantiate()` later adds the scheduler to the repository singleton:

https://github.com/quartznet/quartznet/blob/7d0dc1b6d9189b836d889e9be590d0ce380b5635/src/Quartz/Impl/StdSchedulerFactory.cs#L1099-L1113

In the intervening time another thread may call GetScheduler, resulting in an exception from SchedulerRepository.Bind:

https://github.com/quartznet/quartznet/blob/7d0dc1b6d9189b836d889e9be590d0ce380b5635/src/Quartz/Impl/SchedulerRepository.cs#L53-L64

This is problematic under dependency injection where the ISchedulerFactory seems to be the point of access for the scheduler. I see #1265, however this does not address the issue as the race occurs within StdSchedulerFactory prior to acquiring the semaphore.

Also related: #1453.

jcracknell:
fixes #1587

I moved the call to SchedulerRepository.Bind to GetScheduler, as this makes the registration process clearer.

You can't write tests against this as it would pollute the SchedulerRepository singleton as noted in #1453.

### DIFF
--- a/src/Quartz.Extensions.DependencyInjection/ServiceCollectionSchedulerFactory.cs
+++ b/src/Quartz.Extensions.DependencyInjection/ServiceCollectionSchedulerFactory.cs
@@ -20,8 +20,8 @@ namespace Quartz
         private readonly IServiceProvider serviceProvider;
         private readonly IOptions<QuartzOptions> options;
         private readonly ContainerConfigurationProcessor processor;
-        private bool initialized;
         private readonly SemaphoreSlim semaphore = new SemaphoreSlim(1, 1);
+        private IScheduler? initializedScheduler;
 
         public ServiceCollectionSchedulerFactory(
             IServiceProvider serviceProvider,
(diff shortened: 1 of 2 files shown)


jcracknell:
I thought about this some more last night and there's a second bug here where the ServiceCollectionSchedulerFactory would fail to initialize a new scheduler instance produced by StdSchedulerFactory in the event that the previously initialized factory was shut down:

https://github.com/quartznet/quartznet/blob/2a5e43d5df3ac1bfa9d43f7e65284db11bb512d8/src/Quartz/Impl/StdSchedulerFactory.cs#L1103-L1119

lahma (approved):
Looks good, thank you!

lahma:
@jcracknell I hate to ask, but would you be willing to create another PR against main as this doesn't merge cleanly due to some structural changes done there.

PR #1592, opened 2022-04-26
Add TryGet functions to JobDataMap and StringKeyDirtyFlagMap

asherber:
Added `TryGet` versions for `GetXxx`, `GetXxxValueFromString`, and `GetXxxValue` methods.

*   I was tempted to refactor many of the existing `GetXxx` and `GetXxxValue` methods, because most of each group follows the same pattern. I left them alone for now, and I similarly used a repetitive pattern for the new methods instead of refactoring.
*   Two of the types don't have the standard complement of three methods. I assumed there was a reason for this and did not add missing methods.
    *   `char` has `GetCharFromString` (only method named `GetXxxFromString`) and `GetChar`, but it does not have `GetCharValueFromString` or `GetCharValue`
    *   `Guid?` does not have `GetNullableGuidFromString`
*   There were no unit tests for the existing methods, so I did not add any for the new ones.
*   I'm happy to revise as needed.

#1582

### DIFF
--- a/src/Quartz/JobDataMap.cs
+++ b/src/Quartz/JobDataMap.cs
@@ -472,5 +472,301 @@ namespace Quartz
 
             return GetNullableGuid(key);
         }
+
+        /// <summary>
+        /// Try to retrieve the identified <see cref="int" /> value from the <see cref="JobDataMap" />.
+        /// </summary>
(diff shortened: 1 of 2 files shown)


lahma (approved):
This is great addition, thank you!

PR #1598, opened 2022-05-07
Allow benchmarks to run against .NET 6.0

drieseng:
If I do not set **ProduceReferenceAssembly** to **false**, I get the following error when I attempt to run our benchmarks against .NET 6.0:

> CSC : error CS0006: Metadata file 'C:\development\quartznet\src\Quartz.Benchmark\bin\Release\net6.0\ref\Quartz.Benchmark.dll' could not be found [C:\development\quartznet\src\Quartz.Benchmark\bin\release\net6.0\eabb1d91-56a8-446e-b662-2e305185d641\BenchmarkDotNet.Autogenerated.csproj]
Build FAILED.
> CSC : error CS0006: Metadata file 'C:\development\quartznet\src\Quartz.Benchmark\bin\Release\net6.0\ref\Quartz.Benchmark.dll' could not be found [C:\development\quartznet\src\Quartz.Benchmark\bin\release\net6.0\eabb1d91-56a8-446e-b662-2e305185d641\BenchmarkDotNet.Autogenerated.csproj]

I also added a test that I forgot to commit in my previous PR.

### DIFF
--- a/src/Quartz.Benchmark/Directory.Build.props
+++ b/src/Quartz.Benchmark/Directory.Build.props
@@ -1,0 +1,7 @@
+<Project>
+  <Import Project="..\Directory.Build.props" />
+
+  <PropertyGroup>
+    <ProduceReferenceAssembly>false</ProduceReferenceAssembly>
+  </PropertyGroup>
+</Project>
--- a/src/Quartz.Tests.Unit/Collections/OrderedDictionaryTest.cs
+++ b/src/Quartz.Tests.Unit/Collections/OrderedDictionaryTest.cs
@@ -1,0 +1,36 @@
+using NUnit.Framework;
+using Quartz.Collections;
+using System;
+
+namespace Quartz.Tests.Unit.Collections
+{
+    [TestFixture]
+    internal class OrderedDictionaryTest
+    {
+        [Test]
+        public void Values_Empty_ToArray()
+        {
+            var dictionary = new OrderedDictionary<string, int>();
+            var valuesCollection = dictionary.Values;
+
+            var valuesArray = valuesCollection.ToArray();
+            Assert.AreSame(Array.Empty<int>(), valuesArray);
+        }
+
+        [Test]
+        public void Values_NotEmpty_ToArray()
+        {
+            var dictionary = new OrderedDictionary<string, int>();
+            dictionary.Add("A", 5);
+            dictionary.Add("Z", 2);
+
+            var valuesCollection = dictionary.Values;
+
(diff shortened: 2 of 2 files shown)


src/Quartz.Benchmark/Directory.Build.props
```diff
@@ -0,0 +1,7 @@
+<Project>
+  <Import Project="..\Directory.Build.props" />
+
+  <PropertyGroup>
+    <ProduceReferenceAssembly>false</ProduceReferenceAssembly>
```
lahma, line 5: I guess setting directly in csproj would be easier?

src/Quartz.Tests.Unit/Collections/OrderedDictionaryTest.cs
```diff
@@ -0,0 +1,36 @@
+using NUnit.Framework;
+using Quartz.Collections;
+using System;
+
+namespace Quartz.Tests.Unit.Collections
```
lahma, line 5: For new code can use file-scoped namespaces

src/Quartz.Tests.Unit/Collections/OrderedDictionaryTest.cs
```diff
@@ -0,0 +1,36 @@
...
+{
+    [TestFixture]
```
lahma, line 7: This has been  redundant for a good while now, Test attribute is enough for discovery.

lahma (approved):

src/Quartz.Tests.Unit/Collections/OrderedDictionaryTest.cs
drieseng: I'm getting out of touch, it seems. Thanks for the info.

src/Quartz.Tests.Unit/Collections/OrderedDictionaryTest.cs
drieseng: Indeed, it's still growing on me. Other "new" C# features like **using declarations** will always get lost while travelling the distance between my brain and my keyboard :p

src/Quartz.Benchmark/Directory.Build.props
drieseng: Yeah, I don't see why not. I'll change this in my next PR (unless you want a isolated PR for this).

lahma: No worries, can be part of some future pr

PR #1600, opened 2022-05-09
Optimize notify of trigger listeners by only locking and allocating w…

drieseng:
Optimize notify of trigger listeners by only locking and allocating when non-global trigger listeners have been added.
No longer define properties and methods for accessing or manipulating internal trigger listeners.

**.NET 4.8**

|                                                      Method | Branch |      Mean |     Error |    StdDev |  Gen 0 | Allocated |
|------------------------------------------------------------ |-------:|-----------|----------:|----------:|-------:|----------:|
| Success_NoTriggerListenersAndSingleJobListener_MayFireAgain | main   |  1322  ns | 2.62 ns   | 2.45 ns   | 0.1163 |     489 B |
| Success_NoTriggerListenersAndSingleJobListener_MayFireAgain | pr     |  936.7 ns | 3.30 ns   | 2.76 ns   | 0.0858 |     361 B |

**.NET 6.0**

|                                                      Method | Branch |      Mean |     Error |    StdDev |  Gen 0 | Allocated |
|------------------------------------------------------------ |-------:|-----------|----------:|----------:|-------:|----------:|
| Success_NoTriggerListenersAndSingleJobListener_MayFireAgain | main   | 902.0 ns  | 3.47 ns   | 3.24 ns   | 0.0668 |     280 B |
| Success_NoTriggerListenersAndSingleJobListener_MayFireAgain | pr     | 617.6 ns  | 1.12 ns   | 0.99 ns   | 0.0362 |     152 B |

### DIFF
--- a/src/Quartz/Core/ListenerManagerImpl.cs
+++ b/src/Quartz/Core/ListenerManagerImpl.cs
@@ -12,14 +12,12 @@ namespace Quartz.Core
     public class ListenerManagerImpl : IListenerManager
     {
         private readonly object globalJobListenerLock = new object();
-
         private OrderedDictionary<string, IJobListener>? globalJobListeners;
-
-        private readonly OrderedDictionary<string, ITriggerListener> globalTriggerListeners = new OrderedDictionary<string, ITriggerListener>(10);
-
         private Dictionary<string, List<IMatcher<JobKey>>>? globalJobListenersMatchers;
 
-        private readonly Dictionary<string, List<IMatcher<TriggerKey>>> globalTriggerListenersMatchers = new Dictionary<string, List<IMatcher<TriggerKey>>>(10);
+        private readonly object globalTriggerListenerLock = new object();
+        private OrderedDictionary<string, ITriggerListener>? globalTriggerListeners;
+        private Dictionary<string, List<IMatcher<TriggerKey>>>? globalTriggerListenersMatchers;
 
         private readonly List<ISchedulerListener> schedulerListeners = new List<ISchedulerListener>(10);
 
@@ -34,17 +32,18 @@ namespace Quartz.Core
         {
             if (string.IsNullOrEmpty(jobListener.Name))
             {
-                throw new ArgumentException("JobListener name cannot be empty.");
+                throw new ArgumentException($"{nameof(jobListener.Name)} cannot be empty.", nameof(jobListener));
             }
 
             lock (globalJobListenerLock)
             {
+                // Add or replace the job listener
                 globalJobListeners ??= new OrderedDictionary<string, IJobListener>();
                 globalJobListeners[jobListener.Name] = jobListener;
 
                 if (matchers != null && matchers.Count > 0)
                 {
-                    // Add or replace matchers for the job listener
+                    // Add or replace the matchers for the job listener
                     globalJobListenersMatchers ??= new Dictionary<string, List<IMatcher<JobKey>>>();
                     globalJobListenersMatchers[jobListener.Name] = new List<IMatcher<JobKey>>(matchers);
                 }
@@ -175,6 +174,7 @@ namespace Quartz.Core
                 }
                 else
                 {
+                    // Add or replace the matchers for the job listener
                     globalJobListenersMatchers ??= new Dictionary<string, List<IMatcher<JobKey>>>();
                     globalJobListenersMatchers[listenerName] = new List<IMatcher<JobKey>>(matchers);
                 }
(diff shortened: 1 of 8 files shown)


drieseng:
Marked as draft as I'm still working on adding more unit tests and improving documentation.

drieseng:
@lahma This one is ready for review now.

Once this is merged, I see two remaining candidates for optimization for this specific scenario.
I'm just laying out further options to improve performance, not trying to convince you to take these changes.
The **ExecutingJobsManager** "proposal" is definitely not one to rush.

### Remove use of **Quartz.Core.Context**

*Context.CallerId* is an *AsyncLocal<Guid?>* that we assign a new **Guid** to in `JobRunShell.Run(CancellationToken cancellationToken = default)` and `QuartzSchedulerThread.Run()`.

The assigned value is used by **JobStoreSupport**. I can't immediately tell if the assign in `JobRunShell.Run(CancellationToken cancellationToken = default)` is actually necessary.

Removing it (branch 'experiment') gives us a nice win:

|                                                      Method | Branch |     Mean |   Error |  StdDev |  Gen 0 | Allocated |
|------------------------------------------------------------ |-------:| --------:|--------:|--------:|-------:|----------:|
| Success_NoTriggerListenersAndSingleJobListener_MayFireAgain | 559fd40c29f3e63ca19659ee1e995c50440bec02 | 1566,0 ns | 1.14 ns | 1.01 ns | 0.2403 |   1,008 B |
| Success_NoTriggerListenersAndSingleJobListener_MayFireAgain | main   | 902.0 ns | 3.47 ns | 3.24 ns | 0.0668 |     280 B |
| Success_NoTriggerListenersAndSingleJobListener_MayFireAgain | pr     | 617.6 ns | 1.12 ns | 0.99 ns | 0.0362 |     152 B |
| Success_NoTriggerListenersAndSingleJobListener_MayFireAgain | experiment | 446.2 ns | 2.86 ns | 2.54 ns | 0.0114 |      48 B |

### Ship **ExecutingJobsManager** as an opt-in **IJobListener**

I'm not sure you would want to consider this as we'd have to remove the **NumJobsExecuted** and **CurrentlyExecutingJobs** properties from **QuartzScheduler**, and users would have to obtain this information directly from **ExecutingJobsManager**.

Doing this on top of the removal of / alternative for **Quartz.Core.Context** gets us to zero allocations and another nice performance boost:

|                                                      Method | Branch |     Mean |   Error |  StdDev |  Gen 0 | Allocated |
|------------------------------------------------------------ |-------:| --------:|--------:|--------:|-------:|----------:|
| Success_NoTriggerListenersAndSingleJobListener_MayFireAgain | 559fd40c29f3e63ca19659ee1e995c50440bec02 | 1566,0 ns | 1.14 ns | 1.01 ns | 0.2403 |   1,008 B |
| Success_NoTriggerListenersAndSingleJobListener_MayFireAgain | main   | 902.0 ns | 3.47 ns | 3.24 ns | 0.0668 |     280 B |
| Success_NoTriggerListenersAndSingleJobListener_MayFireAgain | pr     | 617.6 ns | 1.12 ns | 0.99 ns | 0.0362 |     152 B |
| Success_NoTriggerListenersAndSingleJobListener_MayFireAgain | experiment | 301.0 ns | 2.77 ns | 2.45 ns | -      |         - |

src/Quartz/Core/ListenerManagerImpl.cs
```diff
@@ -12,14 +12,12 @@ namespace Quartz.Core
...
-
         private OrderedDictionary<string, IJobListener>? globalJobListeners;
-
-        private readonly OrderedDictionary<string, ITriggerListener> globalTriggerListeners = new OrderedDictionary<string, ITriggerListener>(10);
-
         private Dictionary<string, List<IMatcher<JobKey>>>? globalJobListenersMatchers;
 
-        private readonly Dictionary<string, List<IMatcher<TriggerKey>>> globalTriggerListenersMatchers = new Dictionary<string, List<IMatcher<TriggerKey>>>(10);
+        private readonly object globalTriggerListenerLock = new object();
+        private OrderedDictionary<string, ITriggerListener>? globalTriggerListeners;
```
lahma, line 19: This is quite the micro-optimization for applications that usually have single scheduler

lahma (approved):
Thanks for working out the PR. Generally I think reducing allocations for core services can be somewhat time wasted compared to the time building the actual scheduler can take.

drieseng:
> Thanks for working out the PR. Generally I think reducing allocations for core services can be somewhat time wasted compared to the time building the actual scheduler can take.

Building the scheduler is a one-time cost. Paying extra for each job that is triggered is a cost that you pay time and time again.
In my project, we actually schedule millions of jobs a day. Personally, I don't consider my effort to be "time wasted".

Do you see a way to avoid the use of assigning of Context.CallerId in `JobRunShell.Run(CancellationToken cancellationToken = default)`?

lahma:
> > Thanks for working out the PR. Generally I think reducing allocations for core services can be somewhat time wasted compared to the time building the actual scheduler can take.
> 
> Building the scheduler is a one-time cost. Paying extra for each job that is triggered is a cost that you pay time and time again. In my project, we actually schedule millions of jobs a day. Personally, I don't consider my effort to be "time wasted".

I think my comment came out a bit wrong. The thing was about trying to avoid allocation lists/dictionary instance members for core services which are instantiated only once per scheduler lifetime. Reducing recurring runtime allocations is of course awesome!

> Do you see a way to avoid the use of (assigning of) **Context.CallerId** in `JobRunShell.Run(CancellationToken cancellationToken = default)`?

Because it's used as locking handle in downstream ado job store it might be a bit tricky to get to flow downstream without changing all the APIs. I think `JobRunShell` and `QuartzSchedulerThread` would need to pass the value explicitly. Been quite the while I've been in contact with that async local so my memory isn't serving me the best here.

PR #1609, opened 2022-05-20
Fix pathname in job-stores documentation.

merlinschumacher:
The path in the section about the ADO storage is not correct. This fixes it.


lahma:
Thank you!

Issue #1523, opened 2022-02-21
Remote Microservices Code - JobBuilder.Create method should support string for fully qualified assembly name


siva538:
**Is your feature request related to a problem? Please describe.**
Currently Quartz supports multiple schedulers. In our current scenario, the setup of the jobs is happening from a location where the code access do not present for the remote Job types. This is causing massive over engineering of dynamically generating the class type etc. that is not needed in the first place.

i.e. public static JobBuilder Create(Type jobType) or the generic method, mandatorily needs Type available during the time of creation of the Job. 

**Describe the solution you'd like**

public static JobBuilder Create(string jobType); with this string overload, the string passed in to this method can directly be inserted into the backend database, instead of the quartz library trying to use - AssemblyQualifiedNameWithoutVersion()

**Describe alternatives you've considered**
As I mentioned we had to generate a type dynamically with the system.reflection support, which is not only difficult to implement but also not maintainable.

**Additional context**
NA

lahma:
Should be supported scenario with v4 now.

PR #1610, opened 2022-05-21
Serialized field for Job Type to enable maintenance on Jobs that cannot be constructed to their Type

### LINKED ISSUE #1560: Introduce JobType to allow storing job's type information without actual Type instace
**Describe the bug**

In JobDetailImpl, the constructor accepts a `Type? jobType` 

https://github.com/quartznet/quartznet/blob/9e29c25f7d245eab50ba5a8bd5bd4d66bc05cf27/src/Quartz/Impl/JobDetailImpl.cs#L184-L200

Yet, if the value is null, when it is assigned in the constructor, it will throw an exception.  

https://github.com/quartznet/quartznet/blob/9e29c25f7d245eab50ba5a8bd5bd4d66bc05cf27/src/Quartz/Impl/JobDetailImpl.cs#L312-L329
**Version used**
V4

## Repro

Using QuartzScheduler.UnscheduleJob 
A call to `SelectJobForTrigger` with `loadJobType = false`   will cause an exception due to JobType being set to null.

## Questions

Should this continue to operate like in 3.3 and allow the null jobType through, perhaps will a null check in JobDetailImpl like

```c#
  if (jobType != null) //cannot set JobType to null will throw, yet unset is ok.
      JobType = jobType!;
```

or is the intent to not allow null Jobtypes into JobDetailImpl?

jafin:
Reduce the need to construct the stored job `Type` , allowing maintenance tasks such as `UnschduleJob` to operate when the stored job Type can no longer be constructed.
JobTypeSerialized is stored and can be retrieved as Type still via a Lazy getter

It should be a `Non-breaking change`, but if external code depended on exceptions being thrown for invalid job type retrieval this might introduce unexpected results

Should resolve #1560  - (V4) Error in JobDetailImpl due to null JobType being passed

fixes #1560

### DIFF
--- a/src/Quartz.Tests.Integration/JobMaintenanceTests.cs
+++ b/src/Quartz.Tests.Integration/JobMaintenanceTests.cs
@@ -1,0 +1,81 @@
+using System.Threading;
+using System.Threading.Tasks;
+
+using FluentAssertions;
+
+using Microsoft.Data.SqlClient;
+
+using NUnit.Framework;
+
+using Quartz.Tests.Integration.TestHelpers;
+
+namespace Quartz.Tests.Integration;
+
+[TestFixture]
+public class JobMaintenanceTests : IntegrationTest
+{
+    [Test]
+    public async Task CanUnscheduleNonConstructableTypeJobs()
+    {
+        var scheduler = await GetACleanScheduler();
+
+        const string jobKey = "CanDeleteUnknownTypeJobs";
+        const string triggerKey = "CanDeleteUnknownTypeJobsTrigger";
+        var jobDetail = JobBuilder.Create<KnownJobType>()
+            .WithIdentity(jobKey)
+            .UsingJobData("jfoo", "bar")
+            .Build();
+
+        var trigger = TriggerBuilder.Create()
+            .WithIdentity(triggerKey)
+            .WithCronSchedule("0 0 0 * * ?")
+            .UsingJobData("tfoo", "bar")
+            .Build();
+
+        await scheduler.ScheduleJob(jobDetail, trigger);
+
+        RenameJobClass("Quartz.Tests.Integration.JobMaintenanceTests+UnKnownJobType");
+
+        // assert job is stored
+        var storedJobDetail = await scheduler.GetJobDetail(new JobKey(jobKey));
+        var storedJobMap = storedJobDetail.JobDataMap;
+        Assert.That(storedJobMap.Dirty, Is.False);
+
+        var storedTrigger = await scheduler.GetTrigger(new TriggerKey(triggerKey));
+        var storedTriggerMap = storedTrigger.JobDataMap;
+        Assert.That(storedTriggerMap.Dirty, Is.False);
+
+        // assert can unSchedule
+        var unscheduleResponse = await scheduler.UnscheduleJob(new TriggerKey(triggerKey), CancellationToken.None);
+        unscheduleResponse.Should().BeTrue();
+    }
+
+    private async Task<IScheduler> GetACleanScheduler()
+    {
+        var scheduler = await SchedulerHelper.CreateScheduler(nameof(JobMaintenanceTests));
+        await scheduler.Clear();
+        return scheduler;
+    }
+
+    // Rename the Job Type ClassName
+    private static void RenameJobClass(string jobClassName)
+    {
+        using var dbConnection = new SqlConnection(TestConstants.SqlServerConnectionString);
+
+        var sql = $@"
+update {SchedulerHelper.TablePrefix}JOB_DETAILS
+set JOB_CLASS_NAME = '{jobClassName}'
+where SCHED_NAME = 'JobMaintenanceTestsScheduler'";
+        dbConnection.Open();
+        using var command = new SqlCommand(sql, dbConnection);
+        command.ExecuteNonQuery();
+    }
+
+    public class KnownJobType : IJob
+    {
+        public Task Execute(IJobExecutionContext context)
+        {
+            return Task.CompletedTask;
+        }
+    }
+}
--- a/src/Quartz.Tests.Integration/JobMaintenanceTests.cs
+++ b/src/Quartz.Tests.Integration/JobMaintenanceTests.cs
@@ -11,7 +11,6 @@ using Quartz.Tests.Integration.TestHelpers;
 
 namespace Quartz.Tests.Integration;
 
-[TestFixture]
 public class JobMaintenanceTests : IntegrationTest
 {
     [Test]
--- a/src/Quartz.Tests.Integration/Quartz.Tests.Integration.csproj
+++ b/src/Quartz.Tests.Integration/Quartz.Tests.Integration.csproj
@@ -39,7 +39,4 @@
     <Reference Include="Microsoft.CSharp" />
     <PackageReference Include="Oracle.ManagedDataAccess" Version="19.7.0" />
   </ItemGroup>
-  <ItemGroup>
-    <Folder Include="TestHelpers\" />
-  </ItemGroup>
 </Project>
--- a/src/Quartz.Tests.Integration/Quartz.Tests.Integration.csproj
+++ b/src/Quartz.Tests.Integration/Quartz.Tests.Integration.csproj
@@ -13,6 +13,7 @@
     <PackageReference Include="FakeItEasy" Version="7.3.1" />
     <PackageReference Include="GitHubActionsTestLogger" Version="2.0.0-alpha" PrivateAssets="all" />
     <PackageReference Include="FirebirdSql.Data.FirebirdClient" Version="8.5.4" />
+    <PackageReference Include="FluentAssertions" Version="6.7.0" />
     <PackageReference Include="MELT" Version="0.8.0" />
     <PackageReference Include="Microsoft.Data.SqlClient" Version="2.0.0" />
     <PackageReference Include="Microsoft.Data.SQLite" Version="3.1.6" />
@@ -37,5 +38,8 @@
   <ItemGroup Condition=" '$(TargetFramework)' == 'net472' ">
     <Reference Include="Microsoft.CSharp" />
     <PackageReference Include="Oracle.ManagedDataAccess" Version="19.7.0" />
+  </ItemGroup>
+  <ItemGroup>
+    <Folder Include="TestHelpers\" />
   </ItemGroup>
 </Project>
--- a/src/Quartz/IJobDetail.cs
+++ b/src/Quartz/IJobDetail.cs
@@ -19,8 +19,6 @@
 
 #endregion
 
-using System;
-
 using Quartz.Impl;
 namespace Quartz
 {
@@ -62,14 +60,9 @@ namespace Quartz
         string? Description { get; }
 
         /// <summary>
-        /// Get or sets the instance of <see cref="IJob" /> that will be executed.
+        /// Get the instance of <see cref="IJob" /> that will be executed.
         /// </summary>
-        Type JobType { get; }
-
-        /// <summary>
-        /// JobType type in serialized format.
-        /// </summary>
-        JobType JobTypeWithStorage { get; }
+        JobType JobType { get; }
 
         /// <summary>
         /// Get or set the <see cref="JobDataMap" /> that is associated with the <see cref="IJob" />.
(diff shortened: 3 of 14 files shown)


jafin:
For review, if there are better names to use than StorableTypeName etc let me know.

src/Quartz.Tests.Integration/JobMaintenanceTests.cs
```diff
@@ -0,0 +1,81 @@
...
+
+using Microsoft.Data.SqlClient;
+
+using NUnit.Framework;
+
+using Quartz.Tests.Integration.TestHelpers;
+
+namespace Quartz.Tests.Integration;
+
+[TestFixture]
```
lahma, line 14: nit: no need for TestFixture

src/Quartz.Tests.Integration/Quartz.Tests.Integration.csproj
```diff
@@ -37,4 +38,7 @@
     <Reference Include="Microsoft.CSharp" />
     <PackageReference Include="Oracle.ManagedDataAccess" Version="19.7.0" />
   </ItemGroup>
+  <ItemGroup>
```
lahma, line 41: looks unnecessary

src/Quartz/IJobDetail.cs
```diff
@@ -65,6 +66,11 @@ public interface IJobDetail
         /// </summary>
         Type JobType { get; }
 
+        /// <summary>
```
lahma, line 69: I think we should just break the API and change `JobType` to be of type `JobType`, maybe even change the property name to `Type`?

src/Quartz/Impl/JobType.cs
```diff
@@ -0,0 +1,75 @@
+using System;
+
+using Quartz.Util;
+
+namespace Quartz.Impl;
+
+/// <summary>
+/// Store the Job Type and backing field name for serialization
+/// </summary>
+public class JobType
```
lahma, line 10: this should be sealed, I think also allowing implicit conversion for `Type` might be a good thing?

src/Quartz/Impl/JobType.cs
```diff
@@ -0,0 +1,75 @@
...
+{
+    private string storableTypeName = string.Empty;
+    private Lazy<Type> lazyType = new(() => throw new InvalidOperationException("Type not defined"));
+
+    public JobType()
```
lahma, line 15: Probably doesn't make sense to allow without parameter, should be private

src/Quartz/Impl/JobType.cs
```diff
@@ -0,0 +1,75 @@
...
+    {
+    }
+
+    public JobType(string? typeStorageName)
```
lahma, line 19: I think the logical name would be `typeName`, like it's named in `Type.GetType`?

src/Quartz/Impl/JobType.cs
```diff
@@ -0,0 +1,75 @@
...
+    /// </summary>
+    public JobType(Type type)
+    {
+        Type = type;
+    }
+
+    /// <summary>
+    /// Type as a string represented storable name.
+    /// </summary>
+    public string StorableTypeName
```
lahma, line 35: I think this is just `Name`, or `FullName`, should follow same sort of naming logic in `Type`

src/Quartz/JobBuilder.cs
```diff
@@ -157,7 +159,7 @@ public IJobDetail Build()
             }
 
             return new JobDetailImpl(Key ?? new JobKey(Guid.NewGuid().ToString()),
-                                     jobType,
+                                     new JobType(jobTypeSerialized),
```
lahma, line 162: implicit conversion operator between Type and and JobType should make this redundant

src/Quartz/JobBuilder.cs
```diff
@@ -269,6 +271,17 @@ public JobBuilder WithDescription(string? description)
             return this;
         }
 
+        /// <summary>
+        /// Set the JobType Serialized name
+        /// </summary>
+        /// <param name="jobTypeSerialized">the serialized job Type</param>
+        /// /// <returns>the updated JobBuilder</returns>
+        public JobBuilder WithJobTypeSerialized(string? jobTypeSerialized)
```
lahma, line 279: this should probably be `OfType(string typeName)` to use same naming and overloads as passing Type, JobType and Type should be interchangeable concepts

lahma (changes requested):
Great initiative! I think we can break things with v4 so I don't see why we would need to have two competing concepts (Type and JobType), everything should be JobType.

jafin:
@lahma thanks I'll update based on your feedback.

Are you ok with `JobDetailImpl()` constructor being removed?  The only usage in the library is via 2 tests, which can be corrected. 
I can't see a reason to construct this without specifying a name/category and Type (which is done via the other constructors)
It will avoid making JobType property nullable is all.

lahma:
Removing constructor allowing invalid state should be fine. Only thing that springs to my mind is whether this can break serialization? Like full framework remoting probably allows to pass job impls.

jafin:
@lahma  I've updated code with suggested changes for review.

I noted a request in #1523 to allow the creation of a JobType remotely, where the type is not known to the remote code. This should enable this.

### Remoting Serialization
Tests exist in `JobDetailTest -> JobDetailsShouldBeSerializable` to validate remoting serialization.

> Constructors are not called when objects are deserialized.
[src](https://docs.microsoft.com/en-us/dotnet/standard/serialization/serialization-guidelines)

### Breaking changes

- JobType parameterless constructor removed, construct with either `Type FullName` or Job `Type` now.

lahma (approved):

lahma:
Great work and looks good, I did couple small tweaks I didn't want to bother you with and slow down the process of getting into main.

src/Quartz/Impl/JobType.cs
```diff
@@ -0,0 +1,97 @@
...
+
+namespace Quartz.Impl;
+
+/// <summary>
+/// Store the Job Type and FullName for serialization
+/// </summary>
+[Serializable]
+public sealed class JobType
+{
+    private Lazy<Type> type = new(() => throw new InvalidOperationException("Type not defined"));
```
drieseng, line 13: Do we want to pay this extra cost for creating a job type:

|        Method |        Mean |     Error |    StdDev |      Median |  Gen 0 | Allocated |
|-------------- |------------:|----------:|----------:|------------:|-------:|----------:|
| Type |  16.9039 ns | 0.1538 ns | 0.1364 ns |  16.8752 ns | 0.0057 |      24 B |
| JobType | 162.0058 ns | 2.1970 ns | 1.9476 ns | 161.9113 ns | 0.1185 |     497 B |

Once initialized the extra cost is negligable:

|        Method |        Mean |     Error |    StdDev |      Median |  Gen 0 | Allocated |
|-------------- |------------:|----------:|----------:|------------:|-------:|----------:|
|      Type |   0.0031 ns | 0.0120 ns | 0.0106 ns |   0.0000 ns |      - |         - |
|      JobType |   4.1255 ns | 0.1113 ns | 0.2646 ns |   4.0576 ns |      - |         - |

Use of **Lazy\<T>** should also have impact on concurrency. Haven't had time to benchmark this.

src/Quartz/Impl/JobType.cs
```diff
@@ -0,0 +1,97 @@
...
+    {
+        if (jobType.AssemblyQualifiedName == null)
+        {
+            throw new ArgumentException("Cannot determine job type name when type's AssemblyQualifiedName is null");
+        }
+
+        return jobType.AssemblyQualifiedNameWithoutVersion();
+    }
+
+    public static implicit operator Type(JobType jobType) => jobType.Type;
```
drieseng, line 79: I don't like implicit conversions ...

drieseng (changes requested):

drieseng:
@lahma I was too slow, it seems ....

drieseng:
@lahma The way I see it, we only have a single use case for building a **IJobDetail** without a concrete job type: remove a trigger for a job type that cannot be resolved.

Why not introduce another method that can be used by `JobStoreSupport.RemoveTrigger(...)` that does not return an *IJobDetail**, but only a **JobKey** (or a tuple representing the **JobKey** and its **durable** characteristic)?

lahma:
Retrieve list of job details to a process which does not have job type library. That's the main use case.

drieseng:
Can you have a look at my remarks. I think most of them remain valid.

lahma:
I'll have a closer look when I find the time, thanks for doing the review.

jafin:
Hi @drieseng / @lahma ,  Thanks for the continued review.
Just to note all points raised:

> @lahma The way I see it, we only have a single use case for building a IJobDetail without a concrete job type: remove a trigger for a job type that cannot be resolved.

@drieseng There is another use case presented in #1523 , ability to create jobs remotely without having access to the Type.  This sounds like a valid use case.

> I don't like implicit conversions ...

I added this at the request of @lahma ,  I don't have a strong opinion on this, I think it makes it easier for the consumer, but uncertain what you both agree on.

> Do we want to pay this extra cost for creating a job type:

I can't answer that, I was implementing a feature fix, not directly focused on performance.  If there is a better/faster way while still achieving the intent of the change, please advise.  
I see that [Lazy<T> performance is attributed to Activator.CreateInstance](https://stackoverflow.com/questions/11236855/is-using-lazyt-bad-for-performance)

Would removing the Lazy<T> and constructing the Type once on first get, then storing in a backing field be more appropriate?


> public string FullName
> Please remove default, and always assign in actors.

Yes, I'll fix that.

> public JobType(string? fullName)
> Should not be nullable (since it is in fact not allowed to be null).

Yes will fix.

I'll do another fixup PR with the issues above.  Leaving out changes on implicit conversions, and await further advice on performance implications for now.

drieseng:
@jafin Let me have a look at what we can do for performance, if necessary. The cost of creating an **IJobDetail* is not that important.

A few options to consider to reduce the cost of **JobType**:
** Allow **JobType** to have a reference to both a type or a full name.  If constructed using a **Type**, have **JobType.FullName** just return the **FullName** of the **Type** (or the assembly qualified name without version, more on this below) and have **Type** return the **Type** instance that was used to construct the **JobType**.
When constructed using an assembly qualfified name (without version) have, then lazily create the **Type** when the **JobType.Type** is accessed. This can be done with or without **Lazy\<T>**, to be investigated.
* Redefine **JobType** as struct instead of a reference type. Not too sure I like this.

I gave it a little more though, and really think we were too fast in merging this PR.
On the surface it gives you the impression that nothing changed (eg. by having an implicit conversion to **Type**), and it's that what makes me uneasy with it.

For example, we have the following code in **QuartzDiagnostic** that is "broken" by this PR:
```cs
.WithTag(DiagnosticHeaders.JobType, jobContext.JobDetail.JobType.ToString())
```

Same for someone that currently uses **IJobDetail.JobType.FullName**.  I'm sure they do not expect the result to be the assembly qualified name (without version).

Of course it's ok to introduce breaking changes in v4. But with this PR we're introducing breaking changes while - at the same time - we try (and somehow fail) to hide these breaking changes.

As always, it's far easier to identify what is "wrong" than to come up with a solution. I'll give this some thought and try to give more constructive feedback later.

lahma:
> I gave it a little more though, and really think we were too fast in merging this PR.
> On the surface it gives you the impression that nothing changed (eg. by having an implicit conversion to **Type**), and it's that what makes me uneasy with it.
> 
> For example, we have the following code in **QuartzDiagnostic** that is "broken" by this PR:
> ```cs
> .WithTag(DiagnosticHeaders.JobType, jobContext.JobDetail.JobType.ToString())
> ```

That's a good example where version should not matter. Assembly qualified name without version is correct. This PR was very important progress to remedy a long-standing issue.

> Of course it's ok to introduce breaking changes in v4. But with this PR we're introducing breaking changes while - at the same time - we try (and somehow fail) to hide these breaking changes.

V3 was about keeping Java version resemblance, but as there has been very little progress on that front, probably a lost cause. Binary/database compatibility is important, API can be improved with breaking changes.

drieseng:
@lahma jobContext.JobDetail.JobType.ToString() will always return **Quartz.Impl.JobType**. I don't think that's what was intended. This can be fixed of course, but I won't necessarily match what was previous returned (by **Type.ToString()**).

It's just an example of where these changes are breaking, but not in a way that is immediately clear. 

I have no problem at all with breaking API changes. It's just that here the breaking changes are perhaps a little too subtle.
It's of course up to you to decide how to proceed with these changes. Maybe I'm being too cautious here.

lahma:
Current support contracts allow quite a lot of breaking changes 😉

PR #1613, opened 2022-05-22
use property names instead of reflection

### LINKED ISSUE #1407: Replace use of SchedulerDetailsSetter
We currently use reflection to set the name and id of the scheduler instance on the job store and the threadpool.
The corresponding properties are part of the **IJobStore** and **IThreadPool** interfaces, and as such there's no need to use reflection.

@lahma, is it ok to create a PR for this?

tonyqus:
fixes #1407


lahma (changes requested):
Can you also delete the now obsolete `SchedulerDetailsSetter`? In the future, it would nice if you could create a separate branch for each PR in your fork - this allows me to to rebase etc the PR if needed - I usually cannot push to anyone's main branch.

**EDIT** I also edited the description with "fixes #1407" , this single line comment makes sure we have PR and issue in sync. After merge the issue gets closed.

tonyqus:
I've deleted `SchedulerDetailsSetter`. 

Ok. I'll use branch instead of main in the future.

lahma (approved):

lahma:
Thank you!

PR #1621, opened 2022-05-28
Add UseMySqlConnector overload for DB configuration

### LINKED ISSUE #1616: Add UseMySqlConnector overload for MS DI database configuration
`UsePersistentStore` enables implementation-specific extension methods, such as `UseMySql`. These dictate both the dialect and the specific database provider used. We can do better.

As a rule, we should depend on abstractions, not specific third-party packages. The developer should choose the concrete database provider. All that Quartz needs is a way to produce a `DbConnection` on demand.

Concretely, the issue I am running into is a code base that uses the truly async [MySqlConnector](https://github.com/mysql-net/MySqlConnector/) (for good reasons), with Quartz's MySQL implementation being tightly coupled to MySql.Data. This results in the need for two different MySQL connectors in one application. Additional complications arise, such as one connector needing `UseXaTransactions=False` to exhibit the desired behavior, but the other rejecting that option, defaulting to that behavior instead.

**Scope**

Although I am requesting this feature primarily for the MySql implementation, it would make sense to have for all the ADO.NET-based implementations.

**Proposed Solution**

Next to `SchedulerBuilder.AdoProviderOptions.ConnectionString`, we could introduce `SchedulerBuilder.AdoProviderOptions.ConnectionFactory`. A `Func<DbConnection>` type would do nicely.

If only `ConnectionString` is set, we maintain the existing behavior. If only `ConnectionFactory` is set, we use that to provide connections, and these are expected to come with their connection strings preassigned.

If _both_ properties are set, then the connection factory is used to produce a `DbConnection`, and then the connection string is used to assign `DbConnection.ConnectionString`. Although this may seem unimportant, it provides a sensible way of handling the presence of _both_ properties, and it fulfills a niche use case where the developer wants to separately provide `DbConnection` instantiation vs. `ConnectionString` selection.

**Alternative Solution**

An alternative solution might be to specify the assembly name, namespace, and type name of the concrete `DbConnection` type to use. We would rely on it having a default constructor. However, I consider this solution inferior. Not only are the strings obscure and error-prone, but they lack the `Func<DbConnection>`'s ability to configure the `DbConnection` in different ways, should the developer so desire.

lahma:
fixes #1616

### DIFF
--- a/src/Quartz/SchedulerBuilder.cs
+++ b/src/Quartz/SchedulerBuilder.cs
@@ -555,9 +555,24 @@ namespace Quartz
             this SchedulerBuilder.PersistentStoreOptions options,
             Action<SchedulerBuilder.AdoProviderOptions> configurer)
         {
+            UseMySqlInternal(options, "MySql", configurer);
+        }
+
+        public static void UseMySqlConnector(
+            this SchedulerBuilder.PersistentStoreOptions options,
+            Action<SchedulerBuilder.AdoProviderOptions> configurer)
+        {
+            UseMySqlInternal(options, "MySqlConnector", configurer);
+        }
+
+        internal static void UseMySqlInternal(
+            this SchedulerBuilder.PersistentStoreOptions options,
+            string provider,
+            Action<SchedulerBuilder.AdoProviderOptions> configurer)
+        {
             options.SetProperty("quartz.jobStore.driverDelegateType", typeof(MySQLDelegate).AssemblyQualifiedNameWithoutVersion());
             options.SetProperty("quartz.jobStore.dataSource", SchedulerBuilder.AdoProviderOptions.DefaultDataSourceName);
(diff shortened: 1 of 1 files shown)


johnwc:
What release do you think this will go into?

lahma:
It's already available on MyGet feed as preview, will be published as part of 3.5 at some point.

johnwc:
Is there a ETA for 3.5?

lahma:
Not at the moment, you should be able to see the MyGet 3.5.0-preview build which contain everything that will be in next release though.

johnwc:
We're not allowed to use preview builds unfortunately.

johnwc:
No, they're are for OSS. They just don't believe we should be running preview code in prod environment. We also tend to wait at least a month or two before using just released builds, unless the software/library has a good track record of fully testing the release before official marking it as stable.

PR #1624, opened 2022-05-29
fix: (Config) Use configuration value quartz.jobStore.dbRetryInterval to set DbRetryInterval

### LINKED ISSUE #1623: Property quartz.scheduler.dbFailureRetryInterval not used
### Describe the bug

property `quartz.scheduler.dbFailureRetryInterval` was [deprecated](https://github.com/quartznet/quartznet/blob/35b8532dc005af2fc907b6d2516c2e4bf40f833b/docs/_posts/2016-08-18-quartznet-2.4-released.md)

It is still being referenced in code, and when set does not modify the expected dbRetry value. 
Should be replaced with `quartz.jobStore.dbRetryInterval` instead.

### Reproduce
Set the property `quartz.scheduler.dbFailureRetryInterval`
Write a job that will trigger a JobPersistenceException
Note the times the job is retried against the value set for the retry interval. (will be the default of 15 seconds, not the value configured)

### Expected behaviour

When setting the dbRetryValue it should be updated correctly in the JobStore configuration.

### Versions
- Quartz 3.4
- 4.[55834f7](https://github.com/quartznet/quartznet/commit/55834f7613e6c612a58aa1d5fa700465baad9def)

jafin:
Use configuration value `quartz.jobStore.dbRetryInterval` instead of deprecated `quartz.scheduler.dbFailureRetryInterval`, which is not used to set the property  of `DbRetryInterval` in JobStoreSupport

Resolves #1623

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -60,6 +60,11 @@
     **ListenerManager** on **QuartzScheduler** allows more control over the events that a **ITriggerListener** will
     receive.
 
+* FIXES
+  * Configuration property `quartz.jobStore.dbRetryInterval` will be correctly set when constructing the Scheduler JobStore.
+    If you previously had configuration with the key `quartz.scheduler.dbFailureRetryInterval` please change to the above mentioned key. 
+  
+
 ## Release 3.4.0, Mar 27 2022
 
 This release has Quartz jobs start executing only after application startup completes successfully, unless QuartzHostedServiceOptions are used to specify otherwise.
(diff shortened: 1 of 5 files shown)


lahma:
Should this go to v3 branch too? Next release will be 3.5 so I think we could fix this with note of possible breaking change in the changelog?

jafin:
Sure, did you want me to just do another PR targeted to 3.x?

lahma:
I can do it manually, just making sure that this probably is a good fit for 3.x too.

lahma (approved):
Thanks for fixing this!

PR #1626, opened 2022-05-30
refactor (JobType)

jafin:
based on feedback from #1610 PR

- Improve documentation
- Move logic into the constructor
- Remove ignored field initializer


lahma:
I think we could still change the implicit conversion logic so that only `Type -> JobType` is automatic, because that can never fail and is unambiguous conceptually. I *think* I had  that in mind in the beginning, but didn't communicate well or... maybe just gave bad guidance in hindsight.

lahma:
Thanks!

PR #1629, opened 2022-06-04
Refactor - Move Lock Sqls  into XXXDelegate class

tonyqus:
- Add [StdAdoDelegate.Locks.cs](https://github.com/quartznet/quartznet/pull/1629/files#diff-18a57e50fb5569e11585631ec05d84bc789eb6fa91ad5ca64d3651cfc7ed2430) to enable overriding select and insert lock sqls
- Update StdRowLockSemaphore and UpdateLockRowSemaphore to remove constant lock sqls. Instead, it will get sqls from delegates
- UpdateLockRowSemaphore is a special case. SqlUpdateForLock constant remains inside the class because it's used for SQLite case only.


lahma:
I don't understand, [lock SQL is already configurable via properties](https://www.quartz-scheduler.net/documentation/quartz-3.x/configuration/reference.html#quartz-jobstore-selectwithlocksql). What' the use case to change insert sql?

tonyqus:
Ok. I don't know there is a "selectWithLockSQL" configuration entry .Where is "quartz.jobStore.selectWithLockSQL" loaded? I don't see the code.

lahma:
It was here https://github.com/quartznet/quartznet/pull/1629/files#diff-e31658076b059505e3a9fe7922abe093035be9f89b3243ed8ffe5473e58a2f46L295-L301

Quartz sets property values via reflection based on on configuration.

Issue #1628, opened 2022-06-02
Allow stores to receive injectable dependencies


cadilhac:
**Is your feature request related to a problem? Please describe.**
Well, sort of. I was trying to use the [Quartz.NET-RavenDB store](https://github.com/ravendb/quartznet-RavenDB) and there was no way to make the store use an existing dependency (a db session instance). So I asked the author for this feature and [he added an option](https://github.com/ravendb/quartznet-RavenDB/issues/19#issuecomment-1112004596) to have a store where the needed dependency is injected. Unfortunately, it doesn't work. A custom store does not use MS dependency injection and I imagine he thought intuitively that this would work.

**Describe the solution you'd like**
Like jobs, I would like a store instance to have constructors where parameters are injected dependencies.

**Describe alternatives you've considered**
I'm obliged to inject my dependency through a static member.

Thanks

lahma:
Thanks for the report, I'm scheduling this for V4, without any promises...

Matthewsre:
I just bumped into this as well. Started working on a custom IJobStore implementation to use Cosmos, but quickly ran into this hurdle when trying to resolve dependencies for retrieving secrets from KeyVault, configuration values from persistent store, etc. Looks like a potential place to add support for this would be in the StdSchedulerFactory InstantiateType<>. 

Alternatively I would be able to set options using IServiceProvider, but I couldn't find anywhere in the framework to support that either.

Matthewsre:
I just came across the "ServiceCollectionSchedulerFactory", that looks like might already solve this problem?

lahma:
The `ServiceCollectionSchedulerFactory` used by DI tries to resolve from service provider first and then does normal instantiation. So if you register your IJobStore into services you could explore if Quartz gets the hang of it. Note that the `InstantiateType` method uses concrete types.

lahma:
This should work quite naturally with 3.7 API (also with earlier if manual job store registration has been done), see https://github.com/quartznet/quartznet/commit/133c5a34cb39a38c1060ce98c96794bd75344917 .

PR #1630, opened 2022-06-06
Better IoC integration with JobSchedulerFactory

ayende:
This is a rough implementation that is sufficient to provide JobSchedulerFactories with the ability to resolve their own dependencies via IoC. See: https://github.com/quartznet/quartznet/issues/1628

The idea is that we now support accepting an `IServiceProvider` in the constructor, and then it is in the hands of the factory.
See example usage here: https://github.com/ravendb/quartznet-RavenDB/blob/master/src/Quartz.Impl.RavenDB/RavenJobStoreWithInjectDocumentStore.cs#L14

### DIFF
--- a/src/Quartz.Extensions.DependencyInjection/ServiceCollectionSchedulerFactory.cs
+++ b/src/Quartz.Extensions.DependencyInjection/ServiceCollectionSchedulerFactory.cs
@@ -94,6 +94,12 @@ namespace Quartz
             var service = serviceProvider.GetService<T>();
             if (service is null)
             {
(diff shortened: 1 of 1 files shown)


drieseng:
@ayende, why don't you inject **IDocumentStore** into **RavenJobStoreWithInjectDocumentStore**?

ayende:
There were... reasons. I can't figure out why it didn't work out before. 
Sorry, you are correct, there is a simple change I can make to make it happen far more seamlessly.

PR #1710, opened 2022-08-15
Support DataSource name configuration

### LINKED ISSUE #1709: Allow changing DataSource name using built-in extension methods on SchedulerBuilder.PersistentStoreOptions
I'm trying to setup aspnetcore integration tests with multiple databases in a single testhost (using xUnit) and noticed that some of the Quartz internals (e.g. SchedulerRepository, DbConnectionManager) are using static members which are shared between test runs.

In case of SchedulerRepository, there is pretty simple workaround - just generate unique SchedulerName per test.

The situation is more complicated with DbConnectionManager. Currently there's no possibility to setup name of the DataSource, it is hardcoded to 'default'.
<br />
I would like an overload for UseSqlServer extension method with DataSource name parameter.
<br />
Copy-paste Quartz private/protected code and configure needed settings manually.
Perhaps I'm missing something and there is an easy way to implement the tests as I need.
<br />
**Additional context**
Reference to code:
https://github.com/quartznet/quartznet/blob/6c81befdb55bda55b6dbc596630c2a19c1aef973/src/Quartz/SchedulerBuilder.cs#L519-L520
https://github.com/quartznet/quartznet/blob/6c81befdb55bda55b6dbc596630c2a19c1aef973/src/Quartz/SchedulerBuilder.cs#L452-L458

devbased:
This PR resolves #1709

### DIFF
--- a/src/Quartz/SchedulerBuilder.cs
+++ b/src/Quartz/SchedulerBuilder.cs
@@ -381,16 +381,14 @@ namespace Quartz
             /// </summary>
             /// <param name="provider">Valid provider name to configure driver details.</param>
             /// <param name="configurer">Callback to refine configuration.</param>
+            /// <param name="dataSourceName">Name of the database. Defaults to '<see cref="AdoProviderOptions.DefaultDataSourceName"/>'.</param>
             /// <returns></returns>
             public void UseGenericDatabase(
                 string provider,
-                Action<AdoProviderOptions>? configurer = null)
+                Action<AdoProviderOptions>? configurer = null,
+                string? dataSourceName = null)
             {
-                SetProperty("quartz.jobStore.driverDelegateType", typeof(StdAdoDelegate).AssemblyQualifiedNameWithoutVersion());
-                SetProperty("quartz.jobStore.dataSource", AdoProviderOptions.DefaultDataSourceName);
-                SetProperty($"quartz.dataSource.{AdoProviderOptions.DefaultDataSourceName}.provider", provider);
-
-                configurer?.Invoke(new AdoProviderOptions(this));
+                UseGenericDatabase<StdAdoDelegate>(provider, configurer, dataSourceName);
             }
 
             /// <summary>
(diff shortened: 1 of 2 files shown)


src/Quartz/SchedulerBuilder.cs
```diff
@@ -408,6 +406,28 @@ public void UseBinarySerializer()
...
             }
+
+            /// <summary>
+            /// Configures persistence to use custom ADO provider.
+            /// </summary>
+            /// <typeparam name="TAdoDelegate">Type of the <see cref="StdAdoDelegate"/> inheritor.</typeparam>
+            /// <param name="provider">Valid provider name to configure driver details.</param>
+            /// <param name="configurer">Callback to refine configuration.</param>
+            /// <param name="dataSourceName">Name of the database. Defaults to '<see cref="AdoProviderOptions.DefaultDataSourceName"/>'.</param>
+            public void UseCustomDatabase<TAdoDelegate>(
```
lahma, line 417: I think this should be `UseGenericDatabase` too, it's just an overload with generic support, don't you think?

lahma (changes requested):
I think we shouldn't introduce a new method name unless it's really necessary, it still seems Generic like before and not more custom.

src/Quartz/SchedulerBuilder.cs
devbased: @lahma Yeah.. But introducing overloads is always risky, especially for popular projects like this one, so I chose the safe option.
For example, if you rename UseCustomDatabase, then there will be two methods in the type metadata with the same name and set of parameters, only one will be IsGeneric==true, this can break the code with reflection (if there is one, I don’t know) if this is not taken into account. 
Another option is to make UseCustomDatabase internal.
Maybe I'm overthinking this though..

lahma (approved):
Looks good!

lahma:
Thanks for your contribution, always appreciated!

PR #1714, opened 2022-08-19
add "UsePersistentStore<T> where T : IJobStore" in DI Extension to main

kondr1:
I need to use custom job store, but DI extension don't have any method for that.


lahma:
Can you change this to target v3 branch?

kondr1:
Make new PR #1715

lahma:
Thanks!

PR #1761, opened 2022-09-28
Actually return Guid in JobDataMap.TryGetGuidValue

aheubusch:
Amend #1592


drieseng:
@aheubusch, could you also add (both positive and negative) unit tests? Thanks!

PR #1797, opened 2022-11-03
Add how to schedule a one off job

drusellers:
The goal of this PR is to document how to schedule a one-off job. I would also like this page to include any other considerations when building a one-shot job.

Notably, how to best deal with failure. I've been running into one shot jobs that are hanging around in an error state. I'm not sure how to best handle that. So please lay the comments down on me in this PR and I'll happily update the docs!

### DIFF
--- a/docs/documentation/quartz-3.x/how-tos/one-off-job.md
+++ b/docs/documentation/quartz-3.x/how-tos/one-off-job.md
@@ -1,0 +1,87 @@
+# One-Off Job
+
+You can run the simplest job this way.
+
+:::tip
+
+Misfire Mode: Smart
+
+:::
+
+## Ahead of Time Job Registration
+
+If you have a static set of jobs, you can register them ahead of time using someting like this.
+If the `durable` flag is `true`, then the job will stay dormant until its triggered.
+
+```csharp
+public Task DoSomething(IScheduler schedule, CancellationToken ct)
+{
+    var job = JobBuilder.Create<AnExampleJob>()
+                        .WithIdentity("name", "group")
+                        .Build();
+    
+    var replace = true;
+    var durable = true;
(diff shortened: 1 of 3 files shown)


docs/documentation/quartz-3.x/how-tos/one-off-job.md
```diff
@@ -0,0 +1,44 @@
...
+:::
+
+```csharp
+public Task DoSomething(IScheduler schedule)
+{
+    var job = JobBuilder.Create<AnExampleJob>()
+                        .WithIdentity("name", "group")
+                        .Build();
+
+    var trigger = TriggerBuilder.Create()
```
lahma, line 18: If this is only about triggering a job, maybe `IScheduler.TriggerJob` should also be mentioned as that's the easiest option. in this example you have paired a job so this will also work job hasn't been registered yet. 

For `ISchduler.TriggerJob` comes into play whether job has `Durable` flag set - can it be dangling without triggers - otherwise Quartz would remove it as orphaned. Of course a job having triggers like nightly and calling `TriggerJob` would just trigger it "out-of-schedule".

docs/documentation/quartz-3.x/how-tos/one-off-job.md
```diff
@@ -0,0 +1,44 @@
...
+        .WithIdentity("name", "group")
+        .StartNow()
+        .Build();
+
+    await scheduler.ScheduleJob(job, trigger, ct);
```
lahma, line 23: CancellationToken `ct` might not be obvious here, e.g. where it comes from

lahma (approved):
Looks good, thank you! I'll merge this so we'll see it live and future refinements are always welcome 🚀

PR #1798, opened 2022-11-04
Another Pass on Docs

drusellers:
This time, I'm taking a stab at improving my understanding of using multiple triggers for one job registration. My mental model today is very one-to-one but I'm starting to see the value in really using both of these concepts.

Step one: A little how to on using multiple triggers. I would love any commentary about practices that should be done when working with multiple triggers.

I also reworked the initial tutorial to use the Microsoft.Extensions.Hosting. If you are ok with that, I'll be taking a large sweep over the docs as I go, switching to the new hosting framework.

Again, all the feedback please!

### DIFF
--- a/docs/documentation/best-practices.md
+++ b/docs/documentation/best-practices.md
@@ -14,7 +14,8 @@ Only store primitive data types (including strings) in JobDataMap to avoid data 
 ### Use the Merged JobDataMap
 
 The JobDataMap that is found on the `JobExecutionContext` during Job execution serves as a convenience.
-It is a merge of the JobDataMap found on the JobDetail and the one found on the Trigger, with the value in the latter overriding any same-named values in the former.
+The data in the JobDataMap is a merger of the JobDetail and the Trigger, with the data in the Trigger overriding
+any same-named value in the Job.
 
 Storing JobDataMap values on a Trigger can be useful in the case where you have a Job that is stored in the scheduler for regular/repeated use by multiple Triggers,
 yet with each independent triggering, you want to supply the Job with different data inputs.
--- a/docs/documentation/quartz-3.x/how-tos/multiple-triggers.md
+++ b/docs/documentation/quartz-3.x/how-tos/multiple-triggers.md
@@ -1,0 +1,59 @@
+---
+title: Multiple Triggers
+---
+
+One nice thing about Quartz is that you can split the job registration from the 
+triggers. One nice use case is being able to split out the JobData for each
+trigger. 
+
+Below, we have two triggers, each with their own set of data, but
+we only had to register the one job.
+
+
+```csharp
+public Task DoSomething(IScheduler schedule, CancellationToken ct)
+{
+    var job = JobBuilder.Create<AnExampleJob>()
+                        .WithIdentity("customer-process", "group")
+                        .Build();
+    
+    var replace = true;
+    var durable = true;
+    await schedule.AddJob(job, replace, durable, ct);
+
+    // Trigger 1
+    var jobData1 = new JobDataMap { { "CustomerId", 1 } };
+    await scheduler.TriggerJob(new JobKey("customer-process", "group"), jobData1, ct);
+
+    // Trigger 2
+    var jobData2 = new JobDataMap { { "CustomerId", 2 } };
+    await scheduler.TriggerJob(new JobKey("customer-process", "group"), jobData2, ct);
+}
+```
+
+You could even set common data parameters on the job itself. Here
+we are adding some job data to the job itself.
+
+```csharp
+public Task DoSomething(IScheduler schedule, CancellationToken ct)
+{
+    var job = JobBuilder.Create<AnExampleJob>()
+                        .WithIdentity("customer-process", "group")
+                        .UsingJobData("batch-size", 50)
+                        .Build();
+    
+    var replace = true;
+    var durable = true;
+    await schedule.AddJob(job, replace, durable, ct);
+
+    // Trigger 1
+    var jobData1 = new JobDataMap { { "CustomerId", 1 } };
+    await scheduler.TriggerJob(new JobKey("customer-process", "group"), jobData1, ct);
+
+    // Trigger 2
+    var jobData2 = new JobDataMap { { "CustomerId", 2 } };
+    await scheduler.TriggerJob(new JobKey("customer-process", "group"), jobData2, ct);
+}
+```
+
+Using this flexbility in the data model should make working with Quartz a flexible solution.
--- a/docs/documentation/quartz-3.x/tutorial/using-quartz.md
+++ b/docs/documentation/quartz-3.x/tutorial/using-quartz.md
@@ -12,70 +12,83 @@ in the paused state.
 
 Here's a quick snippet of code, that instantiates and starts a scheduler, and schedules a job for execution:
 
-__Using Quartz.NET__
+### Install Quartz.NET NuGets
+
+```sh
+Install-Package Microsoft.Extensions.Hosting
+Install-Package Quartz
+Install-Packake Quartz.Extensions.DependencyInjection
+Install-Packake Quartz.Extensions.Hosting
+Install-Packake Quartz.Serialization.Json
+```
+
+### Configure `Program.cs`
+
+A minimal example of configuring Quartz.NET with the Microsoft Hosting framework
+looks like this.
 
 ```csharp
-// construct a scheduler factory using defaults
-StdSchedulerFactory factory = new StdSchedulerFactory();
+using Microsoft.Extensions.Hosting;
+using Quartz;
 
-// get a scheduler
-IScheduler scheduler = await factory.GetScheduler();
-await scheduler.Start();
+var builder = Host.CreateDefaultBuilder()
+    .ConfigureServices((cxt, services) =>
+    {
+        services.AddQuartz(q =>
+        {
+            q.UseMicrosoftDependencyInjectionJobFactory();
+        });
+        services.AddQuartzHostedService(opt =>
+        {
+            opt.WaitForJobsToComplete = true;
+        });
+    }).Build();
+
+await builder.RunAsync();
+```
+
+Let's add a job to this.
+
+
+```csharp
+
+using Microsoft.Extensions.Hosting;
+using Quartz;
+
+var builder = Host.CreateDefaultBuilder()
+    .ConfigureServices((cxt, services) =>
+    {
+        services.AddQuartz(q =>
+        {
+            q.UseMicrosoftDependencyInjectionJobFactory();
+        });
+        services.AddQuartzHostedService(opt =>
+        {
+            opt.WaitForJobsToComplete = true;
+        });
+    }).Build();
+
+var schedulerFactory = builder.Services.GetRequiredService<ISchedulerFactory>();
+var scheduler = await schedulerFactory.GetScheduler();
 
 // define the job and tie it to our HelloJob class
-IJobDetail job = JobBuilder.Create<HelloJob>()
+var job = JobBuilder.Create<HelloJob>()
     .WithIdentity("myJob", "group1")
     .Build();
 
 // Trigger the job to run now, and then every 40 seconds
-ITrigger trigger = TriggerBuilder.Create()
+var trigger = TriggerBuilder.Create()
     .WithIdentity("myTrigger", "group1")
     .StartNow()
     .WithSimpleSchedule(x => x
         .WithIntervalInSeconds(40)
         .RepeatForever())
-.Build();
-    
+    .Build();
+
 await scheduler.ScheduleJob(job, trigger);
 
-// You could also schedule multiple triggers for the same job with
-// await scheduler.ScheduleJob(job, new List<ITrigger>() { trigger1, trigger2 }, replace: true);
-```
 
-__Configuring scheduler with fluent API__
-
-You can also use SchedulerBuilder fluent API to programmatically configure different aspects of the scheduler.
-
-```csharp
-var sched = await SchedulerBuilder.Create()
-    // default max concurrency is 10
-    .UseDefaultThreadPool(x => x.MaxConcurrency = 5)
-    // this is the default 
-    // .WithMisfireThreshold(TimeSpan.FromSeconds(60))
-    .UsePersistentStore(x =>
-    {
-        // force job data map values to be considered as strings
-        // prevents nasty surprises if object is accidentally serialized and then 
-        // serialization format breaks, defaults to false
-        x.UseProperties = true;
-        x.UseClustering();
-        x.UseSqlServer("my connection string");
-        // this requires Quartz.Serialization.Json NuGet package
-        x.UseJsonSerializer();
-    })
-    // job initialization plugin handles our xml reading, without it defaults are used
-    // requires Quartz.Plugins NuGet package
-    .UseXmlSchedulingConfiguration(x =>
-    {
-        x.Files = new[] { "~/quartz_jobs.xml" };
-        // this is the default
-        x.FailOnFileNotFound = true;
-        // this is not the default
-        x.FailOnSchedulingError = true;
-    })
-    .BuildScheduler();
-
-await scheduler.Start();
+await builder.RunAsync();
 ```
 
 As you can see, working with Quartz.NET is rather simple. In [Lesson 2](jobs-and-triggers.md) we'll give a quick overview of Jobs and Triggers, so that you can more fully understand this example.
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -101,6 +101,7 @@ module.exports = {
         collapsable: false,
         children: [
           ['/documentation/quartz-3.x/how-tos/one-off-job', 'One-Off Job'],
+          ['/documentation/quartz-3.x/how-tos/multiple-triggers', 'Multiple Triggers'],
         ]
       },
       {


docs/documentation/quartz-3.x/how-tos/multiple-triggers.md
```diff
@@ -0,0 +1,59 @@
+---
+title: Multiple Triggers
+---
+
+One nice thing about Quartz is that you can split the job registration from the 
+triggers. One nice use case is being able to split out the JobData for each
```
lahma, line 6: "One nice" a bit repetition 😉

docs/documentation/quartz-3.x/how-tos/multiple-triggers.md
```diff
@@ -0,0 +1,59 @@
...
+
+
+```csharp
+public Task DoSomething(IScheduler schedule, CancellationToken ct)
+{
+    var job = JobBuilder.Create<AnExampleJob>()
+                        .WithIdentity("customer-process", "group")
+                        .Build();
+    
+    var replace = true;
```
lahma, line 20: I don't mind the style, other alternative I like in code reviews especially is supplying as named parameter like `await schedule.AddJob(job, replace: true, durable: true, ct);`

docs/documentation/quartz-3.x/how-tos/multiple-triggers.md
```diff
@@ -0,0 +1,59 @@
...
+    var durable = true;
+    await schedule.AddJob(job, replace, durable, ct);
+
+    // Trigger 1
+    var jobData1 = new JobDataMap { { "CustomerId", 1 } };
+    await scheduler.TriggerJob(new JobKey("customer-process", "group"), jobData1, ct);
```
lahma, line 26: no applicable to this example as job types are not part of it, but I also like to have the the key bound to job as `public static readonly JobKey` when applicable, then it's easy to reference `AnExampleJob.Key`

docs/documentation/best-practices.md
```diff
@@ -14,7 +14,8 @@ Only store primitive data types (including strings) in JobDataMap to avoid data
 ### Use the Merged JobDataMap
 
 The JobDataMap that is found on the `JobExecutionContext` during Job execution serves as a convenience.
-It is a merge of the JobDataMap found on the JobDetail and the one found on the Trigger, with the value in the latter overriding any same-named values in the former.
+The data in the JobDataMap is a merger of the JobDetail and the Trigger, with the data in the Trigger overriding
```
lahma, line 17: this is quite powerful concept, basically you can have one job and users could define in their settings that "I want daily digest". Would one model this with trigger per customer or one trigger or trigger per alphabet to split customer data processing to concurrent workloads that be run on different nodes, there are always tradeoffs.

Million triggers running at the same time is a lot  (trigger per user) - but always has per-trigger recovery, compared to handling all the hundreds (of thousands?) of emails to be sent and failing mid-process and retrying and sending multiples over and over again (did this once). State might need to be in application database (email processing queue with flags) + Quartz in this case.

Just a note.

docs/documentation/quartz-3.x/how-tos/multiple-triggers.md
```diff
@@ -0,0 +1,59 @@
...
+
+    // Trigger 2
+    var jobData2 = new JobDataMap { { "CustomerId", 2 } };
```
lahma, line 29: Quartz has configuration recommendation to have `useProperties` in configuration which validates that only string keys and values are being used for serialization safety (false by default as legacy). It's debatable if this should not raise an error if primary value types like int are being used. Maybe would need three modes (string-only, value-type-only, anything-and-prepare-for-serialization-issues).

lahma (commented):
Looking good, thanks for this! I added some comments - I'm happy with the current version, please see if anything warrants for action.

docs/documentation/quartz-3.x/how-tos/multiple-triggers.md
drusellers: Well snap! That's gold.

docs/documentation/quartz-3.x/how-tos/multiple-triggers.md
```diff
@@ -0,0 +1,59 @@
...
+trigger. 
+
+Below, we have two triggers, each with their own set of data, but
+we only had to register the one job.
+
+
+```csharp
+public Task DoSomething(IScheduler schedule, CancellationToken ct)
+{
+    var job = JobBuilder.Create<AnExampleJob>()
```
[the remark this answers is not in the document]
drusellers: In MassTransit we have a common idiom of trying to use "orders" and "order placement" all the time. I find it helps keep the documentation and the content in a cohesive direction. Do you have a simple domain that would work well for scheduling?

Something that we can do at both a simple level for the intro docs, but has some good complexity for the harder / richer features?

docs/documentation/best-practices.md
drusellers: Yes! I want to have a page some place where we take your years of wisdom and try to distill it down into "things to think" about. As I'm working through the docs, I'm actively going back to my project and making improvements to the app and my underlying mental model.

docs/documentation/quartz-3.x/tutorial/using-quartz.md
```diff
@@ -12,70 +12,83 @@ in the paused state.
 
 Here's a quick snippet of code, that instantiates and starts a scheduler, and schedules a job for execution:
 
-__Using Quartz.NET__
+### Install Quartz.NET NuGets
+
+```sh
+Install-Package Microsoft.Extensions.Hosting
```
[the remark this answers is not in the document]
drusellers: I had to add this one oddly to my project to get it to find the `Host` class. 🤔

docs/documentation/quartz-3.x/how-tos/multiple-triggers.md
drusellers: I'll change my code to use `string`. I'd like to keep the examples as much in the happy path as possible.

docs/documentation/quartz-3.x/tutorial/using-quartz.md
```diff
@@ -12,70 +12,83 @@ in the paused state.
...
+Install-Package Quartz
+Install-Packake Quartz.Extensions.DependencyInjection
+Install-Packake Quartz.Extensions.Hosting
+Install-Packake Quartz.Serialization.Json
+```
+
+### Configure `Program.cs`
+
+A minimal example of configuring Quartz.NET with the Microsoft Hosting framework
+looks like this.
```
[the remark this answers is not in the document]
drusellers: I sure can. How much of your user base is .NET Standard versus .NET Core?

docs/documentation/quartz-3.x/tutorial/using-quartz.md
```diff
@@ -12,70 +12,83 @@ in the paused state.
...
+            opt.WaitForJobsToComplete = true;
+        });
+    }).Build();
+
+var schedulerFactory = builder.Services.GetRequiredService<ISchedulerFactory>();
+var scheduler = await schedulerFactory.GetScheduler();
 
 // define the job and tie it to our HelloJob class
-IJobDetail job = JobBuilder.Create<HelloJob>()
+var job = JobBuilder.Create<HelloJob>()
```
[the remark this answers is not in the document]
drusellers: Happy to switch back. My mental approach was that "knowing" it was `IJobDetail` didn't really help tell the story, of how to get started. It was one more thing that my brain had to process. Additionally, since there are no alternative interfaces that we can compare and contrast `IJobDetail` too - it again felt like an extra detail not helping me understand the very beginning of my Quartz journey. 

Thoughts? Again happy to switch back.

docs/documentation/quartz-3.x/how-tos/multiple-triggers.md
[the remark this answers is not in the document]
lahma: [MassTransit also has the scheduler-backed feature of scheduled send](https://masstransit-project.com/advanced/scheduling/) which I like, but you probably are quite familiar with this library 🙂 

I might have used Quartz in limited domain as there's been people complaining about couple millisecond imprecision for some near real-time processing.. But I think many have been using for things like:

* subscriptions (like daily digest), either user-configured or registering to predefined "triggers" like "at most daily", "at most weekly"
* synchronizing data between systems when there's no real activity/trigger/observer communication - load/sync changed content from CMS daily, sync user preferences nightly etc
* cleanup / maintenance - [OpenIddict cleans some expired data for example](https://kevinchalet.com/2020/10/02/introducing-quartz-net-support-and-new-languages-in-openiddict-3-0-beta4/)) , [Bitwarden uses in multiple ways](https://github.com/bitwarden/server/search?q=quartz)
* generate/prepare reports for past day nightly that would otherwise be expensive to do real-time

Not sure what would be easy one to extend as these are usually highly use case specific and great in that way that they follow single responsibility principle. Following the MS weather forecast service example it could be as simple as getting forecast data and persisting hourly - basically the in-memory caching version without the in-memory caching 🤷🏻‍♂️

docs/documentation/quartz-3.x/tutorial/using-quartz.md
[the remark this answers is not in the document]
lahma: yeah, such direct reference might need one. in a way it's of course good to list everything needed, but also might make it look complex. compared to some other solutions Quartz has tried to keep dependencies to minimum (see main Quartz package which also supports DB persistence), but nowadays probably would be OK to have a lot more dependencies especially if they are from "Microsoft dependency chain".

docs/documentation/quartz-3.x/tutorial/using-quartz.md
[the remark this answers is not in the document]
lahma: I likea your reasoning, `var` should be fine. I guess interfaces will help with testing, hopefully. 

Many of the things are inherited from Java to make porting and syncing between releases easier. Java version hasn't really been updated in years, except their stale bot running rampant closing issues (and some very valid ones too). This is why Quartz 4 (the Duke Nukem Project) will have less worries in that regard and I'm open to change internal code flows and naming when it's more practical.

docs/documentation/quartz-3.x/tutorial/using-quartz.md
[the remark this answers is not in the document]
drusellers: I'll try it again and see what I can get. I can prolly remove the JSON reference too. Right?

Have you considered going all in on the `Microsoft.Extensions.Hosting` for v4? I'm of the opinion that all new projects are working this way.

That could let you get rid of `Quartz.Extensions.DependencyInjection`, `Quartz.Extensions.Hosting`

```csharp
var builder = Host.CreateDefaultBuilder()
    // makes quartz feel like a top level concern
    .UsingQuartz((cxt, options) =>
    {
        // does the "UseMicrosoftDependencyInjectionJobFactory" by default
        // services.AddQuartzHostedService();
    })
    .Build();
```

docs/documentation/quartz-3.x/tutorial/using-quartz.md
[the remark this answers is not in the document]
drusellers: Well snap. I'm looking forward to v4 then. Def feeling the `Java` vibe and looking forward to making quartz a streamlined silky library of awesomeness.

docs/documentation/best-practices.md
```diff
@@ -14,13 +14,72 @@ Only store primitive data types (including strings) in JobDataMap to avoid data
...
+
+### Static Job Key
+
+To simplify `JobKey` access we recommend defining a static field that allows
+easy access to the job's key.
+
+```csharp
+public class SomeJob : IJob 
+{
+    public static readonly JobKey Key = new JobKey("job-name", "group-name");
```
lahma, line 54: If I recall right, we had an earlier convo that might benefit others too, basically a top of having well thought out groups and some operations for scheduler support group-based of partial group match kind of operations. "Suspend all integration jobs", "Start all maintenance jobs again". Might be another tip but just wanted to add a mental note

drusellers: Ok, so how do you manage your triggers?

Do you have a static class full of common groups? so that you can easily rename, etc?

I'm def chewing on how to best use the group of both triggers and jobs.

docs/documentation/quartz-3.x/tutorial/using-quartz.md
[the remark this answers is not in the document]
lahma: And hey don't hesitate to suggest API changes via PRs! Some persisted models like configuration, keys, serialization format I'm more hesitant to change but anything that makes the experience more streamlined like you mentioned is always appreciated. `main` branch is the playground to go wild, there's separate `3.x` maintenance branch to go slow.

The docs come from main so need to remember sync to 4.x folder too at some point, but probably later when iterations have been done.

docs/documentation/best-practices.md
```diff
@@ -14,13 +14,72 @@ Only store primitive data types (including strings) in JobDataMap to avoid data
...
+    public Task Execute(IJobExecutionContext context) { /* elided */ }
+}
+```
+
+then later you can trigger the job directly with
+
+```csharp
+public async Task DoSomething(IScheduler schedule, CancellationToken ct)
+{
+    await schedule.TriggerJob(SomeJob.Key, ct)
```
lahma, line 65: this is exactly what I use, especially if I have an API endpoint like "trigger data synchronization" for our development UI, when you have knowledge of job and the configuration phase has been done already, when configuring you can always have the `.ForJob(theJobInstanceFromJobBuilder)`

docs/documentation/best-practices.md
```diff
@@ -14,13 +14,72 @@ Only store primitive data types (including strings) in JobDataMap to avoid data
...
+```csharp
+public class SomeJob : IJob 
+{
+    public static readonly JobKey Key = new JobKey("job-name", "group-name");
+
+    public Task Execute(IJobExecutionContext context) 
+    {
+        // don't do this
+        var badMethod = context.JobDetail.JobDataMap.GetString("a-value");
+        var alsoBadMethod = context.Trigger.JobDataMap.GetString("a-value");
```
lahma, line 36: one tip could be handling the job data map, like the string-only and then there's the helpers to get typed values if you use the `SetDateTime`/`GetDateTime` etc helpers (from top of my head as I don't have the code open).

drusellers: Yeah, I want to have a specific page on working with Job Data.

Show the different methods, also discuss how it can map to properties on the job itself. etc.

show the TryGet's  etc.

docs/documentation/quartz-3.x/tutorial/using-quartz.md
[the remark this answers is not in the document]
lahma: > I'll try it again and see what I can get. I can prolly remove the JSON reference too. Right?

JSON probably needed, but here's the gist: I'd like people to know about the extension and use it (OK Quartz DI examples are full of using it, and so it should be). Another point is that it's sadly it's *the JSON package* instead of STJ and NSJ separation. 

This probably needs to be addressed in v4 timeframe or even 3.x. Deprecate/redirect JSON to `.Newtonsoft` and somehow endorse the STJ as default, even though the default name is already being used. I'm also quite scared of the migration path as people have data stored in database based on serialization format that also supports inheritance, e.g. `$type`-

> Have you considered going all in on the Microsoft.Extensions.Hosting for v4? I'm of the opinion that all new projects are working this way.

Yes, definitely. `Microsoft.Extensions.Hosting` (MEH?) is natural addition for such service like Quartz. I'd like to see Quartz 4 as meant to / designed for NET 6 and forward era, while still supporting full framework.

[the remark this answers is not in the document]
drusellers: Moving forward I think you can have two packages

`Quartz.Serialization.Newtonsoft`
`Quartz.Serialization.STJ`


That let's you leave the old "JSON" alone, and it doesn't survive into v4. For the v3 -> v4. You could have a 
 `Quartz.Serialization.UpgradeSupport` or some such that can be used to read the old format and rewrite it to the new format? or something. 🤔 

Also in a v3, version start detecting those issues and dropping warnings in the logs. ?? You could have some code that would attempt to upgrade to a newer model. Then in V4 you can do an upgrade. You could also start looking at a policy that you have to do v3 to v4, v4 to v5. No v3 to v5 type approach.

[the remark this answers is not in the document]
lahma: Thanks for the ideas, definitely something to look into. `Quartz.Serialization.Json` could probably be an empty package that depends on `Quartz.Serialization.Newtonsoft` for deprecation. For greenfield projects I would probably already recommend using 3.x with (non-existent) `Quartz.Serialization.STJ` and gather feedback. What I've heard from serialization heavy projects like Marten, it might not be all that easy in all cases.

docs/documentation/best-practices.md
lahma: I don't think there's a right answer there. For systems I know of, it has been quite easy to group triggers based on logical grouping (which might have proven wrong later). It might be tenant, it might be business activity - and hard to change later on (due to DB design 🙍🏻‍♂️). For in-memory scheduling it's always easy to change course.

drusellers: Ok, no worries. I think the PR is good to go at this point. Let me know if there are any other changes. Going to close the conversations. :)  👍

lahma:
Thank you for the conversation and improving the documentation once again!

PR #1803, opened 2022-11-06
First version of HTTP API

VilleHakli:
Here goes nothing...

This PR adds HTTP API endpoints and `HttpScheduler` to use those endpoints through `IScheduler`.

The endpoints are implemented using ASP.NET Core minimal API. I chose minimal API because it felt like the easiest way to implement the endpoints in a library. Obvious downside of minimal API is that those are .NET6+ only feature and thus I changed the `Quartz.AspNetCore` to target .NET6 and Microsoft.AspNetCore.App.

The endpoints can be included into ASP.NET Core app by calling `AddHttpApi` (registers required services) and `MapQuartzApi` (maps the endpoints) extension methods:
```csharp
services.AddQuartz(q =>
{
    q.AddHttpApi();
});
...
app.UseEndpoints(endpoints =>
{
    endpoints.MapQuartzApi();
});
```

To keep the required implementation to a minimum, the endpoints do not include anything that can be configured by consumer of the library. This includes e.g. authentication & authorization, JSON serialization settings and so on. The ASP.NET Core sample includes example how the endpoints can be configured to require authorization using API key.

The endpoints are implemented to fulfill the need of `HttpScheduler`, so those do not return anything other than what is required by the `HttpScheduler`. Because of this the endpoints might not be nicest when used with other clients, but I think that this can be improved later on.

I created a new project `Quartz.HttpClient` for the `HttpScheduler` to not add new dependencies into Quartz project. The project includes only the scheduler and requires a bit more work to make using the `HttpScheduler` nicer. This might mean adding configuration extension methods (`services.AddHttpScheduler()`), adding support in `RemotingSchedulerProxyFactory` and so on.

Because of the limitations of System.Text.Json inheritance support, there was a need to create custom System.Text.Json converters for `ICalendar` and `ITrigger`. These were added into separate `Quartz.Serialization.SystemTextJson` project. This PR only includes the pieces needed by the API endpoints and `HttpScheduler`.

The ASP.NET Core sample app also includes SwaggerUI to test the endpoints. I used NSwag for the generation of OpenAPI document and added required attributes into the endpoints for the generated SwaggerUI to be somewhat usable 😄 There is also simple console app example about how to use the HttpClient.

I also updated Nuke to run tests for Quartz.AspNetCore as those do not require external services.


src/Quartz.AspNetCore/AspNetCore/HttpApi/Util/BadRequestException.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace Quartz.AspNetCore.HttpApi.Util;
+
+internal class BadRequestException : Exception
```
lahma, line 3: nit: I think this is missing some constructors usually present with exceptions, also all internal classes should be `sealed` by default

src/Quartz.AspNetCore/Quartz.AspNetCore.csproj
```diff
@@ -2,27 +2,16 @@
 
   <PropertyGroup>
     <Description>Quartz.NET ASP.NET Core integration; $(Description)</Description>
-    <TargetFrameworks>netcoreapp3.1;netstandard2.0</TargetFrameworks>
+    <TargetFramework>net6.0</TargetFramework>
     <RootNamespace>Quartz</RootNamespace>
     <GenerateDocumentationFile>True</GenerateDocumentationFile>
     <Authors>Lewis Zou, Marko Lahma, Quartz.NET</Authors>
```
lahma, line 8: I think you've deserved your name here 😉

src/Quartz.HttpClient/HttpApiContract/Dtos.cs
```diff
@@ -0,0 +1,37 @@
...
+    public TriggerKey AsTriggerKey() => new(Name, Group);
+
+    public override string ToString() => Group + '.' + Name;
+}
+
+internal record NamesDto(IReadOnlyCollection<string> Names);
+
+internal record SchedulerContextDto(Dictionary<string, string> Context)
+{
+    public static SchedulerContextDto Create(SchedulerContext context)
```
lahma, line 20: I like this `Create` pattern!

src/Quartz.HttpClient/HttpClient/HttpClientException.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace Quartz.HttpClient;
+
+public class HttpClientException : SchedulerException
```
lahma, line 3: probably should be sealed and have some extra constructors, does this need to be public?

src/Quartz.HttpClient/HttpClient/Extensions.cs
```diff
@@ -0,0 +1,37 @@
...
+    public static string ToUrlParameters<T>(this GroupMatcher<T> groupMatcher) where T : Key<T>
+    {
+        if (groupMatcher == null || string.IsNullOrWhiteSpace(groupMatcher.CompareToValue))
+        {
+            return "";
+        }
+
+        if (groupMatcher.CompareWithOperator.Equals(StringOperator.Equality))
+        {
+            return $"groupEquals={groupMatcher.CompareToValue}";
```
lahma, line 17: probably not worth the change, but this could maybe also be `group=*{input}`, probably forbidden by URI schema

src/Quartz.HttpClient/HttpClient/HttpScheduler.cs
```diff
@@ -0,0 +1,476 @@
+﻿using System.Text.Json;
+
+using Quartz.HttpApiContract;
+using Quartz.Impl.Matchers;
+using Quartz.Spi;
+
+namespace Quartz.HttpClient;
+
+public class HttpScheduler : IScheduler
```
lahma, line 9: does this need to be public?

src/Quartz.Serialization.SystemTextJson/Converters/CalendarConverter.cs
```diff
@@ -0,0 +1,99 @@
...
+using Quartz.Impl.Calendar;
+using Quartz.Util;
+
+namespace Quartz.Converters;
+
+internal class CalendarConverter : JsonConverter<ICalendar>
+{
+    private static readonly Dictionary<string, ICalendarSerializer> converters = new()
+    {
+        { BaseCalendarSerializer.CalendarTypeKey, new BaseCalendarSerializer() },
```
lahma, line 14: nit: for thread-safe types could probably also have `BaseCalendarSerializer.Instance`, but the dictionary already is static so does not matter that much

lahma (commented):
This looks great, added some notes but nothing preventing from merging. If I don't hear back many complaints, I'll merge this. Thank you for doing this!

src/Quartz.AspNetCore/AspNetCore/HttpApi/Util/BadRequestException.cs
VilleHakli: I only added the constructors that are needed at the moment. I will add the common constructors and mark these sealed.

src/Quartz.HttpClient/HttpApiContract/Dtos.cs
VilleHakli: I copied it from the preliminary HTTP API work done by you 😄

src/Quartz.HttpClient/HttpClient/Extensions.cs
VilleHakli: Asterisk seems to be reserved character in urls, so might not make things easier.

src/Quartz.HttpClient/HttpClient/HttpScheduler.cs
VilleHakli: At the moment we do not have any factory method to create this, so need to be public for consumers to be able to create new instance.

src/Quartz.Serialization.SystemTextJson/Converters/CalendarConverter.cs
VilleHakli: Was thinking about that also but didn't implement for some reason. I will add the instance methods.

VilleHakli:
Committed review fixes.

While I was doing the fixes, I realized that the endpoints probably do not work correctly with invalid request json. I added tests for that and those showed that there are issued. I fixed the issue with custom STJ converters in separate commit as that was easy fix. The validation of requests deserialized by ASP.NET Core requires a bit more work, so the fixes are not included in this PR. I think that those are something we can patch later on in separate PR.

lahma (approved):
Looking good. Thanks for getting this thing working! One later thing that has to be considered is the proxying feature that scheduler factory has, basically new remote scheduler proxy builder against HTTP API that allows `schedulerFactory.GetScheduler()` which returns proxy for the remote instance.

PR #1815, opened 2022-11-12
Add request validation into HTTP API

VilleHakli:
As discussed in #1803 the validation in HTTP API is still lacking a bit. This PR adds validation into the HTTP API endpoints. This might not cover every possible case, but it should be at least better than before.

I also added some basic null checks for HttpClient and replaced usages of BadRequestException with BadHttpRequestException as ASP.NET Core throws the latter for binding errors.


lahma (approved):
Looking good, thanks for making the improvements!

PR #1831, opened 2022-11-19
Add support for creating HttpScheduler with StdSchedulerFactory

VilleHakli:
This PR adds support for creating `HttpScheduler` using `StdSchedulerFactory`.

I changed the `IRemotableSchedulerProxyFactory` to return `IScheduler` instead of `IRemotableQuartzScheduler` and implemented `HttpSchedulerProxyFactory` which creates `HttpScheduler`. If the default implementation is used through `StdSchedulerFactory`, only base address can be configured for the `HttpClient`. However, one can inherit their own implementation from `HttpSchedulerProxyFactory` which can configure the `HttpClient`. I added examples for both of these cases into `Quartz.Examples.HttpClient`.

I also added extension methods to register `HttpScheduler` into service collection. As it is registered as `IScheduler`, only single scheduler can be added this way. There is also example about this in `Quartz.Examples.HttpClient`.

I also found a bug in the error handling of `HttpScheduler` and fixed it also in this PR. I can split the change into separate PR if that is better. The issue was that JSON deserialization exception was thrown if HTTP response with non OK status code did not have JSON body. Now in these cases exception by `HttpClient` is thrown which contains details about the error (e.g. HTTP status code of the response).

### DIFF
--- a/src/Quartz.Examples.HttpClient/Program.cs
+++ b/src/Quartz.Examples.HttpClient/Program.cs
@@ -1,5 +1,27 @@
-﻿using Quartz.HttpClient;
+﻿using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
 
+using Quartz;
+using Quartz.Simpl;
+
+// Using HttpClientFactory with host builder
+var host = Host.CreateDefaultBuilder(args)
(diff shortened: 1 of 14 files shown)


lahma (approved):
Looking good, thank you!

PR #1845, opened 2022-11-26
Add support for registering multiple HttpSchedulers

VilleHakli:
Current implementation of `services.AddQuartzHttpClient()` will register the `HttpScheduler` as `IScheduler` and thus only single client can be added.

This PR adds support for using marker interfaces to register multiple `HttpScheduler`s by using `services.AddQuartzHttpClient<IMyScheduler>()`. Quartz will dynamically create implementation for the given interface and register it into the service collection. The code to generate the dynamic types is based on code from MassTransit.

I tested the implementation only by using the tests and examples included in this PR. So, there might be some corner cases where the current implementation will not work, but we can fix those as they come up.


lahma (approved):
Looks really nice, great work on test coverage too! 👍🏻

PR #1854, opened 2022-12-01
Cleanup JobTemplate docs

BraedonWooding:
Noticed the typo + the prefix of the class.

C# standard typically uses `A` prefix for abstract classes and I think as a standard (based on other jobs created) `SampleJob` is a cleaner name.


lahma:
Thank you!

PR #1887, opened 2022-12-29
Remove unused `HashHelpers.PowerOf2` function

stevenaw:
I found this unused internal function and have put up a PR to remove it. It doesn't appear to be used in the containing assembly or the assemblies referenced via `InternalsVisibleTo`. I don't know the history of it but I'm guessing it may've been from a previous resizing strategy.


lahma:
Thank you!

Issue #1911, opened 2023-02-08
(Quartz. Net) It will conflict with Nlog


panjonsen:
When you install the Nlog framework, the console will be called (Quartz. Net), causing your console to output a lot of debugging information you don't want.
![image](https://user-images.githubusercontent.com/61653451/217517606-312f11cf-ef3f-4566-abb7-722f85aca3f0.png)

panjonsen:
Ver:3.6.0

jafin:
Perhaps you can use nlog filters?  From my brief reading, you should be able to filter out the Quartz namespace and set a minLevel of ERROR  (or something to that effect) i.e. https://stackoverflow.com/questions/62915844/nlog-ignore-certain-messages-using-filters

panjonsen:
Perhaps there is a (Quartz. Net) command that can decide whether it is better to enable output. Of course, this is just what I think. After all, I don't need to configure Nlog. The program should be responsible for managing some of its own operations.

panjonsen:
I don't know where the nlog.config file used by Quartz is. Or can you provide a code to solve this problem?

报错
笔记

jafin:
@panjonsen  I updated the example to use NLog in #1912
The example uses the standard nlog.config file and shows filters to exclude the Quartz.* namespace logs
So your program will control whatever config you desire when you wire up your NLog in your app.

Change the filters to your desired level.  Nlog appears to match in a way I didn't expect.  The null logger catches Quartz.* and Trace/Debug and logs to nothing.

```xml
 <rules>
    <logger name="Quartz.*" maxLevel="Debug" final="true"/> <!-- send to null logger-->
    <logger name="Quartz.*" minLevel="Info" writeTo="console" final="true"/> <!-- final don't process any further rule if matched-->
    <logger name="*" minLevel="Info" writeTo="console" />
  </rules>
```

lahma:
Closing as there's now example provided how to configure NLog, not an Quartz issue.

PR #1912, opened 2023-02-08
Add logging example using NLog

jafin:
Add an example on using NLog and filtering out Quartz.* namespace messages   relates to #1911 
- Bump packages in example project.
- Microsoft.Extensions.Logging was 5 is now 6


lahma (approved):
Great, thank you!

PR #1918, opened 2023-02-22
Change Dockerfile to use WORKDIR instead of manually creating it

KhanbalaRashidov:


lahma:
@KhanbalaRashidov why?

KhanbalaRashidov:
> @KhanbalaRashidov why?

When you do this, if the folder doesn't exist, it creates it. This is how you do the operation in one operation

lahma:
Thank you!

PR #1923, opened 2023-02-24
Add missing "disallow concurrency" materialization for jobs

IlyaVassyutovich:
Here's the [discussion](https://github.com/quartznet/quartznet/discussions/1921). This is a PR for `3.x`-branch.

> one option is to add `internal` setter for the property

I've managed without adding new setters and just use existing internal ctor which suits me just well.


lahma (approved):
Looks good, thanks for pushing this for 3.x too!

PR #1937, opened 2023-03-05
Fix integer overflow in `QuartzRandom.Next(int, int)`

stevenaw:
An integer overflow is possible in the `Next(int minValue, int maxValue)` if the difference between those two values is greater than 2^31. It doesn't looks like this can happen in the library itself since all usages have hard-coded safe values, but since the class is public it's possible someone else may hit this.

I've also added a few other tests to validate the happy path of this function

### DIFF
--- a/src/Quartz.Tests.Unit/Core/QuartzRandomTest.cs
+++ b/src/Quartz.Tests.Unit/Core/QuartzRandomTest.cs
@@ -1,6 +1,5 @@
 using NUnit.Framework;
 using Quartz.Core;
-using System.Globalization;
 
 namespace Quartz.Tests.Unit.Core
 {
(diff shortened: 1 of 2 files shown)


src/Quartz.Tests.Unit/Core/QuartzRandomTest.cs
```diff
@@ -0,0 +1,57 @@
+using NUnit.Framework;
+using Quartz.Core;
+
+namespace Quartz.Tests.Unit.Core
+{
+    [TestFixture]
+    public class QuartzRandomTest
+    {
+        [Test]
+        public void TestNextValidatesPositiveRange()
```
stevenaw, line 10: I'm not sure if you have any preferences for naming conventions here, but I tried to copy the approach I saw in `QuartzSchedulerTest`. Feel free to let me know if you'd like me to change this

lahma (approved):
Looks good, thank you for your contribution!

Issue #1948, opened 2023-03-14
QuartzServiceCollectionExtensions ambiguous between Quartz.AspNetCore and Quartz.Extensions.Hosting


romantitov:
There is a conflict between QuartzServiceCollectionExtensions from `Quartz.AspNetCore` and `Quartz.Extensions.Hosting` which does not allow to use of the extension from `Quartz.AspNetCore` since it has reference to   `Quartz.Extensions.Hosting` 
![image](https://user-images.githubusercontent.com/11888679/225019085-dd0acfc1-981e-46a1-9e85-224790c51266.png)

**Ver 3.4.0 -  3.6.2**

**To Reproduce**

1. Add `Quartz.AspNetCore` package
2. Just write anywhere in code `QuartzServiceCollectionExtensions` 

**Expected behavior**

Methods from the extension from different packages should be possible to use

romantitov:
Here is the PR with fix: https://github.com/quartznet/quartznet/pull/1949

PR #1949, opened 2023-03-14
Fix for #1948 Ambiguous references 

romantitov:
namespace changed to Quartz.AspNetCore


lahma:
Thanks for the PR, as you are targeting main this will be part of 4.x preview releases, if you'd wish this to be included in 3.x series then also 3.x branch should be targeted.

romantitov:
Thanks @lahma I guess I will create one more PR based on 3.x since there are too many differences if I just change target branch from `main` to `3.x`

PR #1951, opened 2023-03-15
A couple of amendments to the documentations

montoner0:
Fixed a typo and added a crucial tip about configuration through appsettings.json


lahma:

Issue #1939, opened 2023-03-06
Support for specifying 'L' and 'LW' with other days of the month is not implemented


Ghyath-Serhal:
Is there any possibility in the future to implement this feature?

**Support for specifying 'L' and 'LW' with other days of the month is not implemented**

lahma:
Would you be interested in implementing it?

jafin:
@Ghyath-Serhal change is in main (4.x)

Issue #1288, opened 2021-09-01
Support for 'L' in cron expression combinations


i8beef:
**Is your feature request related to a problem? Please describe.**

We require the ability to schedule arbitrary schedules on specific days of the month + "last day of month". I notice in the cron expression builder this is expressly disallowed: https://github.com/quartznet/quartznet/blob/29ba6a0fb6a1d9fb0ec36480a4cd34c3d99e7105/src/Quartz/CronExpression.cs#L614

**Describe the solution you'd like**

Would like to see this available as an option, so someone could do something like "15th and last day of month", etc. in a single cron expression.

**Describe alternatives you've considered**

1. Add a custom calendar that could limit 28th, 29th, 30th, and 31st on demand to black list days until "current month final day", tied with any "last day" selection to a cron date of 28th + this calendar to increment to "real" last day.
2. Using two schedules instead of one. We have some direct dependency on there being a SINGLE schedule right now so this isn't a great option for us.
3. Create a custom trigger that wraps the cron trigger impl and adds this functionality (something like, creating TWO on the fly, splitting the cron expression into two, one every other day and one just the 'L' case, running both and then comparing for "next").

**Additional context**

If you can specify the reasoning this is explicitly disabled right now, I might be able to look and make a PR for this...

jafin:
@lahma consider this complete, merged in v4.

PR #1956, opened 2023-03-20
Cron support for 'L' and 'LW' in expression daysOfMonth

jafin:
- Add cron parser support for `'L'` and `'LW'` in expression combinations daysOfMonth (#1939) (#1288)
- There is a limit of only one 'L' in the expression.  ie 
`"0 15 10 7,15,L,L-2 * ?"` will fail  but 
`"0 15 10 7,15,L-2 * ? "` will pass.

### DIFF
--- a/src/Quartz.Tests.Unit/CronExpressionTest.cs
+++ b/src/Quartz.Tests.Unit/CronExpressionTest.cs
@@ -116,6 +116,40 @@ namespace Quartz.Tests.Unit
             Assert.IsTrue(cronExpression.IsSatisfiedBy(cal));
         }
 
+        [TestCase("0 15 10 6,15 * ? 2010", "0 15 10 6,15 * ? 2010")]
+        public void ExpressionToString(string cronExpression, string expected)
+        {
+            var expr = new CronExpression(cronExpression);
+            expr.ToString().Should().Be(expected);
+        }
+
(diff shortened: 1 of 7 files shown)


src/Quartz.Tests.Unit/CronExpressionTest.cs
```diff
@@ -487,7 +509,7 @@ public void TestGetTimeAfter_QRTZNET149()
         }
 
         [Test]
-        public void TestQRTZNET152()
+        public void TestQRTZNET152_Nearest_Weekday_Expression_W_Does_Not_Work_In_CronTrigger()
```
jafin, line 512: Renamed to better understand intent of test

lahma:
Can you also update change log so that it's easier to keep track of the changes?

lahma (approved):
This required surprisingly lot of changes, but I'm taking somewhat a leap of faith as review is quite difficult for this changeset. Thanks for implementing this!

jafin:
@lahma a lot of the change was refactoring, the cyclomatic complexity of a few of the methods very high, I found it quite challenging to work through, noting that it's been that way since the java port. 

All tests are passing, plus more were added, but I'll see if I can get some more coverage in there for confidence.

In future would it be easier to review, if these types of changes were presented as smaller commits rather than squashed for review?

PR #1964, opened 2023-03-24
Refactor Task to ValueTask (#988)

jafin:
Refactor usages of `Task` with `ValueTask`


drieseng:
@jafin, @lahma: I'm sure you'll burn me for this, but ... should we consider removing use of async (overhead) for methods that are only ever invoked in a "blocking kind of sense" from a dedicated (QuartzScheduler)thread?

jafin:
@drieseng I don't have an opinion really, what you suggest sounds like an improved position and your more aware of the overhead gains. 

Should it be a separate PR to this? 

Concerning overheads, I ran the existing benchmarks in quartz before and after this PR change 
(not really taking an interest in _what_ they covered) and the results from task -> valuetask appeared to be insignificant in performance improvements. From my reading, unless there are serious hot paths with tasks, this aligns with expectations.  

Perhaps the benchmarks don't cover all the hot paths where changes like this would make a difference 🤷🏻

PR #1972, opened 2023-03-26
Refactor CronExpression, add guards for invalid usage, add tests

jafin:
## Features
- Add guard for invalid `L` token location.
- Add guard for `DayOfWeek` on `L` token being out of range
- Add guards for `nthDayOfWeek` being out of range
- Increase test coverage on CronExpression class

## Refactors
- (CronExpression) Extract CronExpressionSummary & CronExpressionConstants.
- (CronExpression) `StoreExpressionVals` refactored 
- (CronExpression) Remove `GetDaysOfMonth`, was refactored out into other methods and is unused.
- (CronExpression) `CheckNext` return values are unused, changed to return void

## Minor (breaking?) change
- CronExpression.GetExpressionSummary() emits `\r\n` for newlines via AppendLine, previously was `\n`  ( I doubt this would cause any concern)


jafin:
Guard tests that fail on 3.6.2 for comparison:

image

lahma (approved):
Thank you for the cleanup!

PR #1975, opened 2023-03-26
Ability to specify an offset from the last weekday of the month (#1287)

### LINKED ISSUE #1287: Ability to  specify an offset from the last weekday of the month
Hello,

I'd like to have a cron expression like **0 0 8 LW-1 * ?** or **0 0 8 LW-2 * ?**

The reason for it is that we want to send some reports to our colleagues. Тhey have to complete certain work based on these reports by the end of the month, and they need more than one day to do this.

Best regards

jafin:
## Feature
- Ability to specify an offset from the last weekday of the month  resolves #1287
  For Example `LW-2`
- If final day value will cross month boundary, ie LW-30 in February would result in possible -2 day. The value for day will reset to 1.  This was preferred over throwing an exception at runtime 

~~NOTE: PR Branch is built on top of #1972~~
EDIT: Rebased on main

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -67,7 +67,9 @@
 
 * NEW FEATURES
 
-  * Add Cron parser support for 'L' and 'LW' in expression combinations for daysOfMonth (#1939) (#1288)
+  * Add cron parser support for 'L' and 'LW' in expression combinations for daysOfMonth (#1939) (#1288)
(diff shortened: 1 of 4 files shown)


lahma (approved):

lahma:
Thank you!

Issue #1954, opened 2023-03-19
ScopedJob is no longer disposed


kmcc049:
**Describe the bug**

We have some jobs that inject some classes that need to be disposed, as they create an outbound socket for operability metrics. After an upgrade from 3.5.0 to 3.6.2 and then running over the weekend our host service began to go OOM and/or to report errors about running out of file handles.

Some debugging has revealed that we are no longer getting disposable objects created via the DI container disposed.

It appears that the JobExecutionContextImpl now returns the wrapped job class inside the scoped job to the JobRunShell, as of #1581. When the JobRunShell invokes [ReturnJob](https://github.com/quartznet/quartznet/blob/b9b455c015032eff974b6e747d74d8e5d95db75d/src/Quartz/Core/JobRunShell.cs#L304)   it is now passing it the actual job type instead of a ScopedJob object. ScopedJob is never having its dispose invoked and thus the DI container scope is not disposed.

**Version used**

3.6.2, but it will be any version with #1581 merged.

**To Reproduce**

Inject any service with the MicrosoftDependencyInjectionJobFactory in use that implements IDisposable inside a job. The dispose will not be invoked for the service, possibly on service shutdown it will, but not within the lifecycle of the host.

**Expected behavior**

After the job has executed the scope should be disposed, and thus all related services.

**Additional context**

If the job itself is IDisposable its dispose will invoke.

kmcc049:
@lahma I don't mind creating a PR to address this, but I'm going to be without access to a computer for the next 4 days. As this seems like a reasonably large problem, you may want to address it more quickly than that?

lahma:
I'm not in a hurry, a PR will be much appreciated!

msa48:
May be it does' not need to use ScopedJob at all? 
MicrosoftDependencyInjectionJobFactory instead of replacing original IJob can use internal Dictionary<IJob, IServiceScope> and do cleanup in RemoveJob method.

kmcc049:
I have a pr up for this [here](https://github.com/quartznet/quartznet/pull/1977)

lahma:
Fixed in `3.x` and main `branches`,

PR #1977, opened 2023-03-26
Introduce an Unwrapped JobInstance property

kmcc049:
this allows us to satisf the requirements of the original PR that introduced the bug( #1581 ) and fix the problem by putting the original JobInstance property back to the correct state. See bug report https://github.com/quartznet/quartznet/issues/1954

@nstdspace you will be interested in this as your code will have to look at the new property.


lahma:
Thanks, I went with slightly different approach in #2059

PR #1980, opened 2023-03-27
Add CRON support for DayOfWeek and DayOfMonth to be set together (inclusive rules)

### LINKED ISSUE #1543: day-of-month and day-of-week both support in CronExpression
A clear and concise description of what the problem is day-of-month and day-of-week can not both specify. Ex. `0 0 12 1/14 * MON, THU, FRI` means '14th day-of-month from 1 through 31 and on MONDAY, THURSDAY, FRIDAY at 12:00:00'

But `CronExpression.IsValidExpression("0 0 12 1/14 * MON, THU, FRI")` will return false, I try many CronExpression, the result is day-of-month and day-of-week is mutual exclusion.

Is it possible to support day-of-month and day-of-week both specify. In Linux Crontab, `0 12 1/14 * MON,THU,FRI` seems legal 

[crontab](https://crontab.guru/#0_12_1/14_*_MON,THU,FRI)

image

AND, `0 0 12 1/14 * MON, THU, FRI` seem not correct expression that I want schedule. 

I want schedule `every two weeks and on MONDAY, THURSDAY, FRIDAY at 12:00:00 `, `MON/14, THU/14, FRI/14` also will express this schedule, But it is invalid increments exceed 7 in day-of-week.

Think you!

jafin:
## Feature

- CronExpression support DayOfMonth and DayOfWeek together (inclusive values)
resolves #1543

## Refactors

Outdated tests removed, condition no longer valid.

```text
public void TestFormatExceptionWildCardDayOfMonthAndDayOfWeek()
public void TestFormatExceptionSpecifiedDayOfMonthAndWildCardDayOfWeek()
public void TestFormatExceptionWildCardDayOfMonthAndSpecifiedDayOfWeek()
```

Below changes were reverted for performance reasons
~~`monthMap` and `dayMap` changed to `IReadOnlyDictionary`~~
~~protected fields initialised instead of null~~

## Note

~~Branched from #1972 - https://github.com/jafin/quartznet/tree/refactor-cron-expressions~~
EDIT: Rebased on main

### DIFF
--- a/src/Quartz/CronExpression.cs
+++ b/src/Quartz/CronExpression.cs
@@ -209,8 +209,32 @@ namespace Quartz
     [Serializable]
     public class CronExpression : ISerializable
     {
-        private static readonly Dictionary<string, int> monthMap = new Dictionary<string, int>(20);
-        private static readonly Dictionary<string, int> dayMap = new Dictionary<string, int>(60);
+        private static readonly Dictionary<string, int> monthMap = new Dictionary<string,int>(12)
+        {
+            {"JAN", 0},
+            {"FEB", 1},
+            {"MAR", 2},
+            {"APR", 3},
+            {"MAY", 4},
+            {"JUN", 5},
+            {"JUL", 6},
+            {"AUG", 7},
+            {"SEP", 8},
+            {"OCT", 9},
+            {"NOV", 10},
+            {"DEC", 11}
+        };
+
+        private static readonly Dictionary<string, int> dayMap = new Dictionary<string, int>(7)
+        {
+            {"SUN", 1},
+            {"MON", 2},
+            {"TUE", 3},
+            {"WED", 4},
+            {"THU", 5},
+            {"FRI", 6},
+            {"SAT", 7}
+        };
 
         private TimeZoneInfo? timeZone;
 
@@ -275,6 +299,7 @@ namespace Quartz
         [NonSerialized] protected bool nearestWeekday;
 
         [NonSerialized] protected int lastdayOffset;
+
         [NonSerialized] protected int lastWeekdayOffset;
 
         /// <summary>
@@ -299,26 +324,6 @@ namespace Quartz
 
         static CronExpression()
         {
-            monthMap.Add("JAN", 0);
-            monthMap.Add("FEB", 1);
-            monthMap.Add("MAR", 2);
-            monthMap.Add("APR", 3);
-            monthMap.Add("MAY", 4);
-            monthMap.Add("JUN", 5);
-            monthMap.Add("JUL", 6);
-            monthMap.Add("AUG", 7);
-            monthMap.Add("SEP", 8);
-            monthMap.Add("OCT", 9);
-            monthMap.Add("NOV", 10);
-            monthMap.Add("DEC", 11);
-
-            dayMap.Add("SUN", 1);
-            dayMap.Add("MON", 2);
-            dayMap.Add("TUE", 3);
-            dayMap.Add("WED", 4);
-            dayMap.Add("THU", 5);
-            dayMap.Add("FRI", 6);
-            dayMap.Add("SAT", 7);
         }
 
         ///<summary>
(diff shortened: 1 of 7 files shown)


src/Quartz/CronExpression.cs
```diff
@@ -209,45 +209,68 @@ namespace Quartz
     [Serializable]
     public class CronExpression : IDeserializationCallback, ISerializable
     {
-        private static readonly Dictionary<string, int> monthMap = new Dictionary<string, int>(20);
-        private static readonly Dictionary<string, int> dayMap = new Dictionary<string, int>(60);
+        private static readonly IReadOnlyDictionary<string, int> monthMap = new Dictionary<string,int>(12)
```
lahma, line 212: should use concrete types for private/internal members, otherwise there will unnecessary interface dispatch cost. new switch dispatch for string (NET7/NET8 compiler) might be faster here, not sure.

src/Quartz/CronExpression.cs
```diff
@@ -209,45 +209,68 @@ namespace Quartz
...
+            {"SAT", 7}
+        };
 
         private TimeZoneInfo? timeZone;
 
         /// <summary>
         /// Seconds.
         /// </summary>
-        [NonSerialized] protected SortedSet<int> seconds = null!;
+        [NonSerialized] protected SortedSet<int> seconds = new();
```
lahma, line 243: if everything is new'ed, there's allocation cost from unused members, are they always needed?

src/Quartz/CronExpression.cs
```diff
@@ -299,26 +323,6 @@ public class CronExpression : IDeserializationCallback, ISerializable
 
         static CronExpression()
         {
-            monthMap.Add("JAN", 0);
```
lahma, line 302: OK I see it was moved from here

src/Quartz/CronExpression.cs
```diff
@@ -1681,265 +1679,289 @@ private SortedSet<int> CalculateDaysOfMonth(DateTimeOffset dt)
...
 
-            // get day
-            var dayOfMSpec = !daysOfMonth.Contains(CronExpressionConstants.NoSpec);
-            var dayOfWSpec = !daysOfWeek.Contains(CronExpressionConstants.NoSpec);
-            SortedSet<int> tailDays;
-            if (dayOfMSpec && !dayOfWSpec)
+            // get day by day of month rule
+            var daysOfMonthCalculated = CalculateDaysOfMonth(d);
+            var tailDays = daysOfMonthCalculated.TailSet(d.Day);
+            var found = tailDays.Any();
```
lahma, line 1692: should avoid LINQ features, `.Count` is always faster, probably was old code

src/Quartz/CronExpression.cs
```diff
@@ -1681,265 +1679,289 @@ private SortedSet<int> CalculateDaysOfMonth(DateTimeOffset dt)
...
-                        day = daysOfMonthCalculated.First(); //for lastDayOfMonth use calculated fields
-                    else
-                        day = daysOfMonth.First(); //if not then initial set of days uncalculated (to avoid issue with stale weekday in wrong month value)
+                    day = daysOfMonthCalculated.First();
                     mon++;
                 }
+            }
+            else
+            {
+                if (lastdayOfMonth)
```
lahma, line 1708: generally should always use braces in code

lahma (approved):
Added some comments, probably some point to old code that has just been moved. We can merge when you say so.

src/Quartz/CronExpression.cs
jafin: Constructor calls through to BuildExpression(...)
So in all cases they are all used because they all initialise to new() in the constructor
But the original justification for the change to new up the members was this:

image


(and probably the tendency for me to avoid nulls where I can, obviously not factoring in performance concerns though)

I'll revert back to the above along with setting to members to `null!` instead of `new()`

src/Quartz/CronExpression.cs
jafin: fixed.

src/Quartz/CronExpression.cs
jafin: Done. thanks for the performance advice!

jafin:
Have addressed the comments. Should be ok to merge now

lahma:
Thank you!

PR #1981, opened 2023-03-27
Documentation: format tables, add additional 3rd party links

jafin:
* Pretty format markdown tables
* Add 3rd party links


lahma:
Thank you!

Issue #1966, opened 2023-03-24
Unable to get any job related information when using IObserver


dima-zhemkov:
**Describe the bug**

Because of changes in [this](https://github.com/quartznet/quartznet/commit/2ffb5a312a734745cf66ee1dba56e3451dacfa5d) commit I can no longer get any information about job or trigger when using `IObserver<in T>.OnNext` since you pass anonymous type to diagnostic listener.

**Version used**

3.6.2

**Expected behavior**

Pass to diagnostic listener some specific type that contains job and trigger details

lahma:
Can you create a PR showing the shape of data you are expecting, the previous version was throwing exceptions and causing problems so it wasn't good either.

PR #1983, opened 2023-03-27
Extend object passed to DiagnosticListener.Start/EndActivity

dima-zhemkov:
Extend object passed to DiagnosticListener.Start/EndActivity.

Fix for #1966


lahma (changes requested):
Added some feedback, also if you target main it will be part of 4.x, you probably want this on 3.x train so this would need to target 3.x branch.

PR #1997, opened 2023-04-01
Fix for deserializing CronExpression using Json Serializer throwing exception on GetNextValidTimeAfter

### LINKED ISSUE #1996: CronExpression.GetNextValidTimeAfter NullReferenceException when JSON Deserialized.
**Describe the bug**

A contract with a CronExpression property serialized to JSON and stored (forexample using Azure storage), then dezerialized, will throw NullReferenceException when GetNextValidTimeAfter is called.

**Version used**

3.6.2

**To Reproduce**

- Create a CronExpression instance with a valid cron string.
- Call GetNextValidTimeAfter on the created.
- Serialize to JSON, deserialize JSON to new CronExpression instance.
- Call GetNextValidTimeAfter on the new instance.

**Expected behavior**

Expectation would be for GetNextValidTimeAfter to not throw exception, instead return proper data.

**Additional context**

jafin:
## Fix

When deserializing a CronExpression outside of builtin Quartz Serializers (for example to store in a custom store). Then calling `GetNextValidTimeAfter` will throw a NullReferenceException

This is due to `BuildExpression` not being called

Resolves #1996

### DIFF
--- a/src/Quartz/CronExpression.cs
+++ b/src/Quartz/CronExpression.cs
@@ -207,7 +207,7 @@ namespace Quartz
     /// <author>Refactoring from CronTrigger to CronExpression by Aaron Craven</author>
     /// <author>Marko Lahma (.NET)</author>
     [Serializable]
-    public class CronExpression : IDeserializationCallback, ISerializable
+    public class CronExpression : ISerializable
     {
         private static readonly Dictionary<string, int> monthMap = new Dictionary<string, int>(20);
         private static readonly Dictionary<string, int> dayMap = new Dictionary<string, int>(60);
@@ -378,6 +378,7 @@ namespace Quartz
                     ThrowHelper.ThrowNotSupportedException($"Unknown serialization version {version}");
                     break;
             }
+            BuildExpression(CronExpressionString);
         }
 
         [System.Security.SecurityCritical]
@@ -2151,11 +2152,6 @@ namespace Quartz
             var copy = new CronExpression(CronExpressionString);
             copy.TimeZone = TimeZone;
             return copy;
-        }
-
-        public void OnDeserialization(object? sender)
-        {
-            BuildExpression(CronExpressionString);
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


src/Quartz/CronExpression.cs
```diff
@@ -433,6 +433,7 @@ protected CronExpression(SerializationInfo info, StreamingContext context)
                     ThrowHelper.ThrowNotSupportedException($"Unknown serialization version {version}");
                     break;
             }
+            BuildExpression(CronExpressionString);
```
lahma, line 436: Because there's already `OnDeserialization` implemented, doesn't it make the logic run twice wasting cycles in non-JSON context?

jafin: 1) Yes, I can confirm what you say that the existing serialization/deserilization will trigger OnDeserilization method, so running BuildExpression twice, (bad).

2) From what I read, System.Text.json and it appears newtonsoft default settings, do not trigger OnDeserialization.  Therefore should a person utilise these to serialize/deserialize the OnDeserilization method are not called. (https://stackoverflow.com/questions/67222737/ondeserialized-is-not-called-in-net-core)

3) can I propose two options:

a) Drop `OnDeserializaiton`, given that it only calls `BuildExpression`, and keep the PR'd BuildExpression in `CronExpression(SerializationInfo info, StreamingContext context)`.

b) Keep `OnDeserialization` as is, utilise the existing `bool expressionParsed` property (which appears unused) to flag Build has occured and conditionally skip in `protected CronExpression(SerializationInfo info, StreamingContext context)`

Thoughts?

lahma: maybe go with A, removing the callback/interface would be breaking change but as this is gargeting main I guess it's fine

jafin: ok done.

lahma (approved):

lahma:
Thank you!

PR #2002, opened 2023-04-06
return PersistJobDataAfterExecution flag in SelectJobDetail

wuuer:
return PersistJobDataAfterExecution flag in SelectJobDetail


wuuer:
ubuntu-latest failed.what's wrong

PR #2004, opened 2023-04-07
JobInterruptMonitor Plugin: Read MaxRunTime from MergedJobDataMap

alanblack:
**Change:**
Reading MaxRuntime from the MergedDataMap instead of JobDataMap.
This allows to set the MaxRunTime also in the trigger configuration. Before this change only configure a 'gracefully' interrupt in JobDetails was possible.

**Reason for change:**
I have the situation to have several (long running) jobs that start at different times with same parameters and need to be finished 'gracefully'  after a different time
This would also work with multiple jobs with one trigger each, but made the configuration bigger.


lahma (approved):
Nice improvement, thank you for the contribution!

alanblack:
> Nice improvement, thank you for the contribution!

You're welcome. 
Very good work you do there - I like quartz.net very much

PR #2014, opened 2023-04-20
V3: return PersistJobDataAfterExecution flag in SelectJobDetail.fixed not…

wuuer:
return PersistJobDataAfterExecution flag in SelectJobDetail.fixed not persisting jobData to database.

fixed incorrect exception message in CronExpression_Throw_Error_Contructed_With_Null
fixed incorrect nextFireTimeUtc in ComputeFirstFireTimeUtc when using SimpleTriggerImpl
fixed ComputeFirstFireTimeUtc for the trigger that will fire once
fixed incorrect time setting in TestAcquireNextTrigger.
fixed incorrect time setting in TestAcquireNextTriggerBatch.
fixed incorrect time setting in TestAcquireTriggers.
fixed incorrect time setting in TestResetErrorTrigger.
fixed incorrect repeatCount  when using RepeatSecondlyForTotalCount and fixed incorrect FireCount in TestTriggerFinalized.

### DIFF
--- a/src/Quartz.Tests.Integration/Core/RecoverJobsTest.cs
+++ b/src/Quartz.Tests.Integration/Core/RecoverJobsTest.cs
@@ -80,7 +80,7 @@ namespace Quartz.Tests.Integration.Core
                     // check that trigger is blocked after fail over situation
                     Assert.AreEqual("BLOCKED", triggerState);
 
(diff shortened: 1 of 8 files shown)


lahma:
Thanks, I've removed changes to simple trigger logic as it should not be a concern for this PR.

PR #2029, opened 2023-05-21
Update nuget packages (this was update hell)

Sicos1977:
- I updated all the nuget packages to the latest versions where possible. Some I just could not update because there were not supported anymore on .NET 4.7.2

The only thing I could nog get fixed was the `Quartz.OpenTelemetry.Instrumentation` project. Classes you used in there are internal in the latest version of `OpenTelemetry` and because I just don't know how this tool works I also have no idea how to fix this. I hope that you can look into this.

This pull request fixes the issue I have with `System.Diagnostics.DiagnosticSource` 7.0.0.0 or higher. The reason why it did not work anymore was because there was a version conflict in your code.

I hope you can merge this pull request so that we can go on with the latest nuget versions. Also one tip... please try to update the nuget packages more often because it took me about 2 hours to solve all the dependencies the package had on each other. I had to figure out the correct order in how to update them to avoid errors like ... this package depends on version x.x.x.x .... etc....

### DIFF
--- a/build/_build.csproj
+++ b/build/_build.csproj
@@ -12,7 +12,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Nuke.Common" Version="6.3.0" />
+    <PackageReference Include="Nuke.Common" Version="7.0.2" />
     <PackageDownload Include="GitVersion.Tool" Version="[5.7.0]" />
   </ItemGroup>
 
--- a/src/Directory.Packages.props
+++ b/src/Directory.Packages.props
@@ -3,65 +3,65 @@
     <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
   </PropertyGroup>
   <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' or '$(TargetFramework)' == 'net472' ">
-    <PackageVersion Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="2.1.1" />
+    <PackageVersion Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="7.0.0" />
     <PackageVersion Include="Microsoft.Extensions.Hosting" Version="2.1.1" />
-    <PackageVersion Include="Microsoft.Extensions.Hosting.Abstractions" Version="2.1.1" />
-    <PackageVersion Include="Microsoft.Extensions.Options" Version="2.1.1" />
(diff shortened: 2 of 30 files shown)


lahma:
> I hope you can merge this pull request so that we can go on with the latest nuget versions. Also one tip... please try to update the nuget packages more often because it took me about 2 hours to solve all the dependencies the package had on each other. I had to figure out the correct order in how to update them to avoid errors like ... this package depends on version x.x.x.x .... etc....

Libraries should depend on the lowest possible dependency as library users can then choose newer version if they need them and binding redirects kick in. Other way around it's not possible when library consumer is not allowed to have newer than X version of some library and thus that would make Quartz unusable.

Looking at the PR there's no changes to Quartz code - only referenced library versions - so you should be able to reference the versions of libraries your code needs/wants alongside Quartz? Microsoft libraries especially have great API surface compatibility.

Sicos1977:
Fair point but your code also used some beta version, that never looks like a good idea to me. Are you still open for a pull request with the lowest versions but still have the issue fixed with the System.Diagnostics.DiagnosticSource.

lahma:
What is the lowest version needed? Is there some reference what is the actual problem between versions? Your PR is targeting main so it 4.x - there's separate 3.x branch that has the current NuGet release code.

Sicos1977:
I did not see that the main branch was version 4.x still have to figure out what the problem is on version 3.x ... when a verison higher then 6.0 of diagnostic source is used all the jobs just get stuck. The job gets fired and never returns a status. After firing it it is just broken. This problem happens on any version higher then 6.0

lahma:
My gut feeling is that on-demand loading of the DLL fails during runtime and it seems that things are stuck.  Fuslogvw might help diagnosing issues https://learn.microsoft.com/en-us/dotnet/framework/tools/fuslogvw-exe-assembly-binding-log-viewer .

Sicos1977:
I'll try to diagnose the problem tomorrow when I'm back at work. I still think that you at least should upgrade most of the nuget packages because you are using very old versions (and beta versions that should never be used on a production environment). I know that you can use binding redirects but using very old versions is not always a good idea (but that is just my opinion). In the company where I work we try to avoid having to old versions because of security leaks and things like that.

lahma:
I think the only beta version reference is in `Quartz.OpenTelemetry.Instrumentation` which has already been deprecated in favor of [the official integration](https://www.nuget.org/packages/OpenTelemetry.Instrumentation.Quartz/1.0.0-alpha.2). Of course Quartz should only reference minimal versions that do not have security vulnerabilities, dependabot check alerts about these.

PR #2039, opened 2023-06-12
Fix persistence of extended properties when the trigger type is changed

philr:
When using JobStoreTX (ADO.NET), changing the type of a trigger (by calling `ScheduleJob` with `replace` set to `true`) will corrupt the trigger state in the database. The extended properties relating to the old state will be left unchanged. Extended properties relating to the new type will not be written to the database. Once the next fire time for the updated trigger is reached, the trigger will fail to be loaded from the database and the scheduler will stop executing _all_ jobs.

This is an issue with version 3.6.2 and the latest code in the main branch.

RAMJobStore works correctly.

Here's an example that will reproduce the problem:

```c#
public class Test
{
    public static async Task Execute()
    {
        var factory = new StdSchedulerFactory();
        var scheduler = await factory.GetScheduler();
        await scheduler.Start();	

        var regularJob = JobBuilder
            .Create<TestJob>()
            .WithIdentity("Regular", "Test")
            .Build();
        var regularTrigger = TriggerBuilder
            .Create()
            .WithIdentity("Regular", "Test").StartNow()
            .WithSimpleSchedule(b => b.WithInterval(TimeSpan.FromSeconds(10)).RepeatForever())
            .Build();
        await scheduler.ScheduleJob(regularJob, regularTrigger);

        var problemJob = JobBuilder
            .Create<TestJob>()
            .WithIdentity("Problem", "Test")
            .Build();
        var problemTrigger = TriggerBuilder
            .Create()
            .WithIdentity("Problem", "Test")
            .StartAt(DateTime.UtcNow.AddSeconds(40))
            .WithSimpleSchedule(b => b.WithRepeatCount(0))
            .Build();
        await scheduler.ScheduleJob(problemJob, problemTrigger);

        await Task.Delay(TimeSpan.FromSeconds(11));

        var changedProblemJob = JobBuilder
            .Create<TestJob>()
            .WithIdentity("Problem", "Test")
            .Build();
        var changedProblemTrigger = TriggerBuilder
            .Create()
            .WithIdentity("Problem", "Test")
            .StartNow()
            .WithCronSchedule("0/15 * * * * ?")
            .Build();
        await scheduler.ScheduleJob(changedProblemJob, new[] { changedProblemTrigger }, true);

        await Task.Delay(TimeSpan.FromMinutes(1));

        await scheduler.Shutdown();
    }
}

public class TestJob : IJob
{
    public Task Execute(IJobExecutionContext context)
    {
[...]

PR #2040, opened 2023-06-12
Fix persistence of extended properties when the trigger type is changed (v3)

philr:
This is the fix from #2039 backported to the 3.x branch. Please see the other PR for details.


philr:
The test failures don't appear to be relevant to the change.

PR #2044, opened 2023-06-18
fix: Update Nuke to 7.0.2 to reduce vulnerabilities

jafin:
The following vulnerabilities are fixed with an upgrade:
- https://snyk.io/vuln/SNYK-DOTNET-SYSTEMDRAWINGCOMMON-3063427
- https://snyk.io/vuln/SNYK-DOTNET-NEWTONSOFTJSON-2774678
- https://snyk.io/vuln/SNYK-DOTNET-SYSTEMTEXTENCODINGSWEB-1253267

PR #2047, opened 2023-06-19
fix(build): Nuke Pack for 7.0.2

jafin:
Fix issue introduced in #2044 with Nuke Pack not working.
- CopyFileToDirectory must be rooted


lahma:
Thank you!

PR #2060, opened 2023-06-25
Optimize reading blobs on SQL Server

### LINKED ISSUE #2054: Performance issues reading large job objects from AdoJobStore on SQL Server
**Describe the bug**

Reading very large JobDetail objects from SQL Server is extremely slow due to https://github.com/dotnet/SqlClient/issues/593

I have changed the read at [this line](https://github.com/quartznet/quartznet/blob/e40294119921ba04f27cb050338dbfff26b566c8/src/Quartz/Impl/AdoJobStore/StdAdoDelegate.Jobs.cs#L130) to be synchronous in order to work around the issue temporarily, which fixed it for me.

To work around this issue, if possible, `StdAdoDelegate` should pass `CommandBehavior.SequentialAccess` to `ExecuteReaderAsync`. [Link to docs](https://learn.microsoft.com/en-us/dotnet/api/system.data.commandbehavior?view=net-7.0)

Additionally, the code of `StdAdoDelegate` could use some optimization, [see here](https://github.com/quartznet/quartznet/blob/e40294119921ba04f27cb050338dbfff26b566c8/src/Quartz/Impl/AdoJobStore/StdAdoDelegate.cs#L591). The bytes are written to a `MemoryStream` which is immediately discarded. Please consider using `MemoryPool` or `RecycleableMemoryStream` instead of allocating byte arrays.

**Version used**

* Quartz 3.6.2
* .NET 6.0
* Microsoft.Data.SqlClient 5.1.1

**To Reproduce**

Schedule any job object that has a size in the order of megabytes, then wait for the trigger to fire.

OronDF343:
Fixes #2054

Use `CommandBehavior.SequentialAccess` to work around performance issues with SqlClient when reading blobs (SQL Server specific), and remove unnecessary copying to `MemoryStream` (All ADO.NET delegates).

### DIFF
--- a/src/Quartz/Impl/AdoJobStore/StdAdoDelegate.Calendars.cs
+++ b/src/Quartz/Impl/AdoJobStore/StdAdoDelegate.Calendars.cs
@@ -1,4 +1,5 @@
 using Microsoft.Extensions.Logging;
+using System.Data;
 
 namespace Quartz.Impl.AdoJobStore
 {
@@ -64,7 +65,7 @@ namespace Quartz.Impl.AdoJobStore
             using var cmd = PrepareCommand(conn, ReplaceTablePrefix(SqlSelectCalendar));
             AddCommandParameter(cmd, "schedulerName", schedName);
             AddCommandParameter(cmd, "calendarName", calendarName);
-            using var rs = await cmd.ExecuteReaderAsync(cancellationToken).ConfigureAwait(false);
+            using var rs = await cmd.ExecuteReaderAsync(CommandBehavior.SequentialAccess, cancellationToken).ConfigureAwait(false);
             ICalendar? cal = null;
             if (await rs.ReadAsync(cancellationToken).ConfigureAwait(false))
             {
--- a/src/Quartz/Impl/AdoJobStore/StdAdoDelegate.Calendars.cs
+++ b/src/Quartz/Impl/AdoJobStore/StdAdoDelegate.Calendars.cs
@@ -1,5 +1,4 @@
 using Microsoft.Extensions.Logging;
-using System.Data;
 
 namespace Quartz.Impl.AdoJobStore
 {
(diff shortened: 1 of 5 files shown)


OronDF343:
Test failure appears to be unrelated to changes, not sure why it is happening

lahma:
Something with this one seems to have a problem, can you use the same logic as in 3.x without the job builder?

OronDF343:
Doesn't seem to have helped... The faked DbDataReader is giving empty columns

lahma:
Thanks for checking, I'll try to have a look later.

OronDF343:
I found what is causing the test to fail. In net472 and in version 3.x, the extension method `Quartz.Util.DataReaderExtensions.GetString` is used, while in `net6.0`, `System.Data.DataReaderExtensions.GetString` is used. The custom extension uses the indexer `[]` while the system extension apparently does not, which means that the faked call to the indexer is not effective.

The reason why the matched method has changed is due to adding a `net6.0` target to Quartz, since that method does not exist in `netstandard2.0` or `net472`. Also, since it is an extension method, it cannot be directly faked.

I have reverted to using the builder and fixed the failing test for now.

src/Quartz/Impl/AdoJobStore/StdAdoDelegate.Calendars.cs
```diff
@@ -1,4 +1,5 @@
 using Microsoft.Extensions.Logging;
+using System.Data;
```
lahma, line 2: the integration tests seem to be showing a real problem, now different extension method is being chosen for `reader.GetString`, I think you need to remove `using System.Data;` additions in this PR and instead refer to `SequentialAccess` with its full name `System.Data.CommandBehavior.SequentialAccess`

lahma: I couldn't easily push a tweak to to this PR as you have created your PR for main - it's better convention to have PR's made from separate branch

lahma:
Nice detective work, the approach still has problem (and now easy to understand why as you checked the extension method). I've commented about the problem and why integration tests are now failing.

src/Quartz/Impl/AdoJobStore/StdAdoDelegate.Calendars.cs
OronDF343: Ok, that makes a lot more sense. I am removing the usings that I added now and rolling back the test

OronDF343:
Alright, looks like it's passing now, thanks for the feedback

lahma:
Thanks for getting this ready for `main` too!

PR #2061, opened 2023-06-25
[3.x] Optimize reading blobs on SQL Server

### LINKED ISSUE #2054: Performance issues reading large job objects from AdoJobStore on SQL Server

OronDF343:
Backport of #2060 to version 3.x

Fixes #2054

Use `CommandBehavior.SequentialAccess` to work around performance issues with SqlClient when reading blobs (SQL Server specific), and remove unnecessary copying to `MemoryStream` (All ADO.NET delegates).

### DIFF
--- a/src/Quartz/Impl/AdoJobStore/StdAdoDelegate.Calendars.cs
+++ b/src/Quartz/Impl/AdoJobStore/StdAdoDelegate.Calendars.cs
@@ -1,5 +1,6 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
+using System.Data;
 using System.Threading;
 using System.Threading.Tasks;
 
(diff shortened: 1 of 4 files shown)


lahma (approved):
Looks good, thank you!

PR #2065, opened 2023-07-04
Update faq.md

ojraqueno:
Change "to" to "so" in 'I know you hate that answer, to here's some information about what it depends "on".'


lahma:
Thank you!

PR #2085, opened 2023-07-19
Mark UseMicrosoftDependencyInjectionJobFactory as obsolete, remove UseMicrosoftDependencyInjectionScopedJobFactory

lahma:


robertodalmonte:
Since UseMicrosoftDependencyInjectionJobFactory is now marked Obsolete, what should I do with my code:

  services.AddQuartz(options =>
  {
      options.UseMicrosoftDependencyInjectionJobFactory();
      options.AddQuartzConfigurationOptions(hostContext);
  });
  
Should I simply remove that line?
  services.AddQuartz(options =>
  {
        options.AddQuartzConfigurationOptions(hostContext);
  });

  Thanks

lahma:
Yes, like the `ObsoleteAttribute` description says "icrosoftDependencyInjectionJobFactory is the default for DI configuration, this method will be removed later on". So it's the default now and you don't need to call it.

robertodalmonte:
Hey Marko,
thanks for your time, competence, and excellent Quartz library.
While I did see that the quartz jobs were behaving as expected after removing the call to 
UseMicrosoftDependencyInjectionJobFactory()

it was somewhat confusing to see the freshly updated documentation here:
Example Startup.ConfigureServices configuration
(https://www.quartz-scheduler.net/documentation/quartz-3.x/packages/microsoft-di-integration.html#persistent-job-stores)
still calling the method.
Since my app heavily relies on your Supercalifragilisticexpialidocious library...a double-check.

nikriaz:
@lahma I join prising Quartz library but as a kind reminder, the link above still [contains ](https://www.quartz-scheduler.net/documentation/quartz-3.x/packages/microsoft-di-integration.html#di-aware-job-factories) UseMicrosoftDependencyInjectionJobFactory as a way to configure DI aware job factories. It's confusing...

MinhThienDX:
> Yes, like the `ObsoleteAttribute` description says "MicrosoftDependencyInjectionJobFactory is the default for DI configuration, this method will be removed later on". So it's the default now and you don't need to call it.

I think it would be great to include the solution in the `ObsoleteAttribute` description.
You know what, I'll try to open a PR.

roblappcoupa:
I am super confused... what is the recommended way to resolve `IJob` instances from DI?

lahma:
Why do you want to resolve them? Quartz will resolve them from the container.  Look at example projects in this repo.

roblappcoupa:
@lahma Apologies, you can disregard my question. I was experimenting around for the first time and was looking at how to run a job on demand given the `IScheduler`. I was looking at a simplistic example that was creating a job and passing it to `IScheduler.ScheduleJob()`:
```
// define the job and tie it to our HelloJob class
var job = JobBuilder.Create<HelloJob>()
    .WithIdentity("myJob", "group1")
    .Build();

// Trigger the job to run now, and then every 40 seconds
    .WithIdentity("myTrigger", "group1")
    .StartNow()
    .WithSimpleSchedule(x => x
        .WithIntervalInSeconds(40)
        .RepeatForever())
    .Build();

await scheduler.ScheduleJob(job, trigger)
```

However, after cloning the repo and going through it, I see you wouldn't need to resolve an instance of the job to run it on demand. Based on what I see, you could use the `ScheduleJob(ITrigger)` overload and let Quartz resolve it. If this is incorrect, please let me know. Thanks for the guidance.

PR #2091, opened 2023-07-21
Add scoped cache for SystemTime.Now and SystemTime.NowUtc values

Keen13:
Resolves https://github.com/quartznet/quartznet/discussions/1806

### DIFF
--- a/src/Quartz/Simpl/RAMJobStore.cs
+++ b/src/Quartz/Simpl/RAMJobStore.cs
@@ -1611,6 +1611,8 @@ public class RAMJobStore : IJobStore
     {
         lock (lockObject)
         {
+            using DateTimeScope _ = new();
+
             // return empty list if store has no triggers.
             if (timeTriggers.Count == 0)
             {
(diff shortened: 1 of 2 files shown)


Keen13:
@lahma, Is this the correct branch for the merge request if I anticipate these changes to be included in the next 3.* release?

lahma:
> @lahma, Is this the correct branch for the merge request if I anticipate these changes to be included in the next 3.* release?

You should make PRs against 3.x branch if you propose them to be included in 3.x line. `main` is for 4.0 development now.

src/Quartz/Simpl/RAMJobStore.cs
```diff
@@ -1611,6 +1611,8 @@ protected virtual bool ApplyMisfire(TriggerWrapper tw)
     {
         lock (lockObject)
         {
+            using DateTimeScope _ = new();
```
lahma, line 1614: isn't this the same thing as just having same `now` value during having the lock?

lahma: had a quick look and doesn't seem to be enough. I'm not fond of adding complexity to our SystemTime hekoer as it will be later replaced with TimeProvider. Should probably investigate further inside RAMJobstore and also how this reflects to AdoJobStore.

Keen13:
I've created a new PR against 3.x.  Please, ignore this one.

src/Quartz/Simpl/RAMJobStore.cs
Keen13: > had a quick look and doesn't seem to be enough. I'm not fond of adding complexity to our SystemTime hekoer as it will be later replaced with TimeProvider. Should probably investigate further inside RAMJobstore and also how this reflects to AdoJobStore.

Do you have a specific vision for updating the code then?

PR #2093, opened 2023-07-24
Add scoped cache for SystemTime.Now and SystemTime.NowUtc values

Keen13:
Resolves https://github.com/quartznet/quartznet/discussions/1806

### DIFF
--- a/src/Quartz.Tests.Unit/Impl/StdSchedulerFactoryTest.cs
+++ b/src/Quartz.Tests.Unit/Impl/StdSchedulerFactoryTest.cs
@@ -21,6 +21,7 @@
 
 using System;
 using System.Collections.Specialized;
+using System.IO;
 using System.Threading.Tasks;
 
 using NUnit.Framework;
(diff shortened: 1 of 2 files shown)


src/Quartz/Simpl/RAMJobStore.cs
```diff
@@ -399,7 +399,7 @@ private bool RemoveJobInternal(JobKey jobKey)
                     StoreJobInternal(triggersByJob.Key, true);
                     foreach (ITrigger trigger in triggersByJob.Value)
                     {
-                        StoreTriggerInternal((IOperableTrigger) trigger, true);
+                        StoreTriggerInternal((IOperableTrigger)trigger, true);
```
Keen13, line 402: It seems, my IDE made unwanted space fixes. Should I revert them?

Keen13:
@lahma,  From what I understand, you are not in favour of making changes to the 'SystemTime' class. I don't see other way to ensure the same  'Now' value throughout the 'ApplyMisfire' method execution, but perhaps you have some ideas to explore?  We really need this fix, and we use a forked version now, but we'd also like to use the latest version of Quartz. 

@Roganik, FYI.

PR #2104, opened 2023-08-03
Reduced nesting levels through block-scoped `using`s in appropriate places

Lehonti:
Specifically where the semantics are identical before and after


Lehonti:
I just updated my branch with the last changes. There doesn't seem to be any conflicts

lahma:
I'm hesitant to take in such a change which doesn't do much more than move code blocks around making diffing and merging between `3.x` and `main` harder.

drieseng:
> I'm hesitant to take in such a change which doesn't do much more than move code blocks around making diffing and merging between `3.x` and `main` harder.

Personally, I'm not very fond of block-scoped usings. Just my 2 cents though.

PR #2111, opened 2023-08-09
Add missing spaces to Examples

CAPCHIK:
The documentation specifies `**` to be separated by a space.

They may have appeared due to the linker trying to display Italic.


lahma (approved):
Thank you!

PR #2113, opened 2023-08-10
Ensure that the DailyCalendar use the same TimeZone offset for all the checks

### LINKED ISSUE #2112: DailyCalendar incorrectly handle timezone
**Describe the bug**
When using a DailyCalendar to prevent jobs to trigger during a period the calendar doesn't respect the interval correctly.
I have an api that use CEST timzone and a Daily Calendar defined like this.
> new DailyCalendar("00:00:00", "04:00:00") { TimeZone = TimeZoneInfo.Utc  }

I expect the job to be executed between 2h and 6h when (on summer saving time).
Here the result of the TriggerUtil calculated fire times.
fireTimes
Like you can see it's not case, the job do not trigger at midnight so 2 hour too early, exactly my local timezone offset between local and utc.

**Version used**
3.7.0

**To Reproduce**
Here a unit test that reproduce the issue.
The test need to be executed on a machine with a different timezone offset than UTC to reproduce the issue.

```cs
 DailyCalendar dailyCalendar = new DailyCalendar("00:00:00", "04:00:00");
  dailyCalendar.TimeZone = TimeZoneInfo.Utc;

  var trigger = (IOperableTrigger)TriggerBuilder
      .Create()
      .WithIdentity("TestTimeZone2Trigger")
      .StartAt(DateBuilder.EvenMinuteDateAfterNow())
      .WithSimpleSchedule(s => s
          .WithIntervalInMinutes(1)
          .RepeatForever())
      .Build();

  var fireTimes = TriggerUtils.ComputeFireTimes(trigger, dailyCalendar, (int)TimeSpan.FromDays(1).TotalMinutes);

  var timeZoneOffset = TimeZoneInfo.Local.BaseUtcOffset;

  // Trigger need to fire during the period when the local timezone and utc timezone are on different day
  if (timeZoneOffset >= TimeSpan.Zero)
  {
      fireTimes.Where(t => t.Hour >= 0 && t.Hour <= timeZoneOffset.Hours).Should().NotBeEmpty();
  }
  else
  {
      fireTimes.Where(t => t.Hour >= 24 - timeZoneOffset.Hours && t.Hour <= 23).Should().NotBeEmpty();
  }
```

**Expected behavior**
The job need to be executed during the correct period like defined in the calendar.

**Additional context**
I made some debug and I think that the problem is that the method GetStartOfDay and GetEndOfDay of the DailyCalendar class don't set the offset of the DateTimeOffset that result in comparaison between DateTimeOffset with different offset.

kvpt:
Fix #2112

### DIFF
--- a/src/Quartz.Tests.Unit/Impl/Calendar/DailyCalendarTest.cs
+++ b/src/Quartz.Tests.Unit/Impl/Calendar/DailyCalendarTest.cs
@@ -117,13 +117,20 @@ namespace Quartz.Tests.Unit.Impl.Calendar
             var timeZoneOffset = TimeZoneInfo.Local.BaseUtcOffset;
 
             // Trigger need to fire during the period when the local timezone and utc timezone are on different day
-            if (timeZoneOffset >= TimeSpan.Zero)
+            if (timeZoneOffset > TimeSpan.Zero)
             {
+                // Trigger must fire between midnight and utc offset if positive offset.
                 fireTimes.Where(t => t.Hour >= 0 && t.Hour <= timeZoneOffset.Hours).Should().NotBeEmpty();
+            }
+            else if (timeZoneOffset < TimeSpan.Zero)
+            {
+                // Trigger must fire between midnight minus utc offset and midnight if negative offset.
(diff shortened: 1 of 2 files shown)


lahma (approved):
LGTM, thank you for the contribution!

PR #2123, opened 2023-08-30
* Create a new Job to scan SFTP folders

ManuelLopezNavarrete:
- Create a base class to share the common keys with DirectroyScaner

### DIFF
--- a/src/Directory.Packages.props
+++ b/src/Directory.Packages.props
@@ -7,6 +7,7 @@
     <PackageVersion Include="Microsoft.Extensions.Hosting" Version="2.1.1" />
     <PackageVersion Include="Microsoft.Extensions.Hosting.Abstractions" Version="2.1.1" />
     <PackageVersion Include="Microsoft.Extensions.Options" Version="2.1.1" />
+    <PackageVersion Include="SSH.NET" Version="2020.0.2" />
     <PackageVersion Include="System.Threading.Tasks.Extensions" Version="4.5.4" />
   </ItemGroup>
   <ItemGroup Condition=" '$(TargetFramework)' == 'net6.0' ">
--- a/src/Quartz.Jobs/Job/BaseDirectoryScanJob.cs
+++ b/src/Quartz.Jobs/Job/BaseDirectoryScanJob.cs
@@ -1,0 +1,56 @@
+using Quartz.Spi;
+
+namespace Quartz.Job;
+/// <summary>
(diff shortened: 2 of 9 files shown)


ManuelLopezNavarrete:
I don't now if there are any issue if i refactor the original DirectroyScaner creating a base class to share the common logic. 
If there isn't any issue i will like to create a base class to DirectoryScanJobModel to.

If you prefer to have separate implementation independence of the duplication of code, i will do a rollback an duplicate the necessary code.

ManuelLopezNavarrete:
I'don't know how to interpret the qodan-scan and reading the log i understand, i need an access Token?

lahma:
You can ignore that check, it has some problems not related to PR content.

A9G-Data-Droid:
I don't recommend using `SSH.NET` at all, let alone including such a large chunk in Quartz. It seems to be way out of scope for a scheduling library. `SSH.NET` has a bunch of problems and is way behind on development. I recently replaced `SSH.NET` with [WinSCP](https://winscp.net/eng/docs/library) in my own projects to solve a connection hang issue. I no longer see the hang issue and the performance is 10x faster.

lahma:
@ManuelLopezNavarrete — thank you for this, and sorry it went unanswered for so long; your question
about extracting a shared base class deserved a reply at the time.

The blocker is the dependency, not the job. `src/Quartz.Jobs/Quartz.Jobs.csproj` has zero
`PackageReference` today — a `TargetFramework`, a README, and a project reference to Quartz, nothing
else. Every job in it (`DirectoryScanJob`, `FileScanJob`, `NativeJob`, `SendMailJob`) uses only the
BCL. Adding SSH.NET would put an SSH client on the dependency graph of everyone who installs
Quartz.Jobs for a directory scan. When we do take a third-party dependency, it gets its own package —
that is what `Quartz.Serialization.Newtonsoft` and `Quartz.Plugins.TimeZoneConverter` are. @A9G-Data-Droid's
note about SSH.NET's maintenance is a second reason to keep that choice out of our graph and in the
hands of whoever needs SFTP.

So: this belongs in a package of its own. If you publish it, open a PR adding it to
`docs/documentation/quartz-4.x/packages/quartz-3rd-party-plugins.md` and we will list it gladly.
`DirectoryScanJob`'s shape — an options type that writes the `JobDataMap` keys, plus a listener
interface — is the pattern worth copying.

Closing as not planned. Thank you for the work.

PR #2125, opened 2023-08-30
Fix for JobDataMap, TryGetValue return false when key is null/doesn't exist

jafin:
When a Key in JobDataMap does not exist/or is null, `TryGetXXXValue` will return false.

Presently, it will return true, due to `Convert.ToInt(null)`  (and other types) returning 0

The follow changes were also done, to adapt to this change:

+ Add `TryGetString` to match other method.
+ Refactor `JobInterruptMonitor` and `SendMailJob` to change to use Try methods now that Gets may throw `KeyNotFoundException`.
+ add a release note for change

### DIFF
--- a/src/Quartz/Util/StringKeyDirtyFlagMap.cs
+++ b/src/Quartz/Util/StringKeyDirtyFlagMap.cs
@@ -208,8 +208,16 @@ namespace Quartz.Util
         /// </summary>
         public virtual int GetInt(string key)
         {
-            var obj = this[key];
+            if (!this.ContainsKey(key))
+            {
+                throw new KeyNotFoundException($"Key {key} not found");
+            }
 
+            var obj = this[key];
+            if (obj is null)
+            {
+                throw new InvalidCastException("Identified object is not an Integer.");
+            }
             try
             {
                 return Convert.ToInt32(obj);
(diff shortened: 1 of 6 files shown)


src/Quartz/Util/StringKeyDirtyFlagMap.cs
```diff
@@ -208,8 +208,16 @@ public virtual void Put(string key, Guid value)
         /// </summary>
         public virtual int GetInt(string key)
         {
-            var obj = this[key];
+            if (!this.ContainsKey(key))
```
lahma, line 211: to avoid doing the lookup twice, should use `!this.TryGetValue(key, out var obj)`

src/Quartz/Util/StringKeyDirtyFlagMap.cs
```diff
@@ -548,6 +604,23 @@ public virtual bool TryGetNullableGuid(string key, out Guid? value)
             }
         }
 
+        /// <summary>
+        /// Try to retrieve the identified <see cref="string" /> value from the <see cref="JobDataMap" />.
+        /// </summary>
+        public virtual bool TryGetNullableString(string key, out string? value)
```
lahma, line 610: I think a better name would be `TryGetString` as string is already nullable, the Guid overload with `Nullable` is a bit of exception to the rule I think, thoughts?

jafin: You make a good point, I was trying to conform to the existing pattern.  Will rename to `TryGetString`..

jafin: I should remove  this from GetString given then that null is valid .
``` c#
 if (obj is null)
            {
                throw new InvalidCastException("Identified object is not a String.");
            }
```

lahma (approved):
LGTM, thank you!

PR #2126, opened 2023-08-31
SendMailJob to throw JobExecutionException on BuildMessage construction failure due to missing mandatory params.

jafin:
When a `SendMailJob` fails to construct message from missing parameters, it should throw a `JobExecutionException`.

Branched from #2125


jafin:
Test fails seem unrelated to change ?!?

lahma (approved):
Nice one, thank you!

PR #2136, opened 2023-09-13
Update multiple-triggers.md

khoitmp:
It seems like one redundant parenthesis


lahma (approved):
Thank you for the corrections!

PR #2159, opened 2023-10-13
Fix a typo in the multiple triggers docs

query-wow:
Fixes a the parentheses on the ` await scheduler.TriggerJob` call


lahma:
Thank you.

Issue #1152, opened 2021-04-07
WithDailyTimeIntervalSchedule


krzysztofsitnik:
**Describe the bug**

TriggerBuilder.WithDailyTimeIntervalSchedule has default interval of 1 minute. 
Wouldn't it be more understandable to have there 24 hours default interval, as the name suggests? 

**Version used**

3.0.4

**To Reproduce**
`var sampleTrigger = TriggerBuilder
                .Create()
                .StartNow()
                .WithDailyTimentervalSchedule(builder =>
                    .StartingDailyAt(TimeOfDay.HourAndMinuteOfDay(23, 56)))
                .Build();`

**Expected behavior**

I expect that trigger to work once a day, not each minute after the time put in StartingDailyAt.

**Additional context**

I'm not the only person which find it as misunderstanding.
[https://stackoverflow.com/questions/12735750/quartz-net-how-to-create-a-daily-schedule-that-does-not-gain-1-minute-per-day](url)

chaquotay:
> I'm not the only person which find it as misunderstanding.

I am also one of them 😢

lahma:
Would it make it more clear to allow passing default interval as part of the method signature, like so (didn't check against IDE so there be dragons):

```csharp
var sampleTrigger = TriggerBuilder .Create()
    .StartNow()
    .WithDailyTimentervalSchedule(IntervalUnit.Day, x => x
        .StartingDailyAt(TimeOfDay.HourAndMinuteOfDay(23, 56)
    )
)
.Build();
```

Just trying to find a solution that wouldn't be breaking for existing users but yet would make API and it's parameters discoverable.

chaquotay:
I'm not sure. I just assumed (my bad) that, given the name `WithDailyTimeIntervalSchedule`, the default interval would be "one day". But I see that changing this is not an option, because it will most likely break the existing code.

Maybe adding clarifiying documentation regarding the default interval to `WithDailyTimeIntervalSchedule` in https://github.com/quartznet/quartznet/blob/35b8532dc005af2fc907b6d2516c2e4bf40f833b/src/Quartz/DailyTimeIntervalScheduleBuilder.cs#L497 is a good and easy step into the right direction?

PR #2160, opened 2023-10-13
Try and clarify WithDailyTimeIntervalSchedule for #1152

lol768:
Previously the WithDailyTimeIntervalSchedule name was confusing, the method lacked inline documentation, and the default interval at 1 minute has led to a non-trivial number of developers accidentally running tasks every minute of every day (when they really meant every day).

I, along with many others, interpreted "daily time interval" as meaning a time interval of 24h. So, I wrote code like this:

```csharp
q.AddTrigger(t => t
    .WithIdentity("process_csv_report_trigger")
    .WithDescription("Generate CSV and upload to upstream SFTP server")
    .ForJob("process_csv_report_trigger")
    .WithDailyTimeIntervalSchedule(DailyTimeIntervalScheduleBuilder.Create().StartingDailyAt(TimeOfDay.HourAndMinuteOfDay(2, 0)).OnEveryDay())
);
```

I then had the horror of experiencing this happen every 60 seconds (yeah I know, it could've been tested a bit more locally). To make matters worse, the SFTP creds were wrong so it's probably generated about 2,500 authentication failure log lines. I'm surprised we didn't get our IP address blocked, to be honest.

Further discussion here: https://github.com/quartznet/quartznet/issues/1152#issue-852435242

StackOverflow confusion (41 upvotes):

image

image

I've added some pointers with `<seealso>` here to try and reduce confusion too. Frankly, I'd say the Quartz.CronScheduleBuilder.DailyAtHourAndMinute method is much more ergonomic to use and developers would probably be better served by using this API for simple daily tasks. I've tried to make the change backwards-compatible, whilst nudging developers to think about their use of this method carefully.

### DIFF
--- a/src/Quartz/Configuration/TriggerExtensions.cs
+++ b/src/Quartz/Configuration/TriggerExtensions.cs
@@ -2,11 +2,45 @@ namespace Quartz;
 
 public static class TriggerExtensions
 {
+    /// <summary>
+    /// Sets up a daily schedule with a default interval of *every-minute* unless overridden by the action
+    /// by calling WithInterval / WithIntervalInHours / WithIntervalInMinutes.
+    ///
+    /// For the sake of clarity, you probably want to use WithDailyConfiguredTimeIntervalSchedule instead.
+    /// </summary>
+    [Obsolete("Use WithDailyConfiguredTimeIntervalSchedule with default interval params instead.")]
     public static ITriggerConfigurator WithDailyTimeIntervalSchedule(
         this ITriggerConfigurator triggerBuilder,
         Action<DailyTimeIntervalScheduleBuilder>? action = null)
     {
         DailyTimeIntervalScheduleBuilder builder = DailyTimeIntervalScheduleBuilder.Create();
+        action?.Invoke(builder);
+        triggerBuilder.WithSchedule(builder);
+        return triggerBuilder;
+    }
+    
+    /// <summary>
+    /// Sets up a trigger schedule for one or more occurrences every day.
+    /// With parameter defaults, sets up a daily interval of every 12h.
+    /// </summary>
+    /// <remarks>
+    /// You need to configure the interval for when the trigger fires the job. If you only want one execution per day,
+    /// call EndingDailyAfterCount(1) or set the interval accordingly.
+    /// </remarks>
+    /// <param name="triggerBuilder"></param>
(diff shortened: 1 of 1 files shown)


src/Quartz/Configuration/TriggerExtensions.cs
```diff
@@ -2,6 +2,13 @@ namespace Quartz;
 
 public static class TriggerExtensions
 {
+    /// <summary>
+    /// Sets up a daily schedule with a default interval of *every-minute* unless overridden by the action
+    /// by calling WithInterval / WithIntervalInHours / WithIntervalInMinutes.
+    ///
+    /// For the sake of clarity, you probably want to use WithDailyConfiguredTimeIntervalSchedule instead.
+    /// </summary>
+    [Obsolete("Use WithDailyConfiguredTimeIntervalSchedule with default interval params instead.")]
```
jafin, line 11: Not sure if obsoleting this is ideal?.  
But adding the doc comments does benefit in describing what is occurring 👍🏻

lol768: My worry is that people won't read the docs comment. The behaviour I ran into (and the folks on StackOverflow also got burnt by) _was_ documented online in some places, but I didn't stumble across it until it was too late. If you obsolete the whole method, you a) don't break backwards-compat _completely_ but more importantly b) warn everyone with a half-decent IDE that what they're doing may be dangerous and prompt them to read the method docs.

My suggestion would be to _either_ change the default interval (to e.g. every 24h) and forego the `[Obsolete]` addition, or keep it because migrating folks over to `WithDailyConfiguredTimeIntervalSchedule` where the interval _has_ to be specified up-front is safer

jafin: I think my concern mainly came about backwards compat, but as you state, obsolete won't break backwards compat.  And perhaps in v4/5 it can be removed.  
Given the confusion you/others have found, makes sense to be more explicit 👍🏻

lol768: Sorry, I know it's a bit messy!

lahma:
Thank you!

lahma:
I tweaked a bit further in `main` (v4) to always require user to provide the values without defaults, now it should be very explicit.

PR #2183, opened 2023-11-10
Update all database scripts to ensure we check for both table and data existence

SeanFarrow-RSK:
This Pr ensures all database scripts do not drop and crate tables if they contain data.


lahma:
@SeanFarrow-RSK — thank you for raising this, and I am sorry it sat without a review.

The concern was right and it is now half handled. `tables_sqlServer.sql`, `tables_sqlServerMOT.sql`,
`tables_sqlServer_Below2016.sql` and `tables_postgres.sql` all carry a `@DropDb` switch — the whole
drop block sits inside `IF @DropDb = 1` (line 12 in the SQL Server script), and the header says "Should
you only require it to create the tables, set @DropDb to 0". Firebird, MySQL, Oracle and SQLite still
drop unconditionally, and Oracle deletes every row before dropping. Closing that gap is on us, and I
have noted it.

I am not going to take the mechanism in this PR, though. Deciding per table means a database with rows
in `QRTZ_TRIGGERS` but not `QRTZ_CALENDARS` gets one dropped and not the other, then fails part-way
through the creates with no clear signal — a whole-script switch is the only safe granularity. And the
`IF ... AND NOT EXISTS(...)` guards are not runnable statements on MySQL or Firebird outside a stored
program.

Closing as not planned; the switch is the shape we will extend. Thank you for the nudge.

PR #2213, opened 2023-12-05
Fix handling of env var quartz.config (#2212)

### LINKED ISSUE #2212: Unknown configuration property 'quartz.config'
**Describe the bug**

If you let the `StdSchedulerFactory` load the configuration from a file specified the environment variable "quartz.config" (defined in `StdSchedulerFactory.PropertiesFile`), then you get a `Quartz.SchedulerConfigException` saying `Unknown configuration property 'quartz.config'`

**Version used**

Appeared in v3.8.0, was not present in v.3.7.0

Seems to have been caused by https://github.com/quartznet/quartznet/commit/5562a9a6221531cb7a569cdc801c73ef06ab22da

That change introduced an explicit allowed list of configuration values, which doesn't include "quartz.config", but "quartz.config" is always added from the environment variables by `OverrideWithSysProps`.

**To Reproduce**

```cs

    [Test]
    public async Task TestFactoryShouldLoadPropertiesFromFileWhosePathIsGivenByEnvVariable()
    {
        var tempFile = Path.GetTempFileName();
        try
        {
            const string InstanceName = "TestInstance";

            File.WriteAllText(tempFile, $"{StdSchedulerFactory.PropertySchedulerInstanceName}={InstanceName}");

            Environment.SetEnvironmentVariable(StdSchedulerFactory.PropertiesFile, tempFile);

            var factory = new StdSchedulerFactory();
            factory.Initialize(); // <- optional, because `GetScheduler` does it anyway
            var scheduler = await factory.GetScheduler();

            Assert.AreEqual(InstanceName, scheduler.SchedulerName);
        }
        finally
        {
            try
            {
                File.Delete(tempFile);
            }
            catch (Exception)
            {
                // ignore temp file delete error
            }

            Environment.SetEnvironmentVariable(StdSchedulerFactory.PropertiesFile, null);
        }
    }
```

**Expected behavior**

Using environment variable "quartz.config" to specify the path to the config file should not cause an exception.

Perhaps  `OverrideWithSysProps` should explicitly skip environment variable "quartz.config" (`StdSchedulerFactory.PropertiesFile`)?

delta-emil:
Fixes #2212 by omitting environment variable "quartz.config" (`StdSchedulerFactory.PropertiesFile`) from the list of environment variable containing possible quartz props which get added by `StdSchedulerFactory.OverrideWithSysProps`

### DIFF
--- a/src/Quartz.Tests.Unit/Impl/StdSchedulerFactoryTest.cs
+++ b/src/Quartz.Tests.Unit/Impl/StdSchedulerFactoryTest.cs
@@ -21,6 +21,7 @@
 
 using System;
 using System.Collections.Specialized;
+using System.IO;
 using System.Threading.Tasks;
 
 using NUnit.Framework;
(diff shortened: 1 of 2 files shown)


lahma (approved):

lahma:
Thank you!

PR #2262, opened 2024-02-01
Make IJobWrapper public

da3dsoul:
This allows accessing the inner job without reflection. The fact that IJobWrapper exists suggests this should be public. In my case, I have extended the JobFactory to have a Post Init stage.

### DIFF
--- a/src/Quartz/IJobWrapper.cs
+++ b/src/Quartz/IJobWrapper.cs
@@ -1,6 +1,6 @@
 namespace Quartz;
 
-internal interface IJobWrapper
+public interface IJobWrapper
 {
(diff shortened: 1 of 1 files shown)


lahma:
I don't quite follow the logic "if something exists, it should be public". I don't see a compelling reason to make this public as it's an *internal* helper.

da3dsoul:
It's an interface. As things stand, there is no way to get the inner job of a Scoped Job or even detect that it's a wrapper without reflection, as both IJobWrapper and ScopedJob are non-public.
There are few reasons to make a non-public interface, and I don't think this is one.

PR #2268, opened 2024-02-04
Use configured type loader in scheduler factory init

lahma:


Trobeloo:
Thank you, now it is possible to add a custom TypeLoaderHelper.

Issue #2273, opened 2024-02-06
Error in ChonExpression DayOfWeek pattern


zorgoz:
**Describe the bug**

If containing a "day of week" part, the expression is not satisfied by a timestamp that should satisfy it. On the other hand, the same timestamp satisfies expressions with specific "day of week" part, that should not be satisfied by it. The `GetNextValidTimeAfter` method return a timestamp that should not satisfy the pattern, but it does.

**Version used**

3.8.0

**To Reproduce**

```
var dto = DateTimeOffset.Parse("2024.02.06 09:35:00");
Console.WriteLine(dto.ToString("dddd")); // yields: Tuesday

var co2 = new CronExpression("0 35 09 ? * 2");
Console.WriteLine(co2.IsSatisfiedBy(dto)); // yields: FALSE, but is should return TRUE

var next = co2.GetNextValidTimeAfter(dto); // it should give a date that is Tuesday
Console.WriteLine(next.Value.ToString("dddd")); // yieldy: Monday
Console.WriteLine(co2.IsSatisfiedBy(next.Value)); // yields: TRUE, should return FALSE

var co3 = new CronExpression("0 35 09 ? * 3"); // note: should match Wednesday only
Console.WriteLine(co3.IsSatisfiedBy(dto)); // yields: TRUE, should return FALSE

var co4 = new CronExpression("0 35 09 ? * 1"); // note: should match Monday only
Console.WriteLine(co4.IsSatisfiedBy(next.Value)); // yields: FALSE, although that's a Monday
```

**Expected behavior**

The pattern should be satisfied by dates that are on the proper days of the week. The next valid date should also match the pattern.

lahma:
As mentioned in the PR, Quartz and Quartz.NET have different cron expression syntax. This cannot be changed that easily.

PR #2274, opened 2024-02-06
Fixed the DayOfWeek handling.

zorgoz:
System DOW is 0-6 based, where 0 is always Sunday. Cron DOW is 1-7 based, where Sunday is 7. Hence simply casting to int will yield wrong results.

When calculating next valid timestamp, it is better to start from the larger offset as the current implementation tends to skip over possible values.

Added test cases.

### DIFF
--- a/src/Quartz.Tests.Unit/Core/CronExpressionTest.cs
+++ b/src/Quartz.Tests.Unit/Core/CronExpressionTest.cs
@@ -1,0 +1,104 @@
+using NUnit.Framework;
+
+namespace Quartz.Tests.Unit.Core;
+
+[TestFixture]
+public class CronExpressionTest
+{
+    private static readonly TimeZoneInfo testTimeZone = TimeZoneInfo.Local;
+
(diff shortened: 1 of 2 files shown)


zorgoz:
Relates to issue: https://github.com/quartznet/quartznet/issues/2273

daniel-white:
this would be a huge breaking change - its documented as valid for 1-7.

zorgoz:
@daniel-white 

Yes, 1-7 should be valid but it isn't. See referred issue amd https://stackoverflow.com/questions/77946937/quartz-net-seems-to-not-work-correctly-with-days-of-week?noredirect=1#comment137416558_77946937

zorgoz:
Ah... I see, this package breaks the CRON "standard"... got it....

jafin:
@zorgoz this package's cron is consistent with the java quartz version.  So I guess there are just variants of cron, UNIX cron and QUARTZ cron (possibly others), each with subtle differences.
I would say this doesn't make it `incompatible` , it just makes it a variant.

lahma:
Can't make such a breaking change, thanks for the PR though.

Issue #2270, opened 2024-02-05
GetNextIncludedTimeUtc method is very slow or crashes


al007:
The `GetNextIncludedTimeUtc` method of ICalendar can be extremely slow in certain conditions. The below example creates a calendar that excludes public holidays, weekends, and nighttime. A call to the `GetNextIncludedTimeUtc` method of such a calendar takes more than 13 seconds on my machine.

Tested version [Quartz 3.8.0](https://www.nuget.org/packages/Quartz) from Nuget as well as the latest version of the `main` branch of this repo.

``` csharp
[Test]
public void GetNextIncludedTimeUtc()
{
    TimeZoneInfo timeZone = TimeZoneInfo.FindSystemTimeZoneById("Pacific Standard Time");

    var holidayCalendar = new HolidayCalendar()
    {
        TimeZone = timeZone,
    };
    holidayCalendar.AddExcludedDate(new DateTime(2024, 2, 19));
    holidayCalendar.AddExcludedDate(new DateTime(2024, 5, 27));
    holidayCalendar.AddExcludedDate(new DateTime(2024, 6, 19));
    holidayCalendar.AddExcludedDate(new DateTime(2024, 7, 4));
    holidayCalendar.AddExcludedDate(new DateTime(2024, 9, 2));
    holidayCalendar.AddExcludedDate(new DateTime(2024, 10, 14));
    holidayCalendar.AddExcludedDate(new DateTime(2024, 11, 11));
    holidayCalendar.AddExcludedDate(new DateTime(2024, 11, 28));
    holidayCalendar.AddExcludedDate(new DateTime(2024, 12, 25));

    var weeklyCalendar = new WeeklyCalendar(holidayCalendar)
    {
        TimeZone = timeZone,
    };
    var calendar = new DailyCalendar(weeklyCalendar, "06:00", "22:00")
    {
        TimeZone = timeZone,
        InvertTimeRange = true,
    };
    var time = new DateTime(2024, 2, 5, 10, 6, 0, DateTimeKind.Utc);
    var expected = new DateTime(2024, 2, 5, 14, 0, 0, DateTimeKind.Utc);

    var d = calendar.GetNextIncludedTimeUtc(time); // takes 13 seconds to execute!
    d.Should().Be(expected);
}
```

al007:
[...]

PR #2284, opened 2024-02-23
Fix issue with Holiday Calendar with nested calendars not identifying NextIncludedTime

jafin:
Fix issue with Holiday Calendar not identifying a date included due to checking modified DateOnly value with base
Resulting in an exception from hitting year 9999.

- Changed to compare original date value with base Calendars, and date only value with current calendar validator (not base calendars). 
- Add scenario described in #2270 

## Notes
Naming is hard, not sure `IsTimeIncludedThisCalendar` is a good name
I did not check other day calendar types, perhaps the scenario exists in other calendars in similar configuration


lahma (approved):

lahma:
Thank you!

PR #2285, opened 2024-02-23
Improve performance of DailyCalendar solving GetNextIncludedTimeUtc  #2270

### LINKED ISSUE #2270: GetNextIncludedTimeUtc method is very slow or crashes

jafin:
## Change
Improve performance of `DailyCalendar` solving `GetNextIncludedTimeUtc` by changing the stepping increment based on the Start/End range of the DailyCalendar parameters.

## Observations
In the original issue OP noted this took >13 seconds.
With a similar test, it is now milliseconds.

fixes #2270

### DIFF
--- a/src/Quartz/Impl/Calendar/DailyCalendar.cs
+++ b/src/Quartz/Impl/Calendar/DailyCalendar.cs
@@ -47,8 +47,6 @@ namespace Quartz.Impl.Calendar
     /// as, for example, WeeklyCalendar defines a set of days that are
     /// excluded <i>every week</i>. Likewise, <see cref="DailyCalendar" /> defines a
     /// set of times that are excluded <i>every day</i>.
-    ///
-    /// The minimum resolution of time supported by this calendar is one second.
     /// </para>
     /// </remarks>
     /// <author>Mike Funk</author>
@@ -63,7 +61,6 @@ namespace Quartz.Impl.Calendar
         private const string InvalidMillis = "Invalid millis: ";
         private const string InvalidTimeRange = "Invalid time range: ";
         private const string Separator = " - ";
-        private const long OneSecondInMillis = 1000;
         private const char Colon = ':';
 
         private const string TwoDigitFormat = "00";
@@ -80,6 +77,7 @@ namespace Quartz.Impl.Calendar
         private int rangeEndingMinute;
         private int rangeEndingSecond;
         private int rangeEndingMillis;
+        private int precisionStepMillis = 1000;
 
         private DailyCalendar()
         {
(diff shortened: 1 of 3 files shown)


src/Quartz/Impl/Calendar/DailyCalendar.cs
```diff
@@ -479,7 +481,7 @@ public override bool IsTimeIncluded(DateTimeOffset timeUtc)
         /// <seealso cref="ICalendar.GetNextIncludedTimeUtc"/>
         public override DateTimeOffset GetNextIncludedTimeUtc(DateTimeOffset timeUtc)
         {
-            DateTimeOffset nextIncludedTime = timeUtc.AddMilliseconds(OneMillis);
+            DateTimeOffset nextIncludedTime = timeUtc.AddMilliseconds(OneSecondInMillis);
```
lahma, line 484: Can we determine the used precision based on configuration? Like if rangeStarting/Ending millis and seconds are zero we can safely search with seconds precision? For v4 we could probably just drop the whole millisecond precision support, if not even seconds support as they both sound a bit too fine-grained for daily operation.

jafin: Good point, I've amended to code to adjust stepping based on start/end range.

lahma (approved):
Awesome work again, thank you!

Issue #2079, opened 2023-07-18
Change SystemTime to TimeProvider / Microsoft.Bcl.TimeProvider


lahma:
* < net8.0 should use NuGet package [Microsoft.Bcl.TimeProvider](https://www.nuget.org/packages/Microsoft.Bcl.TimeProvider)
* net 8.0 build (new feature) should use native version (probably built-in)
* probably easier to to after we get DI working natively

PR #2286, opened 2024-02-24
Refactor code to use TimeProvider.System

jafin:
## Summary
Replace DateTime usage with TimeProvider.System
Replace implicit DateTime to DateTimeOffset conversions with explicit. Add in Analyzer to ensure DateTime isn't used.
Remove a check for IsRUnningOnMono. As it is no longer one of the supported deployment frameworks. Registers a TimeProvider in DI (but is not using it)

`Bcl.TimeProvider` supports Net8, net 4.6.2 ; Netstandard 2.0

❌ This does NOT implement DI for TimeProvider.  A few usage of DateTime are in static variables and private constructors.  I believe this would possibly need a servicelocator to pull in as a dependency.  And would rather see the DI work handle this if possible.

Relates to #2079

### DIFF
--- a/src/Quartz/CronExpression.cs
+++ b/src/Quartz/CronExpression.cs
@@ -261,8 +261,6 @@ public sealed class CronExpression : ISerializable
     /// </summary>
     [NonSerialized] private bool calendarDayOfMonth;
 
-    public static readonly int MaxYear =  TimeProvider.System.GetLocalNow().Year + 100;
-
     private static readonly Regex regex = new("^L(-\\d{1,2})?(W(-\\d{1,2})?)?$", RegexOptions.Compiled | RegexOptions.ExplicitCapture, TimeSpan.FromSeconds(5)); //e.g. LW L-0W L-4 L-12W LW-4 LW-12
     private static readonly Regex offsetRegex = new("LW-(?<offset>[0-9]+)", RegexOptions.Compiled | RegexOptions.ExplicitCapture, TimeSpan.FromSeconds(5));
 
@@ -1219,7 +1217,7 @@ public sealed class CronExpression : ISerializable
         {
             if (stopAt == -1)
             {
-                stopAt = MaxYear;
+                stopAt = CronExpressionConstants.MaxYear;
             }
             if (startAt is -1 or CronExpressionConstants.AllSpec)
             {
@@ -1959,7 +1957,7 @@ public sealed class CronExpression : ISerializable
             }
 
             // test for expressions that never generate a valid fire date,
-            if (nextFireTimeCursor.Date == null || nextFireTimeCursor.Date.Value.Year > MaxYear)
+            if (nextFireTimeCursor.Date == null || nextFireTimeCursor.Date.Value.Year > CronExpressionConstants.MaxYear)
             {
                 return null; // ran out of years
             }
--- a/src/Quartz/CronExpression.cs
+++ b/src/Quartz/CronExpression.cs
@@ -26,6 +26,7 @@ using System.Runtime.Serialization;
 using System.Text;
 using System.Text.RegularExpressions;
 
+using Quartz.Configuration;
 using Quartz.Util;
 
 namespace Quartz;
@@ -260,7 +261,7 @@ public sealed class CronExpression : ISerializable
     /// </summary>
     [NonSerialized] private bool calendarDayOfMonth;
 
-    public static readonly int MaxYear = DateTime.Now.Year + 100;
+    public static readonly int MaxYear =  TimeProvider.System.GetLocalNow().Year + 100;
 
     private static readonly Regex regex = new("^L(-\\d{1,2})?(W(-\\d{1,2})?)?$", RegexOptions.Compiled | RegexOptions.ExplicitCapture, TimeSpan.FromSeconds(5)); //e.g. LW L-0W L-4 L-12W LW-4 LW-12
     private static readonly Regex offsetRegex = new("LW-(?<offset>[0-9]+)", RegexOptions.Compiled | RegexOptions.ExplicitCapture, TimeSpan.FromSeconds(5));
--- a/src/Quartz/DateBuilder.cs
+++ b/src/Quartz/DateBuilder.cs
@@ -63,8 +63,18 @@ public sealed class DateBuilder
     private int second;
     private TimeZoneInfo? tz;
 
-    private DateBuilder(TimeProvider timeProvider)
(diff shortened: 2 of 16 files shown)


src/Quartz/CronExpression.cs
```diff
@@ -260,7 +261,7 @@ public sealed class CronExpression : ISerializable
     /// </summary>
     [NonSerialized] private bool calendarDayOfMonth;
 
-    public static readonly int MaxYear = DateTime.Now.Year + 100;
+    public static readonly int MaxYear =  TimeProvider.System.GetLocalNow().Year + 100;
```
lahma, line 264: I think we can just hard-code this to `2200`

src/Quartz/DateBuilder.cs
```diff
@@ -104,7 +92,7 @@ private DateBuilder(TimeZoneInfo tz)
     /// <returns></returns>
     public static DateBuilder NewDate()
```
lahma, line 93: maybe needs parameter `TimeProvider? timeProvider = null` and then we default to system?

src/Quartz/DateBuilder.cs
```diff
@@ -114,7 +102,7 @@ public static DateBuilder NewDate()
     /// <returns></returns>
     public static DateBuilder NewDateInTimeZone(TimeZoneInfo tz)
```
lahma, line 103: maybe needs parameter `TimeProvider? timeProvider = null` and then we default to system?

src/Quartz/DateBuilder.cs
```diff
@@ -259,7 +247,7 @@ public static DateTimeOffset TomorrowAt(int hour, int minute, int second)
         ValidateMinute(minute);
         ValidateHour(hour);
 
-        DateTimeOffset now = DateTimeOffset.Now;
+        DateTimeOffset now = TimeProvider.System.GetLocalNow();
```
lahma, line 250: maybe needs parameter `TimeProvider? timeProvider = null` and then we default to system?

src/Quartz/Impl/Triggers/CalendarIntervalTriggerImpl.cs
```diff
@@ -54,7 +54,7 @@ namespace Quartz.Impl.Triggers;
 [Serializable]
 public sealed class CalendarIntervalTriggerImpl : AbstractTrigger, ICalendarIntervalTrigger
 {
-    private static readonly int YearToGiveupSchedulingAt = DateTime.Now.AddYears(100).Year;
+    private static readonly int YearToGiveupSchedulingAt = TimeProvider.System.GetLocalNow().AddYears(100).Year;
```
lahma, line 57: maybe just hard-code this to `2200` too?

src/Quartz/Impl/Triggers/CronTriggerImpl.cs
```diff
@@ -386,6 +392,7 @@ public CronTriggerImpl(string name, string group, string cronExpression) : base(
         string cronExpression,
         TimeZoneInfo timeZone) : base(name, group, jobName, jobGroup)
     {
+        timeProvider = TimeProvider.System;
```
lahma, line 395: should the widest constructor take `TimeProvider? timeProvider = null` and then we default to system?

src/Quartz/Impl/Triggers/DailyTimeIntervalTriggerImpl.cs
```diff
@@ -79,7 +79,7 @@ public sealed class DailyTimeIntervalTriggerImpl : AbstractTrigger, IDailyTimeIn
     /// </summary>
     public const int RepeatIndefinitely = -1;
 
-    private static readonly int YearToGiveupSchedulingAt = DateTime.Now.Year + 100;
+    private static readonly int YearToGiveupSchedulingAt = TimeProvider.System.GetLocalNow().AddYears(100).Year;
```
lahma, line 82: maybe `2200` again

src/Quartz/DateBuilder.cs
jafin: As the DateBuilder constructors are private, I've reduced the constructor to one, made TimeZoneInfo? optional, and as per your comment, in the public method `NewDate` and `NewDateInTimeZone` default to TimeProvider.System when not provided.  Take a look at next PR let me know.

src/Quartz/DateBuilder.cs
jafin: Adjusted in next PR, let me know if suitable.

src/Quartz/Impl/Triggers/CronTriggerImpl.cs
jafin: Ok reworked, also trying to constructor chain where possible.

jafin: A more general question overall,  as perhaps I am going about things wrong.  We can inject/pass in a TimeProvider, but there is also this SystemTime with Funcs for Now/UtcNow.      I see these as both conflicting.  

`SystemTime` being static can be set at a global level, good for objects that sit outside of DI etc.  But also being static, are there situations where there is a need for separate `TimeProviders` ?  In which case I would assume the goal is to reduce dependency on `SystemTime` class.    I can't think of a scenario where different TimeProviders would be required?

So just wondering if taking parameters for a TimeProvider actually makes sense and should methods to leaning on `SystemTime` more?

src/Quartz/Impl/Triggers/CronTriggerImpl.cs
```diff
@@ -321,14 +328,15 @@ public CronTriggerImpl(string name, string group, string cronExpression) : base(
     /// <exception cref="ArgumentNullException"><paramref name="name"/>, <paramref name="group"/>, <paramref name="jobName"/> or <paramref name="jobGroup"/> are <see langword="null"/>.</exception>
     public CronTriggerImpl(string name, string group, string jobName,
```
jafin, line 329: This can be removed as it matches line 379.

src/Quartz/Impl/Triggers/CronTriggerImpl.cs
```diff
@@ -345,21 +353,8 @@ public CronTriggerImpl(string name, string group, string cronExpression) : base(
...
```
jafin, line 353: This could be removed as it matches line 379 due to optional parameters.

lahma (approved):
Great step towards the future, thank you!

PR #2312, opened 2024-03-28
Remove deep object cloning from the integration test project to reduce the usage of BinaryFormatter

adamsitnik:
Hello everyone!

My name is Adam, I work on the .NET Team where I am currently working on the removal of `BinaryFormatter` (https://github.com/dotnet/announcements/issues/293). To be exact I am writing a doc about how to migrate from BF.
I am searching for popular NuGet packages that depend on BF and checking if applying the knowledge from my doc would allow to port given library.

While analyzing Quartz.NET source code I've found `DeepClone` method that uses `BinaryFormatter`. After checking where it's used I got to a conclusion that it copies an immutable `GroupMatcher`, which seems to be redundant. I've removed it and made matchers immutable in explicit way by removing the private property setters (they were used only in ctor). I've also removed the empty private parameterless ctors as they required the non-nullable property to be set to `null!` in the property initializer.


lahma:
Thank you for taking the time and investigating the improvements, "popular NuGet packages" is quite the compliment ❤️

PR #2313, opened 2024-03-29
make IJobWrapper public

### LINKED ISSUE #2310: Expose IJobWrapper interface to allow access to the wrapped job
**Describe the bug**

It is not possible to access the inner job when it is wrapped inside of a IJobWrapper (https://github.com/quartznet/quartznet/blob/bf813bc02c9bddc7bd7a50ddc027687563aa657d/src/Quartz/IJobWrapper.cs#L3).

My actual use case is to display some Information about currently running jobs. Most information is directly accessible via the jobContext, but I want to display some public info which is exposed on the Job-Instance itself.
The JobContext exposes the JobInstance, but when using MicrosoftDI, I have access to the JobWrapper only.

Currently I resolve that problem by using Reflection and accessing the public property `Target` at the jobWrapper, but since the Wrapper-class is private and the interface is internal, I can't access the Property directly.

If the IJobWrapper could be changed to be public, it would be possible to check if the JobInstance inherits this interface and access the Target-Job without reflection.

Additionally it would be possible to reuse this interface for custom JobWrappers when creating custom JobFactories. In my actual project I wrap the MicrosoftDI-JobFactory again, to wrap the JobExecution with a Elastic-APM-Transaction. It would be great to resuse the same IJobWrapper interface here.

**Version used**

3.8.1, but it applies to main-branch (4.x) too

**To Reproduce**

Create a new project with MicrosoftDI.
Write the following code:
```
private readonly ISchedulerFactory _schedulerFactory;
async Task Test()
{
    var runningJobs = await scheduler.GetCurrentlyExecutingJobs();
    var target = (runningJobs.First().JobInstance as IJobWrapper).?Target;
}
```
IJobWrapper will not be accessible here.

**Expected behavior**

IJobWrapper (https://github.com/quartznet/quartznet/blob/bf813bc02c9bddc7bd7a50ddc027687563aa657d/src/Quartz/IJobWrapper.cs#L3) is public to allow access to the Target-Job

Philipp-Binder:
Changes
- change IJobWrapper to be public

Fixes #2310

### DIFF
--- a/src/Quartz/Simpl/PropertySettingJobFactory.cs
+++ b/src/Quartz/Simpl/PropertySettingJobFactory.cs
@@ -135,9 +135,16 @@ public class PropertySettingJobFactory : SimpleJobFactory
     /// <param name="data">The data to set.</param>
     public virtual void SetObjectProperties(object obj, JobDataMap data)
     {
-        foreach (string name in data.Keys)
+        if (obj is IJobWrapper jobWrapper)
         {
-            SetObjectProperty(obj, name, data[name]);
+            SetObjectProperties(jobWrapper.Target, data);
+        }
+        else
+        {
+            foreach (string name in data.Keys)
+            {
+                SetObjectProperty(obj, name, data[name]);
+            }
         }
     }
 
--- a/src/Quartz/Simpl/SimpleJobFactory.cs
+++ b/src/Quartz/Simpl/SimpleJobFactory.cs
@@ -88,7 +88,15 @@ public class SimpleJobFactory : IJobFactory
     /// </summary>
     public virtual void ReturnJob(IJob job)
     {
-        var disposable = job as IDisposable;
-        disposable?.Dispose();
+        if (job is IDisposable disposableJob)
+        {
+            disposableJob.Dispose();
+        }
+        // check for wrapped jobs only if the current job is not disposable
+        // disposable wrappers should handle inner disposal on it's own
+        else if (job is IJobWrapper jobWrapper)
+        {
+            ReturnJob(jobWrapper.Target);
+        }
     }
 }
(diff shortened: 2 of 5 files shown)


Philipp-Binder:
I realized that there are some caveats when implementing additional Wrappers around MicrosoftDependencyInjectionJobFactory. (see example here: https://github.com/quartznet/quartznet/discussions/1224#discussioncomment-9024888)
That's why I added some basic recursive implementations for SetObjectProperties and ReturnJob.

Just tell if I should split it, but I think the changes should not break any current behavior and all changes are related to the IJobWrapper functionality.
Instead it should make workarounds for custom JobFactory-Wrappers obsolete, while not breaking the workarounds.

src/Quartz/Simpl/PropertySettingJobFactory.cs
```diff
@@ -135,9 +135,16 @@ protected virtual IJob InstantiateJob(TriggerFiredBundle bundle, IScheduler sche
     /// <param name="data">The data to set.</param>
     public virtual void SetObjectProperties(object obj, JobDataMap data)
     {
-        foreach (string name in data.Keys)
+        if (obj is IJobWrapper jobWrapper)
```
Philipp-Binder, line 138: If we have a IJobWrapper here, we need to drill down until we get to the baseTarget and set the Properties there.

This removes the need for overriding SetObjectProperties for inheritors using the IJobWrapper interface.

src/Quartz/Simpl/SimpleJobFactory.cs
```diff
@@ -88,7 +88,15 @@ public virtual IJob NewJob(TriggerFiredBundle bundle, IScheduler scheduler)
...
     public virtual void ReturnJob(IJob job)
     {
-        var disposable = job as IDisposable;
-        disposable?.Dispose();
+        if (job is IDisposable disposableJob)
+        {
+            disposableJob.Dispose();
+        }
+        // check for wrapped jobs only if the current job is not disposable
+        // disposable wrappers should handle inner disposal on it's own
```
Philipp-Binder, line 96: Correct Disposing via ReturnJob is not really a trivial thing. For things like MicrosoftDI we need to have in mind, that we should dispose only if the scope does not handle the disposal on it's own.

I think we can assume, that the first disposable Job(-Wrapper) should be responsible for disposal down the road, so we just search for the first IDisposable if there is any and then we stop checking.

lahma:
Thank you, I'll migrate the changes to 3.x branch.

PR #2328, opened 2024-04-21
Adding some docs about the database support in quartz

drusellers:
Ok, this is an attempt to put together some docs on the tables. I'm starting to learn them a bit more as I need to debug things and wanted to share my learnings.

1. Where should we put this documentation in the TOC? I just picked a spot but would love any feedback.
2. I for the most part focused on documenting my current questions and how I see things.
3. I'm thinking I'd like more documentation on the trigger types and how they map to the tables. 🤔 

What else can we add?

### DIFF
--- a/docs/documentation/quartz-3.x/db/index.md
+++ b/docs/documentation/quartz-3.x/db/index.md
@@ -1,0 +1,37 @@
+---
+title: Database Schema
+---
+
+When using the JobStoreTX ADO, Quartz.Net will create 11 tables.
+
+
+| Table | Brief Description |
+| -- | -- |
+| qrtz_calendars | Stores non-standard calendars |
+| qrtz_job_details | Stores `IJobDetail` data |
+| qrtz_locks | locks used by quartz |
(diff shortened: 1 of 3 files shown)


docs/documentation/quartz-3.x/db/index.md
```diff
@@ -0,0 +1,37 @@
+---
+title: Database Schema
+---
+
+When using the JobStoreTX ADO, Quartz.Net will create 11 tables.
```
jafin, line 5: Just checking, I don't think Quartz creates tables, the user needs to run the scripts manually. 
Perhaps something like 


"When using JobStoreTX ADO, Quartz requires the creation of a set of tables.  This is a manual step, as Quartz.Net does not create these automatically.  These scripts can be found in the repo [database/tables](https://github.com/quartznet/quartznet/tree/main/database/tables).

lahma: I tweaked the wording a bit.

Issue #2332, opened 2024-04-23
Jobs with injected services that implement IAsyncDisposable are not disposed


williamdenton:
**Describe the bug**

When using Microsoft DI and a service is injected to a job that implements `IAsyncDisposable` the service is not disposed resulting in an InvalidOperationException and being thrown in the Scope finalizer.

The scope created on L30 below is passed to the scoped Job, which takes responsibility for disposing it
https://github.com/quartznet/quartznet/blob/72fce9c599e020f4db0b0c77d277f9971d947a6f/src/Quartz/Simpl/MicrosoftDependencyInjectionJobFactory.cs#L25-L34

However ScopedJob does not implement IAsyncDisposable so is unable to dispose the container correctly
https://github.com/quartznet/quartznet/blob/72fce9c599e020f4db0b0c77d277f9971d947a6f/src/Quartz/Simpl/MicrosoftDependencyInjectionJobFactory.cs#L67

This is where the  Job (scopedjob) is disposed
https://github.com/quartznet/quartznet/blob/72fce9c599e020f4db0b0c77d277f9971d947a6f/src/Quartz/Simpl/MicrosoftDependencyInjectionJobFactory.cs#L62-L65

https://github.com/quartznet/quartznet/blob/72fce9c599e020f4db0b0c77d277f9971d947a6f/src/Quartz/Core/JobRunShell.cs#L301-L306

So to fix this the change should be reasonably simple.

1. ScopedJob needs to implement IAsyncDisposable
2. ReturnJob needs to be converted from void to async task so it can call DisposeAsync on the job

**Version used**
3.8.1

**To Reproduce**

A job that has a service injected that implements IAsyncDisposeable

**Expected behavior**

ScopedJob needs to implement IAsyncDisposable to correctly clean up the scope

**Additional context**

```
System.InvalidOperationException: 'Namespace.AsyncDisposablClient' type only implements IAsyncDisposable. Use DisposeAsync to dispose the container.
  ?, in void ServiceProviderEngineScope.Dispose()
  ?, in void ScopedJob.Dispose()
  ?, in void MicrosoftDependencyInjectionJobFactory.ReturnJob(IJob job)
  ?, in async Task JobRunShell.Run(CancellationToken cancellationToken)
```

williamdenton:
Ive put up 2 PRs, one that fixes main and one for 3.x branch

[...]

PR #2333, opened 2024-04-23
Make ScopedJob implement IAsyncDisposable

williamdenton:
* Implement IAsyncDisposeable
* remove the overload of ReturnJob in MicrosoftDependencyInjectionJobFactory as it is the same as the base

Resolves https://github.com/quartznet/quartznet/issues/2332


src/Quartz/Simpl/MicrosoftDependencyInjectionJobFactory.cs
```diff
@@ -59,12 +59,7 @@ public override void SetObjectProperties(object obj, JobDataMap data)
         return (activatorCache.CreateInstance(serviceProvider, bundle.JobDetail.JobType), false);
     }
 
-    public override void ReturnJob(IJob job)
-    {
-        (job as IDisposable)?.Dispose();
-    }
```
williamdenton, line 65: removed this override as is the same a s the base

PR #2334, opened 2024-04-23
Make ScopedJob implement IAsyncDisposable (3.x branch)

williamdenton:
* Implement IAsyncDisposeable
* remove the overload of ReturnJob in MicrosoftDependencyInjectionJobFactory as it is the same as the base

Resolves https://github.com/quartznet/quartznet/issues/2332

Similar to https://github.com/quartznet/quartznet/pull/2333 but allows multi targeting for older new frameworks while still fixing the issue


williamdenton:
I have pulled this into my local project (replacing the nugets) and confirm it fixes the issue)

PR #2340, opened 2024-04-26
DOCS: Update VuePress to 2.0-RC7

jafin:
- Update VuePress from beta to RC7
- Use redirectPlugin


lahma (approved):
Looking good, thank your for the upgrade!

PR #2342, opened 2024-04-26
Fix lint warning about missing initial value in array.reduce in webpack.config.js Quartz.Web

jafin:
Supply default value for `reduce` to please Sonarlint.


lahma:
Thank you!

PR #2360, opened 2024-05-09
Support negative timezone offsets in TestTimeZone2

A9G-Data-Droid:
- Simplify GetStartOfDay
 - Ensure GetEndOfDay has maximum precision of platform

**Describe the bug**

I am seeing test `TestTimeZone2` introduced in #2113 failing.

**Version used**

Quartz.NET 3.9.0

**To Reproduce**

Have a negative timezone offset on your system clock. 
Run unit tests.

**Expected behavior**

Passing test.

**Additional context**
Looking at the test line where the failure occurs:

https://github.com/quartznet/quartznet/blob/e62b215388d4bba29d0c027c8249bafc6a2b6605/src/Quartz.Tests.Unit/Impl/Calendar/DailyCalendarTest.cs#L128

A negative `timeZoneOffset.Hours` is subtracted from 24 resulting in a time range greater than 24 hours which is more than the available hours in the day.

Adding a negative solves this issue for me. I believe this was the intent.


lahma:
Thank you!

Issue #2354, opened 2024-05-06
asynchronous method 'Quartz.Core.QuartzSchedulerThread.Run()', is waiting on a synchronous call to method 'System.Threading.Monitor.Wait(object, int)'


A9G-Data-Droid:
**Describe the bug**

I have a windows service I wrote in .NET8 that builds Quartz as a hosted service. It works great. It runs things on a schedule. When my service goes to stop it hangs indefinitely, never fully stopping. The final log I see is right before the quartz shutdown line. I have my services `ExecuteAsync` code wrapped in a try...catch with `await _scheduling.Shutdown().ConfigureAwait(false);` in the finally to ensure that this is the last thing it does before returning from `ExecuteAsync` where the host would handle the disposing of Quartz and everything else. 

I attached VS2022 debugger to the hung service, exported a dump, and analyzed it in VS2022. 

It says `AA0001    Thread pool thread or asnchronous task blocked on a synchronous call`

> 1 threads are performing asynchronous work in method 'Quartz.Core.QuartzSchedulerThread.Run()', but are waiting on a synchronous call to method 'System.Threading.Monitor.Wait(object, int)'. This may cause thread pool starvation and hangs.

Potential fix

> Do not synchronously wait on Monitors, Events, Tasks, or any other objects that may block your thread. See if you can update the method to be asynchronous.

**Version used**

Quartz (3.8.1)

**To Reproduce**

1. Follow [guide to make Windows Service in .NET Core](https://learn.microsoft.com/en-us/dotnet/core/extensions/windows-service)
2. Follow guide to [add Quartz to hosted service](https://www.quartz-scheduler.net/documentation/quartz-3.x/packages/hosted-services-integration.html#using)
3. Schedule a job
4. Start Quartz
5. Put `await _scheduling.Shutdown().ConfigureAwait(false);` at the end of your `ExecuteAsync` method inside your Windows service
6. Service is now hung and can't exit gracefully.

**Expected behavior**

`.Shutdown` should either complete or throw.

**Additional context**
[...]

Issue #2133, opened 2023-09-11
Deadlock in MisfireHandler


JustusGreiberORGADATA:
**Describe the bug**

Hi,

we are currently running a lot of test with the Quartz hosted service running in the background as part of our normal host setup.

Sometimes these tests just lock up. With the amount of tests we have, this is pretty much guaranteed to happen once every run. 
I let the test host run in blame mode and produced a minidump after a hang timeout of 40 minutes. Analyzing it, my best guess what happens is as follows:

https://github.com/quartznet/quartznet/blob/3.x/src/Quartz/Impl/AdoJobStore/MisfireHandler.cs creates a QueuedTaskScheduler and schedules the `Run()` method/Task to execute on this scheduler. The result of this operation is saved in the `task` field.
When the host is shutting down, the `Shutdown()` method of the MisfireHandler is called. This method cancels the token source in the handler, disposes the QueuedTaskScheduler and waits for the Task in field `task` to finish. Under normal circumstance the task should finish, because it polls the cancellationToken every once in a while.

The problem is, that the scheduler will never execute the Task, if the scheduler is disposed before it had any chance to schedule the Task. Therefore the await can never "finish".

I tried to recreate the behavior with a simple test: 

```C#
[TestFixture]
public class Class1
{
    [Test]
    public async Task DeadLock_QueuedTaskScheduler()
    {
        var taskScheduler = new QueuedTaskScheduler(threadCount: 1, "test", useForegroundThreads: false);

		// Initialize() in the original
        Task task = Task.Factory.StartNew(Run, CancellationToken.None, TaskCreationOptions.HideScheduler, taskScheduler).Unwrap();

		// Shutdown() in the original
        taskScheduler.Dispose(); // On my machine this happens so fast, that the taskScheduler has no chance to schedule the Run task.
        await task.ConfigureAwait(false); // This will deadlock, because the scheduler is disposed and won't schedule the continuation anymore.
        
        async Task Run()
        {
            await Task.Delay(1000).ConfigureAwait(false); // Fake work that according to the debugger is never reached on my machine.
        }
    }
}
``` 

On my system this test deadlocks consistently and analyzing the async stack traces, seems to be the same behavior that is experienced by my minidump. (Disclaimer: This is my first time working with minidumps so in the end it could turn out, that I am mistaken and this is not the cause 🙃.)

**Version used**

Quartz.AspNetCore Version="3.7.0"

**To Reproduce**

See the test. 

**Expected behavior**

[...]

PR #2365, opened 2024-05-10
Use Microsoft.VisualStudio.Threading to resolve deadlocks 

A9G-Data-Droid:
All tests are passing. I still need to integrate this into my main project and do a prove out there but I wanted to get this PR in front of you so I can get feedback on what I have attempted here.

I apologize for the large PR. As you know, async is contagious and must be implemented across a full chain of events.

I originally set out to remove `Monitor.Wait` from `QuartzSchedulerThread.Run`. I wanted to use a `PauseTokenSource` based on `AsyncManualResetEvent` found in `Microsoft.VisualStudio.Threading`. This library contains `Microsoft.VisualStudio.Threading.Analyzers` which helps resolve deadlocks. Due to the way warnings are configured as errors, I was forced to contend with each one which is how I ended up touching so many files. I attempted to resolve all material warnings without making breaking changes. There are still a couple places where async is waited synchronously or sync method is called in an async method, because I didn't want to change the signature of public methods.

I also used `JoinableTask` in shutdown methods where the intention is to wait for a task started elsewhere to finish. This might end up being the key to solving the shutdown deadlocks I was seeing in edge cases.

RELATED ISSUES:
- #2354
- #2133


Directory.Packages.props
lahma, line 1: This file looks like a merge/rebase gone wrong.

src/Quartz.Benchmark/DefaultThreadPoolBenchmark.cs
```diff
@@ -25,15 +25,15 @@ public void RunInThread_CompletedTask_MaxConcurrencyIsMaxValue_SingleThreaded()
     }
 
     [Benchmark(OperationsPerInvoke = 1_000_000)]
-    public void RunInThread_CompletedTask_MaxConcurrencyIsMaxValue_MultiThreaded()
+    public async Task RunInThread_CompletedTask_MaxConcurrencyIsMaxValue_MultiThreaded()
```
lahma, line 28: maybe these quite safe changes for tests/benchmarks could be a separate PR (calling async versions)

src/Quartz.Server/Program.cs
```diff
@@ -14,14 +15,16 @@ public static class Program
     /// <summary>
     /// Main.
     /// </summary>
-    public static void Main()
+    public static async Task Main()
```
lahma, line 18: Maybe also could be part of a separate cleanup PR as quite safe and outside of core domain

src/Quartz/Core/QuartzSchedulerThread.cs
lahma, line 1: This is the scariest part and requires careful review and testing, crucial parts of Quartz

src/Quartz/Core/QuartzSchedulerThread.cs
```diff
@@ -22,9 +22,10 @@
 using System.Data.Common;
 
 using Microsoft.Extensions.Logging;
-
+using Microsoft.VisualStudio.Threading;
```
lahma, line 25: Core Quartz should not have this kind of dependencies

src/Quartz/Quartz.csproj
```diff
@@ -38,6 +38,8 @@
     <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" />
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" />
     <PackageReference Include="Microsoft.Extensions.Options" />
+    <PackageReference Include="Microsoft.VisualStudio.Threading" />
+    <PackageVersion Include="Microsoft.VisualStudio.Threading" Version="17.10.48" />
```
lahma, line 42: I'm uncomfortable with having this kind of dependency

lahma (commented):
Had a quick read-through and far from proper review. Added some comments.

Directory.Packages.props
A9G-Data-Droid: Yes, that is likely.

src/Quartz.Benchmark/DefaultThreadPoolBenchmark.cs
A9G-Data-Droid: Ok, if I do that I need to add more ignore clauses to stop the warnings from becoming errors and blocking build.

lahma: Well without the analyzer it wouldn't break the build? So my suggestions would be a separate PR for all test/example/benchmark code without analyzers so it should work the same.

src/Quartz/Core/QuartzSchedulerThread.cs
A9G-Data-Droid: Ok, that is the key to what I'm doing here. There are similar tools found in https://github.com/StephenCleary/AsyncEx that we could use instead. I chose to use the Microsoft implementation because I assume they are going to be more popular and better maintained over time. Are you okay with using the Stephen Cleary libraries instead?

lahma: I'm still undecided. Because both are MIT licensed, we could just extract the minimal needed parts and add to Quartz codebase with attribution in each file's header...

A9G-Data-Droid: Ok, I ended up doing that with the PauseToken anyway.

A9G-Data-Droid:
Thank you for the review. This attempt turned out to be a "move fast and break things". I am going to step back and try again, paying attention to your feedback.

PR #2367, opened 2024-05-10
Use async pattern where already available.

A9G-Data-Droid:
These are the lines that were identified as Safe cleanup actions in my other PR.

 - Do not sync wait on ValueTask, always call ToTask first.
 - ContinueWith on TaskScheduler.Default instead of current


lahma:
Excellent, thank you!

PR #2373, opened 2024-05-14
Resolve some deadlocks

A9G-Data-Droid:
I know that you said you are scared of changes to the core of Quartz and for good reason. I am new here and I don't understand this code base. I want to submit my suggestions here but I'm not going to claim this is correct. There are always many ways to solve problems.

My goals here are to implement the two core tenants of async:

1. Do not synchronously block waiting in an async method.
2. Do not call asynchronous code from sync methods.

I have attempted to clarify the pause mechanism by implementing a clearly named async pause token. I replaced some old `lock`s with new SemaphoreSlim, which allows async wait. There is a concern that SemaphoreSlim does not maintain the order of those awaiting. 

I have tested this in my own solution and _"it works on my machine"_. It also appears to pass unit tests. So while it's functional, it may have another set of edge cases. 

I tried to maintain the existing structure and avoid breaking changes.


A9G-Data-Droid:
Spoke too soon. The result of that test here suggests it is still deadlocking.

lahma:
The underlying issue has been fixed

PR #2386, opened 2024-05-23
Added support for IHostedLifecycleService for the QuartzHostedService

GhostlyRaven:
Changes:
1. Added support for IHostedLifecycleService for the QuartzHostedService. This complements the existing methods of configuring related services from the DI container that are used in Quartz jobs with new configuration methods.
2. Fixed preprocessor directive. Old version of the directive: NET6_OR_GREATER. New version of the directive: NET6_0_OR_GREATER.


GhostlyRaven:
@lahma tell me is it normal that Qodana returns an error or is the reason for the error on my side?

lahma:
You could try to rebase against latest main, I tried to tweak Qodana configuration.

lahma:
Seems that it just can be a bit flaky, who knows..

GhostlyRaven:
@lahma Do you mean tweak Qodana configuration or other?

lahma:
I think it is what it is for now, so need to ignore.

GhostlyRaven:
Then I don't see any reason to delay the merger.

hybloid:
@lahma basically, Qodana step can not access the secret, since it's a PR from the user. See [link](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/) for the reference. There is a special trigger that helps to overcome that (some strategies stated there), but they will potentially expose the secret (e.g. your license).

architectbat:
One possible (but not a secure solution) - is to reuse [this example](https://github.com/fluentassertions/fluentassertions/blob/develop/.github/workflows/code_quality.yml). However I recommend to add labeled check to make it more secure.

lahma:
> @lahma basically, Qodana step can not access the secret, since it's a PR from the user. See [link](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/) for the reference. There is a special trigger that helps to overcome that (some strategies stated there), but they will potentially expose the secret (e.g. your license).
> 
> One possible (but not a secure solution) - is to reuse triggers as in [this example](https://github.com/fluentassertions/fluentassertions/blob/develop/.github/workflows/code_quality.yml?rgh-link-date=2024-05-28T21%3A31%3A12Z). However I recommend to add labeled check to make it more secure.

While this is great for personal PRs (trusted), lacking the possibility for having contributor PRs analyzed automatically is somewhat unfortunate. I understand the security perspective, but maybe if there would be a check like "It's a public repository, we can analyze it regardless the license information" - I'm not sure if this is feasible but I guess this is the Sonar cloud approach. 

Qodana is great for pointing out my own flaws, but would probably benefit contributors too if it would be easy to integrate to PR workflow. I'd say it's even more essential to get feedback from public PRs.

lahma:
So what's your use case behind this? I don't fully like the concept of introducing new callback interface when there's already an options object that could be used for callbacks (`Func`). Another option would be to unseal the hosted service and allow extending it with your own version.

GhostlyRaven:
1. My use case is that Quartz should definitely wait for a network connection to a remote server. It should not start and do anything without connection. The time delay available in the options does not work stably in this сase.
2. I agree with you about the availability of options and callbacks via `Func`, but for structuring the source code, this is a bad solution for me.
3. I like the solution with unseal the hosted service. It is a compromise solution in my opinion.

P.S. If you support my 3rd answer, then I could implement it by adding corrections to the preprocessor directives and updating the extension methods to work with Quartz.

lahma:
I tweaked the code a bit and now it's quite minimal allowing you to inherit from default hosted service and override the methods.

lahma:
Thanks for the idea and PR!

GhostlyRaven:
@lahma, I found an error in QuartzHostedService in the StoppedAsync [method](https://github.com/quartznet/quartznet/blob/33e4af7b2c2a1ab8aa34a69b11beb7f7cb3ca880/src/Quartz/Hosting/QuartzHostedService.cs#L126). You have lost the modifier virtual.

lahma:
> @lahma, I found an error in QuartzHostedService in the StoppedAsync [method](https://github.com/quartznet/quartznet/blob/33e4af7b2c2a1ab8aa34a69b11beb7f7cb3ca880/src/Quartz/Hosting/QuartzHostedService.cs#L126). You have lost the modifier virtual.

Please send a PR 🙂

PR #2394, opened 2024-06-05
escape asterisk so it will display in markdown file

bretthoes:


lahma (approved):
Thank you, I sure hope that you aren't using the v1 anymore but migrate to 3.x as soon as possible 🙂

bretthoes:
Unintentionally ended up on v1 doc page and noticed this :sweat_smile: Love using quartz, thank you!

PR #2401, opened 2024-06-14
Fix ScheduledFireTimeUtc for CronTrigger (#605)

### LINKED ISSUE #605: IJobExecutionContext.ScheduledFireTimeUtc is set to the the same value is FireTimeUtc
### Expected behavior
Based on the help description for the **ScheduledFireTimeUtc** property, I would expect the **ScheduledFireTimeUtc** property to contain the time the trigger was _expected_ to start the job. If the trigger fired 1 second late, 2 seconds late, or 10 seconds late for a given time, I would expect the **ScheduledFireTimeUtc** to have contained the same value, regardless of how much the trigger was delayed.

### Actual behavior
The **ScheduledFireTimeUtc** property contains the same value as the **FireTimeUtc** property (i.e. the time the trigger _actually_ ran).

This is preventing my code that (appeared to) work in v2 from working in v3. I have code that prevents a single IJobDetail from running multiple times for the same scheduled fire time (i.e. when multiple triggers are present and overlap with the exact same time).

### Steps to reproduce
Define two triggers that trigger at the same time for a single IJobDetail. Put a breakpoint in the **Execute** method of your IJob implementation, and wait for 10 seconds or so when the breakpoint is hit for the first trigger. When you continue execution and the second trigger his fired late, you will notice that the **ScheduledFireTimeUtc** is showing the time that the trigger actually fired.

This is occurring with the **DailyTimeIntervalTriggerImpl** trigger type.

DPDmancul:
This fixes #605 for CronTrigger using the suggestion of @jayquest https://github.com/quartznet/quartznet/issues/605#issuecomment-830077468

### DIFF
--- a/src/Quartz/Impl/Triggers/CronTriggerImpl.cs
+++ b/src/Quartz/Impl/Triggers/CronTriggerImpl.cs
@@ -818,8 +818,8 @@ namespace Quartz.Impl.Triggers
         /// <seealso cref="JobExecutionException" />
         public override void Triggered(ICalendar? cal)
         {
-            previousFireTimeUtc = nextFireTimeUtc;
-            nextFireTimeUtc = GetFireTimeAfter(nextFireTimeUtc);
+            previousFireTimeUtc = GetFireTimeAfter(previousFireTimeUtc ?? startTimeUtc);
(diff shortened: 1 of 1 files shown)


lahma:
superseded by #2899

Issue #2405, opened 2024-06-19
NativeJob crashes when workingDirectory parameter is missing


viebrix:
After Upgrading Quartz.NET, I have the issue that in NativeJob the handling of workingDirectory is not the same as before.

In older versions providing a workingDirectory wasn't needed. It was an optional parameter.
Now with version 3.9.0 NativeJob breaks with an exception: 
`Quartz.SchedulerException: Job threw an unhandled exception. ---> System.Collections.Generic.KeyNotFoundException: Key workingDirectory not found`

I suspect, that this line in [NativeJob.cs](https://github.com/quartznet/quartznet/blob/b9ea62626c5581ee5402c733ab61630fe75fc894/src/Quartz.Jobs/Job/NativeJob.cs#L83) produces the error.
`        var workingDirectory = data.GetString(PropertyWorkingDirectory);`
In older versions GetString behaved in another way. Null values in GetString were no problem. Now it seems to break something. See also https://github.com/quartznet/quartznet/pull/2125 

This [line of code](https://github.com/quartznet/quartznet/blob/b9ea62626c5581ee5402c733ab61630fe75fc894/src/Quartz.Jobs/Job/NativeJob.cs#L140)  with string? seems to show that null values are allowed and it is intended to skip the workingDirectory parameter:
`private int RunNativeCommand(string command, string parameters, string? workingDirectory, bool wait, bool consumeStreams)`

Thank you in advance for investigating.

viebrix:
https://github.com/quartznet/quartznet/pull/2406

lahma:
Fixed in 3.x, main does not suffer from this.

PR #2406, opened 2024-06-19
Bugfix if workingDirectory is missing in MergedJobDataMap

viebrix:
This could be one possible correction for issue:
https://github.com/quartznet/quartznet/issues/2405
But there are also other possible solutions... but also feel free to ignore this pull request

(I wasn't able to test this code, because other changes in quartz.net were not compatible with by project and dependencies)

### DIFF
--- a/src/Quartz.Jobs/Job/NativeJob.cs
+++ b/src/Quartz.Jobs/Job/NativeJob.cs
@@ -131,7 +131,11 @@ public class NativeJob : IJob
             consumeStreams = data.GetBooleanValue(PropertyConsumeStreams);
         }
 
-        var workingDirectory = data.GetString(PropertyWorkingDirectory);
+        string? workingDirectory = null;
+        if (data.ContainsKey(PropertyWorkingDirectory))
+        {
+            workingDirectory = data.GetString(PropertyWorkingDirectory);
+        }
         int exitCode = RunNativeCommand(command, parameters, workingDirectory, wait, consumeStreams);
         context.Result = exitCode;
         return default;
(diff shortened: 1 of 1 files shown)


src/Quartz.Jobs/Job/NativeJob.cs
```diff
@@ -131,7 +131,11 @@ public virtual ValueTask Execute(IJobExecutionContext context)
             consumeStreams = data.GetBooleanValue(PropertyConsumeStreams);
         }
 
-        var workingDirectory = data.GetString(PropertyWorkingDirectory);
+        string? workingDirectory = null;
```
jafin, line 134: You should be able to shorten this to just
```c#
data.TryGetString(PropertyWorkingDirectory, out var workingDirectory);
```

jafin:
Thanks for the fix.  This highlighted another potential issue when adding a quick test, may I suggest adding the following test code to your PR

```c#
// Tests\Job\NativeJobTests.cs
[TestFixture]
public class NativeJobTests
{
    [Test]
    public void TestNativeJob()
    {
        var job = new NativeJob();
        var context = TestUtil.NewJobExecutionContextFor(job);
        context.MergedJobDataMap.Put(NativeJob.PropertyCommand, "Test");
        //When
        Action act = () => job.Execute(context);
        act.Should().NotThrow<Exception>();
    }
}
```

and NativeJob line 122: `PropertyParameters` is also optional and suffers from the same issue you've identified, so something like:

```c#
	data.TryGetString(PropertyParameters, out var parameters);
	parameters ??= string.Empty;
```

viebrix:
Thanks for the fast answer and your help! In meantime I have added default values for workingDirectory in my project. I didn't know that you react as fast as you did - and I needed a solution till today. Thanks a lot!

lahma:
I'll backport to `main` manually, thanks for the 3.x fix.

PR #2407, opened 2024-06-20
Update NativeJob.cs

viebrix:
Behavier of GetString has changed. Now NativeJob raised an exception if parameters or workingDirectory wasn't included in MergedJobDataMap. See issue 2406
https://github.com/quartznet/quartznet/pull/2406


lahma:
Thank you!

PR #2410, opened 2024-06-23
in HttpClientOptions changed the AssertValid if condition  

AWasseem:
(string.IsNullOrWhiteSpace(HttpClientName) || HttpClient == null)

insetead of
if ((string.IsNullOrWhiteSpace(HttpClientName) && HttpClient == null) ||
    (!string.IsNullOrWhiteSpace(HttpClientName) && HttpClient != null))

### DIFF
--- a/src/Quartz.HttpClient/HttpClientOptions.cs
+++ b/src/Quartz.HttpClient/HttpClientOptions.cs
@@ -37,8 +37,7 @@ public class HttpClientOptions
             throw new InvalidOperationException("Scheduler name required");
         }
 
-        if ((string.IsNullOrWhiteSpace(HttpClientName) && HttpClient == null) ||
-            (!string.IsNullOrWhiteSpace(HttpClientName) && HttpClient != null))
+        if (string.IsNullOrWhiteSpace(HttpClientName) || HttpClient == null)
         {
             throw new InvalidOperationException("Either http client name or http client instance required");
         }


lahma:
I don't this change is correct, the original code does what it is supposed to.

AWasseem:
try to initialize HttpClient like this despite of filling HttpClient and HttpClientName it will throw exception **Either http client name or http client instance required**
```
builder.Services.AddQuartzHttpClient(e => {
    e.HttpClient = new HttpClient
    {
        BaseAddress = new Uri("http://localhost:5000/quartz-api/"),
        DefaultRequestHeaders = { { "X-Quartz-ApiKey", "MySuperSecretApiKey" } }
    };
    e.SchedulerName = "HRMSScheduler";
    e.HttpClientName = "HRMSSchedulerHttpClient";
});
```

jafin:
Is the error message to ambiguous?  Maybe separating to be more clear about both error scenarios?
```c#
  if (string.IsNullOrWhiteSpace(HttpClientName) && HttpClient == null) 
        {
            throw new InvalidOperationException("Either http client name or http client instance required");
        }
        
        if (!string.IsNullOrWhiteSpace(HttpClientName) && HttpClient != null)
        {
            throw new InvalidOperationException("You have set both an HttpClientName and an HttpClient instance. Please specify only one.");
        }
```

lahma:
Maybe improving the error reporting is a good option.

PR #2414, opened 2024-06-25
HttpClient construction. Add separate error messages to better indicate the Invalid state.

jafin:
HttpClient construction. Add separate error messages to better indicate the Invalid state.

Relates to #2410


lahma:
Thank you!

PR #2421, opened 2024-06-27
Refactor code and implemet primary constructor

KhanbalaRashidov:


lahma:
Thank you but I don't see a need for a such large change with little benefits.

PR #2434, opened 2024-07-03
Fix issues with STJ serializer

### LINKED ISSUE #2426: BlobTriggers do not serialize NextFireTime, resulting in them never firing
**Describe the bug**

Attempting to convert from Quartz2 binary serialization to Quartz3 JSON serialization using the serializer linked [here](https://www.quartz-scheduler.net/documentation/quartz-3.x/packages/system-text-json.html#migrating-from-binary-serialization). When a Blob Trigger is serialized using the JSON serializer, the `nextFireTime` value is not included, resulting in this value not being loaded and the [trigger never firing](https://github.com/quartznet/quartznet/blob/main/src/Quartz/Impl/AdoJobStore/JobStoreSupport.cs#L2757-L2766), since, for Blob Triggers, the value in the QRTZ_TRIGGERS table does not overwrite the serialized value in the QRTZ_BLOB_TRIGGERS table when `Delegate.SelectTrigger` is [called](https://github.com/quartznet/quartznet/blob/main/src/Quartz/Impl/AdoJobStore/StdAdoDelegate.Triggers.cs#L708-L718)

**Version used**

3.10

**To Reproduce**

Create a blob trigger, and write it to the database using JSON Serialization

**Expected behavior**

The trigger should fire

**Additional context**

I don't know if blob triggers are no longer supported, or if they are and simply do not work as expected.

VilleHakli:
- Serializing all details for triggers that AdoJobStore stores
- Improve JSON serialization tests
- Add Verify & TimeProvider fake to make testing easier
- Remove outdated comments about JsonPropertyAttribute

fixes #2426


src/Quartz.Serialization.SystemTextJson/Calendars/DailyCalendarSerializer.cs
```diff
@@ -19,9 +19,9 @@ protected override DailyCalendar Create(JsonElement jsonElement, JsonSerializerO
 
     protected override void SerializeFields(Utf8JsonWriter writer, DailyCalendar calendar, JsonSerializerOptions options)
     {
+        writer.WriteBoolean(options.GetPropertyName("InvertTimeRange"), calendar.InvertTimeRange);
```
VilleHakli, line 22: This change was done to generate exactly the same JSON between STJ & Newtonsoft. This makes it possible to test generated JSON with Verify, to assert the same output is generated.

Might be something that is not possible in the long run, but for now only small changes were needed.

src/Quartz.Serialization.SystemTextJson/Converters/CalendarConverter.cs
```diff
@@ -88,6 +86,11 @@ public override void Write(Utf8JsonWriter writer, ICalendar value, JsonSerialize
 
             writer.WriteString(options.GetPropertyName("Description"), value.Description);
 
+            if (value is BaseCalendar baseCalendar)
```
VilleHakli, line 89: This change was done to generate the same JSON between STJ & Newtonsoft. See previous comment for more info.

src/Quartz.Tests.Unit/Quartz.Tests.Unit.csproj
```diff
@@ -5,6 +5,9 @@
     <Nullable>disable</Nullable>
     <NoWarn>SYSLIB0011;CS0618</NoWarn>
     <EnableUnsafeBinaryFormatterSerialization>true</EnableUnsafeBinaryFormatterSerialization>
+
+    <!-- Microsoft.Extensions.TimeProvider.Testing 8.6.0 doesn't support net472 and has not been tested with it... -->
+    <SuppressTfmSupportBuildWarnings>true</SuppressTfmSupportBuildWarnings>
```
VilleHakli, line 10: Microsoft.Extensions.TimeProvider.Testing has net462 target and seems to work when running with net 472. Not sure if we want to disable all the warnings as SuppressTfmSupportBuildWarnings is not M.E.T.T specific.

src/Quartz.Tests.Unit/Simpl/JsonObjectSerializerTest.cs
```diff
@@ -1,232 +1,644 @@
...
-        private static void WriteJson(byte[] bytes)
+        public override string CalendarTypeName => "TestCalendar";
+    }
+}
+
+public class JsonSerializationTestTrigger : SimpleTriggerImpl
+{
+    public int CustomProperty { get; set; }
+
+    public sealed class SystemTextJsonSerializer : TriggerSerializer<JsonSerializationTestTrigger>
```
VilleHakli, line 606: I'm not happy with the amount of code needed to support a "simple" custom trigger like in this case. However, there might not be many cases where a custom trigger is required to have serialization support so left the code as is.

Some ideas I had for improvements:
* Change the CreateScheduleBuilder to Create which returns IMutableTrigger. The TriggerConverter would then use the StaticScheduleBuilder. This way, there would be no need for the added DeserializeFields method as all custom fields could be initialized in the create method.
* Change the serializers for built-in triggers to be public and extendable (somehow...). This would make it easier to create serializers for triggers inherited from the built-in triggers. This might also be the most likely case how custom triggers are implemented.

src/Quartz.Serialization.SystemTextJson/JsonConfigurationExtensions.cs
```diff
@@ -2,14 +2,16 @@
...
 using Quartz.Simpl;
+using Quartz.Triggers;
 
 namespace Quartz;
 
 public static class JsonConfigurationExtensions
 {
     /// <summary>
-    /// Use Newtonsoft JSON as data serialization strategy.
+    /// Use System.Text.Json as data serialization strategy.
```
lahma, line 14: 😅

src/Quartz.Tests.Unit/Simpl/JsonObjectSerializerTest.cs
lahma: Yes, should be quite rare case and should probably be left for the custom trigger implementer.

lahma (approved):
Great work, as always. Thank you!

src/Quartz.Serialization.SystemTextJson/JsonConfigurationExtensions.cs
VilleHakli: Copy & paste for the win!

PR #2444, opened 2024-07-04
Fix issues with STJ serializer

VilleHakli:
- Serializing all details for triggers that AdoJobStore stores
- Improve JSON serialization tests
- Add Verify & TimeProvider fake to make testing easier
- Remove outdated comments about JsonPropertyAttribute
- Fix TODOs in tests related to STJ serialization which were added in PR #2435

#2426


VilleHakli:
This has all the same changes as in #2434 but this time targeting main. Also contains fixes for TODOs added in #2435

Verify.NUnit adds implicit using for NUnit.Framework and as main has implicit usings enabled, IDE shows a lot of errors related to unneeded usings after these changes. I did not remove the unneeded usings in this PR to not bloat it and will create a second PR once this hits the main to fix the issues.

src/Quartz.Tests.Integration/Impl/SmokeTestPerformer.cs
```diff
@@ -516,4 +513,24 @@ protected override void DeserializeFields(CustomCalendar calendar, JObject sourc
     {
         calendar.SomeCustomProperty = source["SomeCustomProperty"]!.Value<bool>();
     }
+}
+
+internal sealed class CustomSystemTextJosnCalendarSerializer : Serialization.Json.Calendars.CalendarSerializer<CustomCalendar>
```
lahma, line 518: `Josn` 😉

lahma (approved):
Awesome, thank you!

src/Quartz.Tests.Integration/Impl/SmokeTestPerformer.cs
VilleHakli: That's the advanced version of JSON :)

PR #2449, opened 2024-07-05
Allow MSSQL Init Database script to be configured to not delete existing tables

### LINKED ISSUE #2179: Databases get recreated when data exists
**Describe the bug**
When using the sql scripts to create databases to use with the Ado job store, the databases get recreated whether or not data exists in the tables.

I suspect this is due to the way tables are checked for existence. Currently, the existence of the raw table is checked, but we don't check whether any records exist in the table that may or may not exist.
**Version used**
Latest in the main branch.
With which version did you encounter this behavior.
3.5.0+
**To Reproduce**
Run the database scripts to create a database for hte ado job store (the database provider should not matter here). Then use quartz and add some jobs and triggers. Next, re-run the program, as the tables exist they will be dropped and re-created.

**Expected behavior**
If tables exist with data in, I would expect them not to be dropped and recreated.

jafin:
Adjust MSSQL DB Init script to allow user to set conditional flag to NOT delete existing Quartz tables if desired.
Resolves #2179

No change to existing functionality, as DbScript still set to delete enabled by default


lahma:
There are 3 different schema init files for MS SQL Server, should this be implemented in all of them?

jafin:
SQLMOT and SQLBelow2016 scripts now updated.

lahma (approved):
Great work thank you!

lahma:
I'll bring this to main too.

PR #2451, opened 2024-07-05
Update Build actions to use checkout/@v4 & cache/@v4

jafin:
(v3 is deprecated)


lahma:
Can you change this to target only docs and qodana, they are not auto-generated?

lahma (approved):
Looks good, thank you!

PR #2453, opened 2024-07-05
Fix DbDataSource connection creation

Hawxy:
My bad on this. I was testing out the change today and ran into an exception due to the existing ADO implementation assigning a connection & transaction to the command. This'll resolve that issue.


lahma:
Now worries, thanks for fixing!

PR #2456, opened 2024-07-05
Add Async DB transactions for modern targets

Hawxy:


lahma (approved):
Great, thank you. One of the many reasons that I would like to ditch whole full framework (https://github.com/quartznet/quartznet/discussions/2080) and enjoy better APIs without pre-processor directives...

Do you think this would be something you'd like to see backported to 3.x?

Hawxy:
> Do you think this would be something you'd like to see backported to 3.x?

It's a nice-to-have but given it's a bit of a messy API change I'll leave to your discretion.

Issue #2330, opened 2024-04-22
CronExpresion.GetTimeAfter caught exception 


lynnz11:
**Describe the bug**

CronExpression cronExpression = new CronExpression("0 0 12 31W * ?");
DateTime cal = new DateTime(2024, 1, 31, 12, 0, 0).ToUniversalTime();
var nextFireTime = cronExpression.GetTimeAfter((DateTimeOffset)cal);

The exception: System.ArgumentOutOfRangeException
Year, Month, and Day parameters describe an un-representable DateTime.

**Version used**

3.8.1

**To Reproduce**

See Describe the bug

**Expected behavior**

No exception.

**Additional context**
CronExpression Class line 1747
private SortedSet<int> CalculateDaysOfMonth(DateTimeOffset dt)
{
   if.....
//line 1510
  else if (nearestWeekday) //AND not lastDay
 {
      var lastDayOfMonth = GetLastDayOfMonth(dt.Month, dt.Year);
      // current
           var day = daysOfMonth.Min;
      // update to following
           var day = Math.Min(daysOfMonth.Min, lastDayOfMonth);
      var tcal = new DateTimeOffset(dt.Year, dt.Month, day, 0, 0, 0, dt.Offset);
  .....
}
.....

lynnz11:
Also in version 3.8.1

PR #2341, opened 2024-04-26
Fix CronExpression not always returning correct next schedule for weekdayonly and month flags.  Relates to #2331

jafin:
Fix an issue whereby setting a monthly `CronExpression` with flag `weekdayonly`  `W` , 
when the scheduled day falls on a weekend (Sun/Sat), 
Quartz would incorrectly skip ahead to the next month (or non weekend date month when calling `GetTimeAfter`

Full detail in #2331 
- Targets v4

Incorporate test and fix from #2330 by @Skimmenthal13 (credit PR #2490) relating to NextFireTime bug when next month is shorter and using `W` modifier at end of month with greater days than next month

PR #2490, opened 2024-07-25
Issue #2330 I Add a check in nearestWeekday (CronExpression.cs) secti…

Skimmenthal13:
…on in order to prevent exeption. If DaysOfMonth variable is greater than lastDayOfTheMonth I use the second to populate the variable day.  Without this check we tryed to istantiate invalid DateTimeOffset like 02/31/2025 now quartz will return 02/28/2025. I also added the test.


jafin:
@lahma, @Skimmenthal13  this should be resolved via #2341  as I added the test case developed by @Skimmenthal13  and is passing.

lahma:
Thanks, should be fixed already in main.

PR #2502, opened 2024-08-02
Change scheduler repository to be singleton scoped to DI under MS DI integration

### LINKED ISSUE #1453: Use injected repository service instead of static singleton

lahma:
Make `SchedulerRepository` internal

fixes #1453

### DIFF
--- a/src/Quartz/Core/QuartzScheduler.cs
+++ b/src/Quartz/Core/QuartzScheduler.cs
@@ -264,7 +264,7 @@ namespace Quartz.Core
         /// properties.
         /// </summary>
         /// <seealso cref="QuartzSchedulerResources" />
-        public QuartzScheduler(QuartzSchedulerResources resources, TimeSpan idleWaitTime) : this()
+        internal QuartzScheduler(QuartzSchedulerResources resources, TimeSpan idleWaitTime) : this()
         {
             this.resources = resources;
 
--- a/src/Quartz.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
+++ b/src/Quartz.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
@@ -58,6 +58,7 @@ namespace Quartz
                 configure(target);
             }
 
+            services.AddSingleton<ISchedulerRepository>(new SchedulerRepository());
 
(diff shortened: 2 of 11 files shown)


src/Quartz/Core/QuartzScheduler.cs
```diff
@@ -264,7 +264,7 @@ protected QuartzScheduler()
         /// properties.
         /// </summary>
         /// <seealso cref="QuartzSchedulerResources" />
-        public QuartzScheduler(QuartzSchedulerResources resources, TimeSpan idleWaitTime) : this()
+        internal QuartzScheduler(QuartzSchedulerResources resources, TimeSpan idleWaitTime) : this()
```
smbecker, line 267: Can I ask why the visibility of this constructor was changed? This breaks my code and makes this class impossible to derive from given that `resources` is assumed to be not-null but I can't see a way to set it without the visibility of this constructor being restored.

lahma: Can you show the code?

smbecker: I just create an instance of the scheduler using a custom configured `QuartzSchedulerResources`, `new QuartzScheduler(resources, waitTime);`. Aside from how I am using this class, with this change, this class becomes impossible to derive from. It has an empty protected constructor, but developers can't actually use that since that would leave `resources` as null, which the rest of the class assumes that it isn't.

Issue #2349, opened 2024-04-30
CalendarIntervalSchedule is firing non-stop for 10 mins on the day that DST begins (and we lose an hour).


mwithum-paylocity:
**Describe the bug**

**Version used**

3.3.2

**To Reproduce**

Consider a Calendar Interval Schedule that:

* Runs every other Sunday at 2:01 AM Central
* Starts on 2/25/2024 2:01 AM
* And lines up to run at 2:01 AM on 3/10/2024
 * but that is the day that the clock rolls forward, so 2:01 AM doesn't exist
* Quartz will (on the previous fire in 2/25/2024) correctly set the next fire time to 3/10/2024 3:00 AM -- the earliest time that exists after it was "supposed to fire"

So far all seems correct. However, when 3/10/2024 3:00 AM comes around - the trigger fires non-stop for 10 mins exactly. It is almost like the next fire time never calculates a good time after 3/10/2024 3:00 AM. We had one trigger fire > 30k times.

I can repro this in the `TriggerUtils.ComputeFireTimes` method as well. I've attached an annotated sample where `TriggerUtils.ComputeFireTimes` is doing exactly what I am seeing when Quartz do - it gets stuck on 3/10/2024 3:00 AM.

[QuartzDstRepro.zip](https://github.com/quartznet/quartznet/files/15169513/QuartzDstRepro.zip)

**Expected behavior**

The trigger should only fire once and the `TriggerUtils.ComputeFireTimes` should calculate the proper next fire time.

jafin:
> Quartz will (on the previous fire in 2/25/2024) correctly set the next fire time to 3/10/2024 3:00 AM -- the earliest time that exists after it was "supposed to fire"

Should the next fire to be `3/10/2024 3:01`
Given the StartTime has a 1min offset,  Quartz to should attempt to jump forward the DST interval amount and trigger
Thoughts?

I'm attempting a PR, just trying to establish what everyone thinks is the expected result as I'm in a bit of a grey zone.
The issue I found, is Quartz in this scenario, is leaping ahead in 1 min intervals to find the next possible fire time after the DST event, but this causes it to not use the original startDate minute offset, so it appears to get stuck in a bit of a loop because nextFireTime is one minute less than a calculated next firetime of of 1min past the hour.

i8beef:
Has there been any movement on this? I am seeing similar behavior at last night's DST cross over on 3.8.0...

lahma:
[...]

PR #2515, opened 2024-08-11
Fix issue where CalendarIntervalSchedule is not correctly advancing next-fire time when cross DST event

jafin:
Fix an issue whereby `CalendarIntervalSchedule` `NextFireTime` was getting 'stuck' on a date when passing through a DST zone event.

The change adjusts the increment to get to the next firetime in intervals of 15min, to align to the minimum DST intervals of 30,45, 1hr)    The existing code would increment in 1 minute intervals to reach the next viable fire time.  This causes issues when comparing against the origin starttime if the startTime had a minute schedule that was not on the hour.

-Add test case from #2349 

This is still a WIP, as although it solves the #2349 case, I'm sure there is a can of worms here that are not tested.


src/Quartz/Impl/Triggers/CalendarIntervalTriggerImpl.cs
```diff
@@ -62,6 +62,7 @@ public sealed class CalendarIntervalTriggerImpl : AbstractTrigger, ICalendarInte
     private DateTimeOffset? previousFireTimeUtc;
     private int repeatInterval;
     internal TimeZoneInfo? timeZone;
+    private const int TimeZoneAdjustmentIntervalMinutes = 15;  //timezones can be multiples of 15 minutes (30min, 45min, 1hr)
```
jafin, line 65: I wonder instead of this interval, if we just find the DST difference and add that amount, so if previous DST to new DST is +1hr , add 60 minutes ?

PR #2516, opened 2024-08-11
Refactor UnitTests

jafin:
- Update Assertions to use constraint model ([Nunit 4 suggested approach](https://docs.nunit.org/articles/nunit/release-notes/Nunit4.0-MigrationGuide.html))
  - i.e. `Assert.True` becomes `Assert.That( value, Is.True)`
  - `Assert.Equal` => `Assert.That(val, Is.EqualTo(expected)`
  - `Assert.IsNotNull` => `Assert.That(val, Is.Not.Null)`
  - etc
- Wrap multiple asserts inside [Assert.Multiple](https://docs.nunit.org/articles/nunit/writing-tests/assertions/multiple-asserts.html)
- Minor refactor on language optimizations (object initializer, empty braces) in test lib.
- Add FluentAssertions.Analyzer package
- Add NUnit.Analyzers to package
- Remove the global using Assert for ClassicAssert, no longer required.
- Add `[OneTimeTearDown]` to dispose of objects Nunit.Analyzer requested. (DisallowConcurrentJobExecutioNTests)

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -10,6 +10,7 @@
     <PackageVersion Include="FakeItEasy" Version="8.3.0" />
     <PackageVersion Include="FirebirdSql.Data.FirebirdClient" Version="10.3.1" />
     <PackageVersion Include="FluentAssertions" Version="6.12.0" />
+    <PackageVersion Include="FluentAssertions.Analyzers" Version="0.33.0" />
     <PackageVersion Include="log4net" Version="2.0.17" />
     <PackageVersion Include="MELT" Version="0.9.0" />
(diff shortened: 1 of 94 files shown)


lahma:
Wow, you really did bite the bullet with this one. Is this ready for review?

jafin:
@lahma yes ready for review.

lahma (approved):
Great work here, a lot of grind required to improve these and thank you for taking the time. Ideally we would use FluentAssertions for everything, but that would be a huge undertaking.

jafin:
> Ideally we would use FluentAssertions for everything, but that would be a huge undertaking.

Maybe next wet uneventful day I have free. 🥲

Issue #2475, opened 2024-07-18
Infinite trigger fires in the first hour after switching from DST to ST


kevinkloet-koronet:
**Describe the bug**

We have a trigger with a cron schedule that should fire once every minute at second 0. During the first hour after switching from daylight saving time to standard time, the trigger fires non-stop. The NextFireTimeUtc seems to be 1 hour too early on the IJobExecutionContext.

**Version used**

3.11.0 (also tested against 3.8.0 and 2.6.1)

**To Reproduce**

I have created a project which I have used to test: https://github.com/kkloet1/QuartzNetDST

When run just before the switch from DST to ST we get regular behaviour (with the 1 hour jump as described in the documentation)
```
---------
Recovering False; FireInstanceId 638341378807257526; FireTimeUtc 29/10/2023 00:58:00 +00:00; NextFireTimeUtc 29/10/2023 00:59:00 +00:00; PreviousFireTimeUtc ; RefireCount 0; ScheduledFireTimeUtc 29/10/2023 00:58:00 +00:00;
---------
Recovering False; FireInstanceId 638341378807257527; FireTimeUtc 29/10/2023 00:59:00 +00:00; NextFireTimeUtc 29/10/2023 02:00:00 +00:00; PreviousFireTimeUtc 29/10/2023 00:58:00 +00:00; RefireCount 0; ScheduledFireTimeUtc 29/10/2023 00:59:00 +00:00;
```

But when we start the application in the hour after the switch from DST to ST we get this:
```
---------
Recovering False; FireInstanceId 638341380475357880; FireTimeUtc 29/10/2023 01:00:49 +00:00; NextFireTimeUtc 29/10/2023 00:01:00 +00:00; PreviousFireTimeUtc 29/10/2023 01:00:49 +00:00; RefireCount 0; ScheduledFireTimeUtc 29/10/2023 01:00:49 +00:00;
---------
Recovering False; FireInstanceId 638341380475357881; FireTimeUtc 29/10/2023 01:00:49 +00:00; NextFireTimeUtc 29/10/2023 00:01:00 +00:00; PreviousFireTimeUtc 29/10/2023 01:00:49 +00:00; RefireCount 0; ScheduledFireTimeUtc 29/10/2023 01:00:49 +00:00;
---------
[...]

Issue #2497, opened 2024-07-30
crossing DST start on a CalendarIntervalSchedule trigger results in inaccurate response to `GetFireTimeAfter()`


ewortzman:
**Describe the bug**

When creating a CalendarIntervalTrigger with an interval of 1 day and a start time during the skipped DST hour, the result of `GetFireTimeAfter()` when called with the day before DST will round the result to the moment of DST. I assume this is intended.  However, calling the method on that moment will return the same value, rather than the next day.

**Version used**

3.11

**To Reproduce**

[See here](https://github.com/ewortzman/quartz-blob-bug/blob/DST-bug/Test/Program.cs#L32-L34).  This loop will never terminate, as the value of `next` never advances past 3/30/2025 1:00:00.

**Expected behavior**

With `SkipDayIfHourDoesNotExist` set to false, the value should be rounded to either the moment before or the moment after DST when the trigger time would fall during the skipped hour, but after rounding once, subsequent calls should return the following day, rather than the same time over and over.

**Additional context**

ewortzman:
I believe the issue might be here: https://github.com/quartznet/quartznet/blob/v3.11.0/src/Quartz/Impl/Triggers/CalendarIntervalTriggerImpl.cs#L754.

When passed `3/30/2025 01:30+0000`, it returns `3/30/2025 01:30+0100`.  Obviously this is incorrect.  We've lost an hour somehow. It should return `3/30/2025 02:30+0100`. This method simply calls `System.TimeZoneInfo.ConvertTime`, so I'm not sure why it doesn't work properly.  But that's what I'm observing.

ewortzman:
written some [test code](https://github.com/ewortzman/quartz-blob-bug/blob/dst-test/Test/Program.cs). The trigger does, in fact, break upon crossing over DST.  It fires fine leading up to DST, then fires many times in quick succession as Quartz believes the nextFireTime always to be the same value, then stops when the threshold gets far enough into the future.

ewortzman:
Some additional testing shows that this problem only arises when both `PreserveHourOfDayAcrossDaylightSavings` is true _and_ `SkipDayIfHourDoesNotExist` is false.

In the case where PreserveHour is false, the job simply fires at 01:30 UTC every day, ignoring DST entirely

In the case where SkipDay is true, the job skips the relevant day then begins firing at the correct time on the subsequent day. 

It appears only to be the rounding logic that attempts to execute at a different time that breaks the trigger.

[...]

PR #2520, opened 2024-08-14
Fix for DST NextFireTime being stuck when issued on DST boundary

jafin:
Add `GetTimeAtOrAfter` to CronTriggerImpl, which avoids having move aftertime back 1 seconds, which may put it into different DST zone.
Use this when calculating `ComputeFirstFireTimeUtc` only,

CronExpression
- Add `GetTimeAt` which calls an overloaded `GetTimeAfter` with 0 second offset, again avoiding the back in time challenge.

Relates to #2475 

Needs review.
Need to ensure test case is sufficient.

### DIFF
--- a/src/Quartz.Tests.Unit/DaylightSavingTimeTests.cs
+++ b/src/Quartz.Tests.Unit/DaylightSavingTimeTests.cs
@@ -19,7 +19,13 @@
 
 using System;
 
+using FluentAssertions;
+using FluentAssertions.Execution;
+
 using NUnit.Framework;
+
+using Quartz.Spi;
+using Quartz.Util;
 
 using TimeZoneConverter;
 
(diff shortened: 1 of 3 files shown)


kkloet1 (changes requested):
@jafin Currently the testcase is what was already working before, needs to test starting in the first hour after switching from DST to ST.

I've tested with below 4 testcases and functionality seems to work well.
- start in DST, computeFireTimes with switch from DST to ST.
- start in first hour of ST.
- start in ST, computeFireTimes with switch from ST to DST.
- start in first hour of DST.

jafin:
@lahma up for review.

lahma:
Closing — the issue this PR targeted (#2497) has been verified as fixed on 3.x by #2837 (the spring-forward infinite loop guard). Both `SkipDayIfHourDoesNotExist` variants now advance correctly past the DST boundary.

The CronExpression DST fall-back work in this PR is a separate concern — if it's still needed, it should be split into a focused PR with the API scoping and cleanup suggestions from the review addressed.

Thanks for the investigation and test cases!

PR #2521, opened 2024-08-14
Fix error messaging to match error conditions in Quartz/CronExpression

KPDeweese:
While working with Quartznet I noticed that some of the error messaging did not seem to be consistent with the behavior I was observing, specifically I was receiving an error that my cron schedule for running every 24 hours was greater than the allowed increment of 24. I have created this pull request to bring the error messaging into alignment with the corresponding error conditions for both this piece of logic, and the similar piece restricting the maximum minute interval to 59.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -10,6 +10,7 @@
     <PackageVersion Include="FakeItEasy" Version="8.3.0" />
     <PackageVersion Include="FirebirdSql.Data.FirebirdClient" Version="10.3.1" />
     <PackageVersion Include="FluentAssertions" Version="6.12.0" />
+    <PackageVersion Include="FluentAssertions.Analyzers" Version="0.33.0" />
     <PackageVersion Include="log4net" Version="2.0.17" />
     <PackageVersion Include="MELT" Version="0.9.0" />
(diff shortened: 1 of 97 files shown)


lahma (approved):
Looks good, thank you!

lahma:
But tests are failing, please take a look 😅

KPDeweese:
Sorry, please disregard last update

jafin:
@KPDeweese some suggestions
- revert the change to static, as the lib won't even compile.
- Update CronExpressionTest and amend the tests to match the change in Expected result now. The tests are still expecting the old value and need to match your change.

```text
Assert.That(e.Message, Is.EqualTo("Increment > 24 : 120"))
  String lengths are both 20. Strings differ at index 13.
  Expected: "Increment > 24 : 120"
  But was:  "Increment > 23 : 120"
  ------------------------^
```
There are other tests that are failing, (you can see from the runner logs ie https://github.com/quartznet/quartznet/actions/runs/10404703998/job/28814535388?pr=2521) but similar to the above will be the fix, (adjusting test expectation)
Are you able to run the unit tests locally? As that would provide a simple way to validate the change and make sure all tests pass before pushing the PR ?

lahma:
Superseded by #2524  , ideally you should always create a new branch for pull requests, I didn't want to start pushing fixes to your `main` branch.

PR #2543, opened 2024-09-02
Adding a new functionality to determine the duration of the delay of a quartz cron job

Ghyath-Serhal:
### New Feature
---
The new feature that I am adding with this PR is to know the duration of the delay of a quartz cron job.

### Scenario
---

I am currently deploying a Quartz application on AKS, where a single node is handling 4 concurrent Quartz jobs by default.
Occasionally when there is too much load on the system, there are a number of misfired jobs that remain in the queue, waiting for an available thread to execute.

To optimize the application's performance and scale it effectively, I need to track the number of misfired jobs and their corresponding wait times.
For instance, if there are five misfired jobs with wait times exceeding 15 minutes, I would like to trigger the creation of an additional node to handle the load.

### Current Behavior:
---
Currently if a certain job gets misfired 5 times before being fired, the next_fire_time is updated 5 times, there is no possibility to know the initial fire time and how long the job is late.

### New Behavior:
---
By adding a new column initial_next_fire_time to the qrtz_triggers that will not be updated on misfire. It will be possible to track the delay of a certain cron job.

> [!NOTE]
> This feature is not available for jobs scheduled With Simple Schedule

### DIFF
--- a/database/schema_30_to_31_upgrade.sql
+++ b/database/schema_30_to_31_upgrade.sql
@@ -1,0 +1,5 @@
+--USE [database_name];
+--GO
+
+ALTER TABLE qrtz_triggers
+ADD initial_next_fire_time bigint NULL;
--- a/src/Quartz.Benchmark/SimpleTriggerImplBenchmark.cs
+++ b/src/Quartz.Benchmark/SimpleTriggerImplBenchmark.cs
@@ -326,6 +326,7 @@ public class SimpleTriggerImplBenchmark
         public const int RepeatIndefinitely = -1;
 
         private DateTimeOffset? nextFireTimeUtc; // Making a public property which called GetNextFireTime/SetNextFireTime would make the json attribute unnecessary
+        private DateTimeOffset? initialNextFireTimeUtc; // Making a public property which called GetNextFireTime/SetNextFireTime would make the json attribute unnecessary
         private DateTimeOffset? previousFireTimeUtc; // Making a public property which called GetPreviousFireTime/SetPreviousFireTime would make the json attribute unnecessary
 
         private int repeatCount;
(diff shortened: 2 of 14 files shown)


lahma:
Still not convinced about making database schema changes, especially for v3 line. How do you monitor and trigger the AKS node size increase?

Ghyath-Serhal:
Regarding AKS I will Kubernetes Event-driven Autoscaling (KEDA), it has a list of [scalers](https://keda.sh/docs/2.15/scalers/) that I can use to scale out/in nodes based on the number of waiting jobs.

Please inform me if the PR will not be approved in order to find a workaround without changing in the Quartz source code.

lahma:
So can you tell me what scaler you were planning to use and with what configuration? How do you query the information?

Ghyath-Serhal:
I am planning to use the [PostgreSQL](https://keda.sh/docs/2.15/scalers/postgresql/) scaler that takes a query. Based on the count returned from the query, AKS can scale out/in.

lahma:
I'm trying to think of alternative ways of achieving this. Here's one idea:

* implement `ITriggerListener.TriggerMisfired()` and register it to the scheduler
* producer/consumer or something similar not to block Quartz itself, so needs to be fast recording and flush outside of processing thread
* insert scheduler name, instance id, trigger key and misfire time into table
* query the table with grouping based on scheduler and `COUNT()` misfires within a time window
* implement cleanup logic to delete data from table which is older than a day or similar

Ghyath-Serhal:
I had considered this solution before but was keeping it as a backup option.

My thinking is that if the library can provide the duration of the delay, it would result in a less complex implementation. In addition, from a performance point of view, reduce the number of database calls.

Do you have any concerns about the PR or you just do not want to change the database schema?

lahma:
> My thinking is that if the library can provide the duration of the delay, it would result in a less complex implementation. In addition, from a performance point of view, reduce the number of database calls.

I still think that ideally Quartz would just output metrics and you can aggregate the data from monitoring tool.

> Do you have any concerns about the PR or you just do not want to change the database schema?

3.x line is in maintenance mode (LTS) and I don't think it's a good idea to start schema changes.

Ghyath-Serhal:
Ok noted.
Thank you for your time.
I will close the PR.

lahma:
@Ghyath-Serhal please note that I appreciate you bringing the problem to the table and showing a possible solution. I've been somewhat exhausted in all mandatory ordeals in life so I can't always commit time as much as people contributing should deserve. The auto-scaling and means to achieve is an important feature to tackle.

PR #2547, opened 2024-09-04
Updated the expression for CronTrigger Example 4 according to description

Shashidanand:
Changed the expression according to description: 
CronTrigger Example 4 - an expression to create a trigger that fires every half hour between the hours of 8 am and 10 am on the 5th and 20th of every month. Note that the trigger will NOT fire at 10:00 am, just at 8:00, 8:30, 9:00 and 9:30

### DIFF
--- a/docs/documentation/quartz-3.x/tutorial/crontriggers.md
+++ b/docs/documentation/quartz-3.x/tutorial/crontriggers.md
@@ -95,7 +95,7 @@ Here are a few more examples of expressions and their meanings - you can find ev
 Note that the trigger will NOT fire at 10:00 am, just at 8:00, 8:30, 9:00 and 9:30**
 
 ```text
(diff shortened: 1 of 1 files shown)


lahma:
Your change makes it fire on 10:00 and 10:30 which is incorrect

Shashidanand:
image

Please check the link provided on the documentation page [Cron Triggers](https://www.quartz-scheduler.net/documentation/quartz-3.x/tutorial/crontriggers.html#cron-expressions) in the TIP section. And please correct me if there is something wrong.

Shashidanand:
> > Note that the trigger will NOT fire at 10:00 am, just at 8:00, 8:30, 9:00 and 9:30

Got it. Description at FreeFormatter.com confused me greatly. Thank you.

PR #2578, opened 2024-10-03
OpenTracing: using IJobDiagnosticData instead of IJobExecutionContext

### LINKED ISSUE #2577: OpenTracing exception: Event-Exception: "Quartz.Job.Execute.Start"
**Describe the bug**

When I add OpenTracing support like this:
```
            services.AddQuartzOpenTracing(options =>
            {
                options.IncludeExceptionDetails = false;
                options.OperationNameResolver = context => $"job: {context.JobDetail.Key.Group}";
            });
```
service begins writing warning message to log on every Quartz Job start:

```
Event: Quartz.Job.Execute.Start
SourceContext : Quartz.OpenTracing.QuartzDiagnostic
exception_message: untypedArg
exception_type: ArgumentException
exception_stacktrace: 
   at Quartz.OpenTracing.QuartzDiagnostic.OnNext(String eventName, Object untypedArg)
   at Quartz.OpenTracing.QuartzDiagnostic.System.IObserver<System.Collections.Generic.KeyValuePair<System.String,System.Object>>.OnNext(KeyValuePair`2 value)
```

Seems like [this line](https://github.com/quartznet/quartznet/blob/3.x/src/Quartz.OpenTracing/QuartzDiagnostic.cs#L83) is producting it. And [here](https://github.com/quartznet/quartznet/blob/3.x/src/Quartz/Logging/JobDiagnosticsWriter.cs#L32) we see the "Start" event writer, that supplies event argument of type `JobDiagnosticData` that wraps `IJobExecutionContext` but not implements it. So seems that every job start event fails because it expects argument of wrong type.

**Version used**

3.13

**To Reproduce**

Register OpenTracing as described above, run a task and see log.

**Expected behavior**

No warning message must be produced, but SpanContext created and started instead.

justmara:
fixes #2577

### DIFF
--- a/src/Quartz.OpenTracing/QuartzDiagnostic.cs
+++ b/src/Quartz.OpenTracing/QuartzDiagnostic.cs
@@ -78,7 +78,7 @@ namespace Quartz.OpenTracing
         {
             if (eventName == jobExecuteStartEventName)
             {
-                var jobContext = untypedArg as IJobExecutionContext;
+                var jobContext = untypedArg as IJobDiagnosticData;
                 if (jobContext is null)
                     throw new ArgumentException(nameof(untypedArg));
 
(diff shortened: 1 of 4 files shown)


lahma:
Thank you!

PR #2606, opened 2024-10-23
Implemented the IReadOnlyDictionary<TKey, TValue?> interface in DirtyFlagMap

perringaiden:
To allow use of `System.Collections.Generic.CollectionExtensions.GetValueOrDefault` extensions, the `DirtyFlagMap` class needed to implement `IReadOnlyDictionary<TKey, TValue?>`, as the .NET Framework assumes (but does not enforce) every `IDictionary<>` also implementing the `IReadOnlyDictionary<>` interface, in the same way that `Dictionary<>` does.

These changes allow the use of `GetValueOrDefault` for `SchedulerContext` and `JobDataMap` to replace the need for 

```
string item = null;
var value = dirtyFlagMap.TryGetValue(key, out item) ? item : null;
```
with
```
var value = dirtyFlagMap.GetValueOrDefault(key)
```

(Note: There's a bunch of auto-formatting applied by Visual Studio that I can try to strip out if you feel it's necessary)

### DIFF
--- a/src/Quartz.Tests.Unit/Utils/DirtyFlagMapTest.cs
+++ b/src/Quartz.Tests.Unit/Utils/DirtyFlagMapTest.cs
@@ -2283,8 +2283,8 @@ public class DirtyFlagMapTest
         Assert.Multiple(() =>
         {
             Assert.That(keys.Count, Is.EqualTo(2));
-            Assert.That(keys, Contains.Value("One"));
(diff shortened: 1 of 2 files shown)


lahma (approved):
Looks good, thank you for implementing this. Would you like to also backport a PR targeting 3.x branch so we could get this to 3.x train too?

PR #2660, opened 2024-12-13
DirtyFlagMap-fixAdd

fbenoit-coexya:
Turns out MaybeNull is an output attribute, and it results in the compiler assuming an object is no longer guaranteed non-null after being added.
The actual input attribute is AllowNull.
- AllowNullAttribute: Specifies that null is allowed as an input even if the corresponding type disallows it.
- MaybeNullAttribute: Specifies that an output may be null even if the corresponding type disallows it.


fbenoit-coexya:
Welp, I think I've found why the output attribute was being used instead of the input one.

lahma:
Thank you.

PR #2665, opened 2024-12-31
Add a link to the Weasel.Quartz library

Hawxy:
I cooked up https://github.com/Hawxy/Weasel.Quartz a few months ago but forgot to add it here.


lahma:
Looks like a great library, thanks for the PR!

PR #2699, opened 2025-02-18
(v4) Fix some .NET 9 issues and ensure OTEL records exceptions.

Hawxy:


lahma:
Thank you, I appreciate the PR but I'd like to discuss the problems that are being fixed. So new dependency on `System.Diagnostics.DiagnosticSource` to get an API I guess? Also pragmas because some types conflict maybe? 

Would NET 9 compile without extra reference as the `System.Diagnostics.DiagnosticSource `comes out of the box?

Hawxy:
> So new dependency on System.Diagnostics.DiagnosticSource to get an API I guess?

Correct, `AddException` is needed so the correct exception tags get set on the trace, otherwise in many tracing platforms you end up with an error span but little metadata around what went wrong. It used to be part of the `OpenTelemetry` package but got added to core library in .NET 9.

> Would NET 9 compile without extra reference as the System.Diagnostics.DiagnosticSource comes out of the box?

Good point, has been fixed.

PR #2724, opened 2025-04-15
make sure maxNumOfCount is not less than 1

### LINKED ISSUE #2723: Edge case for every day trigger causes exception
I need to execute a job once daily at a certain hour and minute. Therefore from my understanding the correct way will be to do like this:

                .WithDailyTimeIntervalSchedule
                  (s => s.OnEveryDay().StartingDailyAt(
                      TimeOfDay.HourAndMinuteOfDay(hours, minutes))
                      .EndingDailyAfterCount(1)
                      .WithMisfireHandlingInstructionDoNothing())
                  .WithIdentity(_triggerKey)
                  .ForJob(_syncJobKey)
                .Build();

But if i select 23:59 I got an exception `The startTimeOfDay is too late with given Interval and IntervalUnit values.`. Can you please explain why? Is there a correct way to do this?
Overall I don't really like the code:

                DateTimeOffset maxEndTimeOfDayDate = TimeOfDay.HourMinuteAndSecondOfDay(23, 59, 59).GetTimeOfDayForDate(today);

When I see 23:59:59 that triggers me more than Quartz )

sherlock1982:
fixes #2723

### DIFF
--- a/src/Quartz/DailyTimeIntervalScheduleBuilder.cs
+++ b/src/Quartz/DailyTimeIntervalScheduleBuilder.cs
@@ -396,7 +396,7 @@ public sealed class DailyTimeIntervalScheduleBuilder : ScheduleBuilder<IDailyTim
             return this;
         }
 
-        if (remainingMillisInDay < intervalInMillis)
+        if (remainingMillisInDay <= intervalInMillis)
         {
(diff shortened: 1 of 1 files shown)


lahma:
Can you please target this against 3.x branch and add a test case?

Issue #1617, opened 2022-05-24
Factory-based AddQuartz() overload


Timovzl:
When we register Quartz.NET, we might do so as follows:

```cs
services.AddQuartz(q =>
{
	q.UsePersistentStore(s =>
	{
		s.UseMySql(mySql =>
		{
			mySql.ConnectionString = "But what if I need to be obtained from a service?";
		});
		// ...
	});
	// ...
```

The configurator lambda is run before the `IServiceProvider` is available to the application, and it does not grant access to it via a parameter. This can be inconvenient, as I will explain.

Sure, a connection string is often injected directly as a single config value, where all that we need to duplicate is _where_ that connection string is obtained (duplicated from where we normally register the database, for example when we register Entity Framework). However, we are currently in a situation where the devops engineers provide us with three database config values: host, username, and password. The application can create its own connection string from those, including any connection options that it deems necessary. It is a sensible setup, in my opinion.

To extract the appropriate config values, add the options, and produce the final connection string, we have a service, `DatabaseConnectionFactory`. _Anything_ requiring database connections (such as respositories) takes a dependency on this service. Quartz needs it as well.

The problem is that we cannot access services from the container in the `AddQuartz` configurator lambda.

**Proposed Solution**

With DI, we can normally do something like this:

```cs
services.AddSingleton<IWhatever>(serviceProvider => ... /* Factory to produce an instance, with access to the DI container */);
```

The same would solve our problem for `AddQuartz`. The overload I propose adding is as follows:

```cs
services.AddQuartz((quartz, serviceProvider) => ...);
```

**Additional Context**

I have also proposed the feature of [making Quartz independent of any specific database connector](https://github.com/quartznet/quartznet/issues/1616), having it rely on abstractions instead, e.g. `Func<DbConnection>`. The current feature makes sense with regards to that feature as well, since the factory specified there may well require a dependency.

siva538:
@Timovzl  - I have already requested this functionality as part - https://github.com/quartznet/quartznet/issues/1529

Please let us know if you have any progress on this, as I see that you are a contributor to this project.

Thanks a lot.

MaxLevs:
Sometimes I need IOptions<T> for configurating jobs. 
I'm looking forward to realease IServiceProvider feature.

lahma:
Implemented in PR #3007 (merged to 3.x).

New `AddQuartz` overloads provide `IServiceProvider` access during configuration:

```csharp
[...]

PR #2787, opened 2025-10-17
Add IServiceProvider overloads

bdovaz:
Partially fixes https://github.com/quartznet/quartznet/issues/1617

Add factories with `IServiceProvider` to all extension methods: AddJob, AddTrigger, ScheduleJob, and AddCalendar.

### DIFF
--- a/src/Quartz.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
+++ b/src/Quartz.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
@@ -104,11 +104,11 @@ public static class ServiceCollectionExtensions
 #if NET6_0_OR_GREATER
             [System.Diagnostics.CodeAnalysis.DynamicallyAccessedMembers(System.Diagnostics.CodeAnalysis.DynamicallyAccessedMemberTypes.PublicConstructors | System.Diagnostics.CodeAnalysis.DynamicallyAccessedMemberTypes.PublicMethods)]
 #endif
-        T>(
+    T>(
         this IServiceCollectionQuartzConfigurator options,
         Action<IJobConfigurator>? configure = null) where T : IJob
     {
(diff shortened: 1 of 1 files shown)


lahma:
I believe the ask was to have `services.AddQuartz((quartz, serviceProvider) => ...);` which is currently quite hard to achieve, how does this tackle this?

bdovaz:
As you say, it doesn't completely solve the problem, but it does partially solve it and enables many use cases that were not possible until now.

I first tried to do what was requested, but as you said it required a major refactoring and I found it very complex.

I'm sure users will at least appreciate having these overloads.

In my case it unlocks many uses cases. Now, for example, if I want to add cron, I can read the expression with IOptions from a JSON file or an env.

lahma (approved):
Looks good for this specific use case, thank you for your contribution!

lahma:
@bdovaz could you also create a PR against main the ensure these will be available there too?

bdovaz:
> @bdovaz could you also create a PR against main the ensure these will be available there too?

Done https://github.com/quartznet/quartznet/pull/2796

SvenIJK:
I was using `options.AddJob(jobInstance.GetType(), jobInstance.JobKey);`, but after update to 3.15.1 (from 3.15.0), I get the following error: The call is ambiguous between the following methods or properties: 'ServiceCollectionExtensions.AddJob(IServiceCollectionQuartzConfigurator, Type, JobKey?, Action<IJobConfigurator>?)' and 'ServiceCollectionExtensions.AddJob(IServiceCollectionQuartzConfigurator, Type, JobKey?, Action<IServiceProvider, IJobConfigurator>?)'
Which should be the default option to use?
- Giving an empty IJobConfigurator
- Giving an empty IServiceProvider AND empty IJobConfigurator?

Issue: https://github.com/quartznet/quartznet/issues/2795

PR #2788, opened 2025-10-23
Fix `StdScheduler.Interrupt` not interrupting all identified jobs

### LINKED ISSUE #757: Unexpected behavior of StdScheduler.Interrupt
### Problem
Having an instance of `StdScheduler`, I expect `Interrupt(jobKey)` call to interrupt all jobs with the same `JobKey`. However, the implementation of wrapped `QuartzScheduler` shows that it looks for the first cancellable job context with the specified `jobKey`.
```
public virtual async Task<bool> Interrupt(JobKey jobKey, CancellationToken cancellationToken = default)
{
    var cancellableJobs = CurrentlyExecutingJobs.OfType<ICancellableJobExecutionContext>();

    bool interrupted = false;

    foreach (var cancellableJobExecutionContext in cancellableJobs)
    {
        cancellationToken.ThrowIfCancellationRequested();

        var jobDetail = cancellableJobExecutionContext.JobDetail;
        if (jobKey.Equals(jobDetail.Key))
        {
            cancellableJobExecutionContext.Cancel();
            interrupted = true;
            break;
        }
    }

    if (interrupted)
    {
        await NotifySchedulerListeners(l => l.JobInterrupted(jobKey, cancellationToken), "job interruption").ConfigureAwait(false);
    }

    return interrupted;
}
```

It seems like the only way to actually interrupt same key jobs is individually by their `fireInstanceId`, using overloaded version of the method:
```
public Task<bool> Interrupt(string fireInstanceId, CancellationToken cancellationToken = default)
{
    var cancellableJobs = CurrentlyExecutingJobs.OfType<ICancellableJobExecutionContext>();

    bool interrupted = false;

    foreach (var cancellableJobExecutionContext in cancellableJobs)
    {
        cancellationToken.ThrowIfCancellationRequested();

        if (cancellableJobExecutionContext.FireInstanceId.Equals(fireInstanceId))
        {
            cancellableJobExecutionContext.Cancel();
            interrupted = true;
            break;
        }
    }

    return Task.FromResult(interrupted);
}
```

It's also interesting that `jobKey`-version of `Interrupt` notifies scheduler listeners, while `fireInstanceId`-version doesn't.

Is it all by design? If it is, doesn't it worth to point out in documentation is some way?

unageek:
Fixes #757.

### DIFF
--- a/src/Quartz/Core/QuartzScheduler.cs
+++ b/src/Quartz/Core/QuartzScheduler.cs
@@ -2038,7 +2038,6 @@ public sealed class QuartzScheduler
             {
                 cancellableJobExecutionContext.Cancel();
                 interrupted = true;
-                break;
             }
         }
 


lahma (approved):

PR #2796, opened 2025-11-13
Add IServiceProvider overloads for AddJob, AddTrigger, ScheduleJob, and AddCalendar

bdovaz:
Same as https://github.com/quartznet/quartznet/pull/2787 but for 4.x


bdovaz:
Tests fail, but they don't seem to be related to the tests I created

lahma (approved):
Thank you!

PR #2805, opened 2025-12-27
Docs: clarify exception re-execution behavior

### LINKED ISSUE #2303: Re-execute after throwing exception in docs
Hi, I'm reading the docs (https://www.quartz-scheduler.net/documentation/best-practices.html#throwing-exceptions) and it contains this:

> If a job throws an exception, Quartz will typically immediately re-execute it, meaning the job can and likely will throw the same exception again.

Is this actually true? I've been both testing this and reading the code, and it seems that when an exception occurs, the `JobRunShell` converts it to `JobExecutionException` with `refireImmediately: false`:

https://github.com/quartznet/quartznet/blob/fe5420a8b206fd4e3954e9921146d42a05c3f345/src/Quartz/Core/JobRunShell.cs#L222-L230

This then means that the trigger will be marked as completed and no re-execution will happen.

Is this indended behavior? If yes, I think the documentation should be fixed. What do you think?

Thanks.

meiranzheng:
Fixes #2303.

The best-practices page currently says Quartz will typically immediately re-execute a job when it throws, but by default unhandled exceptions are wrapped into JobExecutionException with refireImmediately: false.

This updates the wording to match the actual behavior and notes the explicit opt-in via JobExecutionException(refireImmediately: true).

### DIFF
--- a/docs/documentation/best-practices.md
+++ b/docs/documentation/best-practices.md
@@ -171,7 +171,10 @@ A better solution is to release the worker thread (exit the job) and allow other
 
 A Job's execute method should contain a try-catch block that handles all possible exceptions.
 
-If a job throws an exception, Quartz will typically immediately re-execute it, meaning the job can and likely will throw the same exception again. This can lead to wasted resources and, in the worst cases, unstable or crashed applications.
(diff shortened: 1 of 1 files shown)


lahma (approved):
Thank you!

Issue #2469, opened 2024-07-15
STJ serializer does not support objects in `JobDataMap`


OronDF343:
**Describe the bug**

The new `System.Text.Json` serializer does not support non-primitive non-IDictionary values inside `JobDataMap`. This is a regression compares to the Newtonsoft serializer, which used `TypeNameHandling.Auto`, and there is no easy way to override the new behavior (it is set  by new custom converters which are internal).

**Version used**

3.11.0

**To Reproduce**

Create a job with objects inside the job data, for example:

```
    public static IJobDetail CreateJob(string id, ICollection<MyDto> data)
    {
        var jobData = new JobDataMap
        {
            ["JobType"] = JobType.Inline,
            ["MyData"] = data
        };
        return JobBuilder.Create<MyJob>()
            .WithIdentity(id)
            .SetJobData(jobData)
            .RequestRecovery()
            .StoreDurably(false)
            .Build();
    }
```

Schedule the job with any trigger. When the trigger fires, the deserialization will fail with an exception similar to this:

```
Quartz.JsonSerializationException: Could not deserialize JSON: {"JobType":0,"MyData":[{...},{...}]}
 ---> Quartz.JsonSerializationException: Failed to parse JobDataMap from json
 ---> System.Text.Json.JsonException: Unsupported value kind: Array
   at Quartz.Util.Utf8JsonWriterExtensions.GetJobDataMap(JsonElement jsonElement, JsonSerializerOptions options)
   at Quartz.Converters.DictionaryConverter.Read(Utf8JsonReader& reader, Type typeToConvert, JsonSerializerOptions options)
   --- End of inner exception stack trace ---
   at Quartz.Converters.DictionaryConverter.Read(Utf8JsonReader& reader, Type typeToConvert, JsonSerializerOptions options)
   at System.Text.Json.Serialization.JsonConverter`1.TryRead(Utf8JsonReader& reader, Type typeToConvert, JsonSerializerOptions options, ReadStack& state, T& value, Boolean& isPopulatedValue)
   at System.Text.Json.Serialization.JsonConverter`1.ReadCore(Utf8JsonReader& reader, JsonSerializerOptions options, ReadStack& state)
   at System.Text.Json.JsonSerializer.ReadFromSpan[TValue](ReadOnlySpan`1 utf8Json, JsonTypeInfo`1 jsonTypeInfo, Nullable`1 actualByteCount)
[...]

PR #2806, opened 2026-01-03
Improve object deserialization in `JobDataMap`

ricardoboss:
This PR adds special handling when deserializing an object in `JobDataMap`.

More specifically, it checks if the `JsonSerializerOptions` contain a converter that handles deserializing of the current data key.

If no converter has been found, it falls back to the default behaviour, which is trying to deserialize it as `Dictionary<string, string>`.

This change allows developers to register custom converters for complex objects in the options while staying backward compatible.

Related issue: #2469

### DIFF
--- a/src/Quartz.Serialization.SystemTextJson/Util/ICustomJobDataConverter.cs
+++ b/src/Quartz.Serialization.SystemTextJson/Util/ICustomJobDataConverter.cs
@@ -1,0 +1,15 @@
+namespace Quartz.Util;
+
+/// <summary>
+/// Can be used to convert more complex objects in <see cref="JobDataMap"/>s.
+/// </summary>
+public interface ICustomJobDataConverter
+{
+    /// <summary>
+    /// Whether this converter handles properties with the given name.
+    /// </summary>
+    /// <param name="propertyName">The job data property name to check</param>
+    /// <returns><see langword="true"/> if this converter handles the given <paramref name="propertyName"/>,
+    /// <see langword="false"/> if not.</returns>
+    bool HandlesProperty(string propertyName);
+}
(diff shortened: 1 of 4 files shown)


lahma:
I'm not sure about this approach. Relying on random text strings doesn't feel right. With a large installation keys could be repeated and mean different things. This mostly signifies why it's so hard to have key-value pairs with value being an object with type information missing.

ricardoboss:
Yeah I can definitely see what you mean. I agree it's not the best approach. I have another idea: since the type information is missing (and Quartz can put all kinds of constraints on the converter), we could wrap the object-to-be-serialized with a Quartz-provided type that contains exactly this type information. Sure, it would make the job data a bit bigger, but we could preserve type and even converter types (as in, the exact converter that was used to serialize).

What do you think about that approach?

lahma:
Well I think this would approach just having a custom serializer for specific project use case. I don't want depress here but I don't see a easy "fits 80% of use cases" approach here. Is your use case to use in multiple projects/setups or what's the original need? Just tinkering whether creating a custom one would be the easiest and not trying to create fits 'em all solution.

ricardoboss:
My original use case is: send an e-mail. The job data is a serialized version of the mail object (headers + body). My workaround is using a helper method to write/read all the data to/from a job data. I can't simply register a custom converter in the STJ resolver chain, because it will never be used when deserializing. I _could_ create my own serializer for all job data, but this mail job is just one simple instance and most other jobs probably don't need this level of custom data serialization.

That's why I came to the suggested solution in this PR. If the key matches, I know this will be a mail object, which will use that specific converter. Otherwise everything is handled as always.

lahma:
Some things that came to my mind:

* a value of JSON string should work, you could deserialize on job level
* you could go with customization like there's for .NET JSON configuration, where there's paths like `Email.Subject` which maps to options object like `class Email { string Subject { get;set; } }`

I'm just spitballing some  alternatives

ricardoboss:
The second option is the workaround I described. My job data has `Mail.Html`, `Mail.Headers.To`, `Mail.Headers.Bcc`, ...

I also considered  the first approach, but I don't really want my job to have to do the serialization (just a preference, it could work of course).

ricardoboss:
Is there any scenario where you see this getting merged? Otherwise I'll just stick with the workaround (or eventually write a custom serializer).

PR #2874, opened 2026-03-08
Refactor Params to const in StdAdoDelegate

jafin:
Resolves `SonarCloud: Define a constant instead of using this literal 'triggerName' 21 times.`

### DIFF
--- a/src/Quartz/Impl/AdoJobStore/StdAdoDelegate.Triggers.cs
+++ b/src/Quartz/Impl/AdoJobStore/StdAdoDelegate.Triggers.cs
@@ -16,6 +16,48 @@ namespace Quartz.Impl.AdoJobStore;
 
 public partial class StdAdoDelegate
 {
+    // Parameter name constants for trigger-related queries
+    private const string ParamBlob = "blob";
+    private const string ParamCalendarName = "calendarName";
+    private const string ParamFireTime = "fireTime";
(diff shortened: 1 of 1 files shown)


lahma:
@jafin — thank you for the patch, and sorry for the slow reply.

The finding is real: on `3.x`, `StdAdoDelegate.Triggers.cs` still spells `"triggerName"` 32 times,
`"triggerGroup"` 42 and `"schedulerName"` 73. I am not taking it there, though.

Since your base commit, 13 commits have landed on that one file — #3202, #3013, #3004 and the rest of
the 3.x fix stream — and it has grown by 720 lines. Rebasing means re-running the substitution across
all of it and re-reviewing ~240 lines that change no behaviour. The lasting cost is bigger: `3.x` is
the maintenance branch and fixes move between it and `main` constantly, so a whole-file rename makes
every future backport in this file conflict on lines that do not semantically differ. That is a tax on
every later bug fix, for a static-analysis count.

Where the idea survives: on `main`, `AdoConstants` names columns but nothing names a parameter, and
`StdAdoConstants` spells `@triggerName` inline 40 times. A constant on the C# side alone still has to
agree with the SQL text by hand. Interpolate one constant into both, so the two spellings cannot
drift — that is worth doing, and I would review it happily.

Closing as not planned.

Issue #2697, opened 2025-02-14
In a cluster jobs run simultaneously even if they are marked with DisallowConcurrentExecution


pigiax:
**Describe the bug**
When running Quartz.NET in clustered mode, multiple instances of the same job class can execute concurrently on different nodes, even if they are configured with different parameters. Additionally, there is no built-in mechanism to ensure that only one instance of a specific job type is running at a given time across the cluster.

**Version used**
Quartz.NET 3.6.0

**To Reproduce**
1. Set up a Quartz.NET cluster with multiple nodes.
2. Define a job class that should not run concurrently.
3. Schedule multiple triggers with different parameters for the same job class.
4. Observe that multiple instances of the job class are executing simultaneously on different nodes.

Example configuration:
```csharp
var schedulerFactory = new StdSchedulerFactory();
var scheduler = await schedulerFactory.GetScheduler();
await scheduler.Start();

var jobDetail = JobBuilder.Create<MyJob>()
    .WithIdentity("myJob", "group1")
    .Build();

var trigger1 = TriggerBuilder.Create()
    .WithIdentity("trigger1", "group1")
    .StartNow()
    .WithSimpleSchedule(x => x.WithIntervalInSeconds(10).RepeatForever())
    .UsingJobData("param", "value1")
    .Build();

var trigger2 = TriggerBuilder.Create()
    .WithIdentity("trigger2", "group1")
    .StartNow()
    .WithSimpleSchedule(x => x.WithIntervalInSeconds(10).RepeatForever())
    .UsingJobData("param", "value2")
    .Build();

await scheduler.ScheduleJob(jobDetail, new List<ITrigger> { trigger1, trigger2 }, replace: true);
```

**Expected behavior**
- Only one instance of a job should run at a time across the cluster, regardless of different trigger parameters.
- A mechanism should be available to enforce a single execution of a job type at any given time.

**Additional context**
- The current workaround is to implement a custom locking mechanism in the job execution logic.
- It would be useful if Quartz.NET provided built-in support for job-level concurrency control in clustered mode.

lahma:
Seems like some guesswork ahead.. you do have `DisallowConcurrentExecution` attribute in place?

pigiax:
> Seems like some guesswork ahead.. you do have `DisallowConcurrentExecution` attribute in place?

Yes, already in place on all jobs.

But looking at the code, seems clear the issue:

[...]

PR #2914, opened 2026-03-28
Fix false cluster recovery causing DisallowConcurrentExecution violation

### LINKED ISSUE #2817: Jobs executed multiple time in a clustered environment
I’ve been running a Quartz clustering environment in Kubernetes for about three years without issues.
Last Friday, I encountered unexpected behavior.

I have a job **JOB1** scheduled to run every 5 minutes.

The cluster consists of two pods:
- **Pod1** was executing **JOB1**, which took about 1 hour to complete.
- **Pod2** was handling other jobs.

During execution:
- **Pod2** crashed due to an unhandled exception and restarted.
- After the restart, **JOB1** continued running normally on **Pod1**.
- However, Quartz began triggering **JOB1** again every 5 minutes on **Pod1** and **Pod2**, even though the original **JOB1** long‑running instance was still active on **Pod1**.

Key observation:
- The new instances of **JOB1** did not overlap with each other.
- They all overlapped only with the original long‑running execution on **Pod1**.

Note: The **RequestRecovery** option is not enabled.
It appears that after the crash, Quartz no longer recognized the original long‑running **JOB1** instance on **Pod1** as active, which led to new triggers being scheduled despite the job still running.

Actually I have too much load on the database, I am suspecting that the pod is not able to check-in in the table qrtz_scheduler_state, so that other pods are considering this pod is failed. Currently my checkin_interval is 7500, I will increase it.

lahma:
## Summary

Fixes #2817 — When a node misses its check-in (e.g., due to high DB load), other nodes declare it "failed" and `ClusterRecover` immediately deletes the EXECUTING `QRTZ_FIRED_TRIGGERS` record and unblocks sibling triggers. The "dead" node is still alive and running the job, but now other nodes can fire the same `[DisallowConcurrentExecution]` job concurrently.

The `IsJobCurrentlyExecuting` check from #2697/#2911 doesn't help here because recovery already deleted the EXECUTING record.

### Fix

**Preserve EXECUTING records for DisallowConcurrentExecution jobs during recovery of timed-out (non-orphan) instances.** The node may still be alive.

1. **`ClusterRecover`**: For EXECUTING + `DisallowConcurrentExecution` records from timed-out instances, skip the unblock and preserve the fired trigger record. Use selective deletion instead of bulk `DeleteFiredTriggers`. Orphaned instances (no scheduler state record — truly dead) get full cleanup as before.

2. **`FindFailedInstances`**: Run `FindOrphanedFailedInstances` on **every check-in** (not just `firstCheckIn`). This ensures preserved records from truly dead nodes are detected and cleaned up within one check-in cycle (~7.5s).

### How the scenarios play out

| Scenario | What happens |
|----------|-------------|
| **Node was falsely declared dead (comes back)** | EXECUTING record preserved → `IsJobCurrentlyExecuting` prevents new triggers → job completes normally → `TriggeredJobComplete` cleans up |
| **Node truly died** | EXECUTING record preserved → next check-in detects orphan (instance not in scheduler_state) → `ClusterRecover` runs for orphan → full cleanup (unblock + delete) |

### Changes

Single file: `JobStoreSupport.cs` — 48 insertions, 6 deletions. No new SQL, interfaces, or schema changes.

## Test plan

- [x] Build: 0 warnings, 0 errors
- [x] Full unit test suite: 1316 passed (1 pre-existing cron failure)

### DIFF
--- a/src/Quartz.Tests.Unit/Impl/AdoJobStore/JobStoreSupportTest.cs
+++ b/src/Quartz.Tests.Unit/Impl/AdoJobStore/JobStoreSupportTest.cs
@@ -6,6 +6,7 @@ using FakeItEasy;
 using FluentAssertions;
 
 using Quartz.Impl.AdoJobStore;
+using Quartz.Spi;
 
 namespace Quartz.Tests.Unit.Impl.AdoJobStore;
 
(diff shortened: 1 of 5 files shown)


lahma:
Closing — will retarget to 3.x branch.

PR #3044, opened 2026-04-15
Dashboard docs: add note about UseStaticFiles

Foxite:
I have found that when running a non-Blazor API project, turning on `<RequiresAspNetWebAssets>` is not enough to make the dashboard work properly. You also have to ensure that UseStaticFiles is called when configuring the request pipeline, otherwise blazor.web.js will not be served.


lahma:
Thank you!

PR #3141, opened 2026-07-17
Fix "Calender" misspelling of "Calendar" in error messages and docs

net0well:
Fix "Calender" misspelling of "Calendar" in error messages and docs

 - Correct "Calender" -> "Calendar" in the JobPersistenceException thrown by
   RAMJobStore when a calendar is removed while still referenced by a trigger
 - Correct the same misspelling in the equivalent AdoJobStore
   (JobStoreSupport.RemoveCalendar) message so both job stores agree
 - Fix the grammar of both messages: "if it referenced" -> "if it is referenced"
 - Fix the "CalenderIntervalTrigger" -> "CalendarIntervalTrigger" reference in
   docs/documentation/best-practices.md (the real type is CalendarIntervalTrigger)
 - Message/text only: no behavioral change, no public API change, and no tests
   assert on these strings

### DIFF
--- a/docs/documentation/best-practices.md
+++ b/docs/documentation/best-practices.md
@@ -173,7 +173,7 @@ As an example of possible issues, scheduling in the United States within TimeZon
 
 Again, specifics of time and amount of adjustment varies by locale.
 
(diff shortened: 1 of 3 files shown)


lahma (approved):
Thank you!

PR #3142, opened 2026-07-17
Add unit tests for TimeOfDay

net0well:
- TimeOfDay had no dedicated test fixture and was only covered indirectly via
   the DailyTimeIntervalTrigger and schedule-builder tests, leaving its own
   contract (validation, comparison, equality, date projection) untested
 - Add Quartz.Tests.Unit/TimeOfDayTest.cs covering:
   - both constructors and the HourAndMinuteOfDay/HourMinuteAndSecondOfDay
     factories, including the second-defaults-to-zero behavior
   - argument validation of hour/minute/second (valid boundaries and out-of-range)
   - Before() chronological comparison across hour, minute, second and equal times
   - Equals()/GetHashCode() for equal values, differing components, null and a
     non-TimeOfDay argument
   - GetTimeOfDayForDate() applying the time and zeroing milliseconds, and
     returning null for a null date
   - ToString() formatting
 - 30 test cases, all passing; no production code changed

### DIFF
--- a/src/Quartz.Tests.Unit/TimeOfDayTest.cs
+++ b/src/Quartz.Tests.Unit/TimeOfDayTest.cs
@@ -1,0 +1,185 @@
+namespace Quartz.Tests.Unit;
+
+public class TimeOfDayTest
+{
+    [Test]
+    public void Constructor_WithHourMinuteAndSecond_SetsAllComponents()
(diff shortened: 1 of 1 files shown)


lahma (approved):
Thank you!

Issue #2238, opened 2024-01-09
[Proposal] Change IDriverDelegate to Filter Triggers for Acquisition Based on Job Type


da3dsoul:
I am trying to build custom logic for acquiring the next jobs. I have various Job Types that do things like call external APIs, need limited (but not no) concurrency, need a database to be accessible (but not all jobs), and other conditions.
Currently, the solution is to just get more triggers, filter them, get more, filter.... It's not efficient.

My proposal is to add a filter to exclude a set of JobTypes/JobClasses to IDriverDelegate.SelectTriggerToAcquire().
```cs
Task<IReadOnlyCollection<TriggerAcquireResult>> SelectTriggerToAcquire(
            ConnectionAndTransactionHolder conn, 
            DateTimeOffset noLaterThan, 
            DateTimeOffset noEarlierThan, 
            int maxCount,
            Type[] jobTypesToExclude = null,
            CancellationToken cancellationToken = default);
```

It would also change StdAdoDriverDelegate.
Not the most efficient implementation, but it gets the point across.
```cs
    protected virtual string GetSelectNextTriggerToAcquireSql(int maxCount, Type[] jobTypesToExclude)
    {
        return $@"SELECT
                t.{ColumnTriggerName}, t.{ColumnTriggerGroup}, jd.{ColumnJobClass}
              FROM
                {TablePrefixSubst}{TableTriggers} t
              JOIN
                {TablePrefixSubst}{TableJobDetails} jd ON (jd.{ColumnSchedulerName} = t.{ColumnSchedulerName} AND  jd.{ColumnJobGroup} = t.{ColumnJobGroup} AND jd.{ColumnJobName} = t.{ColumnJobName}) 
              WHERE
                t.{ColumnSchedulerName} = @schedulerName AND {ColumnTriggerState} = @state AND {ColumnNextFireTime} <= @noLaterThan AND ({ColumnMifireInstruction} = -1 OR ({ColumnMifireInstruction} <> -1 AND {ColumnNextFireTime} >= @noEarlierThan))
                {(jobTypesToExclude == null || jobTypesToExclude.Length == 0 ? "" : $"AND jd.{ColumnJobClass} NOT IN ({string.Join(",", jobTypesToExclude.Select(a => $"'{GetStorableJobTypeName(a)}'"))})")} 
              ORDER BY 
                {ColumnNextFireTime} ASC, {ColumnPriority} DESC;";
    }
```

[...]

PR #3197, opened 2026-07-28
Replace job store listings with paged query objects

### LINKED ISSUE #206: Fetch a list of IJobDetails
Please consider implementing a better way to fetch a list of IJobDetails.

Currently, I have to use something along the lines of:

```
List<IJobDetail> jobs = new List<IJobDetail>();

scheduler.GetJobKeys(GroupMatcher<JobKey>.AnyGroup()).ToList().ForEach(delegate(JobKey key){

    jobs.Add(scheduler.GetJobDetail(key));

});
```

Now imaging a system with a couple thousand jobs in the database, it would make x thousand round trips to the scheduler service and to the db...

We should have an api to easily fetch jobinfo objects from a scheduler like:

```
List<IJobDetail> jobs = scheduler.GetJobs(GroupMatcher matcher).Take(10).Skip(10);

//Even could be in the already existing method with an overload taking a groupmatcher object:
List<IJobDetail> jobs = scheduler.GetJobDetail(GroupMatcher matcher);

```

take and skip would be nice to fetch only a subset of the jobdetail objects. 

This would fetch the data in 1 single call.

### LINKED ISSUE #1273: Paging support for IScheduler
Current interfaces to fetch jobs and triggers don't support paging. But in case if there are many jobs or triggers the methods take some time to return data. Also it's not feasible to use them for HTTP API for the same reason. 

`IScheduler` should be extended to provide methods for fetching paged data. 

I created #1272 as a draft of how the API should look like. I considered extending existing methods instead of creating overloads but in order to maintain Application Binary Interface compatibility overloads should be used. That's due to the way how default values for methods are used.

lahma:
## Summary

Replaces the job store listing surface with query objects that carry filtering, paging and projection — the redesign asked for in #1273 and sketched in its thread (headers instead of full triggers, bulk fetch by keys, paging flowing through stores and delegates).

- **`PagedQuery`** base (`Skip`, `Take`, opt-in `IncludeTotalCount`) with `JobQuery`, `TriggerQuery` (group/job/calendar/state filters), `JobGroupQuery`/`TriggerGroupQuery`, `CalendarQuery`; results as **`PagedResult<T>(Items, HasMore, TotalCount?)`** — `HasMore` is exact via a read-one-past fetch, `TotalCount` opt-in. Results are always ordered by (group, name).
- **`JobHeader`/`TriggerHeader` projections**: listings never load `JobDataMap`, and the trigger header carries `State` and `ExecutionGroup`, ending the per-trigger round trips.
- **Bulk fetch**: `GetJobDetails(keys)` / `GetTriggers(keys)` retrieve full objects in chunked single round trips (#206).
- **`IScheduler`** −8/+7 members and **`IJobStore`** −12/+7; `SchedulerQueryExtensions` keeps the familiar call shapes compiling as extensions over the query core. Counts are `Take = 0, IncludeTotalCount = true` — the three caller-less `GetNumberOf*` members are gone.
- **`IDriverDelegate` 91 → 66 members**: six dead members deleted (four had latent parameter-name bugs proving they never ran); the fired-trigger select/delete family collapses 7 → 2 onto `FiredTriggerQuery`; acquisition collapses onto the `TriggerAcquisitionCriteria` record — the additive extension point #2238 asks for; paging is modeled per dialect through an `ApplyPaging` seam that generates ORDER BY and limit together (the #413 lesson). `ITriggerPersistenceDelegate` gains a batch loader as a default interface method, so third-party delegates keep working.
- **HTTP API**: listing endpoints are paged behind one wire shape per resource (`{ items, hasMore, totalCount }`) plus `POST .../fetch` bulk routes; **Dashboard** drops every catalogued N+1 (#3008) and its error tile becomes exact instead of silently capped at 25.
- **Schema**: `(SCHED_NAME, <group>, <name>)` indexes in all dialect scripts and `schema_30_to_40_upgrade.sql`; the Postgres script''s index definitions gain their long-missing `SCHED_NAME` columns.

Incidental fixes found along the way: `ResumeAll` deleted its paused-groups sentinel via `LIKE` with `_` acting as wildcards; cluster recovery triggers always got priority 0 (the instance fired-trigger reader never read `PRIORITY`); group matchers now escape LIKE wildcards (escape char `!` — `ESCAPE '\'` is a MySQL syntax error) and use equality SQL for equality matchers everywhere.

Closes #1273, closes #206, closes #3008. Groundwork for #2238; advances the #988 "document database friendly" roadmap item.

## Verification

- Build clean under `TreatWarningsAsErrors`; unit suite 1725 passed / 0 failed; ASP.NET Core suite 197 passed / 0 failed; public API baselines reviewed and accepted; changelog + migration guide + AGENTS.md updated.
- New tests: 30 RAM store query tests, 14 matcher-translation/escaping SQL tests, 13 paged endpoint tests, dashboard paging tests, and a six-database `AdoJobStorePagingTest` integration fixture (needs the Docker CI lane; includes a `pg_B` group guarding the LIKE escaping per dialect).

## Notes for review

- `GetJobKeys(null)` now throws `ArgumentNullException` instead of silently defaulting to the DEFAULT group (extensions behavior).
- ADO ordering follows the server collation rather than ordinal — documented as an inherent store difference.
- The dashboard search box became a server-side group filter; a `NameContains` query filter is a candidate follow-up if name search should return.
- `/jobs|triggers/groups/{group}/paused` routes were kept as a convenience exception to one-shape-per-resource; trivially deletable if strictness is preferred.

https://claude.ai/code/session_015eMEzyDdXgvgxBSuZBJ6mT


lahma:
Follow-up commit 322fa2dc: audited all dialect index scripts for leftmost-prefix overlap and dropped 23 redundant index creations (IDX_QRTZ_J_GRP, IDX_QRTZ_T_G, IDX_QRTZ_T_STATE, IDX_QRTZ_T_NFT_MISFIRE, IDX_QRTZ_FT_TRIG_INST_NAME per dialect where present, plus a MySQL BLOB_TRIGGERS inline index duplicating its own PK). MySQL's FORCE INDEX targets are untouched. The 3.x upgrade script drops the same names with coverer-exists guards, and warns Postgres upgraders that their 3.x idx_qrtz_t_nft_st has reversed columns and needs rebuilding first. Left alone as judgment calls rather than provable redundancy: IDX_QRTZ_T_N_STATE (covering-read value) and Postgres's five single-column fired-trigger indexes that lack sched_name.

lahma:
Commit 0f2855b9 realigns the Postgres index set against the actual query shapes: the five sched_name-less single-column fired-trigger indexes are replaced by the composite set the queries use (ft_inst_job_req_rcvry, ft_j_g, ft_t_g rename), and qrtz_triggers gains t_j and t_c which Postgres alone lacked (those paths were full scans). The MySQL-style misfire indexes were deliberately NOT copied: their misfire_instr column only appears as '<> -1', which btree cannot use as a scan boundary - they exist elsewhere for FORCE INDEX, and the Postgres planner does better with idx_qrtz_t_nft_st, whose ordering also feeds the recovery LIMIT. Upgrade script gained a Postgres realignment section (creates before drops, reasons inline). Flagged for a possible future cross-dialect pass, not acted on here: IDX_QRTZ_J_REQ_RECOVERY and IDX_QRTZ_T_NEXT_FIRE_TIME appear dead under 4.x query shapes in every dialect, and SQL Server still collapses the job/trigger index pairs differently (T_G_J/FT_G_J/FT_G_T) from the other scripts.

lahma:
Commit 23911987 completes the cross-dialect index pass: every dialect now ships the same nine indexes (names and columns identical), with one documented exception (Postgres/SQLite omit T_NFT_ST_MISFIRE - 'MISFIRE_INSTR <> -1' gives a btree no scan boundary; it exists elsewhere as MySQL's FORCE INDEX target). Verified-dead and dropped everywhere: J_REQ_RECOVERY, T_NEXT_FIRE_TIME, the group-only key twins (T_JG/FT_JG/FT_TG), T_NFT_ST_MISFIRE_GRP, and T_N_STATE/T_N_G_STATE (state is the most-updated column; those indexes paid a key-changing write per transition to save one logical read). SQL Server's group-first key indexes renamed to the canonical name-first forms, and SQLite backfilled from 2 to the full set (its acquire poll was scanning the trigger table). Upgrade script documents each drop with its reason and coverer guard. Also completed: a 3.x backport analysis (recomputed against 3.x's own queries) - recommendations P1-P5 with the Postgres index realignment as the standout, plus two trivial bug fixes (ClusterRecover priority, four latent parameter-name bugs); notably IDX_QRTZ_J_GRP must be KEPT on 3.x since its 4.x coverer doesn't exist there.

PR #3203, opened 2026-07-28
Realign Postgres indexes and drop prefix-redundant indexes across dialects

lahma:
Backports the index findings from the 4.x JobStoreQueries campaign (PR #3197), each recomputed against 3.x's own query shapes rather than copied — the verdicts differ from 4.x in several places, deliberately.

## PostgreSQL realignment (the high-value half)

The 3.x Postgres script had 9 of 11 indexes not leading with `sched_name` (which every Quartz statement filters first), and `idx_qrtz_t_nft_st` with its columns **reversed** — `(next_fire_time, trigger_state)` against an acquire query that is two equalities then a range. That index could never bound the scan on state or scheduler name; this is the shape behind slow-acquire-on-Postgres reports. The new set leads everything with `sched_name`, reorders `t_nft_st` to `(sched_name, trigger_state, next_fire_time)`, adds the missing composites (`t_j`, `t_c`, `t_g_n`, `ft_inst_job_req_rcvry`, `ft_j_g`), renames `ft_trig_nm_gp` to the 4.x `ft_t_g`, and drops `t_state` plus the six single-column fired-trigger indexes that could never lead a scan. Names/shapes match 4.x so the eventual 3.x-to-4.x migration is a no-op. No misfire index on Postgres: `MISFIRE_INSTR <> -1` gives a btree no scan boundary, and unlike MySQL there is no FORCE INDEX to satisfy.

**Verified on live PostgreSQL 16**: the upgrade script (old schema + `schema_30_postgres_index_realignment.sql`) and a fresh install converge on byte-identical `pg_indexes` sets, and a 160k-trigger acquire-plan comparison moves all three predicates from post-filter into the index condition (buffer hits 136 to 57, discarded rows 61 to 0).

## Cross-dialect prefix drops

An index whose columns are a leftmost prefix, in order, of a wider same-table index serves no query the wider one cannot. Dropped where present, coverer verified per script: `IDX_QRTZ_T_STATE` (covered by `T_NFT_ST` — every dialect), `IDX_QRTZ_T_G` (by `T_N_G_STATE`), `IDX_QRTZ_T_NFT_MISFIRE` (by `T_NFT_ST_MISFIRE`), `IDX_QRTZ_FT_TRIG_INST_NAME` (by `FT_INST_JOB_REQ_RCVRY`), plus MySQL''s unnamed `BLOB_TRIGGERS` inline index that duplicated its own primary key (create-script only; the PK satisfies the FK — verified on live MySQL 8, where both FORCE INDEX hints still resolve).

**Deliberate differences from 4.x**: `IDX_QRTZ_J_GRP` stays on 3.x (its 4.x coverer does not exist here and it serves the group listings), as do `j_req_recovery` and `t_next_fire_time` (kept for cross-dialect parity; the script comments note honestly that no live 3.x query drives them). SQLite stays index-free as it ships today.

## Upgrade scripts

- `database/schema_30_postgres_index_realignment.sql` — live SQL, creates first; the three indexes that keep their names while changing columns (`j_req_recovery`, `t_next_fire_time`, `t_nft_st`) are drop+recreate, because `CREATE INDEX IF NOT EXISTS` would silently keep the wrong shapes on existing databases; `CONCURRENTLY` note in the header.
- `database/schema_30_drop_redundant_indexes.sql` — guarded SQL Server statements (pre- and post-2016), commented MySQL/Oracle/Firebird sections, the coverer named on every drop, and a do-not-drop warning beside the MySQL FORCE INDEX targets.

A programmatic re-audit of the final state across all eight scripts reports zero leftmost-prefix redundancies among named indexes, and no reference to the removed names exists outside `database/`.

https://claude.ai/code/session_015eMEzyDdXgvgxBSuZBJ6mT


lahma:
Follow-up from review discussion: the 'keep SQLite index-free' verdict was inherited conservatism, not a real trade-off - the query shapes are identical and SQLiteDelegate is a shipped production dialect, so an index-free schema meant full trigger-table scans (blobs included) on every acquire poll. SQLite now gets the same ten-index set as the realigned Postgres script (sched_name-first, 4.x-aligned names, no misfire index for the same btree '<>' reason) plus schema_30_sqlite_indexes.sql for existing databases. Verified via sqlite3: fresh install and old-db+upgrade converge on identical index sets; the upgrade script is idempotent.

PR #3204, opened 2026-07-28
Let AdoJobStore take part in the application's transaction (#2038)

### LINKED ISSUE #2038: Support for TransactionScope to include the scheduling in an existing transaction
I want to schedule jobs within a transaction using the `TransactionScope`:

```c#
// Not working: Making changes and add job in same transaction
public async Task ExampleMethod()
{
    using (var transaction = new TransactionScope(TransactionScopeAsyncFlowOption.Enabled))
    {
        // Make changes to entities
        var entity = _dbContext.ExampleEntities.Add(new ExampleEntity() { Name = new Random().Next(1, 100).ToString() });
        await _dbContext.SaveChangesAsync();

        // Add job for entity
            .WithIdentity("exampleJob")
            .UsingJobData("entity-name", entity.Entity.Name)
            .Build();
            .WithIdentity("exampleTrigger")
            .StartNow()
            .Build();

        transaction.Complete();
    }
}
```

This gives the following exception:
```
System.InvalidOperationException: A transaction is already in progress; nested/concurrent transactions aren't supported.
   at Npgsql.NpgsqlConnection.BeginTransaction(IsolationLevel level, Boolean async, CancellationToken cancellationToken)
   at Npgsql.NpgsqlConnection.BeginTransaction(IsolationLevel level)
   at Npgsql.NpgsqlConnection.BeginDbTransaction(IsolationLevel isolationLevel)
   at Quartz.Impl.AdoJobStore.JobStoreSupport.GetConnection()
```

It can be fixed by suppressing the `TransactionScope` or removing it, but this is not the wanted behavior because everything including the scheduling schould be done in the transaction:
```c#
// This is working - but not same transaction
public async Task ExampleMethodWithWorkaround()
{
    using (var transaction = new TransactionScope(TransactionScopeAsyncFlowOption.Enabled))
    {
        // Make changes to entities
        var entity = _dbContext.ExampleEntities.Add(new ExampleEntity() { Name = new Random().Next(1, 100).ToString() });
        await _dbContext.SaveChangesAsync();

        // Add job for entity
            .WithIdentity("exampleJob")
            .UsingJobData("entity-name", entity.Entity.Name)
            .Build();
            .WithIdentity("exampleTrigger")
            .StartNow()
            .Build();
        
        using (var suppressedTransactionScope = new TransactionScope(TransactionScopeOption.Suppress, TransactionScopeAsyncFlowOption.Enabled))
        {
        }

        transaction.Complete();
    }
}

// This is working - but not same transaction
public async Task ExampleMethodWithoutScope()
{
    // Make changes to entities
    var entity = _dbContext.ExampleEntities.Add(new ExampleEntity() { Name = new Random().Next(1, 100).ToString() });
    await _dbContext.SaveChangesAsync();

    // Add job for entity
        .WithIdentity("exampleJob")
        .UsingJobData("entity-name", entity.Entity.Name)
        .Build();
        .WithIdentity("exampleTrigger")
        .StartNow()
        .Build();
}
```

The complete example can be found here: https://github.com/NilsEngelbach/quartznet-transactionscope-example/blob/main/Source/TransactionScopeExample/ExampleService.cs

I also saw that in the Quartz Java Implementation there is the Option to use the `JobStoreCMT` where the transaction is also managed outside of Quartz.
How could this feature be implemented in Quartz.NET? I would also love to contribute the feature, but i don't really know where to start and there are any special things to consider?

lahma:
Fixes #2038.

## The problem

Scheduling cannot be made atomic with the application's own database work. `JobStoreSupport.GetConnection()` always opens a connection of its own and calls `BeginTransaction` on it. Inside an ambient `TransactionScope` the freshly opened connection auto-enlists, and the `BeginTransaction` that follows is either rejected outright — Npgsql: *"A transaction is already in progress; nested/concurrent transactions aren't supported"*, surfaced as `JobPersistenceException: Failure setting up connection` — or produces a local transaction that commits regardless of the scope's outcome.

Even past that, the job store is using a *second* connection, so the ambient transaction has to be promoted to a distributed one. That needs MSDTC (Windows only, and since .NET 10 also `TransactionManager.ImplicitDistributedTransactions`), and it is impossible with Npgsql, which has no distributed transaction support at all. That is the wall the recent commenters on the issue hit.

The workaround people converge on is subclassing `JobStoreCMT` and overriding `GetNonManagedTXConnection()` to hand back the EF Core `DbContext`'s connection. It works, but it needs a custom job store type, fights the singleton lifetime of `IJobStore`, and leaks Quartz internals into application code.

## The change

Opt in with `quartz.jobStore.acceptEnlistedTransactions`, or `AcceptEnlistedTransactions()` when configuring through `SchedulerBuilder`:

```csharp
q.UsePersistentStore(s =>
{
    s.UsePostgres(connectionString);
    s.AcceptEnlistedTransactions();
});
```

Then hand the job store a connection for the duration of a scope:

```csharp
await using var tx = await dbContext.Database.BeginTransactionAsync();

dbContext.Add(entity);
await dbContext.SaveChangesAsync();

using (scheduler.EnlistTransaction(tx.GetDbTransaction()))
{
    await tx.CommitAsync();
}
```

The application owns the commit, so the job store neither commits nor rolls back. The enlistment flows with the current asynchronous context, the same way `Transaction.Current` itself does, so nothing has to be resolved from the container — which is what made this intractable while `IJobStore` is a singleton and a `DbContext` is scoped. Nothing here is EF Core specific: any `DbConnection` and `DbTransaction` will do.

**Taking part always means handing over a connection.** An ambient `TransactionScope` on its own is not enough — for `JobStoreTX` a connection the job store opens for itself is deliberately kept *out* of one, since joining a scope whose outcome it does not control would also put a second connection in that transaction. Inside a `TransactionScope`, open the connection there and enlist it; sharing the one connection is what keeps the transaction local:

```csharp
using var scope = new TransactionScope(TransactionScopeOption.Required, options, TransactionScopeAsyncFlowOption.Enabled);

using var connection = new NpgsqlConnection(connectionString);
await connection.OpenAsync();

using (scheduler.EnlistConnection(connection))
{
}

scope.Complete();
```

`JobStoreCMT` is left alone — running inside a transaction its container manages is that store's whole contract.

The switch is opt-in because joining the application's transaction changes when, and whether, scheduling commits.

## Supporting changes

- `ConnectionAndTransactionHolder` can hold borrowed resources, in which case commit, rollback, close and dispose are left to the owner. The new constructor and `OwnsResources` are internal.
- Locking moves to database locks unless a lock handler was configured explicitly — `SimpleSemaphore` releases its in-process lock before the application commits, which would let a concurrent caller act on scheduling data it cannot see yet. SQLite cannot be upgraded this way and logs what that means.
- Scheduler-owned work — column probing, cluster check-in, misfire recovery, recovery at start, and completion bookkeeping after a job runs — never borrows the enlisted connection. A probe is *expected* to fail when a column is absent, and on PostgreSQL that aborts the whole transaction.
- Transient retries are skipped inside an application-owned transaction; on most providers the first failure has already doomed it.
- Write operations arrange a scheduling change signal for after the commit — accumulated on the enlistment so the earliest candidate across the whole scope wins, and hooked once per transaction. `QuartzScheduler` otherwise notifies the scheduler thread only *before* the rows are visible, so the trigger would sit until the idle interval elapsed. Reads do not signal, since a null candidate time bounces acquired triggers back to waiting. When an ambient transaction is present it owns the signal, so a rollback raises nothing.
- Misuse is refused rather than left to fail obscurely: a job store that cannot take part (including `RAMJobStore`), a remote scheduler, a connection whose provider does not match the configured one, a transaction begun on a different connection, the feature switched off, a connection with no transaction to join, a transaction that has since completed, an ambient scope that is no longer current, concurrent calls on one enlisted connection, and starting the scheduler from inside an enlistment scope — which deadlocks against the locks that scope holds. The guard resolves the real scheduler by name, so it still applies behind an `IScheduler` decorator.
- `QuartzScheduler.Start` releases its initial-start marker when — and only when — the job store declines to start before creating anything, so a corrected retry runs the full start-up sequence instead of the resume path. Any later failure keeps the marker latched, since re-running would orphan a misfire handler whose foreground thread outlives `Shutdown`.
- `System.Transactions` reference added for the full-framework targets; netstandard2.0 and net8.0+ need none.

Everything is additive: no existing signature changed, and with the switch off behaviour is what it was.

## Trade-offs and known limits

- The job store takes its locks in the caller's transaction, so they are released only when that transaction completes. Long application transactions block trigger acquisition, the misfire handler and cluster check-in. Inherent to container-managed transactions; `lockOnInsert=false` mitigates it for insert-only workloads.
- No savepoint: an operation that fails halfway leaves its statements in the caller's transaction.
- A connection opened *before* the ambient scope it is enlisted under is not actually part of that scope, and ADO.NET offers no portable way to detect it. Open the connection inside the scope.
- The enlistment rides on `AsyncLocal`, so establishing it inside an `async` helper does not carry it back out to the caller — the same constraint `TransactionScope` has.
- `txIsolationLevelSerializable` applies to the job store's own connections; an enlisted operation runs at the caller's isolation level. `Initialize` says so when both are configured.
- The "this transaction already completed" refusal relies on `DbTransaction.Connection` becoming null, which Npgsql only does on dispose rather than on commit. Dispose the enlistment scope right after committing rather than scheduling more inside it.
- `StartDelayed` logs the enlistment-scope refusal rather than surfacing it, as it does for any start-up failure.

## Testing

Unit tests cover holder ownership, the enlistment flowing/nesting/expiring, isolation per scheduler and per async flow, forking inside a scope being refused rather than corrupting the connection, the completed-transaction and concurrent-use refusals, the fail-fast guard against a non-persistent store, suppression of scheduler-owned work, and that a job store connection stays out of an ambient scope.

Integration tests on PostgreSQL and SQL Server cover rollback discarding the schedule — asserting first that it *is* visible inside the transaction, so the write really went through it — commit keeping it, an incomplete `TransactionScope` discarding it, the enlist-refused guard, and, with a two-minute idle wait and a 30-second assertion, that a running scheduler fires the job promptly after the commit, which only passes if the post-commit signal works.

Solution builds warning-free across all target frameworks; 1740 unit tests, 53 `db-postgres` and 49 `db-sqlserver` integration tests pass.

Documentation lands on `main`, together with the 4.x port, since the site is published from there.

### DIFF
--- a/src/Quartz.Tests.Integration/Impl/AdoJobStore/EnlistedTransactionTest.cs
+++ b/src/Quartz.Tests.Integration/Impl/AdoJobStore/EnlistedTransactionTest.cs
@@ -1,0 +1,405 @@
+#region License
+
+/*
+ * All content copyright Marko Lahma, unless otherwise indicated. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy
+ * of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+
+#endregion
+
+using System;
+using System.Collections.Specialized;
+using System.Data.Common;
+using System.Threading;
+using System.Threading.Tasks;
+using System.Transactions;
+
+using Microsoft.Data.SqlClient;
+
+using Npgsql;
+
+using Quartz.Impl;
+using Quartz.Impl.AdoJobStore;
+using Quartz.Job;
+using Quartz.Util;
+
+using IsolationLevel = System.Transactions.IsolationLevel;
+
+namespace Quartz.Tests.Integration.Impl.AdoJobStore;
+
+/// <summary>
+/// Verifies that scheduling can take part in a transaction the application owns: the schedule is
+/// committed together with the application's own work, and is discarded together with it when that
+/// work is rolled back. Reported as https://github.com/quartznet/quartznet/issues/2038.
+/// </summary>
+[NonParallelizable]
+public class EnlistedTransactionTest
+{
+    [Test]
+    [Category("db-postgres")]
+    public Task PostgresRollingBackTheApplicationTransactionDiscardsTheSchedule()
+    {
+        return RollingBackTheApplicationTransactionDiscardsTheSchedule(Postgres(), "EnlistedRollbackPg");
+    }
+
+    [Test]
+    [Category("db-postgres")]
+    public Task PostgresCommittingTheApplicationTransactionKeepsTheSchedule()
+    {
+        return CommittingTheApplicationTransactionKeepsTheSchedule(Postgres(), "EnlistedCommitPg");
+    }
+
+    [Test]
+    [Category("db-postgres")]
+    public Task PostgresIncompleteAmbientScopeDiscardsTheSchedule()
+    {
+        return IncompleteAmbientScopeDiscardsTheSchedule(Postgres(), "AmbientScopePg");
+    }
+
+    [Test]
+    [Category("db-postgres")]
+    public Task PostgresRunningSchedulerFiresTheJobRightAfterTheApplicationCommits()
+    {
+        return RunningSchedulerFiresTheJobRightAfterTheApplicationCommits(Postgres(), "EnlistedRunningPg");
+    }
+
+    [Test]
+    [Category("db-postgres")]
+    public Task PostgresEnlistingIsRefusedWhenTheStoreDoesNotAcceptIt()
+    {
+        return EnlistingIsRefusedWhenTheStoreDoesNotAcceptIt(Postgres(), "EnlistedRefusedPg");
+    }
+
+    [Test]
+    [Category("db-sqlserver")]
+    public Task SqlServerRollingBackTheApplicationTransactionDiscardsTheSchedule()
+    {
+        return RollingBackTheApplicationTransactionDiscardsTheSchedule(SqlServer(), "EnlistedRollbackMssql");
+    }
+
+    [Test]
+    [Category("db-sqlserver")]
+    public Task SqlServerCommittingTheApplicationTransactionKeepsTheSchedule()
+    {
+        return CommittingTheApplicationTransactionKeepsTheSchedule(SqlServer(), "EnlistedCommitMssql");
+    }
+
+    [Test]
+    [Category("db-sqlserver")]
+    public Task SqlServerIncompleteAmbientScopeDiscardsTheSchedule()
+    {
+        return IncompleteAmbientScopeDiscardsTheSchedule(SqlServer(), "AmbientScopeMssql");
+    }
+
+    private static async Task RollingBackTheApplicationTransactionDiscardsTheSchedule(
+        TestProvider provider,
+        string schedulerName)
+    {
+        IScheduler scheduler = await CreateScheduler(provider, schedulerName);
+        JobKey jobKey = new JobKey("enlisted", schedulerName);
+
+        try
+        {
+            await scheduler.Clear();
+
+            using (DbConnection connection = provider.CreateConnection())
+            {
+                await connection.OpenAsync();
+                using (DbTransaction transaction = connection.BeginTransaction())
+                {
+                    using (scheduler.EnlistTransaction(transaction))
+                    {
+                        await scheduler.ScheduleJob(CreateJob(jobKey), CreateTrigger(jobKey));
+
+                        bool visibleInsideTheTransaction = await scheduler.CheckExists(jobKey);
+                        visibleInsideTheTransaction.Should().BeTrue("the job store wrote through the enlisted transaction");
+                    }
+
+                    transaction.Rollback();
+                }
+            }
+
+            bool survivedTheRollback = await scheduler.CheckExists(jobKey);
+            survivedTheRollback.Should().BeFalse("the schedule belongs to the transaction the application rolled back");
+        }
+        finally
+        {
+            await scheduler.Shutdown(false);
+        }
+    }
+
+    private static async Task CommittingTheApplicationTransactionKeepsTheSchedule(
+        TestProvider provider,
+        string schedulerName)
+    {
+        IScheduler scheduler = await CreateScheduler(provider, schedulerName);
+        JobKey jobKey = new JobKey("enlisted", schedulerName);
+
+        try
+        {
+            await scheduler.Clear();
+
+            using (DbConnection connection = provider.CreateConnection())
+            {
+                await connection.OpenAsync();
+                using (DbTransaction transaction = connection.BeginTransaction())
+                {
+                    using (scheduler.EnlistTransaction(transaction))
+                    {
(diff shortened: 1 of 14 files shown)


lahma:
Two `xhigh` review passes over this branch found 30 verified defects, and the second pass caught four that the first pass's own fixes had introduced. That pattern pointed at the design rather than the code, so the feature is now narrower.

## Taking part always means handing over a connection

The ambient-`TransactionScope`-only half is gone. It could never work on Npgsql — a second connection in the scope needs a distributed transaction — and in every degraded configuration it failed *silently* rather than loudly:

* `Enlist=false` in the connection string: every job store write autocommitted while `Rollback` was a no-op.
* `new TransactionScope()` without `TransactionScopeAsyncFlowOption.Enabled`: `Transaction.Current` is null after the first `await`, so scheduling committed separately — nondeterministically, since a synchronously-completing continuation behaved transactionally.
* `txIsolationLevelSerializable` was silently dropped.

Now a connection the job store opens for itself is deliberately kept *out* of any ambient transaction while the feature is on, instead of quietly joining a scope whose outcome it does not control. One mode, identical on every provider, and those three corruption paths no longer exist.

`EnlistConnection(connection)` inside a `TransactionScope` is unchanged and still the recommended shape — the connection carries the transaction, so nothing has to be promoted.

## Other fixes from review

* Scheduler-owned work — column probing, cluster check-in, misfire recovery, recovery at start — no longer borrows the enlisted connection. A probe is *expected* to fail when a column is absent, and on PostgreSQL that aborts the whole transaction, so it was poisoning work the caller had already done.
* Using an enlistment after its transaction completed is refused, rather than running in autocommit where a half-finished write can no longer be rolled back.
* Concurrent scheduler calls on one enlisted connection are refused with an explanation instead of an opaque "a command is already in progress".
* The single-use claim is released through the holder that took it, so nested enlistments cannot release each other's.
* Starting the scheduler from inside an enlistment scope is refused — it deadlocked: startup waits for locks the enlisted transaction holds until the caller commits, which it cannot do while awaiting `Start()`.
* Post-commit signalling was aimed at the wrong path entirely. `SignalSchedulingChangeOnTxCompletion` is only ever set by `TriggeredJobComplete`, which never runs inside an application transaction, while `QuartzScheduler.NotifySchedulerThread` — which does fire, before the commit — was untouched. Writes now signal on both the ambient transaction and the enlistment scope; reads no longer signal at all, since a null candidate time was bouncing acquired triggers back to waiting on every query.
* `JobStoreCMT` signals too, now that it can be enlisted.
* Enlisting is refused up front when it would be ignored: a non-persistent job store (`RAMJobStore` fell straight through the old guard), the feature switched off, or a connection with no transaction and no ambient scope to join.
* The borrowed-holder constructor and `OwnsResources` are internal rather than public.

## Known limits, deliberately not addressed

* **No savepoint.** A job store operation that fails halfway leaves its partial statements in your transaction, and on PostgreSQL a failed statement aborts it. Containing this properly needs savepoint support across the job store (`DbTransaction.Save` is .NET-Core-only and provider-specific).
* **Enlisting inside an `async` helper is silently discarded.** Inherent to `AsyncLocal` — the write never escapes the state machine. `TransactionScope` has the identical constraint. Documented, not detectable.

## Testing

Unit tests cover holder ownership, the enlistment flowing/nesting/expiring, isolation per scheduler and per async flow, the completed-transaction and concurrent-use refusals, suppression of scheduler-owned work, and that a job store connection stays out of an ambient scope. Integration tests on PostgreSQL and SQL Server cover rollback discarding the schedule (asserting first that it *is* visible inside the transaction, so the write really went through it), commit keeping it, an incomplete `TransactionScope` discarding it, the enlist-refused guard, and — with a two-minute idle wait and a 30-second assertion — that a running scheduler fires the job promptly after the commit, which only passes if the post-commit signal works.

Solution builds warning-free across all target frameworks; 1738 unit tests, 53 `db-postgres` and 49 `db-sqlserver` integration tests pass.

PR #3208, opened 2026-07-29
Let AdoJobStore take part in the application's transaction

### LINKED ISSUE #2038: Support for TransactionScope to include the scheduling in an existing transaction

lahma:
Fixes #2038.

Ports the 3.x feature (#3204, merged) to 4.x, and adds the documentation for **both** lines — 3.x and 4.x — since the site is published from this branch. Merging #3204 into `3.x` did not close the issue, so this PR claims it.

## The problem

Scheduling cannot be made atomic with the application's own database work. `JobStoreSupport.GetConnection()` always opens a connection of its own and begins a transaction on it. Inside an ambient `TransactionScope` the freshly opened connection auto-enlists, and the `BeginTransaction` that follows is either rejected outright — Npgsql: *"A transaction is already in progress; nested/concurrent transactions aren't supported"* — or produces a local transaction that commits regardless of the scope's outcome. Even past that, a second connection forces promotion to a distributed transaction: MSDTC-only, Windows-only, and impossible with Npgsql.

## The change

Opt in with `AcceptEnlistedTransactions()` on the persistent store builder, `JobStore:AcceptEnlistedTransactions`, or `quartz.jobStore.acceptEnlistedTransactions`:

```csharp
builder.Services.AddQuartz(q =>
{
    q.UsePersistentStore(store =>
    {
        store.UsePostgres(connectionString);
        store.AcceptEnlistedTransactions();
    });
});
```

Then hand the job store a connection for the duration of a scope:

```csharp
await using var tx = await dbContext.Database.BeginTransactionAsync();

dbContext.Add(entity);
await dbContext.SaveChangesAsync();

using (scheduler.EnlistTransaction(tx.GetDbTransaction()))
{
    await tx.CommitAsync();
}
```

The application owns the commit, so the job store neither commits nor rolls back. The enlistment flows with the current asynchronous context, the same way `Transaction.Current` itself does, so nothing has to be resolved from the container — which is what made this intractable while `IJobStore` is a singleton and a `DbContext` is scoped. Nothing here is EF Core specific: any `DbConnection` and `DbTransaction` will do.

**Taking part always means handing over a connection.** An ambient `TransactionScope` on its own is not enough — for `JobStoreTX` a connection the job store opens for itself is deliberately kept *out* of one, since joining a scope whose outcome it does not control would also put a second connection in that transaction. Inside a `TransactionScope`, open the connection there and enlist it; sharing the one connection is what keeps the transaction local. `JobStoreCMT` is left alone — running inside a transaction its container manages is that store's whole contract.

## What differs from the 3.x version

Mostly mechanical adaptation — `ValueTask` and `CancellationToken` throughout, the sealed `ConnectionAndTransactionHolder`, `Quartz.Extensibility` rather than `Quartz.Spi`, and no optional-column probing or `RemoteScheduler` on this branch. Two things are genuinely different:

- **Configuration is DI-native.** `AdoJobStoreOptions.AcceptEnlistedTransactions`, a `QuartzPropertyBridge` mapping for the legacy property key, and `IPersistentStoreBuilder.AcceptEnlistedTransactions()` — main has no `SchedulerBuilder.PersistentStoreOptions`.
- **The job store itself refuses an enlistment it was not configured to accept**, rather than ignoring it. 3.x resolves a decorated scheduler through the process-wide `SchedulerRepository`; main has none, so the guarantee moved to where it cannot be bypassed. This is stricter than 3.x and worth backporting there.

`DelegatingScheduler` gained a `protected internal InnerScheduler`, so the call-site checks still apply behind a decorator.

## Supporting changes

- `ConnectionAndTransactionHolder` can hold borrowed resources, in which case commit, rollback, close and dispose are left to the owner. The new constructor and `OwnsResources` are internal.
- Locking moves to database locks unless a lock handler was configured explicitly — `SimpleSemaphore` releases its in-process lock before the application commits. SQLite cannot be upgraded this way and logs what that means.
- Scheduler-owned work — cluster check-in, misfire recovery, recovery at start, and completion bookkeeping after a job runs — never borrows the enlisted connection.
- Transient retries are skipped inside an application-owned transaction; on most providers the first failure has already doomed it.
- Write operations arrange a scheduling change signal for after the commit, accumulated on the enlistment so the earliest candidate across the scope wins, and hooked once per transaction. Reads do not signal. When the enlistment was made under an ambient transaction, that transaction owns the signal, so a rollback raises nothing.
- Misuse is refused rather than left to fail obscurely: a job store that cannot take part, a connection whose provider does not match the configured one, a transaction begun on a different connection, a connection with no transaction to join, a transaction that has since completed, an ambient scope that is no longer current, concurrent calls on one enlisted connection, and starting the scheduler from inside an enlistment scope — which deadlocks against the locks that scope holds.
- `QuartzScheduler.Start` releases its initial-start marker when, and only when, the job store declines to start before creating anything, so a corrected retry runs the full start-up sequence rather than the resume path.

## Trade-offs and known limits

- The job store takes its locks in the caller's transaction, so they are released only when that transaction completes. Long application transactions block trigger acquisition, the misfire handler and cluster check-in. Inherent to container-managed transactions; `LockOnInsert = false` mitigates it for insert-only workloads.
- No savepoint: an operation that fails halfway leaves its statements in the caller's transaction.
- A connection opened *before* the ambient scope it is enlisted under is not actually part of that scope, and ADO.NET offers no portable way to detect it. Open the connection inside the scope.
- The enlistment rides on `AsyncLocal`, so establishing it inside an `async` helper does not carry it back out to the caller — the same constraint `TransactionScope` has.
- Disposing an enlistment scope after a rollback still raises one scheduling-change signal, which costs the scheduler a look and nothing more.
- `StartDelayed` logs the start refusal rather than surfacing it, as it does for any start-up failure.
- Only the *first* `Start()` is refused inside an enlistment scope; resuming from standby is not, and the docs say so.
- A decorator that implements `IScheduler` directly rather than deriving from `DelegatingScheduler` bypasses the call-site checks. A persistent ADO.NET store still refuses the enlistment when it reaches one, so only a non-persistent store behind such a decorator goes unreported.
- Two schedulers enlisted on the *same* connection each hold their own claim, so the concurrency guard does not catch that combination.
- `CloseConnection` overrides in another assembly cannot tell a borrowed connection apart, since `OwnsResources` is internal.

## Testing

Unit tests cover borrowed-holder ownership, the enlistment flowing/nesting/expiring, isolation per scheduler and per async flow, forking inside a scope being refused, the completed-transaction and concurrent-use refusals, the store-does-not-accept refusal, suppression of scheduler-owned work, and that a job store connection stays out of an ambient scope.

Integration tests on PostgreSQL cover rollback discarding the schedule — asserting first that it *is* visible inside the transaction — commit keeping it, an incomplete `TransactionScope` discarding it, the refusal when the store is not configured for it, and two started-scheduler tests that assert the job fires promptly after the commit: one for a bare `DbTransaction` and one for an ambient `TransactionScope`, each with a two-minute idle wait and a 30-second assertion, so they only pass if the post-commit signal actually works.

Solution builds warning-free; 2000 unit tests and 58 `db-postgres` integration tests pass. The public API baseline and `changelog.md` are updated together.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -607,6 +607,34 @@
     If the calculated day would cross a month boundary it will be reset to the 1st of the month. (e.g. LW-30 in Feb will be 1st Feb)
   * Add cron parser support for day-of-month and day-of-week together. (#1543)
 
+#### Transactional scheduling
+
+  * AdoJobStore can now take part in a transaction the application owns, so saving your own data and scheduling
+    the job that acts on it commit together or not at all (#2038). Turn it on with
+    `AcceptEnlistedTransactions()` on the persistent store builder, `JobStore:AcceptEnlistedTransactions`, or
+    `quartz.jobStore.acceptEnlistedTransactions`, then hand the store a connection for the duration of a scope:
+
+    ```csharp
+    await using var tx = await dbContext.Database.BeginTransactionAsync();
+    await dbContext.SaveChangesAsync();
+
+    using (scheduler.EnlistTransaction(tx.GetDbTransaction()))
+    {
+        await scheduler.ScheduleJob(job, trigger);
+        await tx.CommitAsync();
+    }
+    ```
+
+    Handing over a connection is the only way to take part: a connection the job store opens for itself is
+    deliberately kept out of any ambient `TransactionScope`, since a second connection in that transaction would
+    require promoting it to a distributed one. New public API: the `SchedulerEnlistmentExtensions` static class with
+    `EnlistTransaction` and `EnlistConnection` extension methods on `IScheduler`,
+    `AdoJobStoreOptions.AcceptEnlistedTransactions`,
+    `IPersistentStoreBuilder.AcceptEnlistedTransactions()` — the last of which is a breaking addition for anyone
+    implementing `IPersistentStoreBuilder` themselves — and `DelegatingScheduler.InnerScheduler`, so a decorated
+    scheduler can still be resolved to the real one. See
+    [Joining an existing transaction](https://www.quartz-scheduler.net/documentation/quartz-4.x/tutorial/job-stores.html#joining-an-existing-transaction).
+
 ### FIXES
 
   * Fix for deserializing CronExpression using Json Serializer throwing error calling `GetNextValidTimeAfter`.  (#1996)
--- a/docs/documentation/quartz-3.x/configuration/reference.md
+++ b/docs/documentation/quartz-3.x/configuration/reference.md
@@ -363,6 +363,7 @@ JobStoreTX can be tuned with the following properties:
 | quartz.jobStore.maxMisfiresToHandleAtATime   | no       | int     | 20                                                                           |
 | quartz.jobStore.selectWithLockSQL            | no       | string  | "SELECT * FROM {0}LOCKS WHERE SCHED_NAME = {1} AND LOCK_NAME = ? FOR UPDATE" |
 | quartz.jobStore.txIsolationLevelSerializable | no       | boolean | false                                                                        |
+| quartz.jobStore.acceptEnlistedTransactions        | no       | boolean | false                                                                        |
 | quartz.jobStore.acquireTriggersWithinLock    | no       | boolean | false (or true - see doc below)                                              |
 | quartz.jobStore.lockHandler.type             | no       | string  | null                                                                         |
 | quartz.jobStore.driverDelegateInitString     | no       | string  | null                                                                         |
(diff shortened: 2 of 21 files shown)


NilsEngelbach:
Nice!

PR #3271, opened 2026-08-11
Pin the SDK to the 10.0.3xx band until 10.0.400's test breakage is fixed

lahma:
CI infrastructure hotfix — blocks nothing but unblocks everything.

The GitHub runner images rolled out .NET SDK **10.0.400** today, and its `dotnet test` fails inside the SDK's own targets before running a single test:

> `InvalidProjectFileException: The expression "[MSBuild]::GetTargetFrameworkIdentifier(net10.0)" cannot be evaluated. Could not load file or assembly 'NuGet.Frameworks, Version=7.9.0.0' … manifest definition does not match` (`sdk\10.0.400\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.TargetFrameworkInference.targets`)

All three OS legs of #3270 failed identically on this; the same will hit every PR as the images finish rolling out. `rollForward: latestMinor` is what floated CI onto the new SDK the day it appeared — `latestPatch` from 10.0.303 stays on the proven feature band and still picks up servicing fixes. Relax back once a fixed 10.0.4xx ships.

The 3.x branch carries the same `global.json` policy and will need the same pin — companion PR to follow if its CI starts failing.

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "10.0.100",
-    "rollForward": "latestMinor"
+    "version": "10.0.303",
+    "rollForward": "latestPatch"
   }
 }


lahma:
Superseded by #3277 (`main`) and #3278 (`3.x`).

Pinning the SDK back to the 10.0.3xx band treats the symptom. The real cause is an assembly-binding
conflict inside the Fallout process: the `UnitTest` target evaluates test projects **in-process**,
which runs the installed SDK's targets, and those bind `NuGet.Frameworks` by strong name. SDK
10.0.400 wants `7.9.0.0`, while `Fallout.Components` resolves `6.14.3` transitively through
`NuGet.Packaging`, and the app-local copy shadows the SDK's.

Pinning `NuGet.Frameworks` forward to `7.9.0` in `build/_build.csproj` fixes it at the cause and
leaves `global.json` free to keep floating, instead of dating the repository to one feature band
that has to be revisited every SDK release.

Issue #3205, opened 2026-07-28
Cluster-aware query for currently executing fire instances


lahma:
### Background

`IScheduler.GetCurrentlyExecutingJobs()` is documented as not cluster aware — it only ever reports what is
running on the node it is called on. Any process that observes a scheduler without executing jobs (a
dashboard, an admin UI, a separate control application) therefore cannot see what is running anywhere else.

#1416 closed the immediate gap by adding `TriggerState.Executing`, which *is* cluster-visible because it is
established from `QRTZ_FIRED_TRIGGERS` rather than from process-local state. That was deliberately the
cheaper answer, and it leaves the richer one on the table.

### The limitation

`TriggerState` is a single-valued scheduling disposition, so collapsing execution into it is lossy by
construction:

* **Multiplicity is erased.** A trigger with a 1-second interval and a 60-second concurrency-allowed job can
  have many executions in flight at once. `Executing` says "at least one".
* **Whichever fact loses the precedence contest is erased.** The order is
  `None > Error > Paused > Executing > Blocked > Complete > Normal`, so a paused-but-still-running trigger
  reports `Paused` and a UI cannot show that its job is still going.
* **Nothing identifies the execution.** There is no fire instance id, no node, no start time — so a stale
  `EXECUTING` row left by a node that died is indistinguishable from a healthy long-running job until
  `ClusterRecover` runs.

### Proposal

A cluster-aware query returning the fire instances themselves, something like:

```csharp
ValueTask<List<ExecutingFireInstance>> GetExecutingFireInstances(
    TriggerKey? triggerKey, CancellationToken cancellationToken = default);

// FireInstanceId, TriggerKey, JobKey, SchedulerInstanceId, FireTimeUtc, ScheduledFireTimeUtc
```

Deliberately **not** `IJobExecutionContext` — a remote node's `JobInstance`, `Result`, `MergedJobDataMap`
and cancellation handle are not reconstructible, so widening `GetCurrentlyExecutingJobs` to cluster scope
would be a worse design than the enum member, not a better one.

That shape lets a UI say **"Paused · 3 running on node-2 since 14:02"** instead of collapsing everything into
one word.

### Cost

* **ADO: small.** `IDriverDelegate.SelectFiredTriggerRecords` already exists, `FiredTriggerRecord` is already
  the right shape minus the ADO-only fields, and `IDX_QRTZ_FT_T_G (SCHED_NAME, TRIGGER_NAME, TRIGGER_GROUP)`
[...]

Issue #1416, opened 2021-11-15
New status of TriggerState to see job execution


Kuznetsov-Alexey:
In our project, there is a situation when a task can be completed within 1 minute or more (reading a file and writing to a database). So it would be nice to add a new status to TriggerState to see if the trigger has fired and is now executing.

We have a client side for triggers with CRUD and a server side to perform them and tried to find workaround way to realize it. There is impossible to get **Executing** status on client side, only actual trigger state, so I tried to realize getting this status with one more database request and if-condition - [Added new TriggerState for tracking job execution status](https://github.com/Kuznetsov-Alexey/quartznet/commit/d55c7c24a9c7e090dd1e82c33448b004a2d97d02).

This be useful for similar long-term operations, because users will be able to see that the job is currently running. Is it possible to add new status? If yes, we can try to find a better solution or I can add tests to the existing commit and create Pull request.

Sicos1977:
You can put your executing state in a job map and read that from the client if you are on the full .net framework.

Something like I do here.

```c#
context.JobDetail.JobDataMap.Put("Status", $"Controleren of groep {group.Name} bestaat in active directory");
```

drieseng:
There's a `IJobStore.GetTriggerState(TriggerKey triggerKey, CancellationToken cancellationToken = default)` method that should allow you to obtain this information, but for some reason it returns **TriggerState.Normal** when the trigger is executing. We do not (yet) have an **Executing** value in our (public) **TriggerState** enum.

If you use an **ADO.NET** based job store, you should be able to get the right status from the **entryState** column of the **FIRED_TRIGGERS** table as we correctly update it [here](https://github.com/quartznet/quartznet/blob/6053359d8756e3dbbd0786272ddf17f9a4bbb28c/src/Quartz/Impl/AdoJobStore/JobStoreSupport.cs#L2892).

For **RAMJobStore**, we are setting the state to **Waiting** when the trigger is being (is bound to be) executed. This, even though we have a **Executing** value in our **InternalTriggerState**.

Perhaps @lahma can shed some light on this.

Sicos1977:
> There's a `IJobStore.GetTriggerState(TriggerKey triggerKey, CancellationToken cancellationToken = default)` method that should allow you to obtain this information, but for some reason it returns **TriggerState.Normal** when the trigger is executing. We do not (yet) have an **Executing** value in our (public) **TriggerState** enum.
> 
[...]

PR #3281, opened 2026-08-19
Add cluster-aware GetExecutingFireInstances API (#3205)

HarnageaGabriel:
## Summary

Implements the `GetExecutingFireInstances` API proposed in #3205: a cluster-aware complement to `GetCurrentlyExecutingJobs` and `TriggerState.Executing` (#1416) that returns the individual fire instances currently executing anywhere in the cluster, rather than collapsing them into a single per-node snapshot or a single enum value.

- **Core model**: `ExecutingFireInstance` (`FireInstanceId`, `TriggerKey`, `JobKey`, `SchedulerInstanceId`, `FireTimeUtc`, `ScheduledFireTimeUtc`), new `IJobStore`/`IScheduler` member `GetExecutingFireInstances(TriggerKey? triggerKey, ct)`.
- **RAMJobStore**: a new `FireInstanceId`-keyed dictionary alongside the existing `TriggerKey`-keyed one that backs `TriggerState.Executing` (left untouched — it's load-bearing). Since RAMJobStore has no constructor-time access to the scheduler's configured instance id, `DefaultSchedulerFactory` now tells it once the id is known, mirroring how `JobStoreSupport` already learns it.
- **ADO job store**: `IDriverDelegate.SelectExecutingFiredTriggers`, a new listing-wide `QRTZ_FIRED_TRIGGERS` query filtered to the `EXECUTING` state (reuses the existing `SCHED_NAME`/`TRIGGER_NAME`/`TRIGGER_GROUP` index — no schema changes needed). `JobStoreSupport.GetExecutingFireInstances` maps the results, following the same `ExecuteWithoutLock` pattern as `GetTriggerState`.
- **HTTP API**: new `GET .../jobs/executing-fire-instances` endpoint (mirrors `/currently-executing`), wired through `Quartz.HttpClient`'s `HttpScheduler`.
- **Blazor dashboard**: the "Currently Executing" stat tile on `Dashboard.razor` now uses the cluster-aware query instead of `GetCurrentlyExecutingJobs` — per the issue's "Bonus" note, that tile was already wrong in a cluster (it only counted the node the dashboard happened to be attached to). `CurrentlyExecuting.razor` (the per-node interrupt page) is intentionally left alone, since interrupting a job only ever works on the local node regardless of what query backs its listing.

This is organized as 5 commits, one per subsystem, in dependency order (core → ADO → HTTP client → HTTP endpoint → dashboard) — happy for it to be reviewed incrementally if that's easier than one pass.

## What's *not* included

- No database migration — the query reuses existing columns/indexes on `QRTZ_FIRED_TRIGGERS`.
- No changes to `docs/documentation/quartz-4.x/migration-guide.md` — this is additive API, not a 3.x→4.x contract change.
- Dialect-specific `IDriverDelegate` subclasses (SqlServer/PostgreSQL/Oracle/MySQL/SQLite/Firebird) inherit `StdAdoDelegate`'s implementation as-is (it's `virtual`, not `abstract`); none of them needed an override since the query has no vendor-specific paging or syntax.

## Test plan

- [x] `dotnet test src/Quartz.Tests.Unit/Quartz.Tests.Unit.csproj` — 2319 passed, 4 skipped (pre-existing skips), 0 failed
- [x] `dotnet test src/Quartz.Tests.AspNetCore/Quartz.Tests.AspNetCore.csproj` — 217 passed, 0 failed
- [x] New unit tests: `RAMJobStoreTest` (fire/complete/filter/multiplicity, plus a regression test tying `GetExecutingFireInstances` to `TriggerState.Executing`), `JobStoreSupportTest` (fake-delegate mapping + defensive null-JobKey skip), `JobEndpointsTest` (end-to-end HTTP round trip: fake scheduler → endpoint → `HttpScheduler`)
- [x] Public API `Verify` baselines (`Quartz`, `Quartz.HttpClient`, `Quartz.Dashboard`) reviewed and accepted — diffs contain exactly the new surface, nothing else
- [ ] Integration tests (`Quartz.Tests.Integration`) not run — they require Docker and weren't exercised in this environment; the ADO delegate change only adds a new read query reusing existing columns, but a maintainer with a DB-backed CI run may want to double check the SQL against a real database
- [ ] No manual verification against the actual Blazor dashboard UI (no browser session in this environment) — the API-level round trip is tested, but the rendered tile itself wasn't visually confirmed

### DIFF
--- a/src/Quartz.AspNetCore/AspNetCore/HttpApi/Endpoints/JobEndpoints.cs
+++ b/src/Quartz.AspNetCore/AspNetCore/HttpApi/Endpoints/JobEndpoints.cs
@@ -33,6 +33,9 @@ internal static class JobEndpoints
         yield return builder.MapGet(patternPrefix + "/currently-executing", CurrentlyExecutingJobs)
             .WithQuartzDefaults(nameof(CurrentlyExecutingJobs), "Get currently executing jobs");
 
+        yield return builder.MapGet(patternPrefix + "/executing-fire-instances", ExecutingFireInstances)
+            .WithQuartzDefaults(nameof(ExecutingFireInstances), "Get executing fire instances");
+
         yield return builder.MapPost(patternPrefix + "/{jobGroup}/{jobName}/pause", PauseJob)
             .WithQuartzDefaults(nameof(PauseJob), "Pause job");
 
@@ -189,6 +192,27 @@ internal static class JobEndpoints
         {
             var currentlyExecutingJobs = await scheduler.GetCurrentlyExecutingJobs(cancellationToken).ConfigureAwait(false);
             var result = currentlyExecutingJobs.Select(CurrentlyExecutingJobDto.Create).ToArray();
+            return result;
+        });
+    }
+
+    [ProducesResponseType(typeof(ExecutingFireInstanceDto[]), StatusCodes.Status200OK)]
+    private static Task<IResult> ExecutingFireInstances(
+        EndpointHelper endpointHelper,
+        ISchedulerRepository schedulerRepository,
+        string schedulerName,
+        string? triggerName = null,
+        string? triggerGroup = null,
+        CancellationToken cancellationToken = default)
+    {
+        TriggerKey? triggerKey = triggerName is not null && triggerGroup is not null
+            ? new TriggerKey(triggerName, triggerGroup)
+            : null;
+
+        return EndpointHelper.ExecuteWithJsonResponse(schedulerName, schedulerRepository, async scheduler =>
+        {
+            var executingFireInstances = await scheduler.GetExecutingFireInstances(triggerKey, cancellationToken).ConfigureAwait(false);
+            var result = executingFireInstances.Select(ExecutingFireInstanceDto.Create).ToArray();
             return result;
         });
     }
--- a/src/Quartz.Dashboard/Components/Pages/Dashboard.razor
+++ b/src/Quartz.Dashboard/Components/Pages/Dashboard.razor
@@ -143,7 +143,7 @@
             JobPageDto jobs = await Api.GetJobs(schedulerName, groupFilter: null, page: 1, pageSize: 0);
             TriggerPageDto triggers = await Api.GetTriggers(schedulerName, groupFilter: null, state: null, page: 1, pageSize: 0);
             TriggerPageDto errorTriggers = await Api.GetTriggers(schedulerName, groupFilter: null, state: TriggerState.Error, page: 1, pageSize: 0);
-            List<CurrentlyExecutingJobDto> executing = await Api.GetCurrentlyExecutingJobs(schedulerName);
+            List<ExecutingFireInstanceDto> executing = await Api.GetExecutingFireInstances(schedulerName);
 
             _totalJobs = jobs.TotalCount;
             _totalTriggers = triggers.TotalCount;
@@ -223,7 +223,7 @@
                 return;
             }
 
-            List<CurrentlyExecutingJobDto> executing = await Api.GetCurrentlyExecutingJobs(schedulerName);
+            List<ExecutingFireInstanceDto> executing = await Api.GetExecutingFireInstances(schedulerName);
             _currentlyExecuting = executing.Count;
         }
         catch (Exception ex)
--- a/src/Quartz.Dashboard/Services/IQuartzApiClient.cs
+++ b/src/Quartz.Dashboard/Services/IQuartzApiClient.cs
@@ -53,6 +53,12 @@ public interface IQuartzApiClient
     ValueTask<List<CurrentlyExecutingJobDto>> GetCurrentlyExecutingJobs(string schedulerName, CancellationToken cancellationToken = default);
 
     /// <summary>
+    /// Gets the fire instances currently executing across every node of the cluster, unlike
+    /// <see cref="GetCurrentlyExecutingJobs" /> which only sees the node it is called on.
+    /// </summary>
+    ValueTask<List<ExecutingFireInstanceDto>> GetExecutingFireInstances(string schedulerName, CancellationToken cancellationToken = default);
+
+    /// <summary>
     /// Pauses the job. Returns <see langword="true" /> when the job existed and was paused,
     /// <see langword="false" /> when there was nothing to pause.
     /// </summary>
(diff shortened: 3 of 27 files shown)


lahma (changes requested):
Thanks for this — and for the quality of it. The RAMJobStore multiplicity tests and the "don't collapse concurrent executions" framing show you understood exactly why #3205 exists, and the `ExecuteWithoutLock` pattern on the ADO side is right. We want this feature in 4.0, and we'd like this PR to be the vehicle.

The reason for **request changes** is that cluster-aware execution listing was already designed in detail during the 4.0 API finalization campaign (it's the feature we've been calling **F1** internally), and the ratified design differs from this PR's shape in ways that are cheaper to adopt now than to migrate later. Since 4.x is pre-release, we get exactly one chance to ship the final shape. Here is the full checklist; happy to expand on any point.

1. **Rebase first.** This branch predates #3270 and its neighbors: `JobStoreSupport` is now `AdoJobStoreBase`, `Quartz.Core` is gone (`QuartzScheduler` moved), and several store members were reshaped. The rebase will be mechanical but noisy — do it before reworking anything else.

2. **Query object + paging, not a filter parameter.** The 4.0 listing convention is query records with paged results (`QueryJobs`/`QueryTriggers` → `PagedResult<T>`). The member should be `QueryFireInstances(FireInstanceQuery query, CancellationToken ct)` on `IScheduler` and `IJobStore`, where `FireInstanceQuery : PagedQuery` carries `Trigger`/`Job`/`SchedulerInstanceId` filters plus a `State` filter. A busy cluster can have thousands of rows in `FIRED_TRIGGERS`; an unpaged `List<>` return is the shape we've been removing everywhere else this cycle. The ordering contract needs an explicit ORDER BY with the fire-instance id as tiebreaker so paging is stable.

3. **The record models both fired-trigger states.** Name it `FireInstance` (it isn't only "executing"): your five fields plus `ExecutionGroup?` and `FireInstanceState State` — a new store-neutral enum `{ Acquired, Executing }`. `JobKey` becomes nullable (it's unknown while a row is merely Acquired). The `State` filter on the query defaults to `Executing` so the common call reads like yours does today.

4. **`GetCurrentlyExecutingJobs` is removed by this feature, not complemented.** That removal is a ratified 4.0 decision — two overlapping listings with different cluster semantics is the confusion #3205 documents. `QuartzScheduler`'s internal uses stay on the internal synchronous member (interrupt must never become a store round-trip), and the in-process live-`IJobExecutionContext` use case gets a documented context-tracking-listener sample instead. Your PR description already says it mirrors the old API — after this change the PR can honestly say `Closes #3205`.

5. **RAM store identity arrives through `Initialize`, not a settable property.** `public string SchedulerInstanceId { get; set; }` mutable-after-start identity is a pattern we've been deleting. The ratified mechanism: `IJobStore.Initialize` gains a `SchedulerIdentity` record parameter (same precedent as `SemaphoreContext`), so every store learns its instance id the same way and `DefaultSchedulerFactory` doesn't special-case RAM. Also, rather than a second `FireInstanceId`-keyed dictionary, keep the existing `TriggerKey`-keyed one (it backs the `TriggerState.Executing` probe) and widen its values to entries carrying `(FireInstanceId, JobKey, FireTimeUtc, ScheduledFireTimeUtc, ExecutionGroup)`; copy-then-release under the lock.

6. **ADO side: one new delegate member with its own SQL.** `IDriverDelegate.SelectFireInstances(conn, FireInstanceQuery, ct)` with its own statement, paging via `ApplyPaging`/`AddPagingParameters`, and the explicit ORDER BY from point 2. Don't extend `FiredTriggerQuery` — it is deliberately unpaged because `ClusterRecover` depends on reading every row; keeping the two queries separate is what protects that. While you're in the statement: `QRTZ_FIRED_TRIGGERS.EXECUTION_GROUP` exists in the schema but is dead today — this feature makes it live (written on insert/update, read back into `FireInstance.ExecutionGroup`).

7. **Wire contract.** The DTO change is a declared breaking change: old→new field table in the PR description plus a JSON snapshot test pinning the new shape. `FireInstanceId`, `ExecutionGroup`, and `State` all go on the wire. One sequencing heads-up: we have a wire-contract consolidation (internally "F5") queued ahead of this feature, so expect a coordinated rebase on the HTTP side — keep that layer as thin as you can until then. `SchedulerMetadata` keeps a node-local count as `LocalExecutingJobs` (named and documented as node-local).

8. **Tests and docs.** Rewrite (don't delete) `ExecutingTriggerStateClusteredPostgresTest` against the new member; your RAMJobStore multiplicity tests carry over nearly as-is and are exactly what we want. Docs caveats worth a sentence each: vetoed fires never appear (they're removed at veto), elapsed time is observer-now minus `FIRED_TIME` so clock skew applies (clamp at zero), and `SCHED_TIME` reflects the post-misfire schedule.

9. **Process bits** (same as on #3282): the `PublicApiTest_*.verified.txt` baselines move with this PR — review and re-approve them deliberately, never hand-edit; add the matching rows to `docs/documentation/quartz-4.x/migration-guide.md` in the same PR (including the `GetCurrentlyExecutingJobs` removal row); and run `.\build.cmd Compile UnitTest PublishAot` locally — plain `Compile` misses the trim-analysis leg CI runs.

This is a bigger reshape than a normal review round, because you happened to build a feature we'd already specced. If you're up for it, we'll review promptly at each step. If it's more than you want to take on, no hard feelings — say so and we'll land the specced version ourselves, crediting this PR and keeping your RAMJobStore tests.

lahma:
Following up on the review: with the alpha window closing we've landed the specced version as #3329 — it credits this PR in its description, and your multiplicity insight (concurrent executions of one trigger key must never collapse) is pinned there by dedicated tests on both stores. Thank you for pushing on #3205; the feature exists because you started it. Closing this PR alongside #3329's merge — and if you'd like to build on the new surface (the `FireInstanceQuery` filters could grow, and the dashboard's fire-instance view has room to get richer), we'd be glad to see more from you.

lahma:
Superseded by the merged #3329 — see the note above; thank you again, and the invitation stands.

HarnageaGabriel:
No hard feelings at all — that's a great outcome, thanks for landing it properly and for the credit. Happy to look at FireInstanceQuery filters or the dashboard fire-instance view next, will take a look at #3329 first to get oriented on the new surface.

PR #3282, opened 2026-08-19
Add job type exclusion filter to trigger acquisition (#2238)

### LINKED ISSUE #2238: [Proposal] Change IDriverDelegate to Filter Triggers for Acquisition Based on Job Type

HarnageaGabriel:
Implements #2238: a job-type exclusion filter for trigger acquisition, pushed down into the SQL instead of over-fetching triggers and discarding unwanted ones in memory.

**⚠️ No prior maintainer confirmation.** I [posted on the issue](https://github.com/quartznet/quartznet/issues/2238#issuecomment-5333611370) proposing this approach and asking for a nod before investing time across all the DB delegates, since it touches `IDriverDelegate`'s data contract. The issue had been open ~2.5 years with zero maintainer response and got none in the time I waited either, so I went ahead per my own judgment rather than leave it stalled indefinitely. Please review the design choices below with extra scrutiny — I'm glad to rework the shape if you'd rather see something else.

### What changed

The original issue (filed against an older `SelectTriggerToAcquire` shape) sketched adding a `Type[] jobTypesToExclude` parameter directly to `IDriverDelegate.SelectTriggerToAcquire` and building the exclusion into `GetSelectNextTriggerToAcquireSql` via string-interpolated SQL.

Since then the delegate signature changed to `SelectTriggersToAcquire(conn, TriggerAcquisitionCriteria, ct)`, and `TriggerAcquisitionCriteria`'s own doc comment already flags itself as "the extension point for acquisition filtering (see issue #2238)" — every future acquisition dimension is meant to land as an additive optional property rather than a signature change. So instead of touching `IDriverDelegate`, this PR:

1. Adds `JobTypesToExclude` (`IReadOnlyCollection<string>?`, job class names — the same format `TriggerAcquireResult.JobTypeName` and `JobHeader.JobTypeName` already store) as an optional property on both `TriggerAcquisitionRequest` (store-level) and `TriggerAcquisitionCriteria` (delegate-level), mirroring exactly how `ExecutionLimits` already flows through both layers. `null`/empty means no filtering, so every existing caller is unaffected.
2. `JobStoreSupport.AcquireNextTrigger` threads `request.JobTypesToExclude` into the criteria it builds, same as it already does for `ExecutionLimits`.
3. `StdAdoConstants.SqlSelectNextTriggerToAcquire` becomes `BuildSqlSelectNextTriggerToAcquire(int jobTypesToExcludeCount)`, which appends a **parameterized** `AND JOB_CLASS_NAME NOT IN (@excludedJobType0, @excludedJobType1, ...)` clause only when the count is non-zero — unlike the string-interpolated `NOT IN` in the issue's sketch, this avoids any injection risk. `jd.JOB_CLASS_NAME` was already joined and selected by this query, so no schema/join changes were needed.
4. `StdAdoDelegate.SelectTriggersToAcquire` binds the `@excludedJobTypeN` parameters (after the existing preferred-node parameters, matching the clause's position in the SQL, for providers that bind positionally).
5. Every dialect override of `GetSelectNextTriggerToAcquireSql` (`SqlServerDelegate`, `PostgreSQLDelegate`, `MySQLDelegate`, `SQLiteDelegate`, `OracleDelegate`, `FirebirdDelegate`) just wraps the same base SQL string to splice in `TOP`/`LIMIT`/`ROWS`/`rownum` paging, so threading the exclusion count through was a mechanical signature change in each — no per-dialect SQL rewrite needed, so the filter is consistent everywhere the same query construction pattern already was.

`IDriverDelegate`'s interface signature is untouched, as is the database schema (no migration needed — `JOB_CLASS_NAME` was already selected).

### Testing

- Unit tests (`StdAdoConstantsTest`) cover `BuildSqlSelectNextTriggerToAcquire` with and without exclusions.
- A new SQLite-backed integration test (`AdoJobStoreTriggerAcquisitionFilterTest`) schedules two jobs of different types against a real SQLite database, calls `AcquireNextTriggers` with one type excluded, and asserts only the non-excluded trigger comes back — verifying the filter actually excludes rows in a real query, not just that the code compiles.
- `PublicApiTest_Quartz.verified.txt` updated for the two new public properties.
- Full `Quartz.Tests.Unit` suite: 2314 passed, 4 skipped (pre-existing, unrelated to this change), 0 failed.
- `dotnet build` on `Quartz`, `Quartz.Tests.Unit`, and `Quartz.Tests.Integration` all succeed with 0 warnings/errors.

Closes #2238 (pending your review of the approach).

### DIFF
--- a/docs/documentation/quartz-4.x/how-tos/custom-job-store.md
+++ b/docs/documentation/quartz-4.x/how-tos/custom-job-store.md
@@ -257,6 +257,45 @@ recomputed, which is deliberate.
 added later will default to "no additional filtering" — an override that starts from `base` and adjusts
 one field keeps working.
 
+### Excluding job types from acquisition
+
+`ExcludedJobTypeNames` is the first of those properties, and it is how a node declines whole classes
+of work: names in the set are kept out of the acquisition query's result set, so an excluded job type
+never occupies one of the `MaxCount` rows a post-filter would have to discard.
+
+<!-- snippet: sample_custom_job_store_excluded_job_types -->
+```csharp
+// JobType.FullName is the spelling the store persists - "Namespace.TypeName, AssemblyName".
+// Type.FullName carries no assembly name and would never match a stored row.
+private static readonly string reportingJobTypeName = new JobType(typeof(ReportingJob)).FullName;
+
+protected override TriggerAcquisitionCriteria CreateAcquisitionCriteria(TriggerAcquisitionRequest request)
+{
+    // Asked again on every acquisition attempt, so a window that opens between two of them takes
+    // effect on the next one without restarting anything.
+    string[]? excluded = this.maintenanceWindow.IsOpen ? [reportingJobTypeName] : null;
+
+    return base.CreateAcquisitionCriteria(request) with { ExcludedJobTypeNames = excluded };
+}
+```
+<!-- endSnippet -->
+
+Two things to get right:
+
+- **Name the type the way the store persists it.** That is `JobType.FullName` —
+  `Namespace.TypeName, AssemblyName`, the same string `TriggerAcquireResult.JobTypeName` carries and
+  the same one the ADO schema keeps in `JOB_CLASS_NAME`. `Type.FullName` has no assembly name and will
+  never match a stored row.
+- **Matching is exact.** There is no prefix or wildcard form. The SQL comparison follows the
+  `JOB_CLASS_NAME` column's collation, so its case sensitivity is the database's, not .NET's; the
+  in-memory store compares ordinally. Rows written by Quartz 2.x or 3.x can carry an older spelling,
+  and the read side never rewrites a stored name, so an exclusion will not match those.
+
+The property is also on `TriggerAcquisitionRequest`, which every shipped store honours — set it there
+when the caller knows the exclusions, and override `CreateAcquisitionCriteria` when the *store* does.
+Entries must be non-blank and there may be at most 1000 of them, both checked at construction; 1000 is
+Oracle's ceiling on an `IN` list.
+
 ## Rebuilding jobs and triggers
 
 A store that reads its data back has to reconstruct `IJobDetail` and `IOperableTrigger`. The two are
--- a/docs/documentation/quartz-4.x/how-tos/dialect-delegate.md
+++ b/docs/documentation/quartz-4.x/how-tos/dialect-delegate.md
@@ -33,7 +33,7 @@ that happens to be `protected virtual` so the class can be composed — treat th
 
 | Member | Override when |
 |---|---|
-| `protected virtual string GetSelectNextTriggerToAcquireSql(int maxCount)` | your database limits rows differently from ANSI |
+| `protected virtual string GetSelectNextTriggerToAcquireSql(int maxCount, int excludedJobTypeBucket)` | your database limits rows differently from ANSI |
 | `protected virtual string GetSelectMisfiredTriggersToRecoverSql(int count)` | the same, for the misfire scan; `count == -1` means "no limit" |
 | `protected virtual string GetCountMisfiredTriggersInStateSql()` | the counting form needs a different shape |
 | `protected virtual string ApplyPaging(string sql, bool takeLimited)` | `OFFSET … FETCH NEXT …` is not understood |
@@ -51,8 +51,8 @@ appear among the shipped dialects:
 <!-- snippet: sample_dialect_delegate_row_limiting -->
 ```csharp
 // append (PostgreSQL, Firebird)
-protected override string GetSelectNextTriggerToAcquireSql(int maxCount)
-    => base.GetSelectNextTriggerToAcquireSql(maxCount) + " LIMIT " + maxCount;
+protected override string GetSelectNextTriggerToAcquireSql(int maxCount, int excludedJobTypeBucket)
+    => base.GetSelectNextTriggerToAcquireSql(maxCount, excludedJobTypeBucket) + " LIMIT " + maxCount;
 
 // splice a prefix (SQL Server: SELECT TOP n)
 // wrap the whole statement (Oracle: SELECT * FROM ( … ) WHERE rownum <= n)
@@ -131,8 +131,9 @@ not a contract; the schema it addresses is, and that lives in `AdoConstants`, wh
 For a delegate in your own assembly this means two things:
 
 - You cannot write `StdAdoConstants.SqlSelectNextTriggerToAcquire`. Derive your statement from what the
-  base returns — `base.GetSelectNextTriggerToAcquireSql(maxCount)` — and transform the string, which is
-  exactly what MySQL's `.Replace("{0}TRIGGERS t", …)` does. Or write the statement whole.
+  base returns — `base.GetSelectNextTriggerToAcquireSql(maxCount, excludedJobTypeBucket)` — and
+  transform the string, which is exactly what MySQL's `.Replace("{0}TRIGGERS t", …)` does. Or write
+  the statement whole.
 - You *can* name tables, columns, trigger types and state values: `AdoConstants.TableTriggers`,
   `AdoConstants.ColumnTriggerName`, `AdoConstants.StateWaiting` and the rest are public precisely so a
   dialect can build its own SQL against the schema.
--- a/docs/documentation/quartz-4.x/how-tos/dialect-delegate.md
+++ b/docs/documentation/quartz-4.x/how-tos/dialect-delegate.md
@@ -33,7 +33,7 @@ that happens to be `protected virtual` so the class can be composed — treat th
 
 | Member | Override when |
 |---|---|
-| `protected virtual string GetSelectNextTriggerToAcquireSql(int maxCount, int excludedJobTypeBucket)` | your database limits rows differently from ANSI |
+| `protected virtual string GetSelectNextTriggerToAcquireSql(TriggerAcquisitionSqlShape shape)` | your database limits rows differently from ANSI |
 | `protected virtual string GetSelectMisfiredTriggersToRecoverSql(int count)` | the same, for the misfire scan; `count == -1` means "no limit" |
 | `protected virtual string GetCountMisfiredTriggersInStateSql()` | the counting form needs a different shape |
 | `protected virtual string ApplyPaging(string sql, bool takeLimited)` | `OFFSET … FETCH NEXT …` is not understood |
@@ -44,21 +44,28 @@ that happens to be `protected virtual` so the class can be composed — treat th
 
 ### Row limiting
 
-The default `GetSelectNextTriggerToAcquireSql` ignores `maxCount` entirely — *"by default we don't
-support limits, this is db specific"* — so a dialect that can limit rows should say so. Four shapes
-appear among the shipped dialects:
+The default `GetSelectNextTriggerToAcquireSql` ignores `shape.MaxCount` entirely — *"by default we
+don't support limits, this is db specific"* — so a dialect that can limit rows should say so. Four
+shapes appear among the shipped dialects:
 
 <!-- snippet: sample_dialect_delegate_row_limiting -->
 ```csharp
 // append (PostgreSQL, Firebird)
-protected override string GetSelectNextTriggerToAcquireSql(int maxCount, int excludedJobTypeBucket)
-    => base.GetSelectNextTriggerToAcquireSql(maxCount, excludedJobTypeBucket) + " LIMIT " + maxCount;
+protected override string GetSelectNextTriggerToAcquireSql(TriggerAcquisitionSqlShape shape)
+    => base.GetSelectNextTriggerToAcquireSql(shape) + " LIMIT " + shape.MaxCount;
 
 // splice a prefix (SQL Server: SELECT TOP n)
 // wrap the whole statement (Oracle: SELECT * FROM ( … ) WHERE rownum <= n)
 // append with an index hint (MySQL: FORCE INDEX (…) … LIMIT n)
 ```
 <!-- endSnippet -->
+
+`TriggerAcquisitionSqlShape` carries everything about an acquisition attempt that changes the text of
+the statement — the row limit, and how many job-type exclusion terms the `NOT IN` clause needs. Read
+`shape.MaxCount` and hand the whole shape to `base`: a dimension added later becomes a property on
+the record, so an override written today keeps compiling and keeps applying it. It is also the key
+the finished statement is cached under, which is why it holds the bucketed exclusion count rather
+than the caller's exact one.
 
 ### Paging
 
@@ -131,9 +138,8 @@ not a contract; the schema it addresses is, and that lives in `AdoConstants`, wh
 For a delegate in your own assembly this means two things:
 
 - You cannot write `StdAdoConstants.SqlSelectNextTriggerToAcquire`. Derive your statement from what the
-  base returns — `base.GetSelectNextTriggerToAcquireSql(maxCount, excludedJobTypeBucket)` — and
-  transform the string, which is exactly what MySQL's `.Replace("{0}TRIGGERS t", …)` does. Or write
-  the statement whole.
+  base returns — `base.GetSelectNextTriggerToAcquireSql(shape)` — and transform the string, which is
+  exactly what MySQL's `.Replace("{0}TRIGGERS t", …)` does. Or write the statement whole.
 - You *can* name tables, columns, trigger types and state values: `AdoConstants.TableTriggers`,
   `AdoConstants.ColumnTriggerName`, `AdoConstants.StateWaiting` and the rest are public precisely so a
   dialect can build its own SQL against the schema.
--- a/docs/documentation/quartz-4.x/migration-guide.md
+++ b/docs/documentation/quartz-4.x/migration-guide.md
@@ -4898,11 +4898,14 @@ which always projects `EXECUTION_GROUP` and always carries the preferred-node fi
 which is the one that was already there:
 
 ```csharp
-protected virtual string GetSelectNextTriggerToAcquireSql(int maxCount, int excludedJobTypeBucket)
+protected virtual string GetSelectNextTriggerToAcquireSql(TriggerAcquisitionSqlShape shape)
 ```
 
-Its `excludedJobTypeBucket` parameter is new — pass it to `base` and the job-type exclusions ride
-along — and it is still the only thing the dialects differed in: `FirebirdDelegate` appends
+Its parameter is new: `TriggerAcquisitionSqlShape` carries everything about an acquisition attempt
+that changes the statement's text — `MaxCount`, and `ExcludedJobTypeBucket` for the job-type
+exclusion clause — so the next acquisition dimension is a property on the record rather than another
+parameter here. Read `shape.MaxCount` and pass the whole shape to `base`. It is still the only thing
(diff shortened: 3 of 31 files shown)


lahma (changes requested):
Thanks for building this — and for flagging the no-prior-confirmation risk yourself. Bad timing on our side: a [design response on the issue](https://github.com/quartznet/quartznet/issues/2238#issuecomment-5346948720) crossed with this PR by a few minutes. Most of it applies verbatim as this review's checklist, so read that first; below is what it means for this diff concretely. The overall verdict: the approach is right, the clause placement and binding order are exactly where they should be, and the SQLite store-level test was the correct instinct — we want to merge this, reshaped.

**1. Rebase onto current main — the branch predates today's #3270.** `JobStoreSupport.cs` is now `AdoJobStoreBase.cs`, `LiveNodeCutoff` became a `required DateTimeOffset`, and the semaphores were renamed, which is why the PR shows as conflicting. More importantly, #3284 just landed the seam this feature needs: `protected virtual AdoJobStoreBase.CreateAcquisitionCriteria(TriggerAcquisitionRequest)`. Rebase once it merges; criteria construction then lives in that one factory.

**2. Drop the `TriggerAcquisitionRequest` property — criteria-level only.** A store-level contract property that only the ADO store honors is a silent no-op on `RAMJobStore`, and "feature wired for one path, silently inert on another" is a failure pattern this codebase has been bitten by more than once. With #3284, the way a deployment sets exclusions is a derived store:

```csharp
protected override TriggerAcquisitionCriteria CreateAcquisitionCriteria(TriggerAcquisitionRequest request)
    => base.CreateAcquisitionCriteria(request) with { ExcludedJobTypeNames = ... };
```

If a store-level need materializes later, the request record can grow the property additively then — with RAM honoring it in the same change. This also improves your integration test: instead of reaching through `((StdScheduler)scheduler).scheduler.resources`, derive a small store with that override — which then proves the real user path end to end.

**3. Hook shape: change it once, to the criteria.** Our issue comment said "don't change `GetSelectNextTriggerToAcquireSql`'s signature" — I'm overriding that here: main is pre-release, so the signature can change, and it should change *once* rather than once per feature. `(int maxCount, int jobTypesToExcludeCount)` re-breaks on the next acquisition dimension; make it

```csharp
protected virtual string GetSelectNextTriggerToAcquireSql(TriggerAcquisitionCriteria criteria)
```

so the SQL layer mirrors the record's additive philosophy — future filters change no signatures anywhere. Dialects read `criteria.MaxCount` for their `TOP`/`LIMIT`/`ROWS`/`rownum` splice (the caller keeps clamping `MaxCount` to ≥ 1 before building).

**4. Naming: `ExcludedJobTypeNames`.** The values are the stored class-name strings; `TriggerAcquireResult.JobTypeName` set that vocabulary.

**5. The parameter names have a real defect: they must be fixed-width, and the list bucketed.** `@excludedJobType1` is a prefix of `@excludedJobType10`, and parameter-name rewriting for providers that adapt names is a plain substring replace — see the remarks above `PreferredNodeWhereClause` in `StdAdoConstants.cs` and `AdoUtil.RewriteParameterName`. Eleven or more exclusions would corrupt the statement on those providers. Use fixed-width names (`@excludedJobType000`…) and pad the list to power-of-two buckets, repeating the last real value in the pad slots (harmless for `NOT IN`); `AdoUtil.BuildJobKeyPredicate` is the in-repo template. Bucketing also keeps `ReplaceTablePrefix`'s statement cache (keyed on raw statement text) and the database's plan cache from growing one entry per distinct exclusion count.

**6. Document the matching semantics on the property.** Exact ordinal match against the stored `JOB_CLASS_NAME`, which 4.x writes as `Namespace.TypeName, AssemblyName`. Rows written by 2.x/3.x can carry older spellings and the read side deliberately never rewrites stored names, so exclusions won't match those rows — say so in the XML docs. Deliberately no `LIKE`.

**7. Tests.** Keep the SQLite test (reshaped per point 2). Add a binding-order assertion (parameters bound in SQL token order after the preferred-node ones — you got the order right; pin it) and a bucket-boundary case. The `StdAdoConstantsTest` additions follow the builder's new criteria shape. A clustered Postgres test modeled on `PreferredNodeFailoverPostgresTest.Failover_ExecutionGroupSaturatedSurvivor_DoesNotClaimTrigger` would be welcome but is not a merge blocker.

**8. Docs ride the PR.** A short section in the execution-limits / node-affinity docs family showing the derived-store override; happy to point you at the exact page once you're rebased.

The parts already right, for the record: parameterized `NOT IN` (no injection), clause between the misfire predicate and `ORDER BY` beside `{PreferredNodeWhereClause}` so every dialect splice keeps working, `jd.JOB_CLASS_NAME` reused from the existing join (no schema change), null/empty = no filtering, and a real-database test rather than a compile check. Ping me on anything that looks wrong once you're in the rebased code.

HarnageaGabriel:
Thanks for the thorough writeup, both here and on the issue — makes sense, and I see #3284 is still open. I'll hold off on the rework until it merges, then rebase onto `CreateAcquisitionCriteria` and go through the checklist: drop the request-level property in favor of the derived-store override, fixed-width bucketed parameter names (good catch on the prefix-collision bug, that's a real defect as written), the criteria-based SQL hook signature, the docs on matching semantics, and the test additions (StdAdoDelegate-level binding-order test, the reshaped SQLite test, bucket-boundary case). Will ping here once I'm back in the rebased code.

lahma:
The seam is on main now — #3284 merged as `9ca2af0e73` (plus the `JobStorePersistent` fix and a CI change, so rebase onto current `main` HEAD `16f9f86f91`). `AdoJobStoreBase.CreateAcquisitionCriteria(TriggerAcquisitionRequest)` is the factory your derived-store test can override. The review checklist above stands; shout if anything in it fights the rebased code.

lahma:
Four amendments to the review after a second design pass — three are edge cases the checklist missed, one softens a claim it overstated.

**9. Reject null/empty entries — a `NULL` poisons the whole query.** One `null` element in the exclusion list becomes `NOT IN (..., NULL)`, which evaluates UNKNOWN for **every** row: the node silently stops acquiring anything at all, which is the worst possible failure mode for a filter. Validate at bind time (or on criteria construction): throw on null/empty/whitespace entries with a message naming the property.

**10. Add a store-level backstop in `AdoJobStoreBase.AcquireNextTrigger`.** The criteria record's contract deliberately allows a custom `IDriverDelegate` to ignore a new property — but for this property, "ignored" means the node silently *runs* job types the deployment excluded. The loop already has `result.JobTypeName` in hand before it marks anything acquired; skip results whose type name is excluded (a `HashSet<string>` built once per attempt). SQL stays the efficiency path; the backstop makes the promise hold for every delegate. Non-cooperating delegates degrade to post-filter starvation instead of silent execution, which is the right trade.

**11. Cap the list size.** Oracle rejects IN-lists over 1,000 entries, and the power-of-two bucketing pads upward. Validate the list length (≤ 1,000 real entries is a natural bound) with a clear error rather than letting Oracle produce ORA-01795 at 2 a.m.

**12. Soften the "exact ordinal" doc wording from point 6.** The SQL comparison uses the column's collation — SQL Server databases are commonly case-insensitive — so the honest phrasing is "exact string equality as the database compares strings; case sensitivity follows the column's collation." (The .NET-side backstop from point 10 compares ordinally; in practice the two never diverge because 4.x writes type names exactly, but the docs shouldn't overclaim.) The pre-4.0-spellings caveat stands unchanged.

One note that's ours rather than yours: `CreateAcquisitionCriteria` overrides should not *raise* `MaxCount` above the request's — the lock-free-vs-locked acquisition decision was already made from the request before the factory runs (raising it is detected by the post-acquisition validation and retried, so it's a performance hazard, not corruption) — we'll add that sentence to the seam's docs on our side; don't worry about it here.

lahma:
Hi @HarnageaGabriel — just checking in, no pressure attached.

Since we last spoke, `4.0.0-alpha.1` shipped and the milestone moved on a fair bit; `main` has had a
lot of churn through the API-finalization and F-waves, so the rebase onto `CreateAcquisitionCriteria`
is a bigger job now than it was on the 19th.

Two things, so you can pick whichever suits you:

1. **If you're still up for it**, it's yours — say the word and I'll leave it alone. Ask me anything
   about the reshape; the nine-point checklist still stands, and I'm happy to talk through any point
   of it rather than have you guess at what I meant.
2. **If you'd rather not carry the rebase**, that's completely fine and no explanation is needed. I'll
   land the change myself with you credited in the commit and the PR, exactly as we did on #3281.

One heads-up either way: [#3335](https://github.com/quartznet/quartznet/issues/3335) (batching the
trigger fire-path SQL) is starting now, and it lives in the same SQL neighbourhood — `TriggersFired`
and the acquisition statements. I've deliberately scoped it to leave your seam alone, but whichever
of the two lands second will need a rebase over the other. I'll handle that reconciliation; you
shouldn't have to.

Thanks again for the work and for the patience — the issue has been open a long time and it's good to
see it moving.

HarnageaGabriel:
Taking option 1 — still up for it, and it's done. Rebased onto current `main` (`9c18da60`); the old commit didn't apply cleanly (files/method moved under #3284 and the F-waves), so I reimplemented fresh against the current shape rather than fight a mechanical rebase.

Against the nine-point checklist plus the four amendments:

1. Rebased onto current `main`, `AdoJobStoreBase`/`CreateAcquisitionCriteria` and all.
2. `ExcludedJobTypeNames` is threaded through `TriggerAcquisitionRequest` → `CreateAcquisitionCriteria` → `TriggerAcquisitionCriteria` — no store-level property.
3. `GetSelectNextTriggerToAcquireSql` now takes the exclusion bucket alongside `maxCount`; all six dialect delegates call `base.GetSelectNextTriggerToAcquireSql(maxCount, excludedJobTypeBucket)` so the clause rides along instead of being duplicated per dialect. The `acquisitionSqlByMaxCount` plan-text cache (new since #3329) is now keyed on `(maxCount, excludedJobTypeBucket)`.
4. Property renamed to `ExcludedJobTypeNames` throughout.
5. Fixed the parameter-naming defect: names are zero-padded fixed-width (`excludedJobType0000`…) and bucketed the same way `AdoUtil`'s job-key predicates are, so no name is ever a string-prefix of another.
6/12. Doc comment states comparison follows the `JOB_CLASS_NAME` column's collation, not guaranteed ordinal; the pre-4.0-spellings caveat carried over.
7. Tests: a `StdAdoDelegateTest` binding-order test (asserts exact parameter order and last-name padding at a bucket boundary), `StdAdoConstantsTest` bucket-boundary and zero-exclusion cases, and a `SQLiteJobStoreContractTest` acquisition-level test.
8. n/a (delegate signature unchanged, already took the whole criteria object).
9. Null/empty/whitespace entries throw at `TriggerAcquisitionRequest` construction.
10. Ordinal `.NET`-side backstop added in `AcquireNextTrigger` — a delegate that ignores the SQL-level filter still gets its result set post-filtered.
11. Capped at 1000 entries (Oracle's IN-list ceiling), validated at construction.
12. Softened per above.

Also re-approved the `PublicApiTest` baseline and added the migration-guide row.

Verified locally (couldn't get Docker for the full integration matrix here, so):
```
dotnet build src/Quartz/Quartz.csproj                                        # 0 errors, 0 warnings
dotnet build src/Quartz.Tests.Unit/Quartz.Tests.Unit.csproj                   # 0 errors, 0 warnings
dotnet test  src/Quartz.Tests.Unit/Quartz.Tests.Unit.csproj                   # 3103 passed, 0 failed, 4 skipped
dotnet build src/Quartz.Tests.Integration/Quartz.Tests.Integration.csproj    # 0 errors, 0 warnings
dotnet test  src/Quartz.Tests.Integration/... --filter SQLiteJobStoreContractTest   # 65 passed, 0 failed
```

Force-pushed the rebase (`fd641fbb` → `0e219f74`) — single commit on top of current `main`, `gh pr view` reports `MERGEABLE`.

Re #3335: nothing here should conflict — my SQL edits are additive (`BuildSqlSelectNextTriggerToAcquire`/bucket cache) rather than restructuring `TriggersFired` or the base acquisition statement shape, but happy to rebase again once it lands if it touches the same lines.

lahma (changes requested):
Thank you — this is a faithful rebuild, and what the 19th asked for is here: the fixed-width bucketed parameters, the ordinal backstop, the cap with construction-time validation, the binding-order test, and the acquisition-level `SQLiteJobStoreContractTest`. Keying the plan-text cache on the bucket was the right call too; that cache did not exist when the checklist was written. I read the whole diff against `9c18da60` and your "no conflict with #3335" holds — #3377 is in that base and the PR is `MERGEABLE`. The CI runs are approved now.

Three things before merge. Two are checklist points that came out differently from what was asked.

**1. Point 2 — `TriggerAcquisitionRequest.ExcludedJobTypeNames` is honoured by the ADO store and silently ignored by `RAMJobStore`.** That is the wired-for-one-path shape the point was about. Now that the property exists at the request level, the cleaner resolution is not to drop it but to make it true for every store: in `RAMJobStore.AcquireNextTriggers`, skip a candidate whose `jobDetail.JobType.FullName` is in the set (the same spelling `TriggerAcquireResult.JobTypeName` carries for a type-constructed job, so one comparison rule covers both stores), and move `AcquisitionExcludesRequestedJobTypes` from `SQLiteJobStoreContractTest` up into the shared `JobStoreContractTest`, so both stores are pinned by one case. The property's remarks can then say that every shipped store honours it.

**2. Point 3 — the hook signature.** `GetSelectNextTriggerToAcquireSql(int maxCount, int excludedJobTypeBucket)` is the shape the review said would re-break on the next acquisition dimension. Make the parameter the thing the cache is keyed on:

```csharp
public readonly record struct TriggerAcquisitionSqlShape(int MaxCount, int ExcludedJobTypeBucket);

protected virtual string GetSelectNextTriggerToAcquireSql(TriggerAcquisitionSqlShape shape)
```

Dialects read `shape.MaxCount` for their `TOP` / `LIMIT` / `ROWS` / `rownum` splice, the `ConcurrentDictionary` is keyed on the record (rename `acquisitionSqlByMaxCount` to `acquisitionSqlByShape` while there), and the next filter adds a property instead of a parameter. Same migration-guide row, different right-hand side.

**3. `StdAdoConstants.BuildSqlSelectNextTriggerToAcquire` is a second copy of the acquisition SELECT** — its own comment says it is kept in sync by eye. Build both from one template with an exclusion-clause slot, the way `{PreferredNodeWhereClause}` already is, so `SqlSelectNextTriggerToAcquire` is the zero-bucket case of the builder and `…_WithNoExclusions_ShouldKeepExistingSqlExactly` holds by construction rather than by vigilance.

Point 8 is still open: a short section showing the derived-store override — `CreateAcquisitionCriteria(request) with { ExcludedJobTypeNames = … }` — belongs on `docs/documentation/quartz-4.x/how-tos/custom-job-store.md`, the SPI how-to hub. A hand-written fence is fine there; that page is not on the compiled-snippet convention yet.

Not blocking: the clustered Postgres case from point 7 is still welcome, though the two-store contract test covers most of what it would prove.

If you are short on time for another pass, say so and I will land these three on top of your commit with your authorship kept, as we did for #3281. Either way this is going into alpha.2.

lahma:
@HarnageaGabriel — thank you for this. Alpha.2 is being tagged and I wanted the filter in it, so I finished the round myself rather than wait: your commit is rebased onto current `main` exactly as you wrote it (author intact — `main` took about thirty merges under it, and the only textual conflict was #3395's `JobTypeInformation` line next to your backstop), and the three asks from the last round are commits on top of it:

- `RAMJobStore` honours `ExcludedJobTypeNames` too (ordinal on `JobType.FullName`, the same string `JOB_CLASS_NAME` holds), and the acquisition test moved into the shared `JobStoreContractTest` so both stores are pinned by one case.
- `GetSelectNextTriggerToAcquireSql` takes a `TriggerAcquisitionSqlShape(MaxCount, ExcludedJobTypeBucket)` record — the plan-text cache key itself — so the next acquisition dimension is a property, not a parameter through seven dialects.
- The acquisition SELECT exists once, with a slot for the exclusion clause; a snapshot recorded *before* the consolidation proves the no-exclusion statement is byte-identical for every dialect.
- The custom-job-store how-to shows a store excluding a job type during a maintenance window, and the same blank/oversized-list validation you put on the request now also guards `TriggerAcquisitionCriteria`, which is the path a derived store actually uses.

Merging on green. The bucketed fixed-width parameters, the ordinal backstop and the construction-time validation were all better than the checklist asked for — much appreciated. One thing left for a later PR, not yours to worry about: the .NET-side backstop runs after `GetTrigger`, so a custom delegate that ignores the SQL filter pays a round trip per excluded candidate; hoisting it to `result.JobTypeName` is a small follow-up.

lahma (approved):
The three asks and the docs section are in (commits on top of the contributor's own); integration suite green on all six dialects.

PR #3295, opened 2026-08-21
A blank calendar name is no calendar name (3.x)

### LINKED ISSUE #3294: Quartz.Dashboard sends empty string instead of null for CalendarName when rescheduling triggers
### Pre-flight checks

- [x] I have searched [open](https://github.com/quartznet/quartznet/issues) and [closed](https://github.com/quartznet/quartznet/issues?q=is%3Aissue+is%3Aclosed) issues and could not find a duplicate.
- [x] I have checked the [documentation](https://www.quartz-scheduler.net/documentation/index.html).
- [x] I can reproduce this against a recent release of Quartz.NET (or the latest pre-release).

### Describe the bug

Updating cron trigger schedule from the dashboard sets `calendar_name` to an empty string instead of null.
This causes Quartz.Impl.AdoJobStore.SqlServerDelegate warning `Couldn't find calendar with name '' ` and the job doesn't run.
If `Description` wasn't set when creating the trigger then it also changes to an empty string, though it doesn't cause any issues.  

Updating cron trigger that already has a calendar works as expected.

### Steps to reproduce

1. Create a job with cron trigger with no calendar and no description
2. Chnage trigger schedule via dashboard

### Expected behavior

`[CALENDAR_NAME]` and `[DESCRIPTION]` should remain unchnaged when changing cron schedule

### Actual behavior

 `[CALENDAR_NAME]` and `[DESCRIPTION]` are set to empty strings

### Code sample

```csharp

```

### Exception / stack trace

```text

```

### Quartz configuration

```csharp

```

### Quartz.NET version

3.19.1

### .NET version

net8.0

### Job store

AdoJobStore — SQL Server

### Serializer

Newtonsoft.Json (Quartz.Serialization.Newtonsoft)

### Clustering

No — single scheduler node

### Hosting / integration

- [ ] Quartz.Extensions.Hosting (IHostedService)
- [ ] Quartz.Extensions.DependencyInjection
- [x] Quartz.AspNetCore (health checks, dashboard)
- [ ] Plain console / manually managed scheduler

### Operating system

Windows

### Regression?

_No response_

### Additional context

_No response_

lahma:
Fixes #3294. Companion to the `main` PR — see the link below.

## The bug

Editing a cron trigger's schedule from the dashboard wrote `CALENDAR_NAME = ''` instead of `NULL`, and the trigger then **stopped firing entirely**. Every job store gates its calendar lookup on `CalendarName is not null`, so the empty string passed the gate, the lookup found nothing, and the fire was silently dropped. The ADO store left a single `Couldn't find calendar with name ''` line from the delegate — which never says a fire was skipped — and `RAMJobStore` left nothing at all. Oracle hides it entirely, since `''` is `NULL` there, so it looks provider-specific.

Two defects compound.

**The dashboard rebuilt the trigger by hand.** `TriggerDetail.razor` re-assembled the whole trigger field by field out of its *display* projection, using `DisplayValueHelper.GetString` — which flattens both "member absent" and "JSON null" to `""`. That is right for rendering and wrong for a payload. The same hand-assembly silently dropped whatever it forgot to list (the node pin) and hardcoded `triggerType = "CronTrigger"`, so a custom `ICronTrigger` came back as a plain one.

**The core treated `""` as a real calendar name.** `AbstractTrigger.CalendarName` was a bare auto-property. The tell that this was an oversight rather than a decision: `ExecutionGroup`, a property on the *same class*, has always normalized blanks to `null` — which is exactly why `executionGroup = ""` from that same payload was harmless while `calendarName = ""` was fatal. The XML and JSON front ends already normalize calendar names too; only the object/JSON API did not, and that is also the HTTP API's path, so `POST .../reschedule` with `"calendarName": ""` reproduces this with no dashboard involved.

## The fix

**`AbstractTrigger.CalendarName` stores a blank name as `null`**, without trimming — the name is a lookup key against whatever `AddCalendar` stored, so trimming `" holidays "` would break a calendar genuinely registered with padding, creating the same bug from the other direction. That one setter is the choke point for `TriggerBuilder`, both JSON converters and the ADO read-back, which means **databases already holding `''` self-heal**: the row rehydrates as "no calendar", the trigger fires again on the next acquisition, and the column is written back as `NULL` next time it is persisted. No migration script.

It uses the `field` keyword rather than a hand-named backing field on purpose. `AbstractTrigger` is `[Serializable]`, `BinaryObjectSerializer` still ships here, and `BinaryFormatter` matches stream members *by field name* — a `private string? calendarName` would rename `<CalendarName>k__BackingField` and silently drop the value out of existing `BLOB_TRIGGERS` payloads. Verified against the built assembly, not just assumed.

**`TriggerDetailsUpdate.WithCalendarName` normalizes separately.** Both stores check the calendar exists *before* the value ever reaches the trigger setter, so `WithCalendarName("")` used to throw `Calendar '' does not exist.` where `WithCalendarName(null)` cleared.

**The dashboard now edits the trigger JSON it was already handed** instead of rebuilding it. A new internal `TriggerPayloadBuilder` copies the serialized trigger and rewrites only `cronExpressionString` — faithful by construction, immune to this whole bug class, and it preserves the node pin and the real trigger type for free. It also nulls `nextFireTimeUtc`, because that value belongs to the expression being replaced and `RescheduleJob` honours a non-null one verbatim; carrying it over would have made the new schedule skip a beat.

**Both stores now log when they skip a fire over a missing calendar.** This is the highest debugging-value line in the change — it turns this class of report from a mystery into a one-line diagnosis.

## Deliberately not here

`RescheduleJob` still accepts a calendar it cannot find. 4.x tightens that to match `ScheduleJob`, but on a released branch it would turn a long-standing silent success into a throw, and Java Quartz carries the same `ScheduleJob`/`RescheduleJob` asymmetry — inherited, not a .NET slip. With the normalization and the new warnings in place, #3294 is fixed here without it. `QuartzScheduler.cs` is untouched on this branch.

`Description` is not normalized either. It has no lookup semantics and `""` is inert; the empty descriptions the issue reports came from the dashboard, and that is fixed at the source.

## Testing

The new tests were checked against the *unfixed* core, not just against the fix — with the setter reverted, the builder test, the JSON round-trip and the `RAMJobStore` firing test all fail.

- `TriggerBuilderTest` — blank names normalize through the builder, through direct assignment, and through `GetTriggerBuilder()`; `" holidays "` is preserved verbatim, pinning the no-trim decision.
- `UpdateTriggerDetailsTest.CalendarName_BlankClearsCalendarRatherThanFailingTheUpdate` — beside the existing `CalendarName_ValidatesExistence`, which must still throw for a real-but-missing name. The pair is the contract.
- `JsonObjectSerializerTest.DeserializeBlankCalendarNameAsNoCalendarAtAll` — a raw `"CalendarName": ""` payload, run through both serializers, which is what is sitting in `BLOB_TRIGGERS` today.
- `RAMJobStoreTest.TriggersFired_BlankCalendarName_StillProducesABundle` — the test that would actually have caught the reported symptom, and the positive twin of the existing missing-calendar test.
- `TriggerPayloadBuilderTest` (new) and `InProcessQuartzApiClientTest.RescheduleFromTriggerDetailPayloadChangesNothingButTheSchedule` — the dashboard round trip end to end, asserting the calendar stays null, the node pin survives, and the trigger still has a next fire time.

The existing `RescheduleJobShouldAcceptCronTriggerPayload` is why nobody caught this: its comment said the payload "mirrors TriggerDetail.razor" and it then hand-copied the builder with `calendarName = (string?) null` — asserting the shape the bug never produces. It stays, re-labelled as what it actually is (a pin on the shape the API accepts), and the real path is covered by the new test.

Green locally: full solution build with 0 warnings, unit tests 1761 (net10.0) + 1750 (net472), AspNetCore tests 122. No Verify baseline moves — `PublicApiGenerator` prints `{ get; set; }` whether or not a property has a backing field.

### DIFF
--- a/src/Quartz.Dashboard/Components/Pages/TriggerDetail.razor
+++ b/src/Quartz.Dashboard/Components/Pages/TriggerDetail.razor
@@ -89,10 +89,6 @@
                OnCancel="CancelUnscheduleAsync" />
 
 @code {
-    private static readonly JsonSerializerOptions requestSerializerOptions = new()
-    {
-        PropertyNamingPolicy = JsonNamingPolicy.CamelCase
-    };
     private bool _isLoading = true;
     private string? _error;
     private object? _trigger;
@@ -236,7 +232,13 @@
             return;
         }
 
-        JsonElement newTrigger = BuildRescheduleTriggerPayload(cronExpression);
+        if (_trigger is not JsonElement triggerJson
+            || !TriggerPayloadBuilder.TryWithCronExpression(triggerJson, cronExpression, out JsonElement newTrigger))
+        {
+            Toasts.Error("This trigger cannot be rescheduled from the dashboard.");
+            return;
+        }
+
         RescheduleRequest request = new(newTrigger);
         _ = await ExecuteMutatingActionAsync(
             () => Api.RescheduleJob(schedulerName, Group, Name, request),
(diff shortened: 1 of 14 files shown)


lahma:
4.x companion: #3296. Land this one first — the 4.x branch is a straight port plus one extra: `RescheduleJob` rejects a calendar it cannot find there, which is deliberately not done on a released branch.

PR #3318, opened 2026-08-22
F5: one wire contract, and its enums have names

lahma:
Feature F5 from the ratified finalization spec — the wire-contract consolidation that F1 builds on. Four commits: the shape-preserving move first (proven — zero snapshot changes before the declared phases), then three declared wire changes.

## The packaging inversion, fixed

`Quartz.AspNetCore` had a ProjectReference on `Quartz.HttpClient` — the server package shipped a NuGet dependency on the client package to read a contract neither owns, through an IVT grant. The contract (`Quartz.HttpApiContract`) now lives internal in core, both ends read it there (existing IVT grants sufficed), the ProjectReference and the grant are gone, and one `HttpApiJson.ConfigureWireFormat` defines the JSON contract for both sides. The dashboard keeps its ratified public projection; the OpenAPI documentation twins stay (they exist so the published schema is useful — deliberate, commented).

## Declared wire changes (old→new tables in the guide; snapshots updated one by one)

1. **Enums travel as names**: `"status": 1` → `"status": "Running"`, trigger states `"Paused"`/`"Blocked"` — per-type `JsonStringEnumConverter<T>` registrations, never a blanket converter (the API adds to the application's shared JsonOptions, and a blanket one would silently re-render the host's own endpoints; per-type is also the trim-safe form). Reading still accepts numbers.
2. **One name per value**: the listing header's `jobTypeName` → `jobType` (value identical to the detail body's).
3. **The dashboard's client speaks one currency (A8-15)**: the three bespoke page DTOs → `PagedQuery`-style Skip/Take + HasMore/TotalCount records; stringly states → `TriggerState` and a new public `SchedulerStatus` enum; fifteen (group, name) string-triplet members → the key DTOs the package already defined. A real bug fell out: the two `IQuartzApiClient` implementations disagreed on a running scheduler — `"Started"` vs `"Running"` — which one shared `IScheduler.GetStatus()` helper now makes impossible.

Snapshot ledger: nine of twelve byte-identical, three changed for the declared reasons, two added (trigger-header state and `/state` body had no pin at all), two status-code rows added documenting the 400 shapes as-is. Docs: the http-api page gains "Enums travel as names" and "Response-shape conventions".

Core baseline +7 (`SchedulerStatus`); dashboard baseline reshaped per A8-15; AspNetCore baseline unchanged.

Unit 2658, AspNetCore 261 (twice, snapshot-stable), basic-leg integration 108, `Compile UnitTest PublishAot` clean — re-verified after rebasing over PR-6.


lahma:
Merging with the Sonar quality gate red on `new_coverage` (52.4% vs 80): the condition is measuring the file MOVE — relocating the internal wire-contract DTOs to core resets their new-code attribution, so their data-record lines count as uncovered new code. No new bugs or vulnerabilities; the wire itself is pinned byte-for-byte by the snapshot suite (nine of twelve unchanged, the rest declared). Same override class as #3267.

PR #3322, opened 2026-08-22
The migration chain finally runs, on two upgrade routes

lahma:
Test-hardening PR 8: the migration scripts have never been EXECUTED by any test on main (VerifyMigrations is a text diff). This ports 3.x's MigrationScriptTest and extends it.

- **Ported**: prefix isolation (migrated schema under `QRTZM_` beside fresh `QRTZ_` in one database — which also proves the replace-the-prefix instruction), six-dialect information_schema diff of tables/columns/indexes, per-dialect batch splitters, per-dialect `db-*` categories.
- **Added — the 4.0 chain** ending at a fresh 4.0 install diff, on TWO routes: stepped (3.16 → 3.17–3.20 → 4.0) and direct (3.16 → 4.0 alone — the database that skipped the optional 3.x migrations; the 4.0 script claims to serve it, nothing ever tested it). Idempotence: every script applied twice (SQLite exception honored).
- **Added — a smoke workload on the migrated store**: one trigger per persisted family incl. recurrence, an execution group, a `PreferredNode` pin — fired, read back, clean shutdown, with row counts proving all six landed in the migrated prefix and schema validation on. This closes the gap 3.x's own remarks admitted.
- **Vendored baselines** (`SchemaBaselines/3.16/`) hash-verified identical to tags v3.16.0/v3.16.1 and to 3.x's copies, with provenance README (CI checkouts are shallow — no runtime `git show`).

**Defect found, deliberately not fixed here** (generated script, byte-identical-across-branches rule): `schema_30_to_40_upgrade_sqlite.sql`'s header claims every statement checks first and the script is re-runnable — its five `ADD COLUMN`s are unguarded and fail on a 3.17+ database with `duplicate column name`. The 3.17–3.19 SQLite files carry the `-- NOT IDEMPOTENT` note this one lacks. Follow-up: fix the header in `Build.DatabaseMigrations.Scripts.cs` + regenerate + 3.x companion PR. The test does the existence check on the script's behalf, documented — an unguarded `CREATE INDEX` would still fail the second pass.

Locally verified: SQLite legs twice, basic leg unchanged at 108, full build + AOT clean, and a deliberate assertion break proving the workload actually executes. The five container dialects are compile-verified locally and proven by this PR's own `db-*` CI legs.


lahma:
Merging with the automatic-analysis gate red: all 59 issues are Sonar's PL/SQL rules critiquing the VENDORED v3.16.1 schema baselines (VARCHAR2 advice against Firebird DDL) plus duplication across six dialects of one schema — files that must stay byte-identical to the tag or the migration chain loses its provenance. #3324 excludes the path in the incoming CI-based scanner. The real verification is this PR's own db-* container legs, all green.

PR #3323, opened 2026-08-22
F3+F7: a set of keys in one call, and calendars filter by name

lahma:
Features F3 and F7 from the ratified finalization spec, with the V6 notification amendments.

## F3 — bulk pause/resume/reset

Five key-set overloads on `IScheduler`/`IJobStore`, each returning the keys the operation actually APPLIED to (order-preserving; missing/ineligible keys are absent, never a throw). RAM coalesces into one lock pass; ADO runs the set inside one `ExecuteInLock` scope reusing the existing per-key statements — no new dialect SQL. Events fire per applied key (never a `TriggersPaused(null)` mass event — the verified false-outage hazard).

**Two spec deviations, both where V6's wording met the code:** the stores never called `SignalSchedulingChange` on any pause/resume path — the observable signal lives in `QuartzScheduler.NotifySchedulerThread` — so "exactly one signal per call" is implemented there and PROVEN where it is real: a counting `DelegatingJobStore` shows one bulk store call and zero single-key calls, and a reflection test proves neither shipped store was left on the looping DIM. And `ResetTriggersFromErrorState` signals and raises nothing, exactly like its singular — the plural-of-the-singular rule beat the literal wording. Both documented.

Five new endpoints under a `keys/` segment with internal contract DTOs (on F5's consolidated source), three new wire snapshots plus two status-code rows.

## F7 — CalendarQuery.Name

`CalendarNameMatcher` (calendars have no key type, so `NameMatcher<TKey>` could not serve; same four factories and wire spellings). RAM filters under its lock; ADO gains name predicates on the page select AND both count selects so totals match pages — including the `!`-escape literal-wildcard discipline, exercised end-to-end by a `50%off`/`50%o_f` contract pair on both stores. One seam reshape with a dialect-author migration row: `ToSqlLikeClause`'s `virtual` moved to a non-generic `(StringOperator, string)` overload so calendar and key matchers escape identically.

Docs: new guide section with the member table and the one-signal rationale; the http-api page's filter table also gains the `name*` rows it had omitted for jobs/triggers/groups.

Unit 2702, AspNetCore 271 (twice, snapshots stable), basic leg 120 + SQLite contract fixture 47, `Compile UnitTest PublishAot` clean. Container dialects proven by this PR's CI legs.


lahma:
Merging with the gate red on new_coverage (66.1% vs 80): zero bugs or vulnerabilities; the shortfall is the forwarder surface (five members × nine implementors + wire DTOs). The feature itself carries six dual-store contract tests, five endpoint tests, and three wire snapshots; all 14 functional legs green.

PR #3329, opened 2026-08-22
F1: every firing in the cluster, one paged query

### LINKED ISSUE #3205: Cluster-aware query for currently executing fire instances

lahma:
Feature F1 (+rider B1) from the ratified finalization spec — the last API change before the 4.0 alpha. Closes #3205. Credit where due: #3281 by @HarnageaGabriel explored this ground first; its multiplicity insight (concurrent executions of one trigger must never collapse) is pinned here by equivalent tests, per the review's follow-through.

## The model

`FireInstance` — `FireInstanceId`, `TriggerKey`, `JobKey?` (null while merely Acquired), `SchedulerInstanceId`, `FireInstanceState { Acquired, Executing }`, `FireTimeUtc`, `ScheduledFireTimeUtc?`, `ExecutionGroup?`. Queried via `QueryFireInstances(FireInstanceQuery)` — paged, filterable by trigger/job/node/state, `State` defaulting to `Executing` so the plain call means what `GetCurrentlyExecutingJobs` used to. **`GetCurrentlyExecutingJobs` is removed** (ratified): interrupt never becomes a store round-trip (the in-process path stays, and `InterruptFireInstance` is now an O(1) dictionary hit), and the live-context use case gets a documented context-tracking listener sample.

## The stores

ADO: new `SelectFireInstances` with its own paged SQL (`FiredTriggerQuery` deliberately untouched — ClusterRecover reads every row); `EXECUTION_GROUP` finally goes live on the fired-trigger insert/update (column shipped in 3.18, written by nothing until now). RAM: the #1416 probe dictionary is untouched; values widen to per-fire-instance entries; Acquired rows come from wrapper state. `IJobStore.Initialize` takes a `SchedulerIdentity` record — every store learns its identity the same way at the same moment, and `DefaultSchedulerFactory`'s `is AdoJobStoreBase` special case is deleted.

## The wire and the dashboard

`…/jobs/currently-executing` → paged `…/jobs/fire-instances` (declared breaking change, old→new table in the guide; enum-as-name per the F5 convention; `?state=Any` widens past the default). `SchedulerMetadata.LocalExecutingJobs` (node-local, named as such). The dashboard's executing page is strongly typed, shows the node, and **interrupts the row, not the job** — fixing interrupt-all-executions.

## B1 — group-partitioned tenants get quotas

`ExecutionLimitsBuilder.UseTriggerGroupWhenUnset()`: a trigger with no explicit `ExecutionGroup` is treated at limit-evaluation time as belonging to its trigger group. Design finding: ADO's limit enforcement was ALWAYS an in-memory post-filter in the acquisition reader (the SQL only projects the column), so derivation needed no dialect work — and `execution-groups.md`'s claim that filtering happens in SQL is corrected. `ExecutionSlots.TryTake` takes the trigger group as a REQUIRED argument so no store can silently opt out. Persisted `EXECUTION_GROUP` stays exactly what the trigger carries; explicit always wins.

Docs caveats each get a paragraph (vetoed fires never appear; elapsed time is observer-clock arithmetic, clamp at zero; `ScheduledFireTimeUtc` reflects the post-misfire schedule). `ExecutingTriggerStateClusteredPostgresTest` rewritten into a positive cross-node assertion.

Unit 2861, AspNetCore 275 (twice, snapshots stable), basic-leg integration 136, `Compile UnitTest PublishAot` clean; the clustered Postgres rewrite is proven by this PR's CI leg.


lahma:
Merging with the gate red on new_coverage (59.7% vs 80): zero bugs or vulnerabilities; the uncovered mass is dashboard component rework and DTO/forwarder plumbing, while the feature core carries dual-store contract tests, delegate SQL-shape tests, endpoint + snapshot tests, and slot-derivation tests. Same override class as #3318/#3323; all 15 functional legs green including the rewritten clustered Postgres test.

PR #3332, opened 2026-08-22
The ADO layer stops paying reflection and string taxes on every statement

lahma:
The internal ADO performance pass from the ratified backlog — zero public API movement (baselines byte-identical, verified).

**The headline (F9)**: parameter-prefix rewriting moved from per-parameter full-text copies into the cached statement path — one rewrite per statement ever, reference-keyed so the delegate's cached texts hit without hashing a kilobyte of SQL. Benchmarked on the real acquisition statement: **2,902→1,071 ns and 14.51→1.01 KB per bind cycle** on `:`-prefix drivers (Npgsql/Oracle); `@`-prefix drivers unchanged. The token-scan rewrite also retires the `@p1`-inside-`@p10` substring hazard.

**The reliability find (F27)**: `DbException.IsTransient` has existed since .NET 6, so the reflection sniff matched EVERY driver exception and short-circuited — `SqlException` does not override it (verified against the shipped assembly), so the Azure SQL error-number list including deadlock 1205 was dead code for any unwrapped `SqlException`, and SQLite's busy/locked retry never ran. The verdict is now inclusive over the inner-exception chain (driver flag, then SQL Server numbers, then SQLite codes, then timeouts), unit-covered, with the four defensive empty catches gone.

Also: `DbMetadata` derivations memoized (ConditionalWeakTable — memo FIELDS would break record equality); per-count memos for the dialect row-limited SQL (signatures untouched); `ConstructorInvoker`/`MethodInvoker` replace raw reflection per command; `ClearData`'s nine leaked commands become one batch; cached ordinals on the three hot row readers (keeping `Convert.ToInt32` — Oracle NUMBERs arrive as decimal); the acquisition reader stops on the FIRST surplus row instead of two late (and `Cancel()` stays — disposing would drain every waiting trigger off the wire on non-row-limited dialects); the ten driver descriptions are now C# literals and both `.properties` files are gone (the 4.1 AOT track's enabler); TPD lookup is a copy-on-write FrozenDictionary honoring legal late registration.

Honest note: the basic integration leg caught two regressions the unit suite missed (an ordinal taken on a statement that doesn't project it; an over-narrowed exception-shape check) — both fixed in-branch, which is exactly why that leg exists.

Unit 2881, AspNetCore 275, basic-leg 136, `Compile UnitTest PublishAot` clean.


lahma:
Merging with the gate red on new_coverage (56.4%) and duplication (4.3%): zero bugs or vulnerabilities. The duplication is the ten driver descriptions as deliberately-parallel C# literals (replacing the runtime-parsed .properties file); the coverage shortfall is internal plumbing whose hot paths are exercised by the container legs — which caught and fixed two real regressions in this very branch. Same override class as #3318/#3323/#3329.

PR #3345, opened 2026-08-22
A release tag is read off the ref and parsed as semver

lahma:
The v4.0.0-alpha.1 publish run skipped Publish (`Tagged build: False`). Root cause: the tag was ANNOTATED — its ref points at the tag object — and Fallout's `GetTagsFromCommit` matches ref contents against the commit SHA, so only lightweight tags were ever visible (every historical release tag is one; this tag was the deviation).

Two-layer fix in our Build.cs, per review:

1. **The ref is the authority on Actions**: `GITHUB_REF_TYPE=tag` → `GITHUB_REF_NAME` carries the name for annotated and lightweight tags alike; `GitRepository.Tags` remains the local fallback.
2. **Real semver parsing** via the [Semver](https://www.nuget.org/packages/Semver) package (3.0.0, strict styles) instead of splitting on the first hyphen: prefix = Major.Minor.Patch, suffix = the prerelease identifiers — and a v-tag that is not valid semver **fails the build loudly** ("fix the tag rather than the build") instead of packing whatever followed the v. Also retires the `SingleOrDefault` throw-on-two-tags.

Verified empirically on all three paths: `GITHUB_REF_NAME=v4.0.0-alpha.1` → Tagged build True, prefix 4.0.0; `vNext` → the loud failure; untagged → preview as before.

After merge: the annotated tag gets replaced by a lightweight one on the fixed head, which reruns publish.

### DIFF
--- a/build/Build.cs
+++ b/build/Build.cs
@@ -4,6 +4,8 @@ using System.Linq;
 using System.Runtime.InteropServices;
 using System.Text.RegularExpressions;
 
+using Semver;
+
 using Fallout.Common;
 using Fallout.Common.CI;
 using Fallout.Common.CI.GitHubActions;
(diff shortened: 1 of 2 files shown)


lahma:
Amended per review discussion: the ref reading now goes through Fallout's typed host (`GitHubActions.Instance is { RefType: "tag" } actions ? actions.RefName : …`) rather than raw environment variables — the host class already exposed RefName/RefType and our Build.cs simply never consulted them. Verified end-to-end under `GITHUB_ACTIONS=true`: `Version: 4.0.0-alpha.1`, `Tagged build: True`; invalid tag still fails loudly; local untagged unchanged. The annotated-tag blind spot in `GitRepository.Tags` itself (peeled refs never dereferenced) is being filed upstream on lahma/Fallout.

Issue #1696, opened 2022-08-09
FIRED_TRIGGERS state EXECUTING


ZLVincent:
**Describe the bug**

When the scheduler executing a longrunning job, I deleted the job halfway through. I have noticed that occasionally there is no deletion in the FIRED_TRIGGERS table, and its STATE is EXECUTING. The next time I try to create a job with the same name, the job starts but never executes.

**Version used**

3.4.0

**Expected behavior**

I don't know what's going on,But I understand that I have removed the job and trigger and they should be removed from FIRED_TRIGGERS as well.

`                           JobKey key = JobKey.Create($"{jobKeyPrefix}{jobId}", typeof(TJobType).Name);
                            if (!_scheduler.IsShutdown && await _scheduler.CheckExists(key))
                            {
                                var triggerKey = new TriggerKey($"Trigger_{typeof(TJobType).Name}_{jobId}");
                                var trigger = await _scheduler.GetTrigger(triggerKey);
                                if (trigger is not null)
                                {
                                    await _scheduler.PauseTrigger(triggerKey);
                                    await _scheduler.UnscheduleJob(triggerKey);
                                }
                                await _scheduler.DeleteJob(key);
                            }`

PR #3376, opened 2026-08-23
The key-set delete and unschedule answer with the keys they removed

### LINKED ISSUE #3360: Key-set DeleteJobs and UnscheduleJobs answer a lossy bool
`IScheduler.DeleteJobs` and `IScheduler.UnscheduleJobs` take a key set and answer with a single
`bool` meaning "every key given was found". That answer is lossy in the one case a caller most needs
to know about: when three of five keys exist, **three jobs are deleted** and the call returns `false`.
Nothing in the answer distinguishes that from "nothing happened", so a caller who retries on `false`
does not learn that most of the work already succeeded.

```csharp
ValueTask<bool> DeleteJobs(IReadOnlyCollection<JobKey> jobKeys, CancellationToken cancellationToken = default);
ValueTask<bool> UnscheduleJobs(IReadOnlyCollection<TriggerKey> triggerKeys, CancellationToken cancellationToken = default);
```

## The shape that already exists

The key-set pause, resume and error-state reset added in 4.0 answer with **the keys they applied to**,
which is exactly the information the delete and unschedule forms throw away:

```csharp
ValueTask<List<JobKey>>     PauseJobs(IReadOnlyCollection<JobKey> jobKeys, …);
ValueTask<List<JobKey>>     ResumeJobs(IReadOnlyCollection<JobKey> jobKeys, …);
ValueTask<List<TriggerKey>> PauseTriggers(IReadOnlyCollection<TriggerKey> triggerKeys, …);
ValueTask<List<TriggerKey>> ResumeTriggers(IReadOnlyCollection<TriggerKey> triggerKeys, …);
ValueTask<List<TriggerKey>> ResetTriggersFromErrorState(IReadOnlyCollection<TriggerKey> triggerKeys, …);
```

A key the operation did not apply to is absent from the list, never an exception, and the order is
the order the keys were given in. Making the two key-set mutations that predate that convention follow
it would leave one rule for the whole family:

```csharp
ValueTask<List<JobKey>>     DeleteJobs(IReadOnlyCollection<JobKey> jobKeys, …);
ValueTask<List<TriggerKey>> UnscheduleJobs(IReadOnlyCollection<TriggerKey> triggerKeys, …);
```

`bool allFound = result.Count == jobKeys.Count` recovers today's answer from tomorrow's, so nothing is
lost.

## What it touches

- **`IScheduler`** — two members change return type. Callers that only test the `bool` need rewriting;
  the compiler finds all of them.
- **`IJobStore`** — the same two members, on both shipped stores. `RAMJobStore` already knows which
  keys it removed. `AdoJobStoreBase` deletes in one pass inside a `SchedulerLock.TriggerAccess` scope
  and would have to collect the keys it actually removed rather than counting them; note that
  `DeleteJobs`' current ADO implementation may take a bulk path that does not report per-key results,
  so this is the part to design first.
- **`IDriverDelegate`** — possibly, depending on how the ADO store gets per-key results back.
- **The HTTP API** — `POST …/jobs/delete` and `POST …/triggers/unschedule` currently answer
  `{"allFound": …}`. They would become `{"jobs": […]}` / `{"triggers": […]}`, matching
  `AppliedJobKeysResponse` / `AppliedTriggerKeysResponse`, and `AllKeysFoundResponse` in
  `src/Quartz/HttpApiContract/RequestAndResponses.cs` would go away with them. `HttpScheduler` follows.
- **Documentation** — `docs/documentation/quartz-4.x/packages/http-api.md` (the response-shape table
  and the note pointing at this issue) and the migration guide's
  "One flag per mutation, named for what it reports" section.

## Why it is not part of #3359

#3359 unified the wire's response shapes and deliberately stopped short of this: it is a scheduler-API
change with store-contract and migration-guide consequences of its own, and burying it inside a
wire-polish PR would have meant it got reviewed as a footnote. That PR named the two endpoints
`{"allFound": …}` rather than `{"applied": …}` precisely so the wire does not claim something the
underlying API cannot support, and points here for the fix.

4.x is unreleased apart from `4.0.0-alpha.1`, so this is cheap now and expensive after 4.0 final.

lahma:
`IScheduler.DeleteJobs` and `UnscheduleJobs` answered one `bool` meaning "every key given was found".
Delete three of five existing jobs and the call answered `false` while **three jobs were deleted** —
nothing in the answer distinguished that from nothing having happened, and a caller who retried on
`false` never learned that most of the work had already succeeded.

They now answer with the keys they applied to, which is what the key-set pause, resume and
error-state reset already did. One rule for the whole family, and
`bool allFound = result.Count == keys.Count` recovers the old answer, so this is a replacement rather
than a trade.

```csharp
ValueTask<List<JobKey>>     DeleteJobs(IReadOnlyCollection<JobKey> jobKeys, …);
ValueTask<List<TriggerKey>> UnscheduleJobs(IReadOnlyCollection<TriggerKey> triggerKeys, …);
```

## The ADO path, which is where this either works or does not

It already walked the set one key at a time under a `// TODO: make this more efficient with a true
bulk operation...`, folding each iteration's result into a boolean with `&&`. **Collecting the keys
instead costs nothing** — the per-key answer was already there and was being thrown away. No extra
round trip, no `IDriverDelegate` change, still one `SchedulerLock.TriggerAccess` scope and one
transaction for the whole set.

The TODO is replaced by a comment saying why the walk stays per key rather than why it should not:
deleting a job is a cascade, not a statement — its triggers and their sub-table rows, the
fired-trigger rows that would otherwise resurrect it (#1696), then the job detail row — and a
set-based `DELETE … WHERE … IN (…)` reports a row count rather than which keys it hit, which is
precisely the answer the member owes its caller.

## Both stores are held to the same answers

`JobStoreContractTest` gains three cases that run against **both** the in-memory store and a real
SQLite ADO one: a partial delete of jobs, a partial unschedule of triggers (including that the
orphaned non-durable job goes with the trigger), and a key repeated in the set — which the old `bool`
hid, because the second pass found nothing left to delete and dragged the whole call's answer to
`false`.

`TheKeySetMembersAreImplementedRatherThanLeftToTheLoopingDefault` now covers both members, and
`SmokeTestPerformer` gains a `TestKeySetDeletes` step so the ADO path is exercised on **every dialect
CI runs**, not just SQLite.

## Behaviour that changed beyond the return type

- **A key that named nothing no longer raises a listener event.** The old implementation raised one
  `JobDeleted` / `JobUnscheduled` per key *given*, so a mistyped key told every scheduler listener
  that a job by that name had been deleted. The single-key `DeleteJob` and the key-set pause family
  never did that.
- **An empty set never reaches the store**; a `null` one is an `ArgumentNullException`.
- **The scheduling change is signalled once**, and not at all when nothing was removed.

## The two HTTP endpoints, under #3359's rule

#3359 established that a `200` carries a body exactly when the operation has something to say the
caller could not work out for itself, and that the flag is named for what it reports. It named these
two `allFound` rather than `applied` precisely because the underlying API could not support
`applied`, and pointed here.

With real applied keys available, `POST …/jobs/delete` and `POST …/triggers/unschedule` become
ordinary members of the key-set family: they answer `{"jobs": […]}` and `{"triggers": […]}`, reusing
`AppliedJobKeysResponse` / `AppliedTriggerKeysResponse`. `AllKeysFoundResponse` is deleted, and the
flag-naming rule loses its only exception — the conventions table in `http-api.md` no longer needs an
"except where…" clause.

## Reviewer decisions

1. **`IJobStore.DeleteJobs` / `DeleteTriggers` became default implementations** (`ApplyToEach`),
   joining the five key-set members that already were. That makes the family uniform and makes
   `custom-job-store.md`'s existing claim about them true — it already said they had defaults, which
   they did not. The cost: a third-party store still declaring `ValueTask<bool> DeleteJobs` **still
   compiles**, silently stops implementing the member, and inherits the per-key default. The
   migration guide says so in as many words, and the contract test's key-set guard catches it for
   anyone running it. The alternative — leaving them required, so the reshape is a compile error —
   was weighed and rejected: the asymmetry it leaves behind has no durable justification, and
   somebody would "finish" it after 4.0.
2. **`DeleteJobsRequest` / `UnscheduleJobsRequest` were left alone.** They are now structurally
   identical to `JobKeySetRequest` / `TriggerKeySetRequest`, and folding them would change no wire
   bytes — only the OpenAPI schema component names. #3359 was about response naming, so this felt
   like scope worth naming rather than taking.

## Baselines and docs

Public API baselines regenerated through Verify; the diff is exactly the four members across
`IScheduler`, `IJobStore`, `AdoJobStoreBase`, `RAMJobStore`, `DelegatingJobStore`,
`DelegatingScheduler` and `HttpScheduler`, with no collateral. Note the generator renders a default
interface member the same as an abstract one, so the baseline cannot show that these two gained
defaults — the migration guide carries that.

The same delta is written up in `migration-guide.md` under "The key-set delete and unschedule answer
with the keys they removed", with the one-line recipe for the old answer, and the wire change is
carried into the "One flag per mutation" section and `packages/http-api.md`.

Closes #3360

https://claude.ai/code/session_01Kr3wNU6NFXyZztWp12PGQc

### DIFF
--- a/docs/documentation/quartz-4.x/migration-guide.md
+++ b/docs/documentation/quartz-4.x/migration-guide.md
@@ -5157,6 +5157,63 @@ Over HTTP the key-set forms are new endpoints, described in
 `…/triggers/keys/reset-from-error-state`. They live under `keys/` because the collection-level `pause`
 and `resume` already belong to the group-matcher forms.
 
+## The key-set delete and unschedule answer with the keys they removed
+
+`DeleteJobs` and `UnscheduleJobs` predate that convention and used to answer one `bool` meaning "every
+key given was found". That answer was lossy in the case a caller most needs to know about: delete
+three of five existing jobs and **three jobs are deleted** while the call answers `false`, which is
+indistinguishable from nothing having happened. A caller who retried on `false` never learned that
+most of the work had already succeeded. They now answer the way the rest of the key-set family does.
+
+| Member | 3.x / 4.0 preview returned | 4.0 returns |
+|---|---|---|
+| `IScheduler.DeleteJobs(IReadOnlyCollection<JobKey>)` | `ValueTask<bool>` — every key was found | `ValueTask<List<JobKey>>` of the keys it deleted |
+| `IScheduler.UnscheduleJobs(IReadOnlyCollection<TriggerKey>)` | `ValueTask<bool>` | `ValueTask<List<TriggerKey>>` of the keys it removed |
+| `IJobStore.DeleteJobs(IReadOnlyCollection<JobKey>)` | `ValueTask<bool>` | `ValueTask<List<JobKey>>` |
+| `IJobStore.DeleteTriggers(IReadOnlyCollection<TriggerKey>)` | `ValueTask<bool>` | `ValueTask<List<TriggerKey>>` |
+
+**To get the old answer back**, compare the counts:
+
+```csharp
+List<JobKey> deleted = await scheduler.DeleteJobs(jobKeys);
+bool allFound = deleted.Count == jobKeys.Count;   // what the bool used to say
+```
+
+That one line is why the change is a replacement rather than a second overload: the old answer is
+still available to anyone who wants it, while the question it could not answer — *which* keys — now
+has one.
+
+Three consequences worth knowing:
+
+* **A key that named nothing no longer raises a listener event.** The old implementation raised one
+  `ISchedulerListener.JobDeleted` / `JobUnscheduled` per key **given**, so a mistyped key told every
+  listener that a job by that name had been deleted. The events now follow the keys the call applied
+  to, which is what the single-key `DeleteJob` and the whole key-set pause family already did. The
+  scheduling change is still signalled once for the call, and not at all when nothing was removed.
+* **An empty key set never reaches the store**, and a `null` one is an `ArgumentNullException` rather
+  than a `NullReferenceException` — again matching the key-set pause family.
+* **Repeating a key in the set no longer poisons the answer.** The second pass finds nothing left to
+  delete; before, that dragged the whole call's `bool` to `false`.
+
+If you implement `IJobStore`, both members are now **default implementations** that walk the set one
+key at a time, so a store that only implements the single-key `DeleteJob` / `DeleteTrigger` is correct
+without writing them. That has one trap: a store that still declares
+`ValueTask<bool> DeleteJobs(IReadOnlyCollection<JobKey>, CancellationToken)` **still compiles** — the
+method simply stops implementing the interface member, and the per-key default silently takes over.
+Change the return type. Both shipped stores override the pair to walk inside one lock and one
+transaction, and `JobStoreContractTest` fails a store that has left the default in place.
+
+The ADO store's walk stays per key on purpose. Deleting a job there is a cascade rather than a
+statement — its triggers and their sub-table rows, the fired-trigger rows that would otherwise
+resurrect it, then the job detail row — and a set-based `DELETE … WHERE … IN (…)` reports a row count
+rather than which keys it hit. Naming the deleted keys therefore costs no extra round trip: each
+iteration's result was already there and was being folded into a boolean and thrown away.
+
+Over HTTP, `POST …/jobs/delete` and `POST …/triggers/unschedule` answer `{"jobs": [ … ]}` and
+`{"triggers": [ … ]}` instead of `{"allFound": …}`, which makes them ordinary members of
+[the key-set family](packages/http-api.md#a-whole-set-of-keys-in-one-call) and removes the one
+exception to the flag-naming rule.
+
 ## One wire contract, and its enums have names
 
 The DTOs that define the HTTP API's JSON used to live in `Quartz.HttpClient`, and `Quartz.AspNetCore`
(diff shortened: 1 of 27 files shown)


lahma:
## Verification actually run locally

| Suite | Result |
|---|---|
| `dotnet build Quartz.slnx` | succeeded, no warnings (`TreatWarningsAsErrors`) |
| `src/Quartz.Tests.Unit` | 3042 passed, 0 failed, 4 skipped (45s) |
| `src/Quartz.Tests.AspNetCore` | 314 passed, 0 failed (5s) |
| `JobStoreContractTest` (RAM + SQLite ADO) | 128 passed, 0 failed — 64 assertions run twice, once per store |
| `AdoJobStoreSmokeTest` SQLite ×2 serializers ×2 providers | 4 passed (51s) |
| `AdoJobStoreSmokeTest` **PostgreSQL** ×2 serializers | 2 passed (33s) |
| `AdoJobStoreSmokeTest` **SQL Server** | 8 passed (46s) |
| `AdoJobStoreSmokeTest` **MySQL** ×2 serializers | 2 passed (35s) |
| `npm run docs:check-links` | 211 pages, 691 links, 386 fragments — no dead links or anchors |
| `dotnet fallout VerifyDocsSnippets` | up to date, 89 markdown files checked |

The ADO bulk-delete path is exercised on **four dialects locally** (SQLite, PostgreSQL, SQL Server,
MySQL) — `SmokeTestPerformer.TestKeySetDeletes` runs inside every dialect leg, so Oracle and Firebird
get it from CI's matrix for free.

CI: 18/18 checks pass. Sonar gate `OK`, new-code coverage **100.0%**.

One environmental note so it is not mistaken for a real failure: the first PostgreSQL attempt stalled
for 45 minutes inside Testcontainers' `ExecScriptAsync` — the container was up and answering `psql`
but its schema had never been created, so nothing of Quartz's had run yet. A managed-stack dump put
the process in `CompositeWorkItem.PerformOneTimeSetUp`, i.e. `TestcontainersDatabaseEnvironment`, well
before any test body. Killed and reran: 33 seconds, green.

## Found while here, deliberately left alone

**`JobDeleted` is raised twice for a non-durable job that has triggers**, and this predates the PR —
it is on the *single-key* path too, which this PR does not touch:

```
SINGLE-KEY DeleteJob raised JobDeleted 2 time(s)
```

`QuartzScheduler.DeleteJob` unschedules the job's triggers first; removing the last trigger of a
non-durable job makes the store delete the job itself and raise `JobDeleted` through the signaler
(`RAMJobStore.RemoveTriggerNoLock`, `AdoJobStoreBase.DeleteTrigger`). `DeleteJob` then raises it again
for the same key. Both stores do it, so it is a scheduler-level double announcement rather than a
divergence.

It is out of scope here and fixing it inside a return-shape change would bury it, so the unit test in
this PR uses durable jobs to isolate what the scheduler announces. Happy to file it separately.

PR #3389, opened 2026-08-24
Every scheduler in a container can be configured at once, and the dashboard uses it

lahma:
A container-wide way to configure every scheduler, the dashboard bug that motivated it, and two
tenancy fixes that ride along.

## `ConfigureAllQuartzSchedulers`

```csharp
public static IServiceCollection ConfigureAllQuartzSchedulers(
    this IServiceCollection services,
    Action<IQuartzBuilder> configure);
```

The options pattern's `ConfigureAll`, for schedulers. The delegate is applied to every scheduler
`AddQuartz()`, `AddQuartz(name, …)` or `AddQuartzSchedulers(…)` registers in the container — the ones
already registered when it is called, and the ones registered after it. **The order of the calls does
not matter**, which is the point: a package that adds something to every scheduler cannot know whether
the application registers its schedulers before or after calling it.

Each scheduler is handed its own `IQuartzBuilder`, so what the delegate registers lands under that
scheduler's own service key — exactly as if it had been written inside that scheduler's
`AddQuartz(name, q => …)` callback. A plugin or listener added this way is one instance per scheduler,
each initialized with the name of the scheduler it belongs to.

Mechanically: the delegate is recorded on `SchedulerNameRegistry`, the registration-time singleton the
container already holds as an `ImplementationInstance`, and `AddQuartz` applies the recorded delegates
after its own configuration callback. Schedulers already registered are configured at the
`ConfigureAllQuartzSchedulers` call site — which is also after their callbacks, so container-wide
configuration always lands in the same position in the order however the calls were written.

A delegate reaches each scheduler **once**. Registering the default scheduler is additive, so
`AddQuartz()` can be called twice for one scheduler, and a listener registration is not a `TryAdd` — it
would otherwise be attached twice.

`KeyedService.AnyKey` is not used: `GetKeyedServices<T>(key)` does not see `AnyKey` descriptors, so it
cannot serve enumerable keyed resolution.

## Bug 1: the dashboard's plugins never reached a named scheduler

`AddQuartzDashboard()` built a `QuartzBuilder(services, schedulerKey: null)` and added
`DashboardLiveEventsPlugin` and `DashboardHistoryPlugin` through it, which registers **unkeyed**.
`DefaultSchedulerFactory` resolves a named scheduler's plugins with `GetKeyedServices`. So a scheduler
registered with `AddQuartz("acme", …)` never had either plugin initialized: it appeared in the scheduler
selector, its jobs and triggers rendered, and its **Live Logs view and History page were silently always
empty** — with nothing to configure that would have filled them.

They now go in through `ConfigureAllQuartzSchedulers`, so every scheduler gets its own instance of each,
initialized with its own name. `DashboardHistoryStore` already partitions by scheduler name and its
entries already carry `context.Scheduler.SchedulerName`, and `DashboardLiveEventsPlugin` already
broadcasts to the SignalR group named after the scheduler, so several instances feeding them need
nothing further — checked rather than assumed.

`DashboardPluginRegistrationTest` hard-coded `new SchedulerKey(null)`, which is why nothing caught it.
It now asserts both keys, on either registration order, and that three schedulers get three instances
rather than one shared. **Three of its four tests fail against the unfixed dashboard** (verified by
stashing the fix).

## Bug 2: `AddQuartzHttpClient` could silently capture the unkeyed `IScheduler`

`AddQuartzHttpClient` registers its remote scheduler unkeyed as well as by name, so a container holding
one remote scheduler and nothing else answers `GetRequiredService<IScheduler>()` with it. `AddQuartz()`
registers the local default scheduler in that same slot. Both use `TryAdd`, so whichever ran first owned
it — and `AddQuartzHttpClient(…)` followed by `AddQuartz()` left "the scheduler" meaning the remote one,
with no error. A program that thought it held its own scheduler was scheduling jobs in another process.

`AddQuartz()` now throws `InvalidOperationException` at registration when an unkeyed `IScheduler` is
already present, **before it registers anything**, so a container that cannot hold the scheduler is not
left half configured by the attempt. The message names `AddQuartzHttpClient` and says the remote
scheduler stays reachable under its keyed name whichever order is written.

The other order is unchanged and is the one to write. `AddQuartz("Local", …)` beside a remote scheduler
is fine either way round — a named scheduler never wanted the unkeyed slot. Both methods' XML docs and
`packages/http-client.md` say so.

## Improvement 3: a logging scope naming the scheduler — **the per-firing half was measured out**

The scheduler thread now opens a logging scope for the lifetime of its loop, carrying
`quartz.scheduler.name` and `quartz.scheduler.id` — the same attribute names the spans and the
measurements use (`ActivityTags`), so one query says "this tenant" against traces, metrics and logs
alike. The state is one immutable `SchedulerLogScope` per scheduler implementing
`IReadOnlyList<KeyValuePair<string, object?>>`, cached on `QuartzSchedulerResources` and dropped only
when the name or instance id is set (a generated instance id arrives after construction).

**There is deliberately no scope per firing in `JobRunShell`.** It was written, measured, and taken back
out under the spec's own rule. `ILogger.BeginScope` pushes onto an `AsyncLocal`, and an `AsyncLocal`
write copies the execution context — so it is not one object, it is about five.

`JobRunShellBenchmark`, `--warmupCount 10 --iterationCount 25 --launchCount 1`, .NET 10.0.11, Ryzen
9 5950X. The benchmark previously ran only with `NullLogger`, whose `BeginScope` returns a cached
singleton and allocates nothing — which would have measured the scope as free, for the one reason that
makes the measurement worthless. A second case was added first (committed separately, so the "before"
numbers come from the same harness) that runs the same firing under a logger factory holding a
scope-supporting provider whose loggers are disabled.

| | before | with `JobRunShell` scope | shipped (thread scope only) |
|---|---|---|---|
| `NullLogger`, mean | 957.9 ns | — | 911.3 ns |
| `NullLogger`, allocated | 776 B | 776 B | **776 B** |
| scope-supporting provider, mean | ~950 ns | 1,087.7 ns (**+13.6%**) | 1,034.0 ns |
| scope-supporting provider, allocated | 776 B | 1,016 B (**+240 B**) | **776 B** |

Both fallback conditions in the spec were met (>2% regression, allocations grew by more than one
object), so the per-firing scope came out. The means are noisy run to run — the two benchmark methods
differ by ~6–13% in either direction on identical code — but **allocation is exact and deterministic,
and is back to 776 B on both paths, byte for byte what it was before this branch**.

## What a reviewer must decide

**The thread-lifetime scope only reaches a job when the application has called
`LogProvider.SetLogProvider`.** `QuartzSchedulerThread` logs through the ambient `LogProvider`, not
through an injected logger, and nothing in the hosting path sets it — so for an application on the
generic host that has not called it, the loop logs to `NullLogger`, its scope is a no-op, and jobs get
nothing. When it *is* set, the thread pools Quartz ships dispatch through a `Task` that captures the
execution context, so a job inherits the loop's scope for free; that is what the test pins.

So item 3 as specified — "a job's own log lines carry it too" — is delivered **only** for applications
that set `LogProvider`. Getting it unconditionally needs one of:

1. **Hand `QuartzSchedulerThread` a container `ILogger`** instead of the ambient one. Then the scope is
   real in every DI application and still flows to jobs at zero per-firing cost. This is the right fix
   and is a change of its own — the thread is constructed by `QuartzScheduler`'s constructor from
   `QuartzSchedulerResources`, which holds no logger factory. Not attempted here.
2. **Re-add the `JobRunShell` scope and accept the 240 B / ~14%.** Note that if it is re-added *and*
   `LogProvider` is set, the scheduler is named **twice** in a job's scope chain (measured: tag count 2)
   — harmless to any structured sink, which flattens a scope into properties where the second write of
   the same key with the same value is invisible, but a console formatter with `IncludeScopes` prints it
   twice. Skipping the redundant push would need an `AsyncLocal<SchedulerLogScope?>` of our own, whose
   write is the very cost being avoided.

Everything above is written up in the comment at the top of `JobRunShell.Run` and in the remarks on
`SchedulerLogScopeTest`, so it is not folklore.

## Baselines moved

One line, in `src/Quartz.Tests.Unit/Verify/PublicApiTest_Quartz.verified.txt`, accepted through the
Verify workflow:

```diff
+ public static Microsoft.Extensions.DependencyInjection.IServiceCollection ConfigureAllQuartzSchedulers(this Microsoft.Extensions.DependencyInjection.IServiceCollection services, System.Action<Quartz.IQuartzBuilder> configure) { }
```

`SchedulerLogScope` is `internal sealed`, so item 3 moves no baseline.

## Documentation

- `migration-guide.md` — a section for `ConfigureAllQuartzSchedulers`, and a three-row table for the
  `AddQuartz()`/`AddQuartzHttpClient` ordering under the existing remote-scheduler section.
- `packages/dashboard.md` — a new section saying plainly that every scheduler in the container gets the
  dashboard's plugins, with a "Changed in 4.0.0-alpha.2" note describing what was broken.
- `packages/http-client.md` — a "Beside a local scheduler" section. The C# is a real compiled snippet
  (`sample_httpclient_beside_local` in `Quartz.Documentation.Samples`), regenerated with
  `dotnet fallout DocsSnippets`.

`multi-tenancy.md`, `multiple-schedulers.md`, `execution-groups.md` and `tenancy-patterns.md` are
deliberately untouched.

## Verification

```
dotnet build Quartz.slnx                                            0 Warning(s)  0 Error(s)
dotnet test src/Quartz.Tests.Unit                     Failed: 0, Passed: 3108, Skipped: 4
dotnet test src/Quartz.Tests.AspNetCore               Failed: 0, Passed:  317, Skipped: 0
dotnet test src/Quartz.Tests.Integration              Failed: 0, Passed:  316, Skipped: 0
dotnet fallout VerifyDocsSnippets                     up to date (89 markdown files checked)
npm run docs:check-links                              no dead links or anchors
```

The integration suite ran against a live Docker daemon — SQL Server 2017 and 2022, PostgreSQL 15,
MySQL 8, Oracle XE 21 and Firebird 4 — in 8m32s. There is no `Quartz.Tests.AOT` project on this base.

https://claude.ai/code/session_01VfNtYcGzWFwJXJNHubegfg

### DIFF
--- a/docs/documentation/quartz-4.x/migration-guide.md
+++ b/docs/documentation/quartz-4.x/migration-guide.md
@@ -1566,6 +1566,47 @@ describes, is unchanged.
 The six phases that decide which of a scheduler's descriptions wins — configuration is last-wins,
 registration is first-wins — are now documented on `AddQuartz` itself rather than in comments inside it.
 
+## Every scheduler in the container can be configured at once
+
+`ConfigureAllQuartzSchedulers(Action<IQuartzBuilder>)` is new. It is the options pattern's `ConfigureAll`,
+for schedulers: the delegate is applied to every scheduler `AddQuartz()`, `AddQuartz(name, …)` or
+`AddQuartzSchedulers(…)` registers in the container.
+
+```csharp
+services.AddQuartz();
+services.AddQuartz("acme", q => q.UseInMemoryStore());
+services.AddQuartz("initech", q => q.UseInMemoryStore());
+
+// Both named schedulers and the default one get their own instance of each.
+services.ConfigureAllQuartzSchedulers(q =>
+{
+    q.AddPlugin<AuditPlugin>("audit");
+    q.AddJobListener<TenantMetricsListener>();
+});
+```
+
+**The order of the calls does not matter.** Schedulers already registered are configured where this is
+called; schedulers registered afterwards are configured by their own `AddQuartz`. That is the point: a
+library that adds something to every scheduler cannot know whether the application registers its
+schedulers before or after calling it.
+
+The delegate is handed a builder *per scheduler*, so everything it registers lands under that scheduler's
+own service key — exactly as if it had been written inside that scheduler's `AddQuartz(name, q => …)`
+callback. A plugin or listener added this way is therefore **one instance per scheduler**, each
+initialized with the name of the scheduler it belongs to, rather than one instance shared between them.
+
+It runs after each scheduler's own configuration callback, which is what makes the order immaterial. The
+usual precedence follows: registration is first-wins, so a job store or thread pool a scheduler chose for
+itself is not replaced by one chosen here; options are last-wins, so a value set here overrides the same
+option set on one scheduler, exactly as `ConfigureAll<TOptions>` overrides an earlier named `Configure`.
+
+Remote schedulers registered with `AddQuartzHttpClient` are not built by a builder and are skipped.
+Calling it when no scheduler is registered at all is not an error.
+
+`AddQuartzDashboard()` is its first caller: it installs its live-events and history plugins this way, so
+a scheduler registered with `AddQuartz("acme", …)` finally has a populated live view and execution
+history rather than two silently empty pages.
+
 ## `IScheduler` is a service, keyed by the scheduler's name
 
 `AddQuartz()` now registers `IScheduler` as well as `ISchedulerFactory`. The default scheduler is
--- a/docs/documentation/quartz-4.x/migration-guide.md
+++ b/docs/documentation/quartz-4.x/migration-guide.md
@@ -1654,6 +1654,14 @@ second remote scheduler needed a marker interface only because the container had
 The marker interfaces themselves can be deleted, and the first remote scheduler registered is still the
 unkeyed `IScheduler` for a container that holds only one.
 
+In a container that also holds a local scheduler, **`AddQuartz()` goes first**:
+
+| Order | What happens |
+|---|---|
+| `AddQuartz()` then `AddQuartzHttpClient(…)` | The local default scheduler owns `GetRequiredService<IScheduler>()`; the remote one is reached by name. This is the arrangement to write. |
+| `AddQuartzHttpClient(…)` then `AddQuartz()` | `AddQuartz()` throws `InvalidOperationException` at registration, naming `AddQuartzHttpClient`. Registration is first-wins, so this used to make "the scheduler" the remote one with no error at all — a program that thought it held its own scheduler was scheduling jobs in somebody else's process. |
+| `AddQuartzHttpClient(…)` then `AddQuartz("Local", …)` | Fine either way round. A named scheduler is keyed by its name and never wanted the unkeyed slot. |
+
 ### A client is named or built, never handed over
 
 `HttpClientOptions.HttpClient` and the `AddQuartzHttpClient(schedulerName, HttpClient, …)` overload are
--- a/docs/documentation/quartz-4.x/packages/dashboard.md
+++ b/docs/documentation/quartz-4.x/packages/dashboard.md
@@ -66,7 +66,7 @@ initialized with its own name, and history entries are attributed to the schedul
 
 It does this with `ConfigureAllQuartzSchedulers`, so nothing extra is written at the call site.
 
-::: warning Changed in 4.0.0-alpha.2
+::: warning Fixed in 4.0.0-alpha.2
 The dashboard's plugins used to be registered without a service key, which meant only a scheduler
 registered by `AddQuartz()` — the unnamed, default one — ever ran them. A named scheduler appeared in the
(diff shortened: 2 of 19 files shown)


lahma:
Review notes, for the record:

- The one red leg was `TestFirebird("newtonsoft")` failing with a Firebird `deadlock` while pausing a trigger group — an integration flake in a path this PR does not touch; re-run requested.
- The per-firing scope being taken back out on the benchmark numbers is the right call, and the comment left in `JobRunShell` says exactly why.
- The caveat that the thread-lifetime scope reaches a hosted application only after `LogProvider.SetLogProvider` is called turned out to be a wider defect than this PR: the scheduler core, the ADO store and its managers, the semaphores and the `Use*<T>()` components all log through the ambient static, while the migration guide says the container injects them. That is #3392 now, sequenced after the listener-signature change; the proof there is that `SchedulerLogScopeTest` stops needing the static.

Merging once the re-run is green.

PR #3397, opened 2026-08-24
Every listener callback is told which scheduler is calling it

### LINKED ISSUE #3063: Add scheduler metadata/context parameter to ISchedulerListener methods
I would like to suggest enhancing the `ISchedulerListener` interface by adding a scheduler-related context parameter to its callback methods.

Currently, methods in `ISchedulerListener`, especially `SchedulerError`, do not provide enough information about the scheduler instance that triggered the event.

For example:

```csharp
Task SchedulerError(string msg, SchedulerException cause, CancellationToken cancellationToken = default);
```

When handling this event in a listener, it can be difficult to determine additional information about the scheduler, such as:

Scheduler name
Scheduler instance ID
Scheduler state
Whether the scheduler is started, paused, or shutdown
Other useful runtime information available from SchedulerMetaData

It would be helpful if these listener methods could include a parameter similar to SchedulerMetaData, or another scheduler context object.
For example:

```csharp
Task SchedulerError(
    string msg,
    SchedulerException cause,
    SchedulerMetaData schedulerMetaData,
    CancellationToken cancellationToken = default);
```

Alternatively, a more general scheduler context object could be introduced if passing SchedulerMetaData directly is not appropriate.

Use Case
In applications that use multiple scheduler instances, a global or shared ISchedulerListener may receive error notifications from different schedulers.
When SchedulerError is triggered, the listener often needs to log, monitor, alert, or diagnose the issue. Without scheduler metadata, the listener has limited context and may not be able to clearly identify which scheduler produced the error.
Having access to scheduler metadata would make it easier to:

Improve logging and diagnostics
Identify the scheduler instance that raised the event
Include scheduler name and instance ID in monitoring or alerting systems
Handle errors differently depending on scheduler state or configuration

Proposed Change
Add a scheduler metadata or scheduler context parameter to ISchedulerListener callback methods, especially SchedulerError.
Possible API example:
```csharp
Task SchedulerError(
    string msg,
    SchedulerException cause,
    SchedulerMetaData schedulerMetaData,
    CancellationToken cancellationToken = default);
```
Or introduce a dedicated context type:
```csharp
Task SchedulerError(
    string msg,
    SchedulerException cause,
    SchedulerListenerContext context,
    CancellationToken cancellationToken = default);
```
Benefit
This would improve the extensibility and observability of scheduler listeners, especially in applications using multiple schedulers or centralized logging/monitoring infrastructure.
Add this to the other methods in the interface as well.

lahma:
Closes #3063. Also answers discussion #3211 — the trigger behind a `SchedulerError`.

## The rule

**A listener reaches the scheduler it serves through its execution context, or as its first argument when there is no execution.**

`IJobListener`'s members and three of `ITriggerListener`'s four already receive an `IJobExecutionContext`, whose `.Scheduler` says who is calling. `ISchedulerListener`'s twenty-three members and `ITriggerListener.TriggerMisfired` received nothing, so a listener shared by several schedulers in one host — which `AddQuartz("acme", …)` plus `AddQuartz("initech", …)` makes ordinary — was told that *a* trigger had been paused and could not ask which scheduler paused it.

Those twenty-four members now take `IScheduler` first. The interfaces state the rule in their XML docs, and so do the two tutorial pages.

It is the `IScheduler` rather than a name or an identity record because a listener that knows *which* scheduler usually wants to *act* on it — pause the trigger it just heard about, read `Status`; identity is `SchedulerName`/`SchedulerInstanceId` on the same object.

## The record

`SchedulerError(string, SchedulerException, ct)` becomes `SchedulerError(IScheduler, SchedulerErrorContext, ct)`:

```csharp
public sealed record SchedulerErrorContext
{
    public required string Message { get; init; }
    public required SchedulerException Exception { get; init; }
    public TriggerKey? TriggerKey { get; init; }
    public JobKey? JobKey { get; init; }
    public string? FireInstanceId { get; init; }
}
```

Filled in where the facts are known:

| Site | What it fills in |
|---|---|
| `JobRunShell` (12 raise sites) | trigger, job and fire instance, from the execution context or — before one exists — from the `TriggerFiredBundle` |
| `SchedulerSignalerImpl`, misfire-notification failure | the trigger that misfired, and its job |
| `QuartzSchedulerThread`, acquisition scan failed | nothing; the scan never reached a trigger |
| `QuartzSchedulerThread`, `TriggersFired` failed | the trigger **only when the batch held one** — see "for a reviewer" below |
| `AdoJobStoreBase`, retry | nothing; the failure is the connection |
| `Xml`/`JsonSchedulingDataProcessorPlugin` | nothing; the failure is the file, which names many jobs |

`ISchedulerSignaler.NotifySchedulerListenersError` takes the record too, so a job store of somebody's own can report the trigger it failed for.

## How `QuartzScheduler` gets its facade

`QuartzScheduler` builds its `StdScheduler` once, in its constructor, and exposes it as `QuartzScheduler.Scheduler`. `DefaultSchedulerFactory.Create` now hands out *that* instance instead of `new StdScheduler(quartzScheduler)`, so the facade the factory returns, the one `IJobExecutionContext.Scheduler` carries, the one `StdJobRunShellFactory.Initialize` receives and the one every listener callback is handed are all the same object. No notification constructs a facade and none resolves one from a container.

*(Deviation from the spec, deliberately: the spec said to hand the facade to `QuartzScheduler` from `DefaultSchedulerFactory`. Constructing it in `QuartzScheduler` instead means it is never null — benchmarks and several unit tests construct `QuartzScheduler` directly and drive the scheduler thread, and a null facade would have been an NRE on the first error notification. Same single instance, same construction point, one fewer way to get it wrong.)*

## Implementers updated

`Core/ErrorLogger` (logs the scheduler name and the keys as structured fields, with a second template for the common no-keys case so a line does not end in three empty fields), `Listeners/BroadcastSchedulerListener`, `Listeners/BroadcastTriggerListener`, `Quartz.Dashboard/Plugins/DashboardLiveEventsPlugin`, `Quartz.Plugins`' `LoggingTriggerHistoryPlugin`, `StructuredLoggingTriggerHistoryPlugin`, `XmlSchedulingDataProcessorPlugin` and `JsonSchedulingDataProcessorPlugin`, `Quartz.Examples.Worker/Listeners`, `Quartz.Examples.AspNetCore`'s `SampleSchedulerListener` and `SampleTriggerListener`, `Quartz.Documentation.Samples/Packages/MicrosoftDiIntegrationSamples`, `Quartz.Benchmark`, and every test implementer the grep found.

`DashboardLiveEventsPlugin` now keys its broadcasts on the *passed* scheduler's name rather than the one it was initialized with (#3389 made it one instance per scheduler anyway), and its `schedulerName` field is gone. `SchedulerErrorDto` gained `TriggerKey` and `JobKey` — added to the existing positional record, no new wire type; `LiveLogs.razor` receives the payload as `object`, so the extra members render without a change there.

## Baselines moved

Three, all through the Verify workflow (run the tests, review the diff, promote `.received.txt`):

- `PublicApiTest_Quartz.verified.txt` — the 23 `ISchedulerListener` members and `BroadcastSchedulerListener`'s 23, `ITriggerListener.TriggerMisfired` and `BroadcastTriggerListener`'s, `ISchedulerSignaler.NotifySchedulerListenersError`, and the new `Quartz.SchedulerErrorContext`.
- `PublicApiTest_Quartz.Plugins.verified.txt` — `TriggerMisfired` on the two trigger-history plugins.
- `PublicApiTest_Quartz.Dashboard.verified.txt` — `DashboardLiveEventsPlugin`'s listener members and `SchedulerErrorDto`'s two new members.

The same delta is in `docs/documentation/quartz-4.x/migration-guide.md`, as a new section with an exhaustive twenty-four-row before/after table.

## For a reviewer to decide

1. **The migration experience is not what the spec assumed, and the guide says so.** The spec expected a listener that overrides a changed member to get a compile error naming it. It does not: every member of these interfaces has a default implementation, so a listener keeping an old signature still compiles — it silently stops implementing anything and its method becomes dead code. Nothing in the language warns. Two analyzers happened to catch Quartz's own (CA1822 on a callback that no longer touches instance state, MA0196 on an `<inheritdoc />` that no longer inherits), and the guide names both, but neither is a guarantee. The guide tells readers to search for the twenty-four names rather than wait to be told. If that is too weak a migration story, the alternative is dropping the default implementations, which would undo [The three `*Support` base classes are gone].

2. **`QuartzSchedulerThread`'s batch-fire failure reports a trigger only when the batch held one.** The spec said this site "sets the trigger", but `TriggersFired` fails for a whole batch and the record has room for one key. Naming one of five would claim the failure was about that trigger. The message still enumerates them all. Say the word if you would rather it were always null, or always `triggers[0]`.

3. **The parameter is `errorContext`, not `error`.** CA1716 rejects `error` on a virtual or interface member — it is a VB keyword — and the rule is on as an error here. Internal implementations still call it `error` where that reads better.

4. **A container-injected `IScheduler` is not reference-equal to the one a callback is handed.** The container hands out `DeferredScheduler`, a proxy that resolves on first use; the callback gets the `StdScheduler` behind it, which is also what `ISchedulerFactory.GetScheduler()` and `IJobExecutionContext.Scheduler` return. That asymmetry predates this PR; the guide now spells it out and says to compare `SchedulerName` in that case.

## Left alone deliberately

- `IJobListener`'s members and `ITriggerListener`'s other three — they already receive an execution context.
- `ISchedulerListener` member *names*.
- `multi-tenancy.md`'s note that a listener "is told which scheduler it is hearing from on every callback" — it now reads literally true.
- `docs/documentation/quartz-3.x/`.

## Verification

```
dotnet build Quartz.slnx                                   Build succeeded. 0 Warning(s) 0 Error(s)
dotnet test src/Quartz.Tests.Unit                          Passed! 0 failed, 3120 passed, 4 skipped
dotnet test src/Quartz.Tests.AspNetCore                    Passed! 0 failed, 328 passed
dotnet test src/Quartz.Tests.Integration -f db-sqlite      Passed! 0 failed, 7 passed      (Docker up)
dotnet test … ~TriggerInErrorNotificationAdoTest           Passed! 0 failed, 1 passed      (Docker up, Postgres)
dotnet fallout VerifyDocsSnippets                          up to date (89 markdown files checked)
npm run docs:check-links                                   211 pages, 712 links, 407 fragments; no dead links or anchors
```

### DIFF
--- a/docs/documentation/quartz-4.x/migration-guide.md
+++ b/docs/documentation/quartz-4.x/migration-guide.md
@@ -2986,6 +2986,133 @@ exception:
 + IJobListener? listener = listenerManager.GetJobListener(name);
 ```
 
+### Listeners are told which scheduler is calling
+
+The listener callbacks that run inside a firing have always been able to say who was calling:
+`IJobExecutionContext.Scheduler`. The rest could not. A listener registered with two schedulers in one
+host — which `AddQuartz("acme", …)` and `AddQuartz("initech", …)` make an ordinary thing to do — was told
+that *a* trigger had been paused, or that *a* scheduler had failed, with no way to ask which one (#3063).
+
+There is one rule now, and it holds for all three listener interfaces: **a listener reaches the scheduler
+it serves through its execution context, or as its first argument when there is no execution.** That makes
+`IScheduler` the first parameter of every one of `ISchedulerListener`'s twenty-three members, and of
+`ITriggerListener.TriggerMisfired` — the one trigger callback with no context to carry it, because a
+misfire is noticed rather than executed.
+
+```diff
+- public ValueTask SchedulerStarted(CancellationToken cancellationToken = default)
++ public ValueTask SchedulerStarted(IScheduler scheduler, CancellationToken cancellationToken = default)
+  {
+-     logger.LogInformation("Scheduler started");
++     logger.LogInformation("Scheduler {SchedulerName} started", scheduler.SchedulerName);
+      return default;
+  }
+```
+
+It is the `IScheduler` rather than a name or an identity record, because a listener that knows *which*
+scheduler usually wants to *act* on it — pause the trigger it was just told about, read `Status`. Identity
+is `SchedulerName` and `SchedulerInstanceId` on the same object.
+
+The instance handed over is the one `ISchedulerFactory.GetScheduler()` returns, and the very one
+`IJobExecutionContext.Scheduler` carries, so those two compare by reference. An `IScheduler` injected from
+the container does not: that is a proxy that resolves the scheduler on first use, so compare a
+`SchedulerName` there rather than the objects.
+
+#### `SchedulerError` says what the error was about
+
+`SchedulerError` took a message and an exception, which said what happened but never what it happened to.
+Most of the places that raise it know: a job that threw is reported from its execution context, a job that
+could not be built from the bundle that was fired. Those facts now travel in a `SchedulerErrorContext`:
+
+```diff
+- ValueTask SchedulerError(string msg, SchedulerException cause, CancellationToken cancellationToken = default);
++ ValueTask SchedulerError(IScheduler scheduler, SchedulerErrorContext errorContext, CancellationToken cancellationToken = default);
+```
+
+```csharp
+public sealed record SchedulerErrorContext
+{
+    public required string Message { get; init; }
+    public required SchedulerException Exception { get; init; }
+    public TriggerKey? TriggerKey { get; init; }
+    public JobKey? JobKey { get; init; }
+    public string? FireInstanceId { get; init; }
+}
+```
+
+`Message` and `Exception` are the two old parameters, so a listener that only logged needs one rename:
+
+```diff
+- public ValueTask SchedulerError(string message, SchedulerException exception, CancellationToken cancellationToken = default)
++ public ValueTask SchedulerError(IScheduler scheduler, SchedulerErrorContext errorContext, CancellationToken cancellationToken = default)
+  {
+-     logger.LogError(exception, "{Message}", message);
++     logger.LogError(errorContext.Exception, "{Message}", errorContext.Message);
+      return default;
+  }
+```
+
+The three keys are nullable because some errors genuinely have no subject: a scan for the next trigger to
+fire that never reached a trigger, a job store retrying a failed connection, a schedule file that names
+many jobs. Read a null as "the scheduler could not say", not as "this concerns no trigger". Where the
+scheduler does know — every failure inside a firing, and a misfire notification that a listener broke —
+all three are filled in, which is what discussion #3211 asked for: the trigger behind a `SchedulerError`.
+
+`ISchedulerSignaler.NotifySchedulerListenersError` takes the record for the same reason, so a job store of
+your own can report the trigger it failed for:
+
+```diff
+- ValueTask NotifySchedulerListenersError(string message, SchedulerException exception, CancellationToken cancellationToken = default);
++ ValueTask NotifySchedulerListenersError(SchedulerErrorContext errorContext, CancellationToken cancellationToken = default);
+```
+
+#### The compiler will not point at every callback you have to change
+
+Every member of these interfaces has a default implementation, which is what lets a listener implement only
+the notifications it cares about — and it is also why a listener that keeps an old signature still compiles.
+It simply stops being an implementation of anything: the default runs instead, and the method becomes dead
+code that is never called.
+
+So the build will not fail. Search your listeners for the twenty-four names in the table below rather than
+waiting to be told, and note that `#pragma`-free hints do exist: a callback that no longer overrides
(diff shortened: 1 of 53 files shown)


lahma:
Merging over the Sonar quality gate, for the record: the failed condition is 54.7 % coverage on new code, and the "new code" is the 24 changed member signatures across `Quartz.Plugins`' history and scheduling-data plugins, the dashboard plugin, `BroadcastSchedulerListener`/`BroadcastTriggerListener` and the examples — lines Sonar attributes as new because the signature changed, with no new logic behind them and no coverage before either. The logic this PR adds — `SchedulerErrorContext` filled from the bundle or the execution context, the single facade `QuartzScheduler` hands out, the misfire notification's sender, broadcast forwarding — is covered by `ListenerSenderTest`'s four cases plus the reshaped existing listener tests. Same class of override as #3318 and #3323.

The silent-dead-member hazard called out in the body is tracked as #3398, on the milestone.

Issue #3414, opened 2026-08-26
Logging is 292 unguarded Log* calls; make it source-generated with a pinned event catalogue


lahma:
`[LoggerMessage]` appears nowhere under `src/`. `Quartz` has 224 `logger.Log{Level}(…)` calls, `Quartz.Plugins` 43, `Quartz.Jobs` 14 and `Quartz.Extensions.Redis` 11 — none behind an `IsEnabled` check, several under `#pragma warning disable CA2254` because the message is assembled at the call site, and one (`AdoJobStoreBase.LogWarnIfNonZero`) logging at a level its name does not say. Every call formats and boxes its arguments whether or not the level is on, including the ones on the scheduling loop.

Source-generated logging is the fix the platform ships for exactly this: no work when the level is off, a stable event id per message that an operator can filter or alert on, and a catalogue the documentation can carry. Decision taken 2026-08-26: all four packages, core first, one pull request per package.

**Shape**

- One `internal static partial class <Area>Log` per area, beside its callers, whose members are extension methods on `ILogger` (`logger.TriggerAcquired(key)` at the call site). Levels and message templates are preserved verbatim; a message that was built by concatenation becomes a template, and the `CA2254` pragmas go.
- Event ids are stable and unique within a package, allocated in ranges by area (scheduler core, in-memory store, ADO store, clustering, misfire handling, lock handlers, configuration and hosting, serialization and type loading; the other three packages get ranges of their own). A Verify snapshot per package — `LogEventCatalogTest_<assembly>.verified.txt`: id, name, level, template — pins the catalogue, so adding or renumbering an event is a reviewed diff.
- A source-scanning guard in `Quartz.Tests.Unit` fails a plain `Log{Level}(` call under `src/<package>` the way `SourceEncodingTest` fails a byte-order mark, so the conversion cannot regress one call at a time.
- Not a behaviour change: same levels, same templates, same exceptions attached. `JobRunShellBenchmark` before and after records what the hot path gained.

**Not in scope:** tests, examples and the documentation samples keep their plain calls; the `LogProvider` static stays for the ambient types (listeners, triggers) — its consumers use the same generated methods.

PR #3450, opened 2026-08-26
The core logs through source-generated events with a pinned catalogue

lahma:
Part 1 of 4 for #3414. The `Quartz` package only; `Quartz.Plugins`, `Quartz.Jobs` and
`Quartz.Extensions.Redis` follow in their own pull requests.

## What changed

All 224 `logger.Log{Trace,Debug,Information,Warning,Error,Critical}(…)` calls under `src/Quartz/`
are now calls on source-generated `[LoggerMessage]` methods — 204 events, because a message several
call sites spell identically is one event. Nothing is formatted or boxed when the level is off, and
every message carries a stable event id an operator can filter or alert on.

Nine `internal static partial class <Area>Log` classes, each beside its callers, each drawing from
its area's id range:

| Range | Class | File |
|---|---|---|
| 1000–1999 | `CoreLog` | `src/Quartz/Core/CoreLog.cs` |
| 2000–2999 | `RAMJobStoreLog` | `src/Quartz/Impl/RAMJobStoreLog.cs` |
| 3000–3499 | `AdoJobStoreLog` | `src/Quartz/Impl/AdoJobStore/AdoJobStoreLog.cs` |
| 3500–3599 | `ClusterLog` | `src/Quartz/Impl/AdoJobStore/ClusterLog.cs` |
| 3600–3699 | `MisfireLog` | `src/Quartz/Impl/AdoJobStore/MisfireLog.cs` |
| 3700–3799 | `LockHandlerLog` | `src/Quartz/Impl/AdoJobStore/LockHandlerLog.cs` |
| 4000–4999 | `ConfigurationLog` | `src/Quartz/Configuration/ConfigurationLog.cs` |
| 5000–5099 | `XmlSchedulingLog` | `src/Quartz/Xml/XmlSchedulingLog.cs` |
| 5100–5199 | `UtilityLog` | `src/Quartz/Util/UtilityLog.cs` |

6000–6999, 7000–7999 and 8000–8999 stay reserved for the other three packages, and the catalogue
test fails an id that strays into one.

Two guards, in `Quartz.Tests.Unit`:

* **`LogEventCatalogTest`** reflects over an assembly for every `[LoggerMessage]` method and Verifies
  id, level, name and template as `LogEventCatalogTest_<assembly>.verified.txt`. It also asserts that
  ids are unique and that every one falls in a range allocated to that assembly — which is the same
  thing as saying it stays out of another package's reserved range. The other three packages join
  with one line in `Catalogued` and a snapshot of their own.
* **`LogCallSiteTest`** walks `src/Quartz/` for a plain `.LogInformation(` and its siblings, the way
  `SourceEncodingTest` walks `src/` for a byte-order mark — a conversion is worth nothing if it can
  come back one call at a time. It carries an allow-list, empty, so a justified exception can be
  recorded with its reason rather than silently excluded. Later parts widen the `Converted` list.

`AdoJobStoreBase.LogWarnIfNonZero` is gone; the `CA2254` pragmas are gone; no other structure moved.

## Message templates that changed

Everything else is verbatim — same level, same template, same exception attached. Only these twelve
templates differ: eight where the source built the message rather than declaring it, and four typos
corrected on review, while their ids are new and doing so still costs nobody anything.

| Old | New | Why |
|---|---|---|
| `"ClusterManager: detected " + failedInstances.Count + " failed or restarted instances."` at **Information** (Debug when zero) | `"ClusterManager: detected {Count} failed or restarted instances."` at **Warning**, raised only when non-zero (3503) | `LogWarnIfNonZero` logged at a level its name did not say; the issue asks for the level the name claims. The zero branch said nothing worth a line |
| `"ClusterManager: ......Freed " + acquiredCount + " acquired trigger(s)."` | `"ClusterManager: ......Freed {Count} acquired trigger(s)."` at Warning (3507) | Same |
| `"ClusterManager: ......Deleted " + completeCount + " complete triggers(s)."` | `"ClusterManager: ......Deleted {Count} complete trigger(s)."` at Warning (3508) | Same, plus `triggers(s)` → `trigger(s)`: typo, corrected while the id is new |
| `"ClusterManager: ......Scheduled " + recoveredCount + " recoverable job(s) for recovery."` | `"ClusterManager: ......Scheduled {Count} recoverable job(s) for recovery."` at Warning (3509) | Same |
| `"ClusterManager: ......Cleaned-up " + otherCount + " other failed job(s)."` | `"ClusterManager: ......Cleaned-up {Count} other failed job(s)."` at Warning (3510) | Same |
| `"ClusterManager: ......Deferred recovery of " + deferredCount + " executing [DisallowConcurrentExecution] job(s)."` | `"ClusterManager: ......Deferred recovery of {Count} executing [DisallowConcurrentExecution] job(s)."` at Warning (3511) | Same |
| `PropertySettingJobFactory.HandleError`: `logger.LogWarning(message)` / `logger.LogWarning(e, message)`, the message interpolated at nine call sites, under `#pragma warning disable CA2254` | `"{Problem}"` at Warning (4009) | The message is assembled from the job type, the property and the value that would not go into it, and the same string is what the `Throw` disposition needs. The rendered text is unchanged and the pragma is gone |
| `XmlSchedulingDataProcessor.XmlValidationCallBack`: `logger.LogWarning(e.Message)` under `#pragma warning disable CA2254` | `"{ValidationMessage}"` at Warning (5017) | The text is the validating reader's own, which is what it always was. Rendered text unchanged, pragma gone |
| `"Removed  {Count} 'complete' triggers."` | `"Removed {Count} 'complete' triggers."` (3021) | Two spaces after `Removed`: typo, corrected while the id is new |
| `"Found {TriggerGroupDeleteCount}delete trigger group commands."` | `"Found {TriggerGroupDeleteCount} delete trigger group commands."` (5003) | No space after the placeholder: typo, corrected while the id is new |
| `"Adding trigger: … already existed.␣␣This is likely … cluster.␣␣Will try to reschedule instead."` (5026) and the same sentence single-spaced (5027) | One event, 5026, with the single-space spelling; **5027 is deliberately never allocated** | The two places that add a trigger and find one there spelled one message two ways. Two ids for one message is what the catalogue exists to prevent; typo, corrected while the ids are new |

**Not** in that table, because they are not changes: seven templates that the source spelled as
`"…" + "…"` across several lines were concatenations of *literals*, so they were already compile-time
constants and the attribute carries the identical string — `AdoJobStoreLog.ConnectionCloseFailed`,
`ClusterLog.RecoveredByAnotherInstance`, `ClusterLog.RecoveryDeferred`,
`CoreLog.GaveUpWaitingForRunningJobs`,
`ConfigurationLog.SchedulersShareDatabaseWithDifferentPrefixes`, `UtilityLog.TypeFoundUnderNewName`
and `XmlSchedulingLog.TriggerAlreadyExistedWillReschedule`.

## Benchmark

`dotnet run -c Release -- --filter *JobRunShell*`, same machine, back to back. AMD Ryzen 9 5950X,
.NET 10.0.11, BenchmarkDotNet 0.15.8. This is a developer workstation rather than a quiet box, and
the two runs are one sample each, so read the delta as "no regression, allocation identical" rather
than as a measured 1.5% win.

Base (`6f11175`):

| Method | Mean | Error | StdDev | Gen0 | Allocated |
|--- |---:|---:|---:|---:|---:|
| Success_NoTriggerListenersAndSingleJobListener_MayFireAgain | 520.0 ns | 3.31 ns | 3.10 ns | 0.0458 | 776 B |
| Success_NoTriggerListenersAndSingleJobListener_MayFireAgain_WithScopedLogger | 527.5 ns | 3.23 ns | 3.03 ns | 0.0458 | 776 B |

This branch:

| Method | Mean | Error | StdDev | Gen0 | Allocated |
|--- |---:|---:|---:|---:|---:|
| Success_NoTriggerListenersAndSingleJobListener_MayFireAgain | 512.4 ns | 8.09 ns | 7.57 ns | 0.0458 | 776 B |
| Success_NoTriggerListenersAndSingleJobListener_MayFireAgain_WithScopedLogger | 519.2 ns | 4.88 ns | 4.33 ns | 0.0458 | 776 B |

The shell's own Debug calls already sat behind `IsEnabled` checks and the benchmark's logger is
disabled, so this path was the one with least to gain. What the change buys is everywhere the guard
was missing — the job store, the delegate, the semaphores — which this benchmark does not reach.

## For a reviewer to decide

* **Two classes in the 5000–5999 range, not one.** The area is "serialization, type loading,
  triggers, calendars, utilities and anything else", and its callers span six namespaces, so "one
  class per area, beside its callers" cannot both hold. `XmlSchedulingLog` (5000–5099) sits in
  `Xml/` because it is 36 of the range's 47 events on its own, and `UtilityLog` (5100–5199) takes the
  other 11. Say the word and they collapse into one.
* **`ConfigurationLog` is one class whose callers span two namespaces.** Most of the 4000 range is
  raised from `Quartz.Impl` — the scheduler factory, the thread pools, the job factories — and the
  class lives in `Configuration/` because that is the folder the area is named for. Those callers
  reach it through a `using Quartz.Configuration;`. The alternative was four classes in one range.
* **`"{Problem}"` and `"{ValidationMessage}"` are degenerate templates.** Both messages are built at
  runtime and neither has structure to give named placeholders to. Doing better would mean nine
  events for `HandleError` and rewriting the throw path that shares the string; I did not, because
  the spec said not to restructure for this.
* **Existing `if (logger.IsEnabled(…))` guards are left in place.** They are redundant now — the
  generated method checks for itself — but removing them is a mechanical sweep with its own diff, and
  in `SQLiteSemaphore` the flag is threaded through a parameter. Worth doing; not here.
* **`CA1848` is still in `NoWarn` in `Directory.Build.props`.** It has nothing left to say about
  `Quartz`, but the other three packages still need it. It comes out with part 4.

## Verification

```
dotnet test src/Quartz.Tests.Unit                   Passed: 3294, Failed: 0, Skipped: 4
dotnet test src/Quartz.Tests.AspNetCore             Passed:  433, Failed: 0, Skipped: 0
  src/Quartz.Tests.Integration (non-db categories)  Passed:  158, Failed: 0, Skipped: 0
node docs/.vuepress/check-links.mjs                 no dead links or anchors
grep -rnE "\.Log(Trace|Debug|Information|Warning|Error|Critical)\(|\.Log\(LogLevel" src/Quartz
                                                    no matches
```

Docker was up, so the basic integration leg really ran.

The public API baselines are untouched: every generated class is `internal`. The migration guide
gains the range table, the one level change, and a correction — two entries said
`LogWarnIfNonZero` is private, and it is now gone.

### DIFF
--- a/docs/documentation/quartz-4.x/migration-guide.md
+++ b/docs/documentation/quartz-4.x/migration-guide.md
@@ -2286,10 +2286,18 @@ Ids are allocated in ranges by area, and are stable from 4.0 onwards:
 | 6000–8999 | Reserved for `Quartz.Plugins`, `Quartz.Jobs` and `Quartz.Extensions.Redis`, which are converting to the same shape |
 
 Levels and message templates are otherwise **unchanged from 3.x**, so a log query that matched a
-message before still matches it. The one exception is `JobStoreSupport.LogWarnIfNonZero`, which wrote
-the cluster recovery counts at Information when the count was non-zero and at Debug when it was zero,
-whatever its name said: those six messages are Warning events now, raised only when the count is
-non-zero. If you filtered them at Information, filter them at Warning.
+message before still matches it. Two exceptions:
+
+* `JobStoreSupport.LogWarnIfNonZero` wrote the cluster recovery counts at Information when the count
+  was non-zero and at Debug when it was zero, whatever its name said. Those six messages are Warning
+  events now, raised only when the count is non-zero. If you filtered them at Information, filter them
+  at Warning.
+* Four messages carried a typo, corrected while their ids were new: `Removed  {Count} 'complete'
+  triggers.` lost a double space, `complete triggers(s)` became `complete trigger(s)`, `Found
+  {TriggerGroupDeleteCount}delete trigger group commands.` gained the missing space, and the message
+  about a trigger that already existed — spelled with double spaces in one place and single spaces in
+  the other — is one event with the single-space spelling. A query matching those on text needs
+  updating; one matching on event id does not, which is rather the point.
 
 ## The ambient logger factory stays ambient
 
--- a/docs/documentation/quartz-4.x/migration-guide.md
+++ b/docs/documentation/quartz-4.x/migration-guide.md
@@ -2264,6 +2264,33 @@ inject, which the next section lists.
 
 Further information on configuring Microsoft.Logging can be found [at Microsoft docs](https://docs.microsoft.com/en-us/dotnet/core/extensions/logging).
 
+### Every message carries an event id
+
+The `Quartz` package logs through source-generated `[LoggerMessage]` methods rather than
+`logger.LogInformation(…)` and its siblings. Two things follow. Nothing is formatted or boxed when the
+level is off, including on the scheduling loop. And every message has a stable event id, so an operator
+can filter or alert on the event rather than on its text.
+
+Ids are allocated in ranges by area, and are stable from 4.0 onwards:
+
+| Range | Area |
+|---|---|
+| 1000–1999 | Scheduler core — the scheduler, its firing loop, the job run shell, the signaler, the error listener |
+| 2000–2999 | `RAMJobStore` |
+| 3000–3499 | ADO.NET store, its connections and its driver delegate |
+| 3500–3599 | Clustering — check-in, failed-instance detection and recovery |
+| 3600–3699 | Misfire handling |
+| 3700–3799 | Lock handlers |
+| 4000–4999 | Configuration, dependency injection and hosting, including the thread pools and job factories |
+| 5000–5999 | Serialization, type loading, triggers, calendars, XML scheduling data and the utilities |
+| 6000–8999 | Reserved for `Quartz.Plugins`, `Quartz.Jobs` and `Quartz.Extensions.Redis`, which are converting to the same shape |
+
+Levels and message templates are otherwise **unchanged from 3.x**, so a log query that matched a
+message before still matches it. The one exception is `JobStoreSupport.LogWarnIfNonZero`, which wrote
+the cluster recovery counts at Information when the count was non-zero and at Debug when it was zero,
+whatever its name said: those six messages are Warning events now, raised only when the count is
+non-zero. If you filtered them at Information, filter them at Warning.
+
 ## The ambient logger factory stays ambient
 
 `LogProvider.SetLogProvider(ILoggerFactory)` is the one piece of mutable process-wide state left in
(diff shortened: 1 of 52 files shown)


lahma:
Re-triggering CI: the 15:13–15:33 GitHub startup failures left the SQL Server leg with no run and cancelled the PostgreSQL and Firebird legs mid-flight (concurrency cancel-in-progress). Reopening immediately.

lahma:
Merging with the SonarCloud gate red (46.4% coverage on new code). The nine `*Log.cs` files contribute zero lines to cover — the declarations are not executable — so the "new code" is the 224 converted call-site lines themselves, counted as new because `logger.LogWarning(…)` became `logger.SomethingHappened(…)`. They sit where they always sat, mostly on error and rollback paths the tests do not drive, with the same coverage they had before the change. This is the changed-line attribution artefact already overridden on #3318 and #3397; nothing here can be tested into coverage without inventing failure injection for every log line.

PR #3451, opened 2026-08-26
Every fresh-install script can be told not to drop

### LINKED ISSUE #3417: Four fresh-install table scripts drop an existing schema with no switch to say no
Four of the eight fresh-install scripts under `database/tables/` destroy an existing schema with no way to say no.

`tables_sqlServer.sql`, `tables_sqlServerMOT.sql`, `tables_sqlServer_Below2016.sql` and `tables_postgres.sql` carry a `@DropDb` switch: every `DROP` sits inside `IF @DropDb = 1` (`tables_sqlServer.sql:10-12`; a `DO $$ … DECLARE DropDb INT := 1` block in the Postgres script), and the header says "Should you only require it to create the tables, set @DropDb to 0." The other four do not: `tables_firebird.sql:4-14` is bare `DROP TABLE …`, `tables_mysql_innodb.sql:8-19` and `tables_sqlite.sql:1-12` are `DROP TABLE IF EXISTS …`, and `tables_oracle.sql:11-35` runs `delete from qrtz_*` on all twelve tables and *then* drops them. `database/README.md` describes the scripts as "creates the current schema in full" and says nothing about the drop.

Surfaced by the review of #2183, whose per-table "drop only if empty" guards were declined (a half-dropped schema is worse than either outcome, and the `IF … AND NOT EXISTS(…)` form is not a runnable statement on MySQL or Firebird outside a stored program). The switch the four scripts already have is the right shape; it is just missing on the other four.

**Done means**

- Each of the four scripts gets the same switch in its dialect's idiom — a variable where the dialect has one, and where it does not (SQLite, MySQL outside a procedure) the drop block is clearly delimited and documented as the part to delete — with the same header sentence the SQL Server script carries.
- `database/README.md`'s fresh-install paragraph says the scripts drop by default and how to turn that off, once, for all dialects.
- `MigrationScriptTest`'s fresh-install fixtures still pass on every dialect; a new case runs each script twice and asserts the second run with the drop switched off leaves the data of the first in place.
- `database/tables/` is the *current* schema and differs by design between branches, so this is `main` only; a `3.x` companion is welcome but not required by the byte-identical rule (which covers `migrations/`).

lahma:
Four of the eight fresh-install scripts under `database/tables/` destroyed an existing Quartz schema with no way to say no. `tables_sqlServer.sql`, `tables_sqlServerMOT.sql`, `tables_sqlServer_Below2016.sql` and `tables_postgres.sql` have carried a `DropDb` switch for a while; `tables_firebird.sql`, `tables_mysql_innodb.sql`, `tables_oracle.sql` and `tables_sqlite.sql` did not. Now all eight do, and `database/README.md` says so once for all of them.

## The idiom per dialect, and why

**Firebird** — an `EXECUTE BLOCK` with `DECLARE VARIABLE DropDb INTEGER = 1`, and each drop issued as `EXECUTE STATEMENT` behind an `EXISTS(SELECT 1 FROM RDB$RELATIONS …)` check. This is the shape the repository's own Firebird migrations already use for conditional DDL, so it is proven in this codebase and in these tests. The existence checks are not decoration: Firebird has no `DROP TABLE IF EXISTS`, and without them the block would fail on the fresh database the old bare `DROP TABLE` lines also failed on. They are the reason `StartFirebirdSqlContainerAsync` no longer strips `DROP` lines out of the script before running it — the container now runs it as it ships.

**Oracle** — a PL/SQL block with `DropDb NUMBER := 1` and a nested `DropQuartzTable` procedure that runs `EXECUTE IMMEDIATE 'DROP TABLE …'` and swallows only ORA-00942. Oracle 21c and earlier have no `DROP TABLE IF EXISTS`, and a bare `EXECUTE IMMEDIATE` would abort the whole block on the first missing table. The twelve `delete from qrtz_*` statements are gone: a drop takes the rows with it, so emptying every table first was work for nothing, and it was also the part that made the script dangerous to read as harmless. The drops now go children first, so no foreign key stands in the way (the old order dropped `qrtz_calendars` first, which was only safe because nothing references it).

**MySQL** — a user variable plus `PREPARE`/`EXECUTE`, because MySQL has no statement-level `IF` outside a stored program. `SET @DropDb = 1` selects between the `DROP TABLE IF EXISTS` and `DO 0`, and the chosen statement is prepared and executed. The twelve tables go in one multi-table `DROP`, children before parents, which is both shorter than twelve prepared statements and verified to work through the foreign keys.

**SQLite** — no variables, no statement-level `IF`, no way to fake either. The honest answer is a delimited block: the drops sit between `-- === BEGIN DROP TABLES ===` and `-- === END DROP TABLES ===` markers, and the header tells the reader to delete what is between them. The script also gained the `-- this script is for SQLite` line it was missing.

Every one of these was verified against its real container before it was written into the script — the MySQL `PREPARE` path, the Firebird DDL-inside-`EXECUTE BLOCK` question the issue raised, and the Oracle block all ran end to end on `mysql:8.0`, `jacobalberty/firebird:v4.0` and `gvenzl/oracle-xe:21.3.0` first.

## Tests

`MigrationScriptTest` gains a `<Dialect>FreshInstallHonoursDropSwitch` case for all six dialects it has fixtures for — including SQL Server and PostgreSQL, which already had the switch and now have it pinned. Each case builds a schema from the shipped script under a prefix of its own (`QRTZS_`), inserts a job detail, and then runs the script's **teardown** twice: once with the drop switched off, where the row has to survive, and once as it ships, where every table has to go.

The teardown runs on its own rather than the whole script because it has to: with the drop declined, the `CREATE TABLE` statements that follow it cannot succeed against the schema the drop deliberately left standing. That is inherent to what the switch is for, not something this PR introduces — the second run of a fresh-install script was never a supported thing to do. Slicing at the first `CREATE TABLE` keeps the assertion about the switch and nothing else.

The switch is thrown the way each dialect's header tells a reader to throw it, and the test asserts the substitution actually changed something — so a script that quietly stops carrying the switch fails here rather than going untested. For SQLite the test strips the block by its markers, which pins that the markers the header sends the reader looking for still exist.

## Verification

```
dotnet build Quartz.slnx                       0 Warning(s), 0 Error(s)
dotnet test src/Quartz.Tests.Unit               Failed: 0, Passed: 3292, Skipped: 4  (52 s)
```

Full integration leg per database, `QUARTZ_TEST_DATABASE=<name>` with the `Build.cs` category filter, all against local Docker:

| Leg | Result | Wall clock |
|---|---|---|
| `db-sqlite` | Passed: 10, Failed: 0 | 59 s |
| `db-postgres` | Passed: 70, Failed: 0 | 2 m 23 s |
| `db-sqlserver` | Passed: 63, Failed: 0 | 2 m 28 s |
| `db-mysql` | Passed: 7, Failed: 0 | 1 m 17 s |
| `db-oracle` | Passed: 6, Failed: 0 | 59 s |
| `db-firebird` | Passed: 7, Failed: 0 | 1 m 24 s |

The new cases within those: `SqliteFreshInstallHonoursDropSwitch` 0.4 s, `PostgreSqlFreshInstallHonoursDropSwitch` 0.4 s, `SqlServerFreshInstallHonoursDropSwitch` 1 s, `MySqlFreshInstallHonoursDropSwitch` 1 s, `OracleFreshInstallHonoursDropSwitch` 6 s, `FirebirdFreshInstallHonoursDropSwitch` 3 s. Oracle and Firebird are slow to start but not slow to run.

The new case was confirmed to bite: mutating `SwitchDropOff` so the SQLite block is not actually removed fails it, with the drop taking the table and the count coming back -1.

`npm run docs:check-links` → `212 pages, 737 internal links, 424 fragments — no dead links or anchors`. `dotnet fallout VerifyDocsSnippets` → up to date, 90 markdown files checked.

## For a reviewer to decide

- **The `-1` in `CountJobDetailsAsync`.** A drop that ran when it was told not to takes the table, so the count throws. Catching `DbException` and returning -1 turns a provider stack trace into "expected 1, found -1, because the teardown has to do nothing once the drop is switched off". It is a deliberate trade of precision for a readable failure; say the word and it can go back to throwing.
- **Oracle losing its `delete from` statements** is a behaviour change beyond adding a switch, though only in the sense that the script no longer does something pointless. It is called out here rather than buried.
- **`database/tables/` differs by design between branches, so this is `main` only.** A `3.x` companion would be welcome but is not required — the byte-identical rule covers `migrations/`, which this PR does not touch.

Closes #3417


lahma:
Re-triggering CI: the 15:13–15:33 GitHub Actions startup failures cancelled a leg mid-flight (concurrency cancel-in-progress) and left this head without a complete check set. Reopening immediately.

lahma:
Merging with the SonarCloud gate red. Its two conditions are both artefacts of this change's shape: "0.0% coverage on new code" is the ten changed lines in `TestcontainersDatabaseEnvironment.cs`, integration-harness code that no unit test executes and that the migration legs exercised on every dialect; and the "D security rating" is `plsql:S1523` on `tables_oracle.sql:25` — `EXECUTE IMMEDIATE 'DROP TABLE ' || TableName` inside a local PL/SQL procedure whose only callers pass the script's own twelve literal table names, which is not dynamic input. Recorded here rather than suppressed in the script.

Issue #703, opened 2018-11-01
Concurrent jobs are fired sometimes


leotsarev:
### Actual behavior
Sometimes our cluster starts one job in parallel. They are DisallowConcurrentExecution, and in 99.9% cases that's are working, but _sometimes_ everything goes nuts.

Our system time is not stable, and our sysops explicitly not guarantee time sync between machines. Could it be related? How we can approve or disprove this?
### Steps to reproduce

                    ["quartz.scheduler.instanceName"] = config.SchedulerName,
                    ["quartz.scheduler.instanceId"] = "AUTO",
                    ["quartz.jobStore.clustered"] = "true",

                    ["quartz.threadPool.threadCount"] = "20",
                    ["quartz.jobStore.type"] = "Quartz.Impl.AdoJobStore.JobStoreTX, Quartz",
                    ["quartz.jobStore.useProperties"] = "true",
                    ["quartz.jobStore.dataSource"] = "myDS",
                    ["quartz.dataSource.myDS.connectionString"] = config.ConnectionString,
                    ["quartz.dataSource.myDS.provider"] = "SqlServer-20",

                    // if running MS SQL Server we need this
                    ["quartz.jobStore.lockHandler.type"] = "Quartz.Impl.AdoJobStore.UpdateLockRowSemaphore, Quartz"

alexorekhov:
@lahma Sorry if I bother you, but could you please take a look on this issue. We face it with our cluster too. I hoped that it had been resolved in 3.0.7, but, unfortunately, it's not. 
If it possible for you, please, pay your attention to the @AnSmol question.
Thanks in advance.

alexorekhov:
@lahma, thank you for your response. Could you please provide me a little beat more information about time-sync. What is a permissible discrepancy between hosts in a cluster? 
We use NTP, is that enough?

lahma:
@alexorekhov NTP should suffice, we are talking something more than a second being bad. With NTP you reach something like nano/milliseconds precision I guess. The queries are made relative to DateTime.UtcNow so if that differs between machines confusion ensues.

alexorekhov:
@lahma According to [MS Documentation](https://docs.microsoft.com/en-us/windows-server/networking/windows-time-service/support-boundary) 1-second accuracy is available only in Windows server OS starting on 2016. 
It said:

> Earlier versions of Windows (Prior to Windows 10 1607 or Windows Server 2016 1607) cannot guarantee highly accurate time.

[...]

PR #3452, opened 2026-08-26
Best practices is written for the scheduler we ship

### LINKED ISSUE #3444: best-practices.md is still the Java port, and it is the page both versions link to
`docs/documentation/best-practices.md` is 242 lines carried over from Java Quartz. It sits in the
version-neutral tier and is linked from both sidebars — `docs/.vuepress/configs/sidebar/en.ts:35` for
3.x and `:127` for 4.x — so it is the advice page a 4.x reader lands on. Its C# sample is 3.x
(`Task Execute(IJobExecutionContext)`) with a comment above it explaining that it is written out by
hand rather than compiled from `Quartz.Documentation.Samples`, because the samples project is 4.x.

Its section list is the Java article's: JobDataMap tips, job tips, trigger tips, ADO.NET JobStore,
daylight savings time, jobs, listeners, exposing scheduler functionality. What a reader actually
needs from a 4.x scheduler in production is spread across pages or nowhere.

## Done means

A rewrite covering, at minimum: idempotent jobs and `RequestsRecovery`;
`[DisallowConcurrentExecution]` against execution groups, and when each is the right tool; how to
choose a misfire instruction, by consequence rather than by name; the DST rules a scheduled job has to
live with; pool sizing, thread pool against connection pool, and why the two numbers relate; clock
synchronisation in a cluster; shutdown drain deadlines; and what to watch — the observability the
release now provides.

Written by `quartz-docs-researcher` with prior art surveyed: Java Quartz's own page, Hangfire's
guidance, and what the community has settled on. It stays version-neutral, so any 4.x-only advice is
marked as such. Every C# block becomes a compiled `#region sample_*` in
`Quartz.Documentation.Samples` with `<!-- snippet: -->` markers, per `CONTRIBUTING.md`; the 3.x
hand-written block goes.

`npm run docs:check-links` and `VerifyDocsSnippets` must pass.

Out of scope: 3.x-specific advice, which stays wherever it is, and any change to
`docs/documentation/quartz-3.x/`.

lahma:
`docs/documentation/best-practices.md` was the Java Quartz article's section list, carried across
with one hand-written 3.x sample and a comment explaining why it could not be compiled. It sits in
the version-neutral tier and both sidebars point at it, so a 4.x reader landed on advice that was
never about this implementation.

It is now organised around the decisions a deployment makes rather than the types it calls:

| Section | Answers |
|---|---|
| Designing a job | idempotence, what `RequestsRecovery` re-runs, what happens when a job throws, job data |
| Deciding what may run at the same time | `[DisallowConcurrentExecution]` against execution groups, and why held-back work misfires |
| Choosing a misfire instruction by its consequence | the three consequences, what `SmartPolicy` resolves to per family, and not starting a trigger in the past |
| Daylight saving, clock changes and cluster skew | the two DST rules, naming the zone, and what `CalcFailedIfAfter` assumes |
| Sizing the thread pool and the connection pool | permits against threads, "thread count plus three" re-derived, batching |
| Operating a scheduler | shutdown deadlines, one name per cluster, the trigger states, listeners, job-type security |
| What to watch | only the spans and instruments that exist today |

The page stays version-neutral. 4.x-only advice is marked as such (cluster-scoped execution limits,
the bounded shutdown drain, `MaxBatchSize` validation, metrics, the health-check registration and
its degraded state), and 3.x spellings are given beside the 4.x samples rather than duplicated as a
second code block. `docs/documentation/quartz-3.x/` is untouched.

Every C# block is a compiled `#region sample_*` in `Quartz.Documentation.Samples`; the hand-written
3.x block is gone. `troubleshooting.md`'s four links move to the new anchors.

## What the code said that the old page did not

Every Java-era claim was re-checked against this worktree. Three did not survive:

- **"If a job throws an exception, Quartz will typically immediately re-execute it."** It does not.
  `JobRunShell` catches the exception, wraps it in a `JobExecutionException` with
  `RefireImmediately` unset, and the trigger completes the firing normally. (The existing page had
  already half-corrected this; the mechanism is now stated.)
- **"1:05 AM may occur twice… 2:05 AM may never occur."** `CronExpression.GetTimeAfter` fires a
  fixed-time expression exactly once either way — shifted forward by the transition delta out of a
  spring-forward gap, and at the first of the two occurrences on fall-back. Both branches. Firing
  through *both* passes of a repeated hour is 4.x-only and only for interval expressions (a
  wildcard, step or range in the second, minute or hour field), which is the migration guide's
  documented 4.0 change.
- **Idempotence framed as a consequence of `RequestsRecovery`.** Without the flag an interrupted
  execution is *lost*, not repeated — so the default is at-most-once, not at-least-once. But four
  ordinary situations re-run work regardless of the flag, so the advice holds for a different
  reason than the one Java gives, and the page says which.

"Datasource max connection size of thread count plus three" was re-derived rather than repeated, and
still holds for 4.x: one per worker's `TriggeredJobComplete` write (a burst ceiling, not a steady
state — no connection is held across `Execute`), one for the scheduler thread, one for the misfire
handler (which runs clustered or not), one for the cluster manager. The row-lock semaphore is handed
the caller's connection rather than opening a second one, so a locked operation costs one connection
and not two. `+3` is exact clustered and one spare otherwise.

## Sources

Prior art (read August 2026), listed on the page itself:

- Java Quartz's own [best-practices page](https://www.quartz-scheduler.org/documentation/quartz-2.5.0/best-practices.html)
  — note the URLs under `documentation/best-practices.html` and `quartz-2.3.0/best-practices.html`
  now 404; `2.3.1-SNAPSHOT` and `quartz-2.5.0` are live — plus its configuration reference,
  JDBC-clustering page and tutorial lessons 3–6
- Hangfire: best practices, dealing with exceptions, throttling, cancellation tokens, and the
  `AutomaticRetryAttribute` / `SqlServerStorageOptions` source
- Sidekiq's Best Practices wiki; Kubernetes `CronJob` and pod lifecycle
- Microsoft: Idempotent Consumer, transient fault handling, thread-pool starvation, generic host
  shutdown, Azure VM time sync; NServiceBus/Particular on outbox and idempotence; Celery on
  per-worker rate limits
- HikariCP "About Pool Sizing" and the PostgreSQL wiki it cites; SqlClient and Npgsql pool defaults
- Kleppmann on distributed locking, Marc Brooker on leases and safety time, David Fowler's Async
  Guidance
- This repository's own issues and discussions, swept by search term — clustering and duplicate
  execution (~205 hits), misfires (165), stuck/blocked/deadlock, DST, thread pool, shutdown,
  clock skew

## Where the helper agents disagreed

Four independent agents ran: repository issue history, the Java/Hangfire/Sidekiq survey, the wider
.NET and distributed-systems consensus, and a read-only source audit of concurrent connection use.

1. **`HostOptions.ShutdownTimeout`: 5 seconds or 30?** The issue-history agent had 5s (from a
   discussion thread); the ecosystem agent had 30s (from the docs). Settled by running it:
   `new HostOptions().ShutdownTimeout` prints `00:00:30` on both net8.0 and net10.0. 5 seconds was
   the pre-.NET-6 default. The page says 30.
2. **Can a scheduler stop a job running twice?** Java's tutorial presents
   `@DisallowConcurrentExecution` without caveat; Hangfire says a feature of that shape "cannot be
   made reliable"; Sidekiq refuses to ship one. Neither position is right here, so the page states
   the mechanism instead: the check and the firing commit together under the cluster-wide
   `TRIGGER_ACCESS` lock against `QRTZ_FIRED_TRIGGERS`, which is stronger than a client-side lease —
   and the residual hole is a node wrongly declared dead, which is why the clock section exists.
3. **Is a job that is interrupted mid-flight re-run without `RequestsRecovery`?** The survey agent
   argued Quartz's framing was "arguably wrong" and that a crash after `TriggersFired` re-fires
   regardless. The code says otherwise: `ClusterRecover` counts that row as `otherCount` and deletes
   it. The minority view (Java's framing) was correct on this narrow point, and the page keeps it
   while giving the four other paths that do produce a second run.
4. **Long-running jobs: exit and reschedule, or hold the slot?** Java says exit; Hangfire says hold
   and honour a token. Resolved from `TaskSchedulingThreadPool`: `MaxConcurrency` is a
   `SemaphoreSlim` permit count over `TaskScheduler.Default`, so an awaiting job holds a permit but
   not a thread. Both rules are true of different resources, and the page says which is which —
   which also retires Java's "performance tanks in the several hundreds of threads", written about
   dedicated threads.

## Claims I could not verify, and things left out

- **"Ordinary NTP is within a few milliseconds on a LAN."** From chrony's documentation, not from
  anything in this repository. It is used only to say the requirement is easy to meet.
- **Azure's "VMs are paused for up to 30 seconds" during memory-preserving maintenance.** Microsoft's
  published figure, reported by a helper agent; I did not re-fetch it.
- **Hangfire's retry backoff spanning "about three hours"** is computed from its source's quartic
  formula as reported, not measured.
- **The maintainer's "raise it to a minute" advice** on cluster check-in is paraphrased from
  discussion #703; the page attributes it as a suggestion rather than a default.
- **Discussion #2248** — the clock-synchronisation question the issue pointed at — turns out to have
  no maintainer answer at all, and a second user re-asking the same thing 20 months later. The page's
  "Clocks in a cluster" section is written as the answer, but nothing in that thread was used as a
  source.

Deliberately left out:

- **Trigger priority.** Real and relevant to a full pool, but it is trigger documentation rather than
  a decision this page frames; it belongs with the trigger pages.
- **A "parked jobs" / dead-letter recommendation.** The neighbouring ecosystems all have one and
  Quartz.NET has no such surface, so recommending a pattern the library cannot support would
  oversell it. The page says instead that there is no retry policy at all and points at the
  rescheduling how-to.
- **Fencing tokens.** Correct advice in general and unactionable here — Quartz has no fencing
  surface for a job's own side effects — so the page says "idempotency where you cannot fence" and
  stops.

## Verification

```
dotnet build src/Quartz.Documentation.Samples/Quartz.Documentation.Samples.csproj   Build succeeded, 0 warnings
dotnet fallout DocsSnippets                                                          Updated docs\documentation\best-practices.md
dotnet fallout VerifyDocsSnippets                                                    up to date (90 markdown files checked)
npm run docs:check-links                                                             212 pages, 767 internal links, 436 fragments — no dead links or anchors
npm run docs:lint                                                                    144 findings, identical to the pre-change baseline; none in either changed page
dotnet test --filter SourceEncodingTest|AgentInstructionsTest|PackageReadmeTest      Passed: 58, Failed: 0
```

Closes #3444

## Review corrections (a4bfff9)

**Both "a pattern cron cannot express" examples were expressible**, as the review caught: the second
Monday of the month is `0 0 0 ? * MON#2` and the last weekday of March is `0 0 0 LW 3 ?`. Checking
the replacements against the real parser turned up a third: Quartz.NET cron has a day-of-week `/`
extension, so "every other Monday" — the review's own suggested replacement — is
`0 0 0 ? * MON/2`, and it holds a true fortnightly cadence across month and year boundaries
(2026-01-12, 01-26, 02-09, 02-23, 03-09, …).

What is genuinely out, each confirmed by trying to write it first:

| Pattern | Cron | RRULE |
|---|---|---|
| Second-to-last Friday of the month | `#` counts forwards and only to 5; `L` gives only the last. `0 0 0 ? * 6L-1` parses but returns exactly the `6L` series. | `FREQ=MONTHLY;BYDAY=-2FR` → 01-23, 02-20, 03-20, … |
| Every third day, not restarting monthly | `0 0 0 1/3 * ?` restarts at the 1st: … 01-28, **01-31, 02-01**, 02-04 … | `FREQ=DAILY;INTERVAL=3` → … 01-28, **01-31, 02-03**, 02-06 … |
| Every 90 minutes | rejected — "Increment > 59 : 90" | `FREQ=MINUTELY;INTERVAL=90` |
| Every 45 days | rejected — "Increment > 31 : 45" | `FREQ=DAILY;INTERVAL=45` |

The page now uses the first two, and leads by saying Quartz cron says more than most cron dialects
and should be checked before reaching past it.

**The "worth its own issue" note in my report was wrong, and is withdrawn.** The review is right that
it is not a bug: `InsertFiredTrigger` is called from `AcquireNextTriggers` with `job: null`, so the
row goes in at `ACQUIRED` with a null job name and `REQUESTS_RECOVERY = false`, and only
`ApplyTriggerFired` updates it to `EXECUTING` with `GetDbBooleanValue(update.JobDetail.RequestsRecovery)`.
`SqlSelectInstancesRecoverableFiredTriggers` filters on `REQUESTS_RECOVERY = @requestsRecovery`, so it
cannot select an acquired-but-unfired row and there is no second run. No issue filed. The page's
recovery section previously attributed the acquired/fired distinction to cluster recovery; it now
states the real mechanism — recovery selects executions and never reservations, on both paths — and
keeps only the release back to `WAITING` as cluster recovery's own.

Re-run after the fix: `VerifyDocsSnippets` up to date (90 files); `docs:check-links` 212 pages, 767
internal links, 436 fragments, no dead links or anchors; `docs:lint` 144 findings, unchanged from
baseline, none in either changed page.

### DIFF
--- a/docs/documentation/best-practices.md
+++ b/docs/documentation/best-practices.md
@@ -1,115 +1,640 @@
 ---
-
 title: Best Practices
 ---
 
 # Best Practices
 
-## JobDataMap Tips
+A scheduler is easy to start and hard to run. The API is small enough that the first job is working
+in an afternoon; what takes longer is deciding how often a job may run, which of two concurrency
+mechanisms bounds it, what a missed firing should do, what the clock does to a schedule, how large
+the two pools should be, and what to look at when it stops.
 
-### Only Store Primitive Data Types (including Strings) In the JobDataMap
+This page is those decisions. It is organised by the choice a reader is making rather than by the
+type they are calling, and every claim about Quartz.NET's behaviour has been checked against the
+code that ships. Where practice in the wider field diverges, the divergence is named rather than
+averaged.
 
-Only store primitive data types (including strings) in JobDataMap to avoid data serialization issues short and long-term.
+It applies to **both Quartz 3.x and Quartz 4.x**. Where the two differ — a name, a default, or a
+capability only one of them has — the sentence says so. The C# is 4.x, because that is what the
+samples project compiles; where 3.x needs a different spelling, the prose beside the sample gives
+it.
 
-### Use the Merged JobDataMap
+## Designing a job
 
-The JobDataMap that is found on the `JobExecutionContext` during Job execution serves as a convenience.
-The data in the JobDataMap is a merger of the JobDetail and the Trigger, with the data in the Trigger overriding
-any same-named value in the Job.
+### Assume the job will run more than once
 
-Storing JobDataMap values on a Trigger can be useful in the case where you have a Job that is stored in the scheduler for regular/repeated use by multiple Triggers,
-yet with each independent triggering, you want to supply the Job with different data inputs.
+Start from what Quartz.NET actually promises, because it is narrower than the folklore and wider
+than the Java page's version of it.
 
-We recommend that code within the `IJob.Execute(..)` method should retrieve
-values from the `MergedJobDataMap` on the `JobExecutionContext`, rather than directly
-from the JobDetail or Trigger.
+By default a firing is **at most once**. If a node dies mid-execution, the fired-trigger row it left
+behind is cleaned up and the occurrence is simply gone; nothing re-runs it, and the trigger carries
+on from its next scheduled time. Ask for recovery — the next section — and you have deliberately
+bought **at least once** instead.
 
-<!-- Quartz 3.x code — `Task Execute(IJobExecutionContext)` — so it is written out here rather
-     than compiled from the samples project, which is 4.x. -->
+That is the guarantee. It is not the whole story, because four ordinary situations produce a second
+run of work you thought ran once:
 
+- **A node wrongly declared dead.** Its peers release its acquired triggers and, for jobs that
+  request recovery, schedule the interrupted executions again — while it is still executing them.
+  See [clocks in a cluster](#clocks-in-a-cluster).
+- **Misfire catch-up.** A trigger set to ignore misfires fires every occurrence it missed, as fast
+  as the pool allows. See [choosing a misfire instruction](#choosing-a-misfire-instruction-by-its-consequence).
+- **A refire.** `JobExecutionException.RefireImmediately` re-runs the same firing on the same worker.
+- **Two schedulers that are not the cluster you think they are.** By far the most common cause in
+  practice, and the subject of [one name per cluster, one id per node](#one-name-per-cluster-one-id-per-node).
+
+The rest of the field does not hedge on this at all. Sidekiq: "Sidekiq will execute your job at
+least once, not exactly once. Even a job which has completed can be re-run… Sidekiq makes no
+exactly-once guarantee at all." Hangfire: "your background jobs can still be executed several times,
+due to re-queue on shutdown and other compensation logic that guarantees the *at least once*
+processing." Kubernetes says the same of `CronJob`: "the Jobs that you define should be
+*idempotent*." Java Quartz is the outlier, scoping idempotence to jobs marked recoverable, and a
+production deployment sees more than that.
+
+So write the job so a second run is uneventful. The order to try, which Microsoft's *Idempotent
+Consumer* guidance puts well, is to "design for natural idempotency first, and add deduplication
+techniques only for operations that can't be made naturally idempotent". A job that computes an
+absolute state and upserts it — a rollup, a cache refresh, a regenerated report — needs nothing
+else. A job with an external side effect needs a key.
+
+The key names the **occurrence**, not the firing:
+
+<!-- snippet: sample_best_practices_idempotent_job -->
 ```csharp
-public class SomeJob : IJob 
+public async ValueTask Execute(IJobExecutionContext context, CancellationToken cancellationToken = default)
 {
-    public static readonly JobKey Key = new JobKey("job-name", "group-name");
+    // The key names the occurrence, not the firing. A recovered execution arrives on a new trigger
+    // with a new fire instance id, so a key derived from either of those would never match the
+    // execution it is repeating.
+    string period = context.MergedJobDataMap.GetString("period")!;
+    string idempotencyKey = $"{context.JobDetail.Key}:{period}";
 
-    public Task Execute(IJobExecutionContext context) 
+    // Recording the key and doing the work commit together, and a unique index on the key is what
+    // settles a race between two executions rather than a read followed by a write.
+    await ledger.ChargeOnce(idempotencyKey, period, cancellationToken);
+}
+```
+<!-- endSnippet -->
+
+Two things about that key are worth being deliberate about.
+
+**Do not derive it from `FireInstanceId` or from the trigger.** A recovered execution arrives on a
+newly created trigger in the `RECOVERING_JOBS` group with a fresh fire instance id, so a key built
+from either is different on the run that is repeating the work — which is exactly the run the key
+exists to catch. `FireInstanceId` is the right identifier for *interrupting* a particular execution
+(`IScheduler.InterruptFireInstance`); it is the wrong one for deduplication.
+
+**Write the key and the effect in one transaction, and let a unique index settle the race.** A read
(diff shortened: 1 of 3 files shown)


lahma:
Re-triggering CI: the 15:13–15:33 GitHub Actions startup failures cancelled a leg mid-flight (concurrency cancel-in-progress) and left this head without a complete check set. Reopening immediately.

PR #3467, opened 2026-08-26
The plugins log through source-generated events

lahma:
Part 2 of 4 for #3414. The `Quartz.Plugins` package only; `Quartz.Jobs` and
`Quartz.Extensions.Redis` follow in their own pull requests. #3450 did `Quartz` and is the shape
this copies.

## What changed

Of the 43 `logger.Log{Trace,Debug,Information,Warning,Error,Critical}(…)` calls under
`src/Quartz.Plugins/`, 36 are now calls on source-generated `[LoggerMessage]` methods and 7 stay
plain, deliberately — see [What keeps a plain call](#what-keeps-a-plain-call). That makes **40
events**, more than the 36 converted calls, because the history plugins funnelled seven distinct
occurrences through three private `WriteInformation`/`WriteWarning` helpers, and each occurrence
now has an id of its own.

Five `internal static partial class <Area>Log` classes, each beside its callers, each drawing from
its slice of the 6000–6999 range `LogEventCatalogTest.Ranges` already reserved for this package:

| Range | Class | File | Events |
|---|---|---|---|
| 6000–6199 | `HistoryPluginLog` | `src/Quartz.Plugins/Plugins/History/HistoryPluginLog.cs` | 7 |
| 6200–6299 | `XmlSchedulingPluginLog` | `src/Quartz.Plugins/Plugins/Xml/XmlSchedulingPluginLog.cs` | 8 |
| 6300–6399 | `JsonSchedulingPluginLog` | `src/Quartz.Plugins/Plugins/Json/JsonSchedulingPluginLog.cs` | 14 |
| 6400–6499 | `InterruptPluginLog` | `src/Quartz.Plugins/Plugins/Interrupt/InterruptPluginLog.cs` | 8 |
| 6500–6599 | `ManagementPluginLog` | `src/Quartz.Plugins/Plugins/Management/ManagementPluginLog.cs` | 3 |

Gaps are left inside each range where a class covers two things: 6000–6009 is
`LoggingJobHistoryPlugin` and 6010–6019 is `LoggingTriggerHistoryPlugin`; 6300–6319 is the JSON
processor and 6320–6339 the plugin that feeds it files; 6400–6409 is the interrupt plugin and
6410–6419 the monitor that does the interrupting.

The two guards join up as part 1 laid out: `LogEventCatalogTest.Catalogued` gains one line and
`LogEventCatalogTest_Quartz.Plugins.verified.txt` is a new snapshot, `LogCallSiteTest.Converted`
gains `"src/Quartz.Plugins"`. The `CA2254` pragmas at the converted sites are gone. No public API
moved — every generated class is `internal`, and the public-API baselines are untouched.

## The history plugins took a degenerate template

`LoggingJobHistoryPlugin` and `LoggingTriggerHistoryPlugin` log a `string.Format` template **the
user configures** — `JobSuccessMessage`, `TriggerMisfiredMessage` and their siblings, with
`{0}`-style placeholders — which is documented, settable through `PluginConfigurationExtensions`,
and covered by tests. A template only known at run time cannot be a compile-time `[LoggerMessage]`
template.

So those seven sites format the configured template exactly as before and pass the result through
an event whose own template is `"{Message}"` — part 1's precedent for `"{Problem}"` and
`"{ValidationMessage}"`. The text a sink renders is unchanged. What is new is the id, and I gave
each occurrence its own rather than sharing one, because "the job failed" and "the job was vetoed"
have to stay separately filterable and that is the entire point of the catalogue:

| Id | Level | Occurrence |
|---|---|---|
| 6000 | Information | `JobToBeFiredMessage` |
| 6001 | Information | `JobSuccessMessage` |
| 6002 | Warning | `JobFailedMessage` (carries the `JobExecutionException`) |
| 6003 | Information | `JobWasVetoedMessage` |
| 6010 | Information | `TriggerFiredMessage` |
| 6011 | Information | `TriggerMisfiredMessage` |
| 6012 | Information | `TriggerCompleteMessage` |

Neither plugin has a configurable *level*, so no site needed the `LogLevel`-parameter form of
`[LoggerMessage]`. The existing `IsEnabled` guards stay: here they are not redundant, because they
guard the argument array and the `string.Format` that precede the call.

`LoggingJobHistoryPluginTest` and `LoggingTriggerHistoryPluginTest` gain a test each that
configures all the templates, drives every occurrence and asserts the rendered text and the id —
the `{Message}` indirection is the one place this change could have altered what an operator sees,
so it is the one place with a test that would catch it. `RecordingLoggerProvider` records the
`EventId` for them.

## What keeps a plain call

`StructuredLoggingJobHistoryPlugin` and `StructuredLoggingTriggerHistoryPlugin` — seven calls —
keep their direct `logger.LogInformation(template, args)`, and are recorded in
`LogCallSiteTest.Allowed` with the reason.

These are the structured-logging counterparts of the pair above: their configured templates carry
**named** placeholders (`{JobGroup}`, `{TriggerName}`, `{FireTime}`) which MEL turns into
properties a Serilog or NLog sink indexes. That is the whole reason the two plugins exist, and it
is what the `{Message}` treatment would destroy — and it cannot be done any other way, because
rendering a named template ahead of time needs MEL's internal `LogValuesFormatter`, and
`LoggerMessage.Define` tops out at six parameters where `JobToBeFiredMessage` takes eight. Every
one of the seven calls already sits behind an `IsEnabled` check, so the allocation the conversion
exists to avoid is already avoided; what they lack is an event id, and only dropping a documented,
tested feature would buy one.

`LogCallSiteTest` gains a second test with the first entries in that allow-list: an entry whose
file no longer makes a plain call is deleted, not kept, because from that moment it is a hole
nobody decided to leave open through which the next plain call walks.

## Message templates that changed

One, and it is a typo corrected while the id is new:

| Old | New | Why |
|---|---|---|
| `"Job's Interrupt Monitor has been scheduled to interrupt with the delay :{Delay}"` | `"…with the delay: {Delay}"` (6402) | The space was before the colon rather than after it |

Everything else is verbatim — same level, same template, same exception attached. Two pairs look
like duplicates in the snapshot and are not: `"Error while notifying SchedulerListener of error"`
and `"Error closing jobs file {FileName}"` are each raised from both the XML plugin and the JSON
plugin, and take an id in each range. That is the same call part 1 made for
`RAMJobStoreLog.TriggerReferencesMissingCalendar` (2002) and
`AdoJobStoreLog.TriggerReferencesMissingCalendar` (3032) — an id names one event in one place; only
call sites *within* one class share one. `"File named '{FileName}' does not exist"` (6325, JSON)
and `"File named '{FileName}' does not exist."` (6206, XML) differ by a full stop, which I left
alone: they are separate events either way, and normalising them would be a text change nobody
asked for.

## For a reviewer to decide

* **Seven ids for the history plugins, not one.** The spec I worked from could be read as "one
  shared `{Message}` event". I gave each occurrence its own id because filtering "the job failed"
  apart from "the job was vetoed" is exactly what an operator wants and the ids cost nothing. Say
  the word and they collapse.
* **The `Structured*` pair stays on plain calls.** This is the one place I did not do what the
  package-wide sweep says, and it is the trade recorded above: their event ids would cost a
  documented, tested feature. The alternative I did not take is dropping template configurability
  from those two plugins and minting fixed templates, which is a 4.x API decision, not a logging
  one.
* **`JsonSchedulingPluginLog` covers two types in two files.** The processor and the plugin are one
  area with one range, split by a gap at 6320 rather than into two classes, because 14 events over
  two adjacent files is not two areas.
* **`CA1848` is still in `NoWarn`.** Same as part 1: `Quartz.Jobs` and `Quartz.Extensions.Redis`
  still need it. It comes out with part 4.

## Verification

```
dotnet test src/Quartz.Tests.Unit                   Passed: 3334, Failed: 0, Skipped: 4
dotnet test src/Quartz.Tests.AspNetCore             Passed:  433, Failed: 0, Skipped: 0
  src/Quartz.Tests.Integration (non-db categories)  Passed:  315, Failed: 0, Skipped: 0
npm run docs:check-links                            212 pages, 769 links, 438 fragments, none dead
grep -rnE "\.Log(Trace|Debug|Information|Warning|Error|Critical)\(|\.Log\(LogLevel" src/Quartz.Plugins
                                                    7 matches, all in the two Structured* plugins
```

Docker was up, so the basic integration leg really ran.

The migration guide's range table gains the five plugin areas, the corrected message, and the
paragraph explaining why two history plugins have event ids and two do not.

**Expect the SonarCloud "coverage on new code" gate to go red**, for the reason recorded on #3450:
the `*Log.cs` files contribute no executable lines, so the "new code" is the converted call-site
lines themselves — counted as new because `logger.LogWarning(…)` became
`logger.SomethingHappened(…)` — sitting where they always sat, on error and rollback paths, with
exactly the coverage they had before. Same attribution artefact as #3318, #3397 and #3450; nothing
here can be tested into coverage without inventing failure injection for every log line.

### DIFF
--- a/docs/documentation/quartz-4.x/migration-guide.md
+++ b/docs/documentation/quartz-4.x/migration-guide.md
@@ -2266,10 +2266,10 @@ Further information on configuring Microsoft.Logging can be found [at Microsoft 
 
 ### Every message carries an event id
 
-The `Quartz` package logs through source-generated `[LoggerMessage]` methods rather than
-`logger.LogInformation(…)` and its siblings. Two things follow. Nothing is formatted or boxed when the
-level is off, including on the scheduling loop. And every message has a stable event id, so an operator
-can filter or alert on the event rather than on its text.
+The `Quartz` and `Quartz.Plugins` packages log through source-generated `[LoggerMessage]` methods
+rather than `logger.LogInformation(…)` and its siblings. Two things follow. Nothing is formatted or
+boxed when the level is off, including on the scheduling loop. And every message has a stable event id,
+so an operator can filter or alert on the event rather than on its text.
 
 Ids are allocated in ranges by area, and are stable from 4.0 onwards:
 
@@ -2283,7 +2283,12 @@ Ids are allocated in ranges by area, and are stable from 4.0 onwards:
 | 3700–3799 | Lock handlers |
 | 4000–4999 | Configuration, dependency injection and hosting, including the thread pools and job factories |
 | 5000–5999 | Serialization, type loading, triggers, calendars, XML scheduling data and the utilities |
-| 6000–8999 | Reserved for `Quartz.Plugins`, `Quartz.Jobs` and `Quartz.Extensions.Redis`, which are converting to the same shape |
+| 6000–6199 | The history plugins — `LoggingJobHistoryPlugin` and `LoggingTriggerHistoryPlugin` |
+| 6200–6299 | The XML scheduling data plugin |
+| 6300–6399 | The JSON scheduling data plugin and its processor |
+| 6400–6499 | The job interrupt monitor plugin |
+| 6500–6599 | The management plugins — the shutdown hook |
+| 7000–8999 | Reserved for `Quartz.Jobs` and `Quartz.Extensions.Redis`, which are converting to the same shape |
 
 Levels and message templates are otherwise **unchanged from 3.x**, so a log query that matched a
 message before still matches it. Two exceptions:
@@ -2298,6 +2303,19 @@ message before still matches it. Two exceptions:
   about a trigger that already existed — spelled with double spaces in one place and single spaces in
   the other — is one event with the single-space spelling. A query matching those on text needs
   updating; one matching on event id does not, which is rather the point.
+* One message in `Quartz.Plugins` carried a typo, corrected while its id was new: the interrupt
+  monitor's `…scheduled to interrupt with the delay :{Delay}` had the space before the colon rather
+  than after it.
+
+`LoggingJobHistoryPlugin` and `LoggingTriggerHistoryPlugin` are a special case, because the message
+they log is a template *you* configure — `JobSuccessMessage` and its siblings, with `{0}`-style
+placeholders — and a template only known at run time cannot be a compile-time one. They format it
+exactly as before and pass the result through an event whose own template is `{Message}`: the text a
+sink renders is unchanged, and each occurrence has an id of its own (6000–6003 for the job plugin,
+6010–6012 for the trigger plugin). `StructuredLoggingJobHistoryPlugin` and
+`StructuredLoggingTriggerHistoryPlugin` keep plain `ILogger` calls and have no event ids, because their
+configured templates carry *named* placeholders that a structured sink resolves into properties for
+itself — which is the whole reason those two plugins exist, and which no fixed template can preserve.
 
 ## The ambient logger factory stays ambient
 
--- a/src/Quartz.Plugins/Plugins/History/HistoryPluginLog.cs
+++ b/src/Quartz.Plugins/Plugins/History/HistoryPluginLog.cs
@@ -1,0 +1,70 @@
+#region License
+
+/*
+ * All content copyright Marko Lahma, unless otherwise indicated. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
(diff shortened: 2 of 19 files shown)


lahma:
Merging with the SonarCloud gate red on "45.0% Coverage on New Code" only — the changed-line attribution artefact recorded on #3450: the `*Log.cs` declarations contribute no lines to cover, and the converted call sites are the same error-path lines with the same coverage they had before. No rating, duplication or hotspot condition moved.

PR #3471, opened 2026-08-26
Both stores mean the same thing by a misfire

### LINKED ISSUE #3462: The threshold instant itself is a misfire on one store and not the other
A trigger due at exactly `now - MisfireThreshold` is a misfire on `RAMJobStore` and not on the ADO store's sweep — and the ADO store disagrees with itself.

- `RAMJobStore.ApplyMisfireNoLock` (`src/Quartz/Impl/RAMJobStore.cs:2388`) returns early only when the fire time is strictly *after* `now - MisfireThreshold`, so the threshold instant itself misfires: `<=`.
- `StdAdoConstants.SqlSelectMisfiredTriggersToRecover` (`src/Quartz/Impl/AdoJobStore/StdAdoConstants.cs:481`, and the count at `:248`) asks for `NEXT_FIRE_TIME < @nextFireTime`, so the sweep leaves the instant alone: `<`.
- `AdoJobStoreBase.UpdateMisfiredTrigger`, the ADO store's single-trigger path that a resumed trigger goes through, uses `<=` like the in-memory store.

Pinned by #3461's `MisfireMatrixTest.BothStoresAgreeOnTheThresholdInstant` (`Assert.Inconclusive` today) with the one-tick-either-side cells asserted on both stores. Java has the same asymmetry, which is how it arrived; it is one tick and it will bite nobody in production, but a contract suite that runs the same cases on every store should not have to carry a documented exception for it.

**Done means:** one comparison — `<=`, the in-memory store's and the ADO single-trigger path's, so the sweep's SQL gains an `=` — with the count query kept in step; the Inconclusive becomes an assertion; a sentence in `troubleshooting.md`'s misfire section says a trigger is misfired when its fire time is at or before `now - MisfireThreshold`. No migration: `<` to `<=` on the sweep is a statement change only.

### LINKED ISSUE #3463: RAMJobStore does not apply the misfire policy when it unblocks a trigger; the ADO store does
When a `[DisallowConcurrentExecution]` job completes, the ADO store applies the misfire policy of every trigger it unblocks in the same transaction (`AdoJobStoreBase.TriggeredJobComplete` → `RecoverUnblockedMisfires`, pinned by `TriggeredJobCompleteTest`). `RAMJobStore.TriggeredJobComplete` (`src/Quartz/Impl/RAMJobStore.cs` around `:538`) only returns the unblocked trigger to `timeTriggers`; its misfire policy runs on the next acquisition, where `ApplyMisfireNoLock` (`:2515`) sees it. `ResumeTrigger` (`:2137`) does apply it immediately, so the in-memory store is consistent with the ADO store on resume and not on unblock.

The observable difference: on the ADO store an unblocked trigger's `NextFireTimeUtc` and state are already updated when `TriggeredJobComplete` returns; on the in-memory store they update only when the scheduler thread next acquires, so a `GetTrigger` in between shows the stale, past-due fire time, and a `DoNothing` trigger past its end time is `Waiting` on one store and `Complete` on the other for that interval. Found by #3461's `MisfireWhilePausedOrBlockedTest`, which asserts each store's actual behaviour and reports the divergence as `Assert.Inconclusive`.

**Done means:** `RAMJobStore.TriggeredJobComplete` applies `ApplyMisfireNoLock` to each trigger it unblocks before re-queuing it, mirroring the ADO store; the Inconclusive becomes an assertion; the contract suite (`JobStoreContractTest`, on every dialect) gains the case so the two can never diverge again silently. This is the fourth store-parity finding of the 4.0 line (#3336 was the third); note it in the migration guide's store section only if 3.x behaves the same way — check `origin/3.x` and say which.

lahma:
Two store-parity findings about misfires, both of which made the same trigger mean different things
to the two stores. Based on `dec1bca52` (#3468), whose fixtures carry the store's clock into every
trigger — the assertions below are exact instants because of it.

## The threshold instant is late on every store (#3462)

`RAMJobStore.ApplyMisfireNoLock` declines only a trigger whose fire time is strictly *after*
`now - MisfireThreshold`, so the threshold instant itself is a misfire. `AdoJobStoreBase.UpdateMisfiredTrigger`
(the single-trigger path a resumed trigger goes through) and `RecoverUnblockedMisfires` say the same.
The ADO store's periodic sweep asked for `NEXT_FIRE_TIME < @nextFireTime`, so it disagreed with the
in-memory store — and with itself — about one tick.

One rule now: **a trigger is misfired when its fire time is at or before `now - MisfireThreshold`.**

Every misfire-time comparison in the store, and what happened to it:

| Site | Before | After |
|---|---|---|
| `StdAdoConstants.SqlSelectMisfiredTriggersToRecover` | `t.NEXT_FIRE_TIME < @nextFireTime` | `<=` |
| `StdAdoConstants.SqlCountMisfiredTriggersInStates` (the handler's double-check peek) | `NEXT_FIRE_TIME < @nextFireTime` | `<=` |
| `StdAdoConstants.SelectNextTriggerToAcquire` — the template behind `SqlSelectNextTriggerToAcquire` and every exclusion bucket | `NEXT_FIRE_TIME >= @noEarlierThan` | `> @noEarlierThan` |
| `RAMJobStore.ApplyMisfireNoLock` | `tnft > misfireTime` declines | unchanged — it is the rule |
| `AdoJobStoreBase.UpdateMisfiredTrigger` | `NextFireTimeUtc > misfireTime` declines | unchanged |
| `AdoJobStoreBase.RecoverUnblockedMisfires` | `NextFireTimeUtc > misfireTime` skips | unchanged |
| `SqlServerDelegate`, `PostgreSQLDelegate`, `MySQLDelegate`, `SQLiteDelegate`, `OracleDelegate`, `FirebirdDelegate` | each wraps the constant with its own paging/index hint | unchanged — none carries a comparison of its own, so all six moved with the constant |
| `PreferredNodeWhereClause`'s `LAST_CHECKIN_TIME + CHECKIN_INTERVAL * 10000 >= @liveNodeCutoff`, `AdoJobStoreBase.CalcFailedIfAfter` | node-liveness, on `ClusterCheckinMisfireThreshold` | **left alone** — a different threshold about a different question |

**The acquisition statement is the one a reviewer should weigh.** `@noEarlierThan` is bound to the
very `MisfireTime` the sweep asks about, so the two predicates are meant to partition the waiting
triggers: acquisition takes what is not misfired. Leaving it at `>=` while the sweep became `<=`
would have made the threshold instant both acquirable and misfired, which is how a trigger gets
fired late without its policy ever running. `<=` and `>` are exact complements, and that is what
`RAMJobStore` does (acquisition applies `ApplyMisfireNoLock` first and skips what misfired). It moves
the two `AcquisitionSqlTest` Verify baselines — reviewed line by line, the only delta is `>=` → `>`.

No schema change: `<` to `<=` is a statement change, and no migration was generated.

Tests: `MisfireMatrixTest.ThresholdEdgeCases` says `AdoMisfires: true` at the instant,
`TheAdoStoreLeavesTheThresholdInstantItselfAlone` is renamed to
`TheAdoStoreCountsTheThresholdInstantItselfAsLate`, and `BothStoresAgreeOnTheThresholdInstant`
asserts instead of `Assert.Inconclusive`. `StdAdoConstantsTest` gains
`TheAcquisitionPredicateShouldBeTheComplementOfTheMisfirePredicate`, which is what fails if the two
statements ever drift apart again.

## The in-memory store applies the policy of a trigger it unblocks (#3463)

A trigger blocked behind a `[DisallowConcurrentExecution]` job is out of the acquisition set and out
of the sweep's, so the completion that unblocks it is the first thing that can settle its missed
firing. `AdoJobStoreBase.TriggeredJobComplete` settles it there, in `RecoverUnblockedMisfires` and in
the same transaction. `RAMJobStore.TriggeredJobComplete` returned the trigger to `timeTriggers` and
nothing more, so until the next acquisition a caller reading it got the fire time it had already
missed, and a trigger past its end time read as `Normal` on one store and `Complete` on the other.

`RAMJobStore.TriggeredJobComplete` now runs `ApplyMisfireNoLock` over each trigger it unblocks — the
way `ResumeTriggerNoLock` already does — and re-queues what still has a fire time. What has none is
**removed**, rather than left as a `Complete` entry `GetTrigger` would keep handing back: that is
what the ADO path does with such a row, and it is what `MisfiredBlockedTriggerTest` (a fire-and-forget
trigger that misfires while blocked) has always expected of the ADO store. `PausedBlocked → Paused`
stays a state change and nothing more, since a paused trigger accrues no misfire.

Note this differs from the wording of the issue's done-list: the unblocked trigger with nothing left
to fire is **gone**, not `Complete`, because that is the ADO store's answer and parity is the point.

Worth knowing, though not new: `ApplyMisfireNoLock` notifies the trigger listeners (`TriggerMisfired`)
and, for a finalized trigger, the scheduler listeners, while the store's semaphore is held — so a
listener that calls back into the store deadlocks. That is already true of the two paths that reach
it, acquisition and resume; this makes it three. The notification moves from one moment to another
rather than being added: what the next acquisition used to raise, the completion raises now.

Tests: `MisfireWhilePausedOrBlockedTest`'s two per-store tests and the `Assert.Inconclusive` parity
report collapse into `BothStoresApplyAnUnblockedTriggersMisfireAsTheyUnblockIt`, which runs the same
expectation against both stores, plus `AnUnblockedTriggersMisfireIsNotAppliedTwice`.
`JobStoreContractTest` — which runs on `RAMJobStore` and on every dialect — gains
`UnblockingAnOverdueTriggerAppliesItsMisfirePolicy` (`FireOnceNow`: the fire time written is inside
the `TriggeredJobComplete` call) and `UnblockingATriggerWithNothingLeftToFireRemovesIt` (`DoNothing`
past the trigger's end time: `GetTrigger` is `null`, the state is `None`, and the job survives). Both
fail on `RAMJobStore` without this change.

## 3.x

Checked, and **3.x has both divergences**, in the same shape:

* `origin/3.x:src/Quartz/Impl/AdoJobStore/StdAdoConstants.cs` spells `NEXT_FIRE_TIME < @nextFireTime`
  in five statements (`SqlSelectMisfiredTriggers`, `…InGroupInState`, `…InState`,
  `SqlCountMisfiredTriggersInStates`, `SqlSelectHasMisfiredTriggersInState`), while
  `origin/3.x:src/Quartz/Simpl/RAMJobStore.cs`'s `ApplyMisfire` declines on `tnft.Value > misfireTime`.
* `origin/3.x`'s `RAMJobStore.TriggeredJobComplete` moves a blocked trigger to `Waiting`, adds it to
  `timeTriggers` and stops there, while `JobStoreSupport.TriggeredJobComplete` calls
  `UpdateMisfiredTrigger` for each trigger of the job that is now `WAITING` and removes the ones that
  land `COMPLETE` — with a comment saying it does so "so that GetTrigger returns null and the trigger
  doesn't linger in the database", which is the behaviour `RAMJobStore` now matches here.

No 3.x pull request opened, as instructed. Both are documented as 3.x behaviour in the migration
guide's "The two job stores answer the same way" section, and `troubleshooting.md` (which serves both
versions) says which way each behaves.

## Verification

Run locally on Windows, on `75401b22e`:

* `dotnet build Quartz.slnx` — succeeded, 0 warnings, 0 errors
* `dotnet test src/Quartz.Tests.Unit` — `Failed: 0, Passed: 3415, Skipped: 4, Total: 3419`
* integration `basic` leg — `Failed: 0, Passed: 329, Skipped: 0, Total: 329`
* integration `postgres` leg — `Failed: 0, Passed: 148, Skipped: 0, Total: 148`
* integration `sqlserver` leg — `Failed: 0, Passed: 140, Skipped: 0, Total: 140`
* `dotnet fallout VerifyDocsSnippets` — succeeded (90 markdown files checked)

Each new or changed assertion was also run against the unmodified store, to be sure it is a test:
`TheAdoStoreCountsTheThresholdInstantItselfAsLate(exactly on the threshold)`,
`BothStoresApplyAnUnblockedTriggersMisfireAsTheyUnblockIt`, `AnUnblockedTriggersMisfireIsNotAppliedTwice`,
`UnblockingAnOverdueTriggerAppliesItsMisfirePolicy` and `UnblockingATriggerWithNothingLeftToFireRemovesIt`
all fail without the two production changes.

Public API baselines are untouched, as expected: nothing about the signatures moved.

Closes #3462
Closes #3463

### DIFF
--- a/docs/documentation/quartz-4.x/migration-guide.md
+++ b/docs/documentation/quartz-4.x/migration-guide.md
@@ -2940,6 +2940,17 @@ of these is a 3.x behaviour, not a 4.x regression:
   the practical effect is that a trigger due at exactly that instant misfires rather than being fired
   late without its policy. 3.x behaves the old way.
 
+* **An unblocked trigger's misfire policy is applied as it is unblocked, in memory too.** When a
+  `[DisallowConcurrentExecution]` job finishes, the triggers of that job that sat `Blocked` behind it
+  go back to `Waiting` — and a trigger that passed its fire time while it waited has its misfire
+  policy applied there and then, so its state and `NextFireTimeUtc` are settled by the time
+  `TriggeredJobComplete` returns. The ADO store has always done this; `RAMJobStore` returned the
+  trigger to the acquisition set and left the policy to the next acquisition, so a `GetTrigger`
+  in between reported the fire time the trigger had already missed, and a trigger past its end time
+  read as waiting on one store and finished on the other. As on the ADO store, an unblocked trigger
+  whose policy leaves it with nothing to fire is removed rather than kept in `Complete`, so
+  `GetTrigger` answers `null` for it. 3.x behaves the old way.
+
 ## JobKey and TriggerKey Null Validation
 
 `JobKey` and `TriggerKey` now throw `ArgumentNullException` when you specify `null` for `name` or `group`. Triggers can no longer be constructed with a null group name. If your code was relying on null group names, switch to an explicit group name.
--- a/docs/documentation/quartz-4.x/migration-guide.md
+++ b/docs/documentation/quartz-4.x/migration-guide.md
@@ -2929,6 +2929,17 @@ of these is a 3.x behaviour, not a 4.x regression:
   store and not the other. `ObjectAlreadyExistsException` derives from `JobPersistenceException`, so
   code catching the base type is unaffected.
 
+* **The threshold instant itself is a misfire on the ADO store too.** A trigger is late once its fire
+  time is *at or before* `now - MisfireThreshold`, and that is now one comparison wherever the
+  question is asked. The ADO store's periodic sweep asked for `NEXT_FIRE_TIME < @nextFireTime` while
+  everything else — `RAMJobStore`, and the ADO store's own single-trigger path that a resumed or
+  unblocked trigger goes through — used `<=`, so the ADO store disagreed with the in-memory one and
+  with itself about one tick. The acquisition statement moved with it, from
+  `NEXT_FIRE_TIME >= @noEarlierThan` to `>`, so that a waiting trigger belongs to acquisition or to
+  the misfire handler and never to both. It is a change to the SQL only — no schema migration — and
+  the practical effect is that a trigger due at exactly that instant misfires rather than being fired
+  late without its policy. 3.x behaves the old way.
+
 ## JobKey and TriggerKey Null Validation
 
 `JobKey` and `TriggerKey` now throw `ArgumentNullException` when you specify `null` for `name` or `group`. Triggers can no longer be constructed with a null group name. If your code was relying on null group names, switch to an explicit group name.
--- a/docs/documentation/troubleshooting.md
+++ b/docs/documentation/troubleshooting.md
@@ -86,10 +86,17 @@ A **misfire** occurs when a trigger's scheduled fire time passes without the job
 
 ### How It Works
 
-1. On startup (and periodically during operation), Quartz scans for triggers whose `NEXT_FIRE_TIME` is older than `now - misfireThreshold`.
+1. On startup (and periodically during operation), Quartz scans for triggers whose `NEXT_FIRE_TIME` is at or older than `now - misfireThreshold`.
 2. For each misfired trigger, Quartz applies the trigger's configured misfire instruction.
 3. The default misfire threshold is 60 seconds for a persistent store — `JobStore:MisfireThreshold` in 4.x,
    `quartz.jobStore.misfireThreshold` as a flat key on both versions.
+
+A trigger is misfired when its fire time is at or before `now - misfireThreshold` — the threshold
+instant itself counts as late. In 4.x that is one rule wherever the question is asked: the in-memory
+store, the persistent store's periodic sweep, and the single-trigger path a resumed or unblocked
+trigger goes through all draw the line in the same place. On 3.x the persistent store's sweep is
+strictly *before* the threshold instant, so a trigger due at exactly `now - misfireThreshold` is
+misfired in memory and, for one tick, not in the database.
 
 ### Misfire Instructions by Trigger Type
 
--- a/src/Quartz.Tests.Integration/Impl/JobStoreContractTest.cs
+++ b/src/Quartz.Tests.Integration/Impl/JobStoreContractTest.cs
@@ -1876,6 +1876,130 @@ public abstract class JobStoreContractTest
     protected abstract string StoreInstanceId { get; }
 
     //////////////////////////////////////////////////////////////////////////////////////////////
+    // Unblocking the triggers of a job that forbids concurrent execution
+    //
+    // A trigger blocked behind a running execution is out of the acquisition set and out of the
+    // misfire sweep's, so however late it gets while it waits, the completion that unblocks it is the
+    // first thing that can settle the debt - and it is where both stores settle it. A store that
+    // instead left the policy to the next acquisition would hand a caller a past-due fire time in the
+    // window between, which is the divergence #3463 reported.
+    //////////////////////////////////////////////////////////////////////////////////////////////
+
+    [Test]
+    public async Task UnblockingAnOverdueTriggerAppliesItsMisfirePolicy()
+    {
+        // FireOnceNow: the missed firing is owed and is fired as soon as the scheduler gets to it, so
+        // the unblocked trigger comes back due now rather than due in the past.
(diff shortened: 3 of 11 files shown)


lahma:
Merging with the SonarCloud gate red on "47.6% Coverage on New Code" only. The uncovered lines are the in-memory unblock path in `RAMJobStore.TriggeredJobComplete`; they are executed by `JobStoreContractTest` (on RAM and every dialect) and by `MisfireWhilePausedOrBlockedTest` in the integration legs, which the Sonar scan's coverage import does not include — it reads the unit and AspNetCore runs only. No rating, duplication or hotspot condition moved.

Issue #3440, opened 2026-08-26
The cluster lock is called a semaphore, which is not what it is


lahma:
`ISemaphore` is a mutual-exclusion lock over a named row: one holder, re-entrant per connection,
released on the same connection. A semaphore is a counted permit. The name arrived from the Java port
and has never described the type. Every implementation carries it —
`src/Quartz/Impl/AdoJobStore/ISemaphore.cs:28`, `DbSemaphore.cs:37`, `SelectForUpdateSemaphore.cs:35`,
`UpdateRowSemaphore.cs:46`, `PostgreSqlSelectForUpdateSemaphore.cs:35`,
`SqlServerMemoryOptimizedUpdateRowSemaphore.cs:32`, `SimpleSemaphore.cs:37`, `SQLiteSemaphore.cs:54`,
`src/Quartz.Extensions.Redis/RedisSemaphore.cs:56` — while the how-to that teaches them is already
called `docs/documentation/quartz-4.x/how-tos/lock-handler.md` and the builder method is already
`UseLockHandler`. The vocabulary has moved; the types have not.

Ratified 2026-08-26. The full set:

- `ISemaphore` → `ILockHandler`
- `DbSemaphore` → `DbLockHandler`
- `SelectForUpdateSemaphore` → `SelectForUpdateLockHandler`
- `UpdateRowSemaphore` → `UpdateRowLockHandler`
- `SqlServerMemoryOptimizedUpdateRowSemaphore` → `SqlServerMemoryOptimizedUpdateRowLockHandler`
- `PostgreSqlSelectForUpdateSemaphore` → `PostgreSqlSelectForUpdateLockHandler`
- `SimpleSemaphore` → `InProcessLockHandler`
- `SQLiteSemaphore` → `SqliteLockHandler`
- `RedisSemaphore` → `RedisLockHandler`
- `SemaphoreContext` → `LockHandlerContext`
- `ObtainLock` → `AcquireLock`; `ReleaseLock` unchanged

One pull request: types, members, `SimpleTypeLoader`'s legacy map keeping every 3.x and alpha spelling
resolving (`src/Quartz/Impl/SimpleTypeLoader.cs:115`), a migration-guide row per name, and the
public-API baselines re-approved. The documentation sweep replaces "semaphore" with "lock handler"
across `docs/documentation/quartz-4.x/` — 151 occurrences over six files, of which `migration-guide.md`
keeps the old spelling where it is quoting the old name.

[...]

PR #3473, opened 2026-08-26
Every shipped package logs through source-generated events

### LINKED ISSUE #3414: Logging is 292 unguarded Log* calls; make it source-generated with a pinned event catalogue

lahma:
Parts 3 and 4 of 4 for #3414, and the wrap-up that finishes it. #3450 did `Quartz`, #3467 did
`Quartz.Plugins`, and this is the shape they set.

## What changed

| Package | Plain calls | Events | Classes |
|---|---:|---:|---|
| `Quartz.Jobs` | 14 | 14 | `DirectoryScanJobLog` 7000–7099, `FileScanJobLog` 7100–7199, `NativeJobLog` 7200–7299, `SendMailJobLog` 7300–7399 |
| `Quartz.Extensions.Redis` | 11 | 10 | `RedisLockHandlerLog` 8000–8099 |
| `Quartz.AspNetCore` | 5 | 5 | `HttpApiLog` 9000–9099 |

Redis is ten events for eleven calls because the two sites that report a cancelled acquisition spell
one message; the rest are one for one. `DirectoryScanJobLog` splits its range by a gap the way part 2
did — 7000–7009 is the job, 7010–7019 the model that finds its listener.

`Quartz.AspNetCore` is not one of the four packages #3414 names. It is here because the issue's point
is that logging is unguarded, and finishing four packages while a fifth still ships five plain calls
leaves the guard with something outside it. Five calls, one class, no ambiguity about the range.

## The one template that changed

Everything else is verbatim — same level, same template, same exception attached. Only `NativeJob`'s
process output moved:

| Old | New | Why |
|---|---|---|
| `"{Type}>{Line}"` at Information for stdout and at Warning for stderr, `Type` interpolated from the stream name | `"stdout>{Line}"` at Information (7201) and `"stderr>{Line}"` at Warning (7202) | What follows the `>` is a line of the child process's own text, which has no structure this package can name — the degenerate-template case part 1 met as `{Problem}` and part 2 as `{Message}`. One event per stream is the honest shape: the text a sink renders is unchanged, and which stream a line came out of becomes an id an operator filters rather than a property value they match. A structured sink that indexed the `Type` property no longer gets one; that is the trade, and the migration guide says so |

`NativeJobStreamLoggingTest` runs a process and reads back what was logged. The catalogue snapshot
pins the two templates, but only a test that runs something can say the right one is called for the
right stream — a swapped branch is the single way this could have gone wrong with nothing else
noticing, and both tests fail when the branches are swapped. `RecordingLoggerProvider` records into a
`ConcurrentQueue` now, because `NativeJob` relays its two streams from a thread each.

`"Error consuming {Type} stream of spawned process."` stays one event with a placeholder: that one is
a real template with a real named value, not a line of somebody else's text.

## `RedisLockHandlerLog`, not `RedisSemaphoreLog`

The type that raises these events is still called `RedisSemaphore` until #3440 renames it. A catalogue
named after its caller is a catalogue that gets renamed when the caller does, and an event class is
exactly the thing that must not move — so it is named for the concept and #3440 will not touch it.

Half its vocabulary is word for word what `LockHandlerLog` raises at 3700–3799. They are separate
events, on the call part 1 and part 2 both made: an id names one event raised in one place, and an
operator filtering 8005 is asking about the Redis handler, not about whichever handler happened to say
the same sentence.

## The guards now cover everything that ships

* **`LogEventCatalogTest.Catalogued`** gains `Quartz.Jobs` and `Quartz.Extensions.Redis`, with
  `LogEventCatalogTest_Quartz.Jobs.verified.txt` and
  `LogEventCatalogTest_Quartz.Extensions.Redis.verified.txt`.
* **A companion `LogEventCatalogTest` in `Quartz.Tests.AspNetCore`** snapshots `Quartz.AspNetCore`.
  `Quartz.Tests.Unit` cannot reference an assembly that needs ASP.NET Core, which is the same split
  `PublicApiTest` already makes for the same reason. Each of the two owns the ranges for the
  assemblies it snapshots, so no range is written down twice and neither copy can go stale.
* **`LogCallSiteTest.Converted`** is every packable project under `src/`: `Quartz`, `Quartz.AspNetCore`,
  `Quartz.Dashboard`, `Quartz.Extensions.Redis`, `Quartz.HttpClient`, `Quartz.Jobs`, `Quartz.Plugins`,
  `Quartz.Plugins.TimeZoneConverter`, `Quartz.Serialization.Newtonsoft`. The last four had nothing to
  convert — they make no `ILogger` call at all — and are listed so that the day they start logging is
  the day the guard is already there.
* **`LogCallSiteTest.EveryPackableProjectIsCovered`** is new, and asserts the covered list *is* the
  packable list, in both directions. Without it a package added after this conversion would ship
  unguarded and would look exactly like a package with nothing to convert, because both spend their
  whole life passing every other test in the file. Finding the packable projects moved out of
  `PackageReadmeTest` into `ShippedProjects`, which is now two tests' question.

Not covered, deliberately, and the test's remarks say why: `Quartz.Examples*` and
`Quartz.Documentation.Samples` are application code shown to a reader, and application code logs with
`logger.LogInformation(…)` — a sample that routed its one message through a generated class would be
teaching the wrong lesson about how to use Quartz. `Quartz.Server`, `Quartz.Benchmark`,
`Quartz.Trimming.Canary` and the test projects have no operator to serve at all.

## `CA1848` comes out of `NoWarn`

Part 1 said it comes out with part 4, and it is out of `Directory.Build.props`. It is off again for
`$(IsPackable) != 'true'`, in `Directory.Build.targets` beside the banned-API rule that keys on the
same property and for the same reason — that is the boundary above, said to the compiler instead of to
a source scan. The two are complementary rather than redundant: `CA1848` sees a call in a `.razor`
file's `@code` block, which a scan of `.cs` files does not.

`StructuredLoggingJobHistoryPlugin` and `StructuredLoggingTriggerHistoryPlugin` do ship and do keep
their plain calls — part 2's decision, unchanged — so each carries a `[SuppressMessage]` with the
reason, in the file whose behaviour depends on it. `LogCallSiteTest.Allowed` already recorded the same
decision from the source side; now both sides say it, and neither can quietly drift.

## For a reviewer to decide

* **`Quartz.AspNetCore` was not in the issue's scope.** I converted it and gave it 9000–9099 anyway,
  for the reason above. If you would rather it were its own pull request, the third commit is
  self-contained and lifts out cleanly.
* **`"About to run {Command} {Temp}..."` keeps `{Temp}`.** The placeholder is named after a local
  variable rather than after the thing it holds, which is the command's arguments. I left it, because
  the campaign's rule is templates verbatim and a placeholder rename is a structured-property rename —
  the rendered text would not change but a dashboard querying `Temp` would break. Say the word and it
  becomes `{Arguments}` while the id is still new.
* **`NativeJob`'s two streams are two events.** The alternative is one event keeping `{Type}` as a
  placeholder, which preserves the property and costs the separate filterability; I took the reading
  the spec asked for and wrote the trade into the migration guide.
* **`EveryPackableProjectIsCovered` is more than #3414 asked for.** The issue asked for the covered
  list to grow; this makes it grow by itself. It is the piece I would most expect a reviewer to want to
  argue about, because it means a future package must decide about logging on the day it is added.
* **A second `LogEventCatalogTest`, in `Quartz.Tests.AspNetCore`.** ~90 lines of the reflection and
  formatting are the same as the one in `Quartz.Tests.Unit`. The alternative was making
  `Quartz.Tests.Unit` reference `Quartz.AspNetCore` and so the ASP.NET Core shared framework, which is
  a bigger change to a project that is otherwise plain, and which `PublicApiTest` already declined to
  make.

## Verification

```
dotnet test src/Quartz.Tests.Unit                   Passed: 3395, Failed: 0, Skipped: 4
dotnet test src/Quartz.Tests.AspNetCore             Passed:  448, Failed: 0, Skipped: 0
  src/Quartz.Tests.Integration (non-db categories)  Passed:  318, Failed: 0, Skipped: 0
QUARTZ_TEST_DATABASE=redis dotnet test
  src/Quartz.Tests.Integration --filter db-redis    Passed:   12, Failed: 0, Skipped: 0
grep -rnE "\.Log(Trace|Debug|Information|Warning|Error|Critical)\(|\.Log\(LogLevel" \
  src/Quartz src/Quartz.AspNetCore src/Quartz.Dashboard src/Quartz.Extensions.Redis \
  src/Quartz.HttpClient src/Quartz.Jobs src/Quartz.Plugins \
  src/Quartz.Plugins.TimeZoneConverter src/Quartz.Serialization.Newtonsoft
                                                    7 matches, all in the two Structured* plugins
```

Docker was up, so the basic and redis integration legs really ran. The same grep over `*.razor` finds
nothing.

The public API baselines are untouched: every generated class is `internal`, and both `PublicApiTest`s
pass unchanged.

Expect the SonarCloud coverage gate to go red, for the recorded attribution reason rather than for
anything in this diff.

Closes #3414.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -63,9 +63,6 @@
     <!-- Do not declare static members on generic types -->
     <NoWarn>$(NoWarn);CA1000</NoWarn>
 
-    <!-- For improved performance, use the LoggerMessage delegates -->
-    <NoWarn>$(NoWarn);CA1848</NoWarn>
-
     <!-- Rename type name X so that it does not end in 'Delegate', 'EventHandler', 'Permission' etc -->
     <NoWarn>$(NoWarn);CA1711</NoWarn>
 
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -16,4 +16,24 @@
     <AdditionalFiles Include="$(MSBuildThisFileDirectory)Analyzers/BannedSymbols.txt" Link="BannedSymbols.txt" />
   </ItemGroup>
 
+  <!--
+    CA1848 - "use the LoggerMessage delegates" - is on for every assembly we ship, and off for the ones
+    we do not. A shipped package's messages are somebody's production log, so they go through a
+    source-generated [LoggerMessage] method with an event id (#3414); LogCallSiteTest guards the same
+    boundary from the source side, and CA1848 is what catches a call the compiler sees and that scan
+    does not.
+
+    The examples and the documentation samples are the reason for the exception rather than a leftover.
+    They are application code shown to a reader, and application code logs with
+    logger.LogInformation(...) - a sample that routed its one message through a generated class would be
+    teaching the wrong lesson about how to use Quartz. Quartz.Server, the benchmarks, the trimming canary
+    and the test projects have no operator to serve at all.
+
+    $(IsPackable) is how this repository marks the projects that ship, and it is not final until the
+    project body has been evaluated, which is why this is here and not in Directory.Build.props.
+  -->
+  <PropertyGroup Condition="'$(IsPackable)' != 'true'">
+    <NoWarn>$(NoWarn);CA1848</NoWarn>
+  </PropertyGroup>
+
 </Project>
--- a/docs/documentation/quartz-4.x/migration-guide.md
+++ b/docs/documentation/quartz-4.x/migration-guide.md
@@ -2266,8 +2266,8 @@ Further information on configuring Microsoft.Logging can be found [at Microsoft 
 
 ### Every message carries an event id
 
-The `Quartz` and `Quartz.Plugins` packages log through source-generated `[LoggerMessage]` methods
-rather than `logger.LogInformation(…)` and its siblings. Two things follow. Nothing is formatted or
+Every shipped package logs through source-generated `[LoggerMessage]` methods rather than
+`logger.LogInformation(…)` and its siblings. Two things follow. Nothing is formatted or
 boxed when the level is off, including on the scheduling loop. And every message has a stable event id,
 so an operator can filter or alert on the event rather than on its text.
 
@@ -2288,10 +2288,12 @@ Ids are allocated in ranges by area, and are stable from 4.0 onwards:
 | 6300–6399 | The JSON scheduling data plugin and its processor |
 | 6400–6499 | The job interrupt monitor plugin |
 | 6500–6599 | The management plugins — the shutdown hook |
-| 7000–8999 | Reserved for `Quartz.Jobs` and `Quartz.Extensions.Redis`, which are converting to the same shape |
+| 7000–7399 | `Quartz.Jobs` — the directory scan job (7000–7099), the file scan job (7100–7199), the native job (7200–7299) and the send mail job (7300–7399) |
+| 8000–8099 | `Quartz.Extensions.Redis` — the Redis lock handler |
+| 9000–9099 | `Quartz.AspNetCore` — the HTTP API |
 
 Levels and message templates are otherwise **unchanged from 3.x**, so a log query that matched a
-message before still matches it. Two exceptions:
+message before still matches it. The exceptions:
 
 * `JobStoreSupport.LogWarnIfNonZero` wrote the cluster recovery counts at Information when the count
   was non-zero and at Debug when it was zero, whatever its name said. Those six messages are Warning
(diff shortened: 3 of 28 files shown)


lahma:
Merging with the SonarCloud gate red on "36.7% Coverage on New Code" only — the changed-line attribution artefact recorded on #3450 and #3467: the `*Log.cs` declarations contribute no lines to cover, and the converted call sites are the same lines with the same coverage they had. No rating, duplication or hotspot condition moved. This closes #3414: every shipped package logs through source-generated events and `CA1848` is on for all of them.

PR #3483, opened 2026-08-26
A scheduler can be authorized on its own name, on the API and on the dashboard

### LINKED ISSUE #3423: Authorization is one policy per process, so a tenant reaches every tenant's scheduler
`QuartzDashboardOptions` has a single `AuthorizationPolicy` (`src/Quartz.Dashboard/QuartzDashboardOptions.cs:45`)
and a single `ReadOnly` flag; `QuartzHttpApiOptions` (`src/Quartz.AspNetCore/AspNetCore/HttpApi/QuartzHttpApiOptions.cs`)
has no policy of its own at all — `RequireAuthorization(...)` on the mapped group applies uniformly to
every route, and the scheduler is a route segment inside it. So a user who may see one scheduler may
see all of them. `docs/documentation/quartz-4.x/multi-tenancy.md:702` says exactly this and tells the
reader to enforce it outside Quartz: a process per tenant, or a proxy that authorizes on the
`{schedulerName}` segment.

## Shape

The design is the maintainer's, and it is the framework's own shape rather than a Quartz-specific
delegate. `QuartzHttpApiOptions.SchedulerAuthorizationPolicy` and
`QuartzDashboardOptions.SchedulerAuthorizationPolicy` each name a policy, evaluated per request as

    IAuthorizationService.AuthorizeAsync(user, resource: new SchedulerResource(name), policy)

An application writes one `AuthorizationHandler<TRequirement, SchedulerResource>` and gets the
decision for every route and every page. The dashboard's scheduler selector hides schedulers the user
fails for; the HTTP API answers 403. Null keeps today's all-or-nothing behaviour, so nothing changes
for an application that does not set it.

## Done means

- Tests in `Quartz.Tests.AspNetCore` covering both surfaces: a handler that allows one scheduler and
  refuses another, asserted through the endpoints and through the selector.
- The public-API baselines for `Quartz.AspNetCore` and `Quartz.Dashboard` record the additions.
- `multi-tenancy.md` loses the "enforce that outside Quartz" paragraph and documents this instead.

Out of scope: per-scheduler `ReadOnly`, authentication of any kind, and claim conventions — the policy
is the application's to write.

lahma:
`QuartzDashboardOptions` had one `AuthorizationPolicy` and one `ReadOnly` flag, and `QuartzHttpApiOptions`
had no policy at all — `RequireAuthorization(...)` on the mapped group applies uniformly and the scheduler
is a route segment inside it. So a tenant who could reach one scheduler could reach every tenant's.

Both surfaces now take a `SchedulerAuthorizationPolicy` that is evaluated per request against the
scheduler the request is for. `null` — the default — is exactly today's behaviour on both, so nothing
changes for an application that does not set it.

## As built

**The shape is the framework's.** A new `public sealed record SchedulerResource(string SchedulerName)`
(namespace `Quartz`, in `Quartz.AspNetCore`, which the dashboard already references) is the resource
Quartz passes to `IAuthorizationService.AuthorizeAsync(user, resource, policyName)`. The application
writes one `AuthorizationHandler<TRequirement, SchedulerResource>` and it answers for every route and
every page. There is no delegate option and no claim convention.

**HTTP API.** `MapQuartzHttpApi` applies a convention to every endpoint it maps; the convention reads the
endpoint's own `RoutePattern` and wraps only the routes carrying `{schedulerName}`, so the rule belongs to
the route rather than to any handler and a route added later is covered by carrying the parameter like
every other one. The wrapper runs before the handler's arguments are bound and before `EndpointHelper`
looks the scheduler up, so a refusal is `403` whether or not the scheduler exists — a `404` only ever
answers a name the caller was allowed to ask about. `GET /schedulers` carries no `{schedulerName}`, so it
filters its own answer instead: registrations included, a tenant sees its own and learns nothing about the
others. The refusal body is RFC 7807 problem details with no `Quartz-ExceptionType` — that member names
the exception a failure came from, and a policy that said no is a decision, not a failure. The endpoints
that carry the check also gain `403` in their OpenAPI metadata.

**Dashboard.** Three places reach a scheduler and all three ask, through a scoped
`SchedulerAuthorization` service that reads the visitor from `AuthenticationStateProvider` (a rendered
dashboard is a circuit; its request is long gone):

- all three scheduler listings the dashboard writes into `SchedulerState` are filtered — the picker's, the
  **Schedulers** page's, and the overview's post-action refresh — so a name a visitor cannot use is never
  offered, and the count of tenants in the process is itself not something a tenant should learn;
- `DashboardLayout` renders `SchedulerNotAuthorized` in place of `@Body` for a scheduler the visitor fails
  for. The page component is never created, so its `OnInitializedAsync` never runs and no
  `IQuartzApiClient` member is called on their behalf. The header stays, so the visitor can switch to a
  scheduler they may see;
- `QuartzDashboardHub.JoinScheduler` refuses with a `HubException`. The live-events plugin broadcasts to a
  group named after the scheduler, so joining that group *is* reaching it; a silent no-op would look
  exactly like an idle scheduler. The hub has a caller of its own, so it passes `Context.User` in.

The three settings compose rather than replace: `AuthorizationPolicy` decides who is in,
`SchedulerAuthorizationPolicy` decides which schedulers they see, `ReadOnly` decides what anyone may
change.

**Validation.** `QuartzHttpApiOptionsValidator` refuses a policy name in a container with no
`IAuthorizationService` — a security setting that silently does nothing is worse than one that is missing.
A policy name no `IAuthorizationPolicyProvider` knows still fails at first use, the way ASP.NET Core makes
it fail everywhere else.

**Sample.** `MultiTenancySamples` compiles the handler (`tenant` claim equal to the scheduler name, matched
ignoring case as every scheduler lookup in Quartz is), the policy registration and both options.

## For a reviewer to decide

1. **A refusal is always `403`, never a `401` challenge.** The spec said "the policy decides… do not
   special-case", and the filter does not look at whether the caller is authenticated. It also does not
   call `ChallengeAsync` for an anonymous caller the way `AuthorizationMiddleware` would: with no default
   challenge scheme registered that throws, and an API is a machine interface that cannot be challenged
   interactively. The documented answer is to keep `RequireAuthorization()` on the mapped group, where the
   authorization middleware issues the `401`. Say if you want the challenge inside the filter instead.

2. **No dashboard-side validator.** The spec asked for an `IValidateOptions` refusal on both, but
   `AddQuartzDashboard` calls `AddSignalR()`, which calls `AddAuthorization()` internally (verified: after
   `AddSignalR()` the container already has `IAuthorizationService` and `IAuthorizationPolicyProvider`), so
   a dashboard-side check could never fail. It would be unreachable code, so it is not there. The HTTP API
   validator can and does fail, and is tested.

3. **The dashboard gate lives in `DashboardLayout`, not in each page.** One place, and it guarantees "no
   data fetched" by never creating the page — a per-page check would still let each page's own
   `OnInitializedAsync` run. The consequence: in *integrated* mode the host's router supplies its own
   `DefaultLayout`, so dashboard pages there are not wrapped by `DashboardLayout` and the frame does not
   run. That is now written down where a user reads it, in `packages/dashboard.md` under "One scheduler at
   a time", and linked from `multi-tenancy.md`: every listing the dashboard writes is filtered in both
   modes and so is the hub subscription, so nothing in the dashboard points a visitor at a scheduler they
   fail for — but an application that routes visitors under its own layout owns that part. If integrated
   mode ever grows its own chrome, the gate moves with it.

4. **A narrow fail-open window on the dashboard.** `_access` starts granted while
   `ActiveSchedulerName` is null, and a scheduler that becomes active is re-evaluated asynchronously. A
   foreign scheduler cannot become active through the UI — every listing that can set it is filtered — so
   this is defence-in-depth rather than the enforcement path. Closing it entirely would need a synchronous
   decision, which resource-based authorization does not offer.

## Found while verifying the integrated-mode claim

Checking whether a page could reach a foreign scheduler's data with the frame absent turned up a hole
that was not mode-specific, and it is fixed in `a6a35d2`. `Dashboard.RefreshSchedulersAsync` — the
overview's re-read after a start, standby, shutdown, pause-all or resume-all — was the third listing
written into `SchedulerState` and the only one that then picks the active scheduler itself, and it was
unfiltered. Shutting down the one scheduler a visitor was authorized for made
`SchedulerState.DefaultSchedulerName` fall through to a foreign one, repopulated the picker with every
name in the process, and `LoadDataAsync()` on the next line read that scheduler — all before the page
frame was consulted. `AnActionThatMovesTheOverviewOffItsSchedulerMovesItToOneTheVisitorMaySee` reproduces
it (it fails on the parent commit with `ActiveSchedulerName` `"globex"`, expected `"acme"`).

## Baselines

Re-accepted, with exactly this delta:

`PublicApiTest_Quartz.AspNetCore.verified.txt`

```diff
+        public string? SchedulerAuthorizationPolicy { get; set; }
+    }
+    public sealed class SchedulerResource : System.IEquatable<Quartz.SchedulerResource>
+    {
+        public SchedulerResource(string SchedulerName) { }
+        public string SchedulerName { get; init; }
```

`PublicApiTest_Quartz.Dashboard.verified.txt`

```diff
+        public string? SchedulerAuthorizationPolicy { get; set; }
```

One new wire snapshot, `SchedulerAuthorizationEndpointTest_ForbiddenProblemDetailsBody.verified.json`:

```json
{
  "type": "https://tools.ietf.org/html/rfc9110#section-15.5.4",
  "title": "Forbidden",
  "status": 403,
  "detail": "Not authorized for scheduler globex"
}
```

`SchedulerNotAuthorized.razor` is a Blazor component, which `PublicApiTest` excludes by design.

## Docs

`multi-tenancy.md` loses the "enforce that outside Quartz" paragraph; the honest limit that remains is
stated for what it is (`ReadOnly` is still process-wide, and there is no per-operation policy), and the
feature is a section under **Scheduler per tenant** with the compiled sample.
`packages/http-api.md` gains an options table where its three settings were a bullet list, plus a section
on the option; `packages/dashboard.md` gains the row and a **One scheduler at a time** section, which
carries the standalone-versus-integrated caveat that `multi-tenancy.md` links to rather than repeats; the
migration guide records the two options and the record.

Two pages outside the brief said 4.x has no per-scheduler policy and now would have been wrong:
`best-practices.md` ("Do not let users choose the job type", which the task named) and
`tenancy-patterns.md` ("What Quartz.NET does not give you"), which the first links to. Both now say which
version has what. `docs/documentation/quartz-3.x/` is untouched.

## Verification

```
dotnet build Quartz.slnx                    Build succeeded. 0 Warning(s) 0 Error(s)
dotnet test src/Quartz.Tests.Unit           Passed! Failed: 0, Passed: 3593, Skipped: 4, Total: 3597
dotnet test src/Quartz.Tests.AspNetCore     Passed! Failed: 0, Passed: 547, Skipped: 0, Total: 547
dotnet fallout VerifyDocsSnippets           Documentation snippets are up to date (93 markdown files checked)
npm run docs:check-links                    214 pages, 851 internal links, 502 fragments - no dead links or anchors
```

Rebased onto `843cd347` after #3482 made `EndpointHelper` an instance. The one conflict was
`GetAllSchedulers`'s last line, resolved to keep both sides: `endpointHelper.JsonResponse(result.ToArray())`,
the instance call over the filtered listing. `RequestDelegatesAreSourceGeneratedTest` still counts
`SchedulerEndpoints.cs = 13` — no route was added, and the generator still intercepts the listing after it
grew an `HttpContext` and an `IOptions<QuartzHttpApiOptions>` parameter, which is the thing that test
exists to catch.

`Quartz.Tests.Integration` references neither `Quartz.AspNetCore` nor `Quartz.Dashboard`, so nothing here
can reach it; Docker is available on this machine if you want it run anyway.

The analyzer findings on the first push are cleared in `5771e1e`: the route filter passes
`context.RequestAborted` and the hub passes `Context.ConnectionAborted` (S8949, MA0040), and the
route-pattern scan is an `Any` rather than a loop that returns out of itself (S3267). `LeaveScheduler`
takes the token too, so the two hub methods read alike — one line beyond what was reported, for
consistency within the class. The MA0003 "name the parameter" notes are on `new ValueTask<bool>(true)` and
on calls to local functions whose parameter names already match their arguments; naming them reads worse,
so they stand.

Closes #3423

### DIFF
--- a/docs/documentation/best-practices.md
+++ b/docs/documentation/best-practices.md
@@ -793,11 +793,16 @@ puts it correctly: allowing users to define whatever job they want "effectively 
 all sorts of vulnerabilities comparable/equivalent to Command Injection Attacks as defined by OWASP
 and MITRE". Offer a fixed set of job types and validate their parameters.
 
-The same caution applies to the management surfaces. The dashboard has a single authorization policy
-and a single read-only flag, and 4.x's HTTP API serves every scheduler in the container behind one
-uniformly applied policy. There is no per-scheduler policy; if different people should reach
-different schedulers, enforce that outside Quartz.NET. See
-[Tenancy Patterns](tenancy-patterns.md#what-quartz-net-does-not-give-you).
+The same caution applies to the management surfaces. On 3.x the dashboard has a single authorization
+policy and a single read-only flag, and there is no per-scheduler policy: if different people should
+reach different schedulers, enforce that outside Quartz.NET. On 4.x both surfaces take a
+`SchedulerAuthorizationPolicy` — `QuartzDashboardOptions` and `QuartzHttpApiOptions` — evaluated per
+request against a `SchedulerResource` naming the scheduler, so one
+`AuthorizationHandler<TRequirement, SchedulerResource>` holds each caller to its own scheduler. What
+a caller may *do* to the scheduler it reaches is still process-wide, through the dashboard's read-only
+flag. See
+[Authorizing a tenant on its own scheduler](quartz-4.x/multi-tenancy.md#authorizing-a-tenant-on-its-own-scheduler)
+and [Tenancy Patterns](tenancy-patterns.md#what-quartz-net-does-not-give-you).
 
 ## What to watch
 
--- a/docs/documentation/quartz-4.x/migration-guide.md
+++ b/docs/documentation/quartz-4.x/migration-guide.md
@@ -6230,6 +6230,36 @@ says nothing about a scheduler that has gone quiet — it keeps whatever it last
 shows executions from an arbitrary distance in the past. The window is measured on the scheduler's
 `TimeProvider` and applied when history is read as well as when it is written.
 
+## A scheduler can be authorized on its own name
+
+Both management surfaces served every scheduler in the process behind one uniformly applied policy, so a
+tenant who could reach one scheduler could reach all of them. Each now takes a policy of its own that is
+evaluated per request against the scheduler the request is for.
+
+| 4.0 preview | 4.0 |
+|---|---|
+| — | `Quartz.SchedulerResource(string SchedulerName)`, in `Quartz.AspNetCore` (new) |
+| — | `QuartzHttpApiOptions.SchedulerAuthorizationPolicy` (new, `null`) |
+| — | `QuartzDashboardOptions.SchedulerAuthorizationPolicy` (new, `null`) |
+
+Nothing is removed and nothing changes for an application that leaves them null: the API is still whatever
+`RequireAuthorization(…)` the application put on the mapped group, and the dashboard is still whatever
+`QuartzDashboardOptions.AuthorizationPolicy` says.
+
+Set one and Quartz asks
+`IAuthorizationService.AuthorizeAsync(user, new SchedulerResource(name), policy)` — ASP.NET Core's own
+resource-based authorization, so the application writes one
+`AuthorizationHandler<TRequirement, SchedulerResource>` and there is no Quartz callback type or claim-name
+convention to learn. The API answers `403` with problem details, decided before the scheduler is looked up
+so that a `404` only ever answers a name the caller was allowed to ask about, and filters
+`GET {ApiPath}/schedulers` to what the caller may act on. The dashboard filters its scheduler picker and
+its Schedulers page, renders a *not authorized* frame instead of a page for a scheduler the visitor fails
+for, and refuses the live-events hub group for it. Setting either option in a container with no
+authorization services is refused at startup.
+
+See [Authorizing a tenant on its own scheduler](multi-tenancy.md#authorizing-a-tenant-on-its-own-scheduler)
+for the handler and the registration.
+
 ## `CheckExists` is `Exists`
 
 Both overloads, on `IScheduler` and `IJobStore`:
--- a/docs/documentation/quartz-4.x/multi-tenancy.md
+++ b/docs/documentation/quartz-4.x/multi-tenancy.md
@@ -440,7 +440,10 @@ anonymous caller gets whatever the policy says, which is a `403` when it refuses
 passes for; a page opened on one they do not renders a "not authorized" frame and reads nothing about that
 scheduler; and the live-events hub refuses to subscribe a connection to its group. It composes with what
 was already there: `AuthorizationPolicy` decides who reaches the dashboard at all, this decides which
-schedulers they see once they are in, and `ReadOnly` still decides what anyone may change.
+schedulers they see once they are in, and `ReadOnly` still decides what anyone may change. That frame is
+the dashboard's own layout, so read
+[Standalone hosting is where this applies today](packages/dashboard.md#one-scheduler-at-a-time) before
+relying on it in an application that hosts the components under a layout of its own.
 
 Null — the default — is the behaviour every earlier release had, so nothing changes for an application
 that does not set it. Setting it in a container with no authorization services fails at startup rather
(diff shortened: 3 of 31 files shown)


lahma:
Closing and reopening to re-trigger the workflows, which did not start for 5771e1e27.

PR #3487, opened 2026-08-26
The console tour shows 4.0, and nothing it cannot show

### LINKED ISSUE #3448: Example 12 teaches remoting, which 4.0 removed, and it still ships
`src/Quartz.Examples/12_RemoteClientServerJobScheduling/` contains one file, `SimpleJob.cs`. The
example it is named for cannot exist: remoting is gone from 4.0, and `Quartz.HttpClient` with
`HttpScheduler` is what replaced it. A numbered directory in the tour that teaches nothing is worse
than a gap, because the reader assumes the number means something.

The other fifteen — `01_SimpleJobScheduler` through `16_RunningAsynchronousJobs` — have not been read
as a set since 4.0's API settled. `Quartz.Documentation.Samples` is now the canonical sample set: it
is in the solution, so a rotted sample fails `Compile`, and its regions are what the documentation
renders. The console tour is a second sample set with a different job to do, and it should either do
that job or go.

## Done means

- An audit of all sixteen against 4.x idiom: `ValueTask`, no `Async` suffix, `CancellationToken` on
  every async member, `TimeProvider` rather than `DateTime.Now`, and configuration through
  `AddQuartz` rather than a properties collection. Each example is listed as current, refreshed, or
  deleted, with a sentence for each.
- `12_RemoteClientServerJobScheduling` deleted, with `ExampleScheduler`'s registration and the
  numbering handled deliberately rather than leaving a hole.
- Examples that teach 3.x habits are rewritten.
- The console app survives only as the interactive tour — something a reader runs and watches — and if
  an example does not earn that, it is deleted rather than modernised.

Out of scope: `Quartz.Examples.Worker`, `Quartz.Examples.AspNetCore` and `Quartz.Examples.HttpClient`,
which are integration examples rather than the tour, and `Quartz.Documentation.Samples`, which is
governed by the documentation.

lahma:
`12_RemoteClientServerJobScheduling` held one file, `SimpleJob.cs`, and no example: the example it was
named for was remoting, which 4.0 removed. It is gone, and so are two others that had stopped earning a
number. The remaining thirteen were read as a set against the 4.x tutorial and refreshed.

## The audit

Judged against the tutorial pages under `docs/documentation/quartz-4.x/tutorial/`: configuration through
`QuartzSchedulerBuilder`/`AddQuartz`, `JobBuilder`/`TriggerBuilder`, `ValueTask` jobs with a
`CancellationToken`, `TimeProvider` rather than `DateTime.Now`.

| Was | Example | Teaches | 4.x-idiomatic? | Verdict |
|---|---|---|---|---|
| 01 | `SimpleJobScheduler` | schedule one job, start, watch it fire | builders yes; `DateTime.Now` in the job, no | **refreshed** — it scheduled a job a minute out and waited 65 seconds for one firing; ten seconds out, twenty of watching |
| 02 | `SchedulingCapabilitiesUsingSimpleTriggers` | the `ISimpleTrigger` vocabulary | yes | **refreshed** — three of its eight jobs could not fire inside the run, and the run was five and a half minutes |
| 03 | `SchedulingCapabilitiesUsingCronTriggers` | the `ICronTrigger` vocabulary | yes | **refreshed** — seven expressions, no word about what any of them meant, and five minutes of watching two of them; it now prints each expression beside its meaning and the time it resolves to |
| 04 | `JobParametersAndJobsStateMaintenance` | `JobDataMap` in, `[PersistJobDataAfterExecution]` across firings | builders yes; data set by mutating the built `IJobDetail` rather than `UsingJobData`, no | **refreshed** — it printed "Waiting 60 seconds" and waited 300 |
| 05 | `SchedulingJobsSettingMisfireInstructions` | smart policy vs `NowWithExistingCount` | yes | **refreshed** — it needed ten minutes because the misfire threshold was the default minute; five seconds instead, and the two policies part company inside one. Carried a dead `MisfireJob` no example referenced, and `StatefulDumbJob` names a concept 4.0 does not have (`SlowJob` now) |
| 06 | `JobExecutionExceptions` | `RefireImmediately`, `UnscheduleAllTriggers` | yes | **refreshed** — lightly; the denominator round-tripped through the data map as a string |
| 07 | `InterrupInProgressJobs` | `IScheduler.Interrupt` and the token it delivers | yes | **refreshed** — fifty interruptions seven seconds apart is six minutes; six of them, and the directory loses its typo |
| 08 | `ExcludeTimePeriodsUsingCalendars` | `AnnualCalendar` | yes | **refreshed** — it waited thirty seconds during which, by design, nothing ran at all. It keeps the holiday calendar, whose effect is the printed first fire time, and adds a `CronCalendar` blocking the first half of every minute, whose effect is a job going quiet and coming back while you watch |
| 09 | `TriggeringAJobUsingJobListeners` | an `IJobListener` chaining a second job | listener implemented all three members though 4.x gives each a default, no | **refreshed** — also a comment promising that nothing would run, above code that ran two jobs |
| 10 | `RunningJobsByPlugInXmlConfiguration` | the XML scheduling plugin | **no** — a `NameValueCollection` of `quartz.plugin.*` assembly-qualified type strings, which typed `UseXmlSchedulingConfiguration` replaced | **deleted** — it did what example 15 did, in the 3.x idiom. Its one live idea, that the plugin rescans the file, moved into the XML example |
| 11 | `RunningLargeNumberOfJobs` | throughput, and the thread pool as the limit | yes | **refreshed** → **10**; it spread 500 jobs over fifty seconds and then waited five minutes for them |
| 12 | `RemoteClientServerJobScheduling` | nothing — one orphaned job file | n/a | **deleted** — remoting is gone from 4.0; `src/Quartz.Examples.HttpClient` is the successor, and the tour's readme says so |
| 13 | `ClusteringJobsExecution` | a clustered ADO.NET store | `UsePersistentStore`/`UseClustering` yes; `InstanceId = "instance_one"` and a connection string compiled in from the *unit test project's* `TestConstants.cs`, no | **refreshed** — and it was broken three ways: its first job was built, printed and never scheduled; it waited an hour; and it could not run at all, because nothing in the project referenced a SQL Server driver, so it died on a missing `Microsoft.Data.SqlClient` |
| 14 | `RunJobsByPriorityWithTriggersPriority` | priority breaking a tie between simultaneous triggers | yes | **refreshed** → **11**; lightly — the shared scheduler helper and the token |
| 15 | `ConfigureJobSchedulingByUsingXmlConfigurations` | jobs and triggers from a file | yes | **refreshed** → **12**; absorbs 10's rescanning and job-history logging, and prints the path of the file to edit while it runs |
| 16 | `RunningAsynchronousJobs` | that a job may be asynchronous | yes, and that is the problem | **deleted** — in 4.0 `IJob.Execute` returns `ValueTask` and takes a `CancellationToken`; there is no other kind of job, so the lesson is vacuous. What was worth watching in it — the job continuing on a different pool thread after an await, and the token being what an interrupt delivers — moved into example 7, which already showed the interrupt half |

Thirteen from sixteen, and about 200 fewer lines.

## The numbering: renumbered, no gap

Nothing outside the project pins these numbers. The one "Example 13" in the repository is in
`docs/old_news.html`, a frozen news archive from the 1.x days; no tutorial page, no `quick-start.md`, no
package readme names a tour number. So the survivors renumber down and the tour runs 01 to 13
contiguously, rather than carrying a hole with a note explaining it. Clustering keeps 13 by arithmetic.

The numbers are positions in the tour and nothing more — which is what makes them safe to move. They are
the directory prefixes so that a menu entry leads straight to the code behind it, and `ExampleCatalog.cs`
is the one written-out list they both come from. That replaced a reflection scan that discovered
`IExample` implementations and ordered them by *namespace length*; deleting an example is now one line
rather than a hole nobody notices.

## What else changed

- **`IExample.Run` takes the token Ctrl+C cancels**, and every await forwards it — which `CA2016` enforces
  now that the parameter exists, and which is why this could not be a smaller diff. `Shutdown` is the one
  call that deliberately passes `CancellationToken.None`, with the reason written beside it: an interrupted
  watch should still leave the scheduler stopped in an orderly way.
- **`Watching.For(duration, what)` is the wait**, and it names what there is to see while it lasts. The
  waits are a tour's device, not an application's, and the class comment says so.
- **`Program` takes the example number and the logger as arguments** (`-- 5 --logger serilog`, `--list`,
  `--help`), so a run can be scripted; it still prompts when it has a terminal and was given nothing.
- **Dead weight gone**: `MisfireJob`, `log4net.config` beside a `log4net` package reference no logger in
  the project ever read (`Quartz.Server` still uses log4net; the examples never did), and the
  `<Compile Include="..\Quartz.Tests.Unit\TestConstants.cs" />` that let example 13 borrow the test
  project's connection string.
- **`src/Quartz.Examples/README.md`** is new: what the tour is, how to run it, one line per example, and
  what example 13 needs. The project is `IsPackable=false`, so `ShippedProjects.Find()` skips it and
  `PackageReadmeTest` does not apply — confirmed by running the tests.
- **Docs**: `quick-start.md` ends by saying how to run the tour, `tutorial/README.md` says which lessons
  have a runnable counterpart, and the migration guide's logging section links `Logging.cs` instead of
  naming the project and leaving the reader to find it.

## Two things a reviewer should decide

- **`GenerateInstanceId = true`, not the string `AUTO`.** The issue's sibling spec asked for `AUTO`;
  `advanced-enterprise-features.md` says `GenerateInstanceId = true` "is what the flat value `AUTO` used
  to mean", so the example teaches the typed form. Say the word and it becomes the flat key.
- **Example 13's shape changed.** It used to have each node schedule jobs into a group named after its own
  instance id, so two nodes ran two disjoint sets and demonstrated nothing about a cluster. All nodes now
  schedule the same three jobs into one fixed group with `Replace = true`, so whichever node acquires a
  trigger runs it — which is what a cluster *is*, and what `QueryClusterNodes()`, printed every fifteen
  seconds, then lets you watch.

## Verified

```
dotnet build Quartz.slnx
  Build succeeded.  0 Warning(s)  0 Error(s)

dotnet test src/Quartz.Tests.Unit/Quartz.Tests.Unit.csproj
  Passed! - Failed: 0, Passed: 3496, Skipped: 4, Total: 3500, Duration: 51 s

dotnet test src/Quartz.Tests.AspNetCore/Quartz.Tests.AspNetCore.csproj
  Passed! - Failed: 0, Passed: 523, Skipped: 0, Total: 523, Duration: 15 s

dotnet fallout VerifyDocsSnippets
  Documentation snippets are up to date (91 markdown files checked)

npm run docs:check-links
  docs: 212 pages, 793 internal links, 461 fragments
  no dead links or anchors
```

**All thirteen were run, not just built.** Highlights:

- **03** — the seven cron expressions parse and resolve as documented, and `0/20 * * * * ?` and
  `0,30 * * ? * MON-FRI` fired during the watch. `0 0 10am 1,15 * ?` became `0 0 10 1,15 * ?`: the hour
  field takes a number, and nothing else in the repository spelled an hour that way.
- **05** — the misfire lesson is now legible in under a minute. `slowJob1` on the smart policy shows
  `due at 23:20:10, started at 23:20:10` every time and runs every 12 s; `slowJob2` on
  `NowWithExistingCount` shows `due at 23:20:04, started at 23:20:11` and falls further behind each cycle.
- **08** — `chattyJob` fires at :30, :35, :40, :45, :50, :55 and then goes silent until the next :30.
- **13** — verified against a real SQL Server (`mcr.microsoft.com/mssql/server:2022-latest`, schema from
  `database/tables/tables_sqlserver.sql`) with **two nodes**. They shared the jobs; killing one had the
  survivor report `RYZEN3439229258221: Failed`, then
  `ClusterManager: detected 1 failed or restarted instances` /
  `Scheduled 1 recoverable job(s) for recovery`, then `cluster-demo.job_2 RECOVERING`. Two defects turned up
  in doing it and are fixed here: the cluster listing printed check-in times in UTC beside fire times in
  local (three hours apart on one screen), and `database/tables/tables_sqlserver.sql` opens with
  `USE [enter_db_name_here]`, so following the readme literally produces
  `Database 'enter_db_name_here' does not exist` — the readme says to fill the name in.

Integration tests were not run: they cover the job stores rather than the examples project, and nothing
here touches them. Docker was available, and was used for the clustering verification above.

## Update: the credential is out of the source (1af5d96e9)

SonarCloud blocked the gate on `ClusteringJobsExecutionExample.cs:59` — `secrets:S6703` (BLOCKER) and
`S2068` (MAJOR): the fallback connection string carried the sa password of the container the readme
tells you to start. Fixed the way #3453 settled it, by removing the credential rather than suppressing
the rule.

The example now has **no default**. It reads `QUARTZ_EXAMPLES_SQLSERVER`, and with the variable unset it
prints one line — which variable to set, and what shape the value takes, naming no value of its own —
and returns before building anything:

```
=== Clustering ===
Several scheduler instances sharing one database, load-balancing the work and recovering each other's. Needs SQL Server.

Set QUARTZ_EXAMPLES_SQLSERVER to a SQL Server connection string - server, database and credentials - for a database carrying the Quartz schema, and run this again. src/Quartz.Examples/README.md has the shape of the value and a docker command that produces the database.
Example finished.
```

That is a better failure than the old one regardless: it is immediate, where the compiled-in default
spent fifteen seconds timing out against a server nobody had started.

Two things beyond the literal on line 59:

- **The unreachable-database message no longer echoes the connection string back.** It used to print a
  constant; it would now print whatever the operator set, password and all, onto a console that log
  scrapers and screen recordings both reach.
- **The readme uses placeholders, not values.** It keeps the docker command and gains the `export`
  beside it, both with `<a strong password>` / `<the password you chose>`, so the repository holds no
  working credential even in prose. A `git ls-files src/Quartz.Examples | xargs grep -niE
  'password|pwd=|user id=|secret'` now returns only those two placeholders and one prose comment.

Also taken while there: **S1192**, `"group1"` four times over in each of the misfire and calendar
examples, is now a `const string Group` in each.

Re-verified after the change: `dotnet build Quartz.slnx` clean, `dotnet test src/Quartz.Tests.Unit`
3496 passed / 0 failed, example 13 run with the variable unset (output above), examples 5 and 8
re-run to confirm the `Group` const changed nothing.

Closes #3448


lahma:
Merging over the two SonarCloud checks. The quality gate's one failing condition is coverage on new code (0% against 80%): `Quartz.Examples` is a console tour with no test project, and nothing here changes that. The code-scanning check still reports the S6703/S2068 pair at line 59 of `ClusteringJobsExecutionExample.cs`, but at `1af5d96e9` line 59 is a doc comment — the credential was removed in that commit, the file reads its connection string from `QUARTZ_EXAMPLES_SQLSERVER` only, and SonarCloud's own analysis of this head lists no vulnerability; the alerts are the previous head's, and close with the branch. The sixteen other checks pass.

PR #3498, opened 2026-08-27
Each JSON serializer writes only what it can read back

### LINKED ISSUE #3494: Newtonsoft's plain-object trigger path loses the trigger's time zone
With trigger converters off — which is the default: `UseNewtonsoftJsonSerializer(registerTriggerConverters: false)` at `src/Quartz.Serialization.Newtonsoft/NewtonsoftJsonConfigurationExtensions.cs:30`, on `3.x` as well — a trigger is written as a plain object graph, and its `TimeZoneInfo` does not survive: Json.NET writes the zone as a whole object and reads it back as nothing, so the getter falls through to `TimeZoneInfo.Local`. Measured in #3491: a `CalendarIntervalTriggerImpl` stored with `Tokyo Standard Time` came back as `FLE Standard Time`. The same holds for `DailyTimeIntervalTriggerImpl` and `RecurrenceTriggerImpl`; `CronTriggerImpl` escapes because its zone rides on the `CronExpression`, which has a converter. The private `timeZoneInfoId` helpers on the first two exist for exactly this and are dead code — `DefaultContractResolver` does not serialize private members.

Blast radius is narrower than it reads: the ADO.NET store persists every shipped trigger type through a persistence delegate (`SIMPROP` columns — `RecurrenceTriggerPersistenceDelegate` included, `StdAdoDelegate.cs:147`), so this path is reached only by a trigger the store serializes as a blob — a custom trigger type — and by anything else that hands a trigger object to the serializer. It is silent either way.

**Done means**

- The Newtonsoft serializer carries a `TimeZoneInfo` converter that writes the zone's id and reads both the id and the old object form (which carries `Id`), so a blob written today still reads after the fix.
- Round-trip tests for the three affected trigger types with a non-local zone, both with and without trigger converters, plus a legacy-payload case in the old form.
- The dead `timeZoneInfoId` helpers go, or become the thing the converter uses.
- A migration-guide note. A `3.x` companion is worth having, since the default there is the same; decide it on the pull request.

Found by #3438's round-trip work; #3491 deliberately asserts nothing about the zone so as not to enshrine the bug.

### LINKED ISSUE #3495: System.Text.Json writes job-data values it cannot read back
A `JobDataMap` holding a `List<string>` or a `Dictionary<string, object>` serializes through `SystemTextJsonObjectSerializer` without complaint, and then throws `JsonSerializationException` on read. The blob is in the database by then and stays unreadable.

The read side is closed on purpose — the store's contract is source-generated and there is no polymorphic `object` deserialization, which is what keeps a trimmed publish honest. The write side is not closed to match: it accepts whatever the runtime can serialize, so the two halves of the serializer disagree about what a job-data value may be, and the disagreement surfaces at the worst possible time, when a job fires.

**Done means**

- Writing refuses a value the reader will not accept, at `Serialize` time, with the typed exception naming the key, the value's type, and `SystemTextJsonSerializerRegistry.AddTypeInfoResolver` as the way to declare an application type.
- The set of accepted value types is one list both sides read from, so they cannot drift again.
- `JobDataMapPortabilityTest` (#3491) gains the negative cases: the two types above, and a nested map, each refused on write with that message.
- One sentence on the job-data-map page saying which value types a persistent store accepts and how to add one.

Found by #3438's round-trip work.

lahma:
Two serialization defects #3491's round-trip work turned up. One commit each; both are about the same
thing from opposite ends — a writer that put something in the database the reader could not get back
out.

Closes #3494
Closes #3495

## A trigger written as a plain object graph keeps its time zone (#3494)

`UseNewtonsoftJsonSerializer` leaves `registerTriggerConverters` off by default, and with it off a
trigger is written property by property. Json.NET's default contract wrote a `TimeZoneInfo` as its
whole public surface — `Id`, `DisplayName`, `BaseUtcOffset`, all read-only — so reading the object back
set nothing and the trigger's getter fell through to `TimeZoneInfo.Local`. A trigger stored under Tokyo
fired on whichever zone the reading machine was in, silently.

`TimeZoneInfoConverter` (internal) writes the id and reads both the id and the old object form. It is
attached per property by `QuartzContractResolver` to members *typed* as a `TimeZoneInfo`, and is
deliberately not in the serializer's converter list — that list is consulted for a value's runtime type
wherever it appears, so a zone held in an `object`-typed slot (a job data map value) would have been
written as a bare string and lost the `$type` that path carries. Scoped this way it reaches every
trigger's `TimeZone` and nothing else; a job-data zone is written byte for byte as before.
`CalendarIntervalTriggerImpl`, `DailyTimeIntervalTriggerImpl` and `RecurrenceTriggerImpl` were
affected; `CronTriggerImpl` escaped because its zone rides on the `CronExpression`, which has always
had a converter. The private `timeZoneInfoId` helpers on those trigger implementations were meant for
exactly this and never worked — `DefaultContractResolver` does not serialize private members — so they
are gone and the converter does the job.

**What the tests prove.** `NewtonsoftTriggerRoundTripTest` is now parameterized on
`registerTriggerConverters`, so all five triggers round-trip both ways, and the zone is asserted on
every trigger that carries one against a zone the test machine is not in. Six of the new assertions
fail on the unfixed tree: the three affected triggers' converters-off round trips and the three legacy
reads.

**Legacy fixture provenance.** The payloads in `NewtonsoftLegacyTimeZonePayloadTest` are verbatim
output of the code at `347392bd` — `NewtonsoftJsonObjectSerializer` with its default settings,
serializing a trigger whose zone was `Tokyo Standard Time` — dumped to disk from a throwaway test
*before* the converter was written and pasted in unedited, Windows display names and all. "It still
reads" is therefore a fact about those bytes rather than a reconstruction of what they probably were.
Only `Id` is ever read, which is precisely why the rest of the object was worthless.

**Reviewer decision: the `3.x` companion.** Not opened, per the brief. It is mechanical, with four
things to know: the converter drops into 3.x's `Quartz.Converters` namespace and the same
`CreateSerializerSettings` list; there are **four** dead `timeZoneInfoId` helpers there, not three,
because 3.x's `RecurrenceTriggerImpl` has one too; 3.x source files carry a BOM; and the tests have to
be written fresh, because none of #3491's four exist on that branch. The existing
`JsonObjectSerializerTest_*.verified.txt` snapshots are the converters-*on* form (`"TimeZone": "UTC"`),
so they do not move.

## System.Text.Json refuses a job data value it cannot read back (#3495)

The read side is closed on purpose — string, bool, int, long, double, null or
`Dictionary<string, string>`, with no polymorphic `object` deserialization, which is what keeps a
trimmed publish honest. The write side was not closed to match, so a `JobDataMap` holding a
`List<string>` serialized happily and threw on read, with the blob already in the database.

`JobDataValues` is now the one declaration both halves read from — the set the writer admits and the
shapes the reader hands back, in one file. The refusal runs before the first byte is written, so a map
with one bad value in it puts nothing at all in the column, and the message names the entry, its
runtime type and `SystemTextJsonSerializerRegistry.AddTypeInfoResolver`. Enums are accepted by rule
rather than by name: they go out as a number and come back as one, and no list can enumerate an
application's own.

`DateOnly` and `TimeOnly` join `QuartzStoreJsonContext`. They have `DataMapExtensions` accessors and so
belong to the accepted set, and every type on that set has to be answerable with reflection off —
`StoreFormatSourceGenerationTest` now fails if the list and the context stop agreeing. This also closes
a latent gap: a `DateOnly` in job data would have failed a trimmed publish before.

**What the tests prove.** `JobDataMapPortabilityTest` gains the three negative cases (`List<string>`,
`Dictionary<string, object>`, a nested `JobDataMap`), each asserting the message names the key, the
type and `AddTypeInfoResolver`; and a positive control where an application type is refused without
`AddTypeInfoResolver`, written with it, and read back as the string map the store format gives for any
object. All four fail with the gate neutralised.

### Two things a reviewer should weigh

**1. The refusal is in `JobDataMapConverter`, so it applies to the HTTP API too.** That converter is
registered by `AddQuartzConverters`, which both the store serializer and `HttpApiJson` use, and both
have the same closed reader. Today a `List<string>` in job data means the server writes a response the
client cannot parse; after this the server refuses instead. Both are failures — the new one is earlier
and says what to do — but it is a behaviour change beyond the store, and putting the check only in the
store path would mean two gates instead of one list. The #3495 migration note now states this outcome
explicitly, and the exception says "Quartz's JSON format" rather than naming a database so that it reads
correctly in front of an HTTP caller. `Quartz.Tests.AspNetCore` is green (527 tests).

**2. `StdAdoDelegateTest.TestSerializeJobData` changed its expectation.** It asserted that a private
application class serializes once binary serialization is out of the picture. That still holds for
Newtonsoft and is now deliberately false for System.Text.Json, so the test states each case per
serializer. Flagging it because flipping a green test is worth a second opinion — the alternative is
letting `SystemTextJsonObjectSerializer` keep writing a blob that cannot be loaded.

## Verification

Run on Windows, against a live Docker daemon (29.5.3):

| | |
|---|---|
| `dotnet build Quartz.slnx` | Build succeeded, **0 Warning(s), 0 Error(s)** — trim/AOT analyzers clean, no new baseline entry |
| `dotnet test src/Quartz.Tests.Unit` | **Failed: 0, Passed: 3616, Skipped: 4, Total: 3620** |
| `dotnet test src/Quartz.Tests.AspNetCore` | **Failed: 0, Passed: 527** |
| integration, `basic` leg | **Failed: 0, Passed: 329** |
| integration, `postgres` leg | **Failed: 0, Passed: 149** (`AdoSchedulerTest`, which runs both serializers, is 18 of them) |
| `dotnet fallout PublishTrimmed` | Succeeded — canary passes every trigger, calendar, `JobDataMap`, `NameValueCollection` and `CronExpression`, plus the SQLite store and configuration binding, with `IsReflectionEnabledByDefault: False` |
| `dotnet fallout VerifyDocsSnippets` | Succeeded — 93 markdown files up to date |

The public API baselines did not move: the new converter and `JobDataValues` are internal, the removed
trigger members were private, and the registry gained only an internal method. `PublicApiTest` is green
inside the unit run above.

The `IntegrationTest` target is gated to `Host is GitHubActions && Linux`, so the two legs were
reproduced directly with the same `QUARTZ_TEST_DATABASE` value and category filter the target builds.
There is no `Quartz.Tests.AOT` project on this base — `Quartz.Trimming.Canary`, via `PublishTrimmed`
above, is what covers that ground.

## A `TimeZoneInfo` held as a job data *value*

Scoping the converter leaves this path exactly as it was, which is the point — but "as it was" means
broken, and that is worth knowing rather than assuming. Captured from the base tree: it writes
`{"$type":"System.TimeZoneInfo, …","Id":"Tokyo Standard Time", …}` and **throws on read**.
`QuartzContractResolver` sets `IgnoreSerializableInterface = true`, so Json.NET never uses
`TimeZoneInfo`'s `ISerializable` implementation; the public members it writes instead are all read-only
and there is no constructor to feed. Writing `$type` explicitly does not rescue it either — with the
slot declared `object`, Json.NET resolves the type name and then takes the object-contract path instead
of dispatching to the converter, so no converter can fix this case.

`NewtonsoftJobDataTimeZoneTest` therefore pins what scoping guarantees (the payload is unchanged,
asserted against the captured bytes) and records the read failure as the pre-existing limitation it is.
A job that needs a zone in its data should store the id. Repairing the `object`-typed case needs
something other than a converter and belongs in its own issue.

## Documentation

- `docs/documentation/quartz-4.x/migration-guide.md` — a section per issue, in the JSON Serialization
  part.
- `docs/documentation/quartz-4.x/tutorial/job-data-map.md` — which value types a persistent store
  accepts and how to declare one, under "Why string-safe storage matters".
- `docs/documentation/quartz-3.x/` untouched.

### DIFF
--- a/docs/documentation/quartz-4.x/migration-guide.md
+++ b/docs/documentation/quartz-4.x/migration-guide.md
@@ -2636,6 +2636,36 @@ catch (Newtonsoft.Json.JsonSerializationException e)
 catch (Quartz.JsonSerializationException e)
 ```
 
+### System.Text.Json refuses a job data value it cannot read back
+
+The System.Text.Json read side is closed on purpose — a stored job data value comes back as a string, a
+bool, an int, a long, a double, null or a `Dictionary<string, string>`, and there is no polymorphic
+`object` deserialization, which is what keeps a trimmed publish honest. The write side was not closed to
+match, so a `JobDataMap` holding a `List<string>` or a `Dictionary<string, object>` serialized without
+complaint and threw `JsonSerializationException` on read. The blob was in the database by then, and the
+failure belonged to whoever next ran the job.
+
+Writing now refuses such a value, naming the entry and the type:
+
+```
+Job data entry 'recipients' holds a System.Collections.Generic.List`1[[System.String, …]], which a
+persistent store cannot read back. A job data value has to be one of the types JobDataMap declares an
+accessor for (string, bool, char, int, long, float, double, decimal, DateTime, DateTimeOffset, TimeSpan,
+Guid, DateOnly, TimeOnly or an enum), a Dictionary<string, string>, or a type the application declares
+through SystemTextJsonSerializerRegistry.AddTypeInfoResolver. Anything with structure of its own has to
+be serialized by the job and stored as a string.
+```
+
+The set is the one `DataMapExtensions` declares an accessor for, plus `Dictionary<string, string>`, plus
+enums by rule. A type of your own is declared through `AddTypeInfoResolver` — the same registration a
+trimmed or native-AOT publish already needs — and is then written like any other, coming back as the
+string map the store format gives for any object.
+
+Nothing that used to round-trip stops working: what is refused is exactly what used to be written and
+then failed to load. If you were relying on a nested `JobDataMap` or a collection surviving, serialize it
+in the job and store the result as a string. Newtonsoft is unchanged, and `StoreJobDataAsStrings` was
+never affected.
+
 ### Custom trigger and calendar serializers are no longer static
 
 The static registration methods have been removed, because they wrote into process-global dictionaries: two
--- a/docs/documentation/quartz-4.x/migration-guide.md
+++ b/docs/documentation/quartz-4.x/migration-guide.md
@@ -2722,6 +2722,33 @@ unpinned it, and the same held for a custom `IJobStore` that persists serialized
 store was never affected: it keeps the pin in the `PREFERRED_NODE` and `PREFERRED_NODE_AUTO` columns and
 reapplies them to every trigger it reads, blob triggers included.
 
+### A Newtonsoft-serialized trigger keeps its time zone
+
+`UseNewtonsoftJsonSerializer` leaves `registerTriggerConverters` off by default, and with it off a trigger is
+written as a plain object graph. A `TimeZoneInfo` written that way came out as its whole public surface —
+`Id`, `DisplayName`, `BaseUtcOffset` and the rest, every one of them read-only — so reading the object back
+set nothing, and the trigger's getter fell through to `TimeZoneInfo.Local`. A trigger stored under
+`Tokyo Standard Time` fired on whichever zone the reading machine was in, and nothing said so.
+
+The serializer now carries a `TimeZoneInfo` converter, so a zone travels as its id — the same spelling the
+trigger and calendar serializers have always used:
+
+```diff
+- "TimeZone": { "Id": "Tokyo Standard Time", "DisplayName": "(UTC+09:00) Osaka, Sapporo, Tokyo", … }
++ "TimeZone": "Tokyo Standard Time"
+```
+
+Both forms are read, so blobs already written in the object form still load and pick their zone out of `Id`.
+`CalendarIntervalTriggerImpl`, `DailyTimeIntervalTriggerImpl` and `RecurrenceTriggerImpl` were affected;
+`CronTriggerImpl` was not, because its zone rides on the `CronExpression`, which has always had a converter
+of its own.
+
+The ADO.NET job store persists every trigger type Quartz ships through a persistence delegate rather than as
+a blob, so this reaches a scheduler only through a trigger type the store has to serialize — a custom one —
+or through code that hands a trigger to `IObjectSerializer` itself. The private `timeZoneInfoId` helpers on
+those trigger implementations, which had been meant for exactly this and never worked because Json.NET's
+default contract does not serialize private members, are gone.
+
 ### Newtonsoft types moved out of the core namespaces
 
 The `Quartz.Serialization.Newtonsoft` package used to put types in namespaces that read as if they came from the
--- a/docs/documentation/quartz-4.x/migration-guide.md
+++ b/docs/documentation/quartz-4.x/migration-guide.md
@@ -2666,6 +2666,13 @@ then failed to load. If you were relying on a nested `JobDataMap` or a collectio
 in the job and store the result as a string. Newtonsoft is unchanged, and `StoreJobDataAsStrings` was
 never affected.
 
+The check lives in the job data map converter, which the HTTP API shares with the store serializer, so
+it applies there too: a job whose map holds a `List<string>` — in a `RAMJobStore` as much as a
+persistent one — now fails to serialize into a `GET` response with the message above, where before the
+server sent a payload `Quartz.HttpClient` threw on reading. Both ends use the same closed reader, so a
+value the server cannot read back is one the client cannot either, and `AddTypeInfoResolver` is the
+answer on both.
+
 ### Custom trigger and calendar serializers are no longer static
 
 The static registration methods have been removed, because they wrote into process-global dictionaries: two
@@ -2760,8 +2767,8 @@ written as a plain object graph. A `TimeZoneInfo` written that way came out as i
 set nothing, and the trigger's getter fell through to `TimeZoneInfo.Local`. A trigger stored under
 `Tokyo Standard Time` fired on whichever zone the reading machine was in, and nothing said so.
 
-The serializer now carries a `TimeZoneInfo` converter, so a zone travels as its id — the same spelling the
-trigger and calendar serializers have always used:
+A `TimeZoneInfo` converter is now attached to every property *typed* as one, so a zone on a trigger travels
+as its id — the same spelling the trigger and calendar serializers have always used:
 
 ```diff
 - "TimeZone": { "Id": "Tokyo Standard Time", "DisplayName": "(UTC+09:00) Osaka, Sapporo, Tokyo", … }
@@ -2772,6 +2779,12 @@ Both forms are read, so blobs already written in the object form still load and 
 `CalendarIntervalTriggerImpl`, `DailyTimeIntervalTriggerImpl` and `RecurrenceTriggerImpl` were affected;
 `CronTriggerImpl` was not, because its zone rides on the `CronExpression`, which has always had a converter
 of its own.
+
+The converter is attached per property rather than registered on the serializer, so it changes typed members
+only. A `TimeZoneInfo` stored as a *value* in a `JobDataMap` sits in an `object`-typed slot and is written
+exactly as before, `$type` and all — that payload has never been readable back (a `TimeZoneInfo` has no
+constructor Json.NET can call), and this change neither introduces that nor repairs it. Store the zone's id
+if a job needs one in its data.
 
 The ADO.NET job store persists every trigger type Quartz ships through a persistence delegate rather than as
 a blob, so this reaches a scheduler only through a trigger type the store has to serialize — a custom one —
--- a/docs/documentation/quartz-4.x/tutorial/job-data-map.md
+++ b/docs/documentation/quartz-4.x/tutorial/job-data-map.md
@@ -167,6 +167,13 @@ stay readable by the new version — a renamed property on a stored options clas
 its next fire, months after the deploy that renamed it. Standard framework types are safe; your own
 types are a versioning commitment.
 
+With System.Text.Json — the default — a persistent store accepts exactly the types the accessors above
+cover (`string`, `bool`, `char`, the numeric types, `DateTime`, `DateTimeOffset`, `TimeSpan`, `Guid`,
+`DateOnly`, `TimeOnly` and enums) plus `Dictionary<string, string>`, and refuses anything else when the
+job is stored rather than writing a blob that fails to load later. To store a type of your own, declare
+it with `SystemTextJsonSerializerRegistry.AddTypeInfoResolver` — the same registration a trimmed or
+native-AOT publish needs — or serialize it yourself and store the result as a string.
+
 **String mode.** `AdoJobStoreOptions.StoreJobDataAsStrings` (the flat key is still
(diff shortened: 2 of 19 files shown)


lahma:
Both review points are in `d967d5a6`. One of them came out differently than expected, and the difference is worth reading before merge.

## 1. The converter is now scoped to typed members — done as asked

`QuartzContractResolver.CreateProperty` attaches `TimeZoneInfoConverter` when `property.PropertyType == typeof(TimeZoneInfo)`, and it is out of the serializer's converter list. That still reaches every trigger's `TimeZone` (all of #3494) and leaves `object`-typed slots alone. I captured a job-data payload before and after: **byte for byte identical**, `$type` and read-only properties and all. `NewtonsoftJobDataTimeZoneTest` asserts that payload verbatim, plus the contrast case that a trigger's typed member writes `"TimeZone":"Tokyo Standard Time"`.

The legacy object-form read stays on the converter, as you asked — a trigger blob written before the fix carries the object form on a typed member.

## 2. But the round-trip test cannot be written, because the round trip never worked

The review says a `TimeZoneInfo` job-data value "used to round-trip under Newtonsoft — `TimeZoneInfo` is `[Serializable]`/`ISerializable`, so Json.NET wrote it with `$type` and rebuilt it". I captured the base tree (`347392bdf`, before any of my commits) to get the bytes for the test, and it does not rebuild it:

```
wrote: {"zone":{"$type":"System.TimeZoneInfo, System.Private.CoreLib","Id":"Tokyo Standard Time",
        "HasIanaId":false,"DisplayName":"(UTC+09:00) Osaka, Sapporo, Tokyo", ... }}
READ THREW Quartz.JsonSerializationException: Could not deserialize JSON: {...}
```

The `$type` is written, but the read throws. The reason is one line in `QuartzContractResolver`'s own constructor:

```csharp
IgnoreSerializableInterface = true;
```

That keeps Json.NET off `ISerializable` entirely, so `TimeZoneInfo` gets an ordinary object contract. Every public member it then writes is read-only, and there is no constructor to feed, so the way back in fails. The `$type` is written and then useless — the same reason the *trigger* bug existed.

I checked whether it could be made to work rather than just reporting it. Writing `{"$type":"System.TimeZoneInfo, System.Private.CoreLib","Id":"..."}` — `CronExpressionConverter`'s shape — still throws: with the slot declared as `object`, Json.NET resolves the type name and then goes down the object-contract path instead of dispatching to the converter. A converter cannot fix this case.

So: **no regression to protect against.** The three states are

| | writes | reads |
|---|---|---|
| base `347392bdf` | `$type` + full object | throws |
| global converter (previous head) | bare string | a `string` |
| scoped converter (now) | `$type` + full object | throws |

The scoped version restores base exactly, which is what you wanted. It is worth being explicit that this restores a *broken* path rather than a working one — the previous head was, narrowly, the only one of the three that did not throw.

Rather than assert a round trip that has never held, `NewtonsoftJobDataTimeZoneTest` pins what the scoping actually guarantees (the payload is unchanged) and records the read failure as the pre-existing limitation it is, with a reason string pointing at the fix an application should use: store the zone's id. The migration guide says the same in one sentence. If you want the `object`-typed case genuinely repaired, it needs something other than a converter and I would rather it were its own issue than smuggled in here.

## 3. The HTTP API sentence and the message wording

The #3495 migration note now says the check is shared with the HTTP API, that a `List<string>` in a `RAMJobStore` job's map fails a `GET` response with the typed message, and why that is the consistent outcome — the client reads with the same closed reader, and `AddTypeInfoResolver` is the answer at both ends.

The message said "which **a persistent store** cannot read back", which is wrong in front of an HTTP caller. It now reads:

> Job data entry 'recipients' holds a System.Collections.Generic.List\`1[…], which **Quartz's JSON format** cannot read back. A job data value has to be one of the types JobDataMap declares an accessor for (…), a Dictionary<string, string>, or a type the application declares through SystemTextJsonSerializerRegistry.AddTypeInfoResolver. …

`JobDataValues`' remarks now spell out that the gate is shared and why the wording is what it is.

## Verification (re-run at `d967d5a6`)

| | |
|---|---|
| `dotnet build Quartz.slnx` | Build succeeded, **0 Warning(s), 0 Error(s)** |
| `dotnet test src/Quartz.Tests.Unit` | **Failed: 0, Passed: 3616, Skipped: 4, Total: 3620** |
| `dotnet test src/Quartz.Tests.AspNetCore` | **Failed: 0, Passed: 527** |
| integration, `postgres` leg | **Failed: 0, Passed: 149** (both serializers via `AdoSchedulerTest`) |
| `dotnet fallout VerifyDocsSnippets` | Succeeded — 93 markdown files up to date |
| `dotnet fallout PublishTrimmed` | Succeeded (not requested; re-run because `JobDataValues` changed) |

Pushed as a fast-forward, no force. Not merged.

lahma:
Fixed in `739b3267`. My bug: I asserted a whole captured string that was mostly the capturing OS's own text.

Both write assertions now parse the JSON and check only what means the same thing everywhere — that the job data slot is still an object rather than a bare string, that it carries `$type`, and that `Id` is right. Nothing reads `DisplayName`, `StandardName`, `DaylightName` or `BaseUtcOffset` any more.

## Verified on Linux, not reasoned about

I ran it in a `mcr.microsoft.com/dotnet/sdk:10.0` container rather than argue from the source. The pre-fix commit reproduces your failure exactly:

```
Failed AZoneInAJobDataMapIsWrittenAsItAlwaysWas
  "…TC+09:00) Japan Standard Time (Tokyo)","StandardName":"Japan…"
  "…TC+09:00) Osaka, Sapporo, Tokyo","StandardName":"Tokyo…"
```

At `739b3267`, all 17 tests across the three Newtonsoft zone fixtures pass on Linux.

## Two corrections to what I told you

**1. My second worry was wrong, and I checked instead of guessing.** I had assumed `ATriggerWritesItsZoneAsTheIdInstead` was failing too, on the theory that a Windows id resolves to an IANA one on Unix and the trigger would write `Asia/Tokyo`. It does not — on Linux the lookup answers with a `TimeZoneInfo` that keeps the Windows spelling, so that assertion passed. You were right that only the one test was failing. I still changed it to compare against the resolved zone's `Id`, because the spelling belongs to the platform and the tz database rather than to the test, but the commit message says plainly that this one was not broken.

**2. The full Linux suite shows two other failures, and they are not mine.** `CronExpressionTest.CanGetExpressionSummary` fails in my container — and fails identically at the base commit `347392bdf`, which is how I know it is not from this branch. The cause is my harness, not Linux: I moved the tree in with `git archive`, which applied this repo's `core.autocrlf` and wrote CRLF files, so the verbatim string literal in that test carries `\r\n` while `Environment.NewLine` is `\n`. Real CI checks out LF on Linux and the test passes. Flagging it only so the number below is not mistaken for a regression.

## Verification

| | |
|---|---|
| `dotnet test src/Quartz.Tests.Unit` (Windows) | **Failed: 0, Passed: 3616, Skipped: 4, Total: 3620** |
| the three zone fixtures, Linux container | **Failed: 0, Passed: 17** |
| same fixtures at `d967d5a6`, Linux container | **Failed: 1** — the failure you reported, reproduced |
| `CanGetExpressionSummary` at base `347392bdf`, Linux container | **Failed: 2** — pre-existing, harness CRLF artifact |

Pushed as a fast-forward, no force. New head `739b3267`. Not merged.

PR #3489, opened 2026-08-26
The ADO.NET store is nine files by concern, and every line is where it was

lahma:
`AdoJobStoreBase` was 5,359 lines in one file. It is now a partial class across nine files, one
per concern, following the `StdAdoDelegate` precedent. **This is a pure move** — the point of doing
it on its own is that a reviewer can confirm that mechanically rather than by reading 5,000 lines.

## The partition

| File | Lines | Declarations | What is in it |
|---|---:|---:|---|
| `AdoJobStoreBase.cs` | 780 | 59 | The type declaration, the 13 fields, the constructor, the 38 configuration, identity and collaborator properties (`Meters`, `Delegate`, `DbProvider`, `LockHandler`, `DbMetadata`, `SupportsPersistence`, …), `GetAcquireRetryDelay`, `InitializeDelegate`, `Initialize`, `SchedulerStarted`/`Paused`/`Resumed`, `Shutdown` |
| `AdoJobStoreBase.Acquisition.cs` | 598 | 7 | `AcquireNextTriggers`, `CreateAcquisitionCriteria`, `AcquireNextTrigger`, `TriggersFired`, `TriggerFired`, `GetFiredTriggerRecordId` and its `ftrCtr` counter |
| `AdoJobStoreBase.Completion.cs` | 264 | 6 | `ReleaseAcquiredTrigger` ×2, `TriggeredJobComplete` ×2, `unblockJobTriggersTransitions`, `RecoverUnblockedMisfires` |
| `AdoJobStoreBase.PauseResume.cs` | 791 | 30 | Every `Pause*`/`Resume*`/`PauseAll`/`ResumeAll`, `GetTriggerState`, `ResetTriggerFromErrorState`, and the paused/blocked bookkeeping (`ApplyPausedTriggerGroupState`, `RecordPausedJobGroups`, `CheckBlockedState`) |
| `AdoJobStoreBase.Store.cs` | 772 | 26 | The write side: `ScheduleJob`, `ScheduleJobs`, `AddJob`, `AddTrigger`, `AddCalendar`, `Delete*`, `ReplaceTrigger`, `UpdateTriggerDetails`, `Clear`, and the `NullJobTypeLoader` nested type `DeleteTrigger` uses |
| `AdoJobStoreBase.Query.cs` | 623 | 36 | The read side: `Get*`, `*Exists`, `Exists`, the key-listing members, and every `Query*` including `QueryClusterNodes` and `ClassifyClusterNode` |
[...]

PR #3499, opened 2026-08-27
The ADO.NET store wraps a failure once, and is built from one argument

### LINKED ISSUE #3428: AdoJobStoreBase is 5,380 lines with a twelve-parameter constructor derived stores must chain
`src/Quartz/Impl/AdoJobStore/AdoJobStoreBase.cs` is 5,380 lines in one file, covering lifecycle,
acquisition and firing, completion, pause and resume, store and remove, query, misfire handling and
clustering. Of its members only seven are `virtual`, so the file's size is not a derivation surface —
it is one class that has never been split.

Two costs follow. The first is repetition: the file has 69 `catch (Exception` blocks feeding 79
`Throw.JobPersistenceException` calls, nearly all of them the same three lines around a different
body. The second is the constructor. `AdoJobStoreBase` takes twelve parameters at `:64`, and
`LocalTransactionJobStore` (`:47`) and `ExternalTransactionJobStore` (`:51`) each restate all twelve
and forward them — which is what the custom-job-store how-to asks a reader to do, and what a
thirteenth dependency would make them all do again.

Internal names are of their era: `schedSignaler`, `ftrCtr`, `rcvryTrig`, `cth`, and a family of `Do*`
methods. `LogWarnIfNonZero` does not log at warning level.

## Shape

The design is the maintainer's.

Partial classes by concern — Lifecycle, Acquisition and Fire, Completion, Pause and Resume, Store and
Remove, Query, Misfire, Cluster — as **pure moves first**, one mechanical diff reviewed on its own.
Then the wrap blocks collapse into one `ExecuteInLock`-family helper that wraps once, and the two
constructors take an `AdoJobStoreDependencies` record so a derived store chains one argument. Names
are cleaned in the same pass; `LogWarnIfNonZero` logs at the level its name says.

## Done means

No behaviour change, no public-API change beyond the constructor shape, a migration-guide row for it,
and the how-to updated. Sonar's new-code coverage will report the file move; note it on the pull
request as the documented artefact.

Out of scope: splitting `StdAdoDelegate`, and internalizing anything — that is the
`Quartz.Impl.AdoJobStore` ledger, sequenced after this.

lahma:
PR 2 of 2 for #3428. [#3489](https://github.com/quartznet/quartznet/pull/3489) split `AdoJobStoreBase`
into nine partials by pure moves; this one changes its shape. Four commits, each green on its own.

## One wrap

`Guarded` lives in `AdoJobStoreBase.Execution.cs` beside the lock family, and is `private static` — a
device, not a seam:

```csharp
private static async ValueTask<T> Guarded<T>(
    Func<ValueTask<T>> operation,
    string description,
    Func<Exception, string?>? reason = null)
```

with a `ValueTask` twin. The call site names the operation instead of writing the message:

```csharp
protected ValueTask<bool> JobExists(ConnectionAndTransactionHolder conn, JobKey jobKey, CancellationToken cancellationToken = default)
{
    return Guarded(
        () => Delegate.JobExists(conn, jobKey, cancellationToken),
        $"determine job existence ({jobKey})");
}
```

**49 `try`/`catch` statements are gone, at 51 call sites; 61 catch clauses became 2.** `git diff -w`
on `AdoJobStoreBase.Acquisition.cs` shows `AcquireNextTrigger`'s body unchanged — its guard moved from
inside the retry loop to around it, which is where it already effectively was, since
`Throw.JobPersistenceException` does not return and the loop could never continue past the catch.

### What the guard does with a failure

- **A provider exception becomes a `JobPersistenceException` naming the operation**, as before —
  `Couldn't pause trigger 'g1.t1': …`, with the original as `InnerException`.
- **A `JobPersistenceException` passes through unchanged.** It is already an answer of that kind:
  either one the store raised on purpose (an `ObjectAlreadyExistsException` a caller catches by type,
  `The job (…) referenced by the trigger does not exist.`, `Calendar cannot be removed if it is
  referenced by a trigger!`) or one an inner operation has already described. Wrapping it pushed that
  answer into `InnerException` behind a message naming only whichever outer operation was in progress.
  Two of the fifty blocks already did this by hand; it is now the rule, so `wrapPersistenceFailures`
  and `ShouldWrap` are gone with it.
- **An `OperationCanceledException` passes through unchanged.** Nothing has gone wrong with the
  database, the caller asked to stop, and callers match on the type. `GetEnlistedConnection` has said
  so in a comment since it was written; now the guard agrees with it.
- **A failure about the stored bytes still says which.** `ReadFailureReason` / `WriteFailureReason`
  carry the "because a required type was not found" and "because the BLOB couldn't be (de)serialized"
  clauses that six separate `catch (TypeLoadException)` / `catch (IOException)` blocks used to.

The type of what a caller sees is unchanged in every case, so `catch (JobPersistenceException)` and the
HTTP API's exception mapping — which keys on the type name, not the message — are unaffected.

### Messages

Every message a test or the documentation quotes survives verbatim — including
`Couldn't record the fire of trigger '…' for '…' job`, which `AdoJobStoreBaseTest` asserts on, and
`Couldn't obtain triggers`, which `docs/documentation/troubleshooting.md` quotes. Five changed, three
because they named the wrong thing:

| Was | Is | Why |
|---|---|---|
| `Couldn't check for existence of job`, from `Exists(TriggerKey)` | `Couldn't check for existence of trigger` | a copy of the `JobKey` overload's message |
| `Couldn't remove trigger`, from `ReplaceTrigger` | `Couldn't replace trigger` | a copy of `DeleteTrigger`'s |
| `Couldn't pause trigger group '…'`, from `ResumeTriggers` | `Couldn't resume trigger group '…'` | a copy of `PauseTriggerGroup`'s — it told an operator the opposite of what failed |
| `Error clearing scheduling data` | `Couldn't clear scheduling data` | the lone message outside the family |
| `error validating trigger acquisition` (×2, and it dropped the inner message) | `Couldn't validate trigger acquisition: …` | the two acquisition validators were identical blocks |

Separately, a nested message loses its outer prefix: `Couldn't store trigger 'a.b' for 'c.d' job: The
job (c.d) referenced by the trigger does not exist.` is now just `The job (c.d) referenced by the
trigger does not exist.` The migration guide carries a row for that and one for cancellation.

### Tests

**No existing test had to change**: the double wrapping was never load-bearing for one, in either
suite — `Quartz.Tests.Integration` contains exactly one `WithMessage` assertion and it is about
enlistment configuration. Two tests were added to state the new rule, alongside the ten that state the
rest of the guard's contract in `AdoJobStoreFailureWrappingTest`:

- `ARefusalTheStoreRaisedTravelsWithoutTheOperationAroundIt` — stores a trigger whose job is gone and
  asserts the **whole** message, so a prefix naming the surrounding operation would fail it.
- `ACancelledTokenSurfacesAsCancellationRatherThanAsAPersistenceFailure` — drives a store call with an
  already-cancelled token through a delegate that honours it, which proves both that the token is
  forwarded and that what comes back is not reported as a failure to reach the database.

### What keeps a catch of its own

Five, and each is one the guard cannot express — said here rather than left for a reviewer to notice:

- `FindFailedInstances`, `ClusterCheckIn` and `ClusterRecover` say "Failure …" rather than
  "Couldn't …", and the first also stamps `LastCheckin` on its way out.
- The job-data write inside `TriggeredJobComplete`, whose two failures name *different operations*
  ("serialize job data" / "update job data") rather than one operation with a reason.
- The connection and retry plumbing (`GetConnection`, `RecoverMisfires`, the transient-retry loops),
  which is not of this shape at all.

## `AdoJobStoreDependencies`

```csharp
public sealed record AdoJobStoreDependencies(
    ISchedulerSignaler SchedulerSignaler, ITypeLoader TypeLoader, TimeProvider TimeProvider,
    IOptions<QuartzSchedulerOptions> SchedulerOptions, IOptions<AdoJobStoreOptions> StoreOptions,
    IOptions<ClusteringOptions> ClusteringOptions, IObjectSerializer ObjectSerializer,
    IDbProvider DbProvider, IDriverDelegate DriverDelegate,
    ISemaphore? LockHandler = null,
    IEnumerable<ITriggerPersistenceDelegate>? TriggerPersistenceDelegates = null,
    ILoggerFactory? LoggerFactory = null);
```

Exactly the twelve parameters the three constructors took — nothing added, nothing dropped — so a
derived store chains one argument. The container registers it per scheduler and builds it with
`ActivatorUtilities`, the way it built the store, so every dependency is still resolved from that
scheduler's own registrations and a store that was never given a lock handler still gets none.
`AdoJobStoreDependencies` joins `SchedulerScopedServiceProvider`'s scheduler-scoped list for the reason
`IDriverDelegate` is on it: a named scheduler must not be handed the default scheduler's parts.

Tests build one through `TestJobStores.Dependencies()`, which defaults every collaborator to a double.
Where a fixture means "no lock handler configured" it says `with { LockHandler = null }` rather than
relying on a positional `null`.

## Names

| Was | Is |
|---|---|
| `schedSignaler` | `signaler` |
| `ftrCtr` | `firedTriggerCounter` |
| `rcvryTrig` | `recoveryTrigger` |
| `cth` | `connection` (`CloseConnection`, `RollbackConnection`, `CommitConnection`) |
| `rec`, `ftRec`, `firedTriggerRecs` | `record`, `firedTrigger`, `nodeFiredTriggers` |
| `trig`, `jd`, `tKey`, `jKey`, `lst`, `stat`, `n` | `trigger`, `jobDataMap`, `triggerKey`, `jobKey`, `firedTriggers`, `header`, `deleted` |
| `DoCheckin` | `CheckIn` |
| `DoRecoverMisfires` | `RecoverMisfires` |
| `DoUpdateOfMisfiredTriggerOptimized` | `ApplyMisfiredTriggerUpdate` |

`DoUpdateOfMisfiredTrigger` is deleted — no callers, as PR 1's report said; only a comment in
`StdAdoDelegate` still named it, and that now describes the behaviour instead.

**`LogWarnIfNonZero` needed nothing**: it was already gone before this branch, its callers raising
source-generated cluster events at the level its name always claimed. What *did* still name a method
that no longer exists was event 3500's message, "in DoCheckin". It reads "of the cluster check-in" now,
which is **the only line of the `LogEventCatalogTest` snapshot that moves, and no level moves with it**.

`conn` is untouched: renaming it to `connection` is ~400 occurrences and a dozen more public parameter
names, which belongs in its own mechanical pass if it is wanted at all.

## Public API

<details><summary>The whole baseline delta</summary>

```diff
-protected AdoJobStoreBase(ISchedulerSignaler schedulerSignaler, ITypeLoader typeLoader, TimeProvider timeProvider, IOptions<QuartzSchedulerOptions> schedulerOptions, IOptions<AdoJobStoreOptions> storeOptions, IOptions<ClusteringOptions> clusteringOptions, IObjectSerializer objectSerializer, IDbProvider dbProvider, IDriverDelegate driverDelegate, ISemaphore? lockHandler = null, IEnumerable<ITriggerPersistenceDelegate>? triggerPersistenceDelegates = null, ILoggerFactory? loggerFactory = null) { }
+protected AdoJobStoreBase(AdoJobStoreDependencies dependencies) { }
-public LocalTransactionJobStore(… the same twelve …) { }
+public LocalTransactionJobStore(AdoJobStoreDependencies dependencies) { }
-public ExternalTransactionJobStore(… the same twelve …) { }
+public ExternalTransactionJobStore(AdoJobStoreDependencies dependencies) { }

+public sealed class AdoJobStoreDependencies : IEquatable<AdoJobStoreDependencies>
+{
+    public AdoJobStoreDependencies(ISchedulerSignaler SchedulerSignaler, ITypeLoader TypeLoader, TimeProvider TimeProvider, IOptions<QuartzSchedulerOptions> SchedulerOptions, IOptions<AdoJobStoreOptions> StoreOptions, IOptions<ClusteringOptions> ClusteringOptions, IObjectSerializer ObjectSerializer, IDbProvider DbProvider, IDriverDelegate DriverDelegate, ISemaphore? LockHandler = null, IEnumerable<ITriggerPersistenceDelegate>? TriggerPersistenceDelegates = null, ILoggerFactory? LoggerFactory = null) { }
+    public IOptions<ClusteringOptions> ClusteringOptions { get; init; }
+    public IDbProvider DbProvider { get; init; }
+    public IDriverDelegate DriverDelegate { get; init; }
+    public ISemaphore? LockHandler { get; init; }
+    public ILoggerFactory? LoggerFactory { get; init; }
+    public IObjectSerializer ObjectSerializer { get; init; }
+    public IOptions<QuartzSchedulerOptions> SchedulerOptions { get; init; }
+    public ISchedulerSignaler SchedulerSignaler { get; init; }
+    public IOptions<AdoJobStoreOptions> StoreOptions { get; init; }
+    public TimeProvider TimeProvider { get; init; }
+    public IEnumerable<ITriggerPersistenceDelegate>? TriggerPersistenceDelegates { get; init; }
+    public ITypeLoader TypeLoader { get; init; }
+}

-protected ValueTask<bool> DoCheckin(Guid requestorId, CancellationToken cancellationToken = default) { }
+protected ValueTask<bool> CheckIn(Guid requestorId, CancellationToken cancellationToken = default) { }
-protected ValueTask<RecoverMisfiredJobsResult> DoRecoverMisfires(Guid requestorId, CancellationToken cancellationToken = default) { }
+protected ValueTask<RecoverMisfiredJobsResult> RecoverMisfires(Guid requestorId, CancellationToken cancellationToken = default) { }

-protected DateTimeOffset CalcFailedIfAfter(SchedulerStateRecord rec) { }
+protected DateTimeOffset CalcFailedIfAfter(SchedulerStateRecord record) { }
-protected ValueTask CommitConnection(ConnectionAndTransactionHolder cth, bool openNewTransaction, CancellationToken cancellationToken = default) { }
+protected ValueTask CommitConnection(ConnectionAndTransactionHolder connection, bool openNewTransaction, CancellationToken cancellationToken = default) { }
-protected ValueTask RollbackConnection(ConnectionAndTransactionHolder? cth, Exception cause, CancellationToken cancellationToken = default) { }
+protected ValueTask RollbackConnection(ConnectionAndTransactionHolder? connection, Exception cause, CancellationToken cancellationToken = default) { }
-protected static ValueTask CloseConnection(ConnectionAndTransactionHolder cth, CancellationToken cancellationToken = default) { }
+protected static ValueTask CloseConnection(ConnectionAndTransactionHolder connection, CancellationToken cancellationToken = default) { }
```

(Namespaces elided for width; the file carries them in full. The fourth commit adds no API — `Guarded`
is private.)

</details>

The migration guide carries five new rows — the constructor shape, the two `Do*` renames, the four
parameter names, the unwrapped nested message and cancellation — and the row about the trailing
`ILoggerFactory?` now points at `AdoJobStoreDependencies.LoggerFactory`. The custom-job-store how-to
gains a compiled `sample_custom_job_store_dependencies` snippet under *Deriving from AdoJobStoreBase*.

## One thing found while doing this, deliberately left alone

The pass-through is fully observable for `ExternalTransactionJobStore` and for the protected
connection-taking members. For `LocalTransactionJobStore`, an `OperationCanceledException` raised
inside a **locked, transacted** operation is still converted by `ExecuteInLocalTransactionLock`'s
blanket `catch (Exception e)` into `JobPersistenceException("Unexpected runtime exception: …")` — that
catch is the transaction machinery, not the guard, and it also owns the rollback. Making cancellation
survive that boundary means `when (e is not OperationCanceledException)` plus a decision about the
rollback, which is a change to the retry machinery rather than to the guard, so it is not in here.
Worth noting that the same method already lets an `OperationCanceledException` from its inter-attempt
`Task.Delay` escape as itself, so the boundary is inconsistent today.

## Verification

At `d9e94e7`:

| | |
|---|---|
| `dotnet build Quartz.slnx` | **Build succeeded. 0 Warning(s), 0 Error(s)** |
| `dotnet test src/Quartz.Tests.Unit` | **Passed! Failed: 0, Passed: 3555, Skipped: 4, Total: 3559** |
| Retry/transient subset (`~Transient`, `~Retry`) | **Passed! 51/51** — the classification `ExecuteInLocalTransactionLock` depends on is unmoved |
| `dotnet test src/Quartz.Tests.AspNetCore` | **Passed! Failed: 0, Passed: 527** |
| Integration, `basic` | **Passed! Failed: 0, Passed: 329, Total: 329** (2 m 3 s) |
| Integration, `postgres` | **Passed! Failed: 0, Passed: 149, Total: 149** (3 m 23 s) — see below |
| Integration, `sqlserver` | **Passed! Failed: 0, Passed: 141, Total: 141**, run at `b3d9c8b0`; not re-run for the guard change |
| `dotnet fallout PublishTrimmed` | **Succeeded** — the canary's "a persistent store schedules, fires and reads back" is the DI path that builds the record under trimming |
| `dotnet fallout VerifyDocsSnippets` | **Documentation snippets are up to date (91 markdown files checked)** |

The first run of the `postgres` leg after the guard change failed one case —
`AdoJobStoreSmokeTest.TestPostgreSql("newtonsoft", Name)`, `GenericJobType.TriggeredCount` 0 rather
than 1. That assertion follows `GenericJobType.WaitForTrigger(TimeSpan.FromSeconds(20))`, a wall-clock
wait, on a machine that was running builds and containers throughout. It passes in isolation and the
full leg passes on re-run, and the change cannot reach it: inside a locked, transacted operation every
non-`JobPersistenceException` still arrives at the scheduler thread as a `JobPersistenceException`,
by the paragraph above. Reporting it rather than quietly re-running.

`npm run docs:check-links` was not run — `node_modules` is not installed in this worktree — but the one
new fragment link, `#deriving-from-adojobstorebase`, follows the same rule as `operations.md`'s
existing link into that page.

Closes #3428.


lahma:
Merging over the SonarCloud gate. Its one failing condition is coverage on new code (72.5% against 80%): the lines are the ADO.NET store's, whose coverage comes from the integration legs Sonar does not see, and most of them are renames and the collapse of 49 identical catch blocks into one guard rather than new logic. The five S3776 findings are pre-existing cognitive complexity in methods the renames touched — the same ones #3489 listed — not something this change introduced. The seventeen other checks pass.

PR #3504, opened 2026-08-27
Paging, projections and parameter names are structure rather than convention

### LINKED ISSUE #3427: Dialects splice paging clauses by string arithmetic, and nine statements read SELECT *
Three things in the SQL layer are held together by convention rather than by structure.

**Paging is spliced with `Substring`.** `SqlServerDelegate.GetSelectNextTriggerToAcquireSql` builds
its statement as `"SELECT TOP " + shape.MaxCount + " " + base.GetSelectNextTriggerToAcquireSql(shape).Substring(6)`
— `src/Quartz/Impl/AdoJobStore/SqlServerDelegate.cs:42`, with the same trick at `:52` for the misfire
statement. It is correct only while the base statement begins with exactly `SELECT`, and nothing
enforces that. `OracleDelegate` wraps instead (`SELECT * FROM (...) WHERE rownum <=`, `:35` and `:42`);
`PostgreSQLDelegate`, `SQLiteDelegate` and `FirebirdDelegate` append ` LIMIT ` or ` ROWS `. Six
dialects, three different assumptions about a statement none of them owns.

**Nine constants read `SELECT *`** — `src/Quartz/Impl/AdoJobStore/StdAdoConstants.cs` lines 174, 183,
190, 212, 410, 413, 416, 502 and 509, over `SIMPROP_TRIGGERS`, `CRON_TRIGGERS`, `SIMPLE_TRIGGERS`,
`FIRED_TRIGGERS` and `SCHEDULER_STATE`. A column added by a user or by a migration changes what comes
back.

**`DbMetadata.GetParameterName` allocates on every parameter.** It is `ParameterNamePrefix + parameterName`
(`src/Quartz/Impl/AdoJobStore/Common/DbMetadata.cs:162`), called from `AdoUtil` at `:200` once per
bound parameter, and the set of names is small and fixed.

## Done means

The paging clause is an explicit slot on the statement-shape hooks the dialects already override — the
`TriggerAcquisitionSqlShape` / `GetSelectNextTriggerToAcquireSql(shape)` family from #3282, not
`SqlStatement`, which is the batch record — rather than a substring offset; the nine statements name
their columns; `GetParameterName` caches per name; and the parameter names that `StdAdoConstants`
spells inline (`@triggerName`, forty times) and that `AddCommandParameter` binds by a separate literal
become one constant interpolated into both, so the two spellings cannot drift — the idea that survives
#2874. A snapshot recorded before the
change proves each statement is byte-identical after it, the way #3282's acquisition-SELECT
consolidation did.

Out of scope: changing any statement's semantics, and the acquisition statement's own shape, which
`TriggerAcquisitionSqlShape` already owns.

lahma:
Three things in the SQL layer were held together by convention. Each becomes a structure, and the
wire is proved unchanged where it was meant to be.

## The row limit is a slot

There is no ANSI row limit, so each of the six dialects built its own row-limited statement by editing
finished SQL. `SqlServerDelegate` cut the first six characters off and pasted `SELECT TOP n` back on —
correct only while the statement began with exactly `SELECT`, which nothing enforced; `OracleDelegate`
wrapped the whole thing; the rest appended `LIMIT n` or `ROWS n`. Two statements carry a limit
(acquisition and the misfire scan), so that was twelve overrides, three assumptions about a statement
none of them owns, and a `count == -1` test written out six times.

A new `SqlRowLimit` names the three placements, and one member says which a dialect wants:

```csharp
protected virtual SqlRowLimit GetRowLimit(int count) => SqlRowLimit.Unlimited;

// SqlServerDelegate
protected override SqlRowLimit GetRowLimit(int count) => SqlRowLimit.InProjection("TOP", count);
// PostgreSQLDelegate, MySQLDelegate, SQLiteDelegate
protected override SqlRowLimit GetRowLimit(int count) => SqlRowLimit.AtStatementEnd("LIMIT", count);
// FirebirdDelegate
protected override SqlRowLimit GetRowLimit(int count) => SqlRowLimit.AtStatementEnd("ROWS", count);
// OracleDelegate
protected override SqlRowLimit GetRowLimit(int count) => SqlRowLimit.InEnclosingSelect("rownum", count);
```

Two of the three placements are slots the statement template carries — `AfterSelect` right after the
`SELECT` keyword and `AtEnd` after the last clause, both empty by default so an unlimited statement is
the text it always was. The third, `Enclose`, wraps a finished statement and assumes nothing about it
beyond its being one. The `-1` that means "every row" is turned into `SqlRowLimit.Unlimited` before the
member is called, so an override never tests for it.

Five of the six dialect delegates now override nothing else. `FirebirdDelegate` is four lines.
`MySQLDelegate` keeps its two overrides, for the `FORCE INDEX` hint a row limit cannot express.

### The wire-identity proof

`AcquisitionSqlTest` already snapshotted the acquisition statement per dialect. The first commit adds
the same for the misfire scan, in both its limited and unlimited forms, **before** anything changed —
so the fourteen statements are recorded as they were and the refactoring commit has to leave them
alone. It does, with one deliberate exception, which is the entire diff of that verified file:

```diff
             FROM
-                {0}TRIGGERS t
+                {0}TRIGGERS t FORCE INDEX (IDX_{1}T_NFT_ST_MISFIRE)
```

**A reviewer should decide about this one.** MySQL's misfire scan used to lose its index hint when the
recovery batch was unlimited, because the hint and the row limit shared an early return — an artefact,
not a decision. Which index the sweep reads does not depend on how many rows come back, and the
unlimited sweep is the one that reads the most, so the hint is applied unconditionally now.
`MySQLDelegateTest.GetSelectMisfiredTriggersToRecoverSql_WithMinusOne_*` changed with it. Say the word
and I will restore the old behaviour instead.

## Nine statements name their columns

`SELECT *` over `SIMPROP_TRIGGERS`, `CRON_TRIGGERS`, `SIMPLE_TRIGGERS`, `FIRED_TRIGGERS` and
`SCHEDULER_STATE` meant a column added by a user or by a migration changed what came back, and the
fired-trigger read was fetching `EXECUTION_GROUP` for a record that has no such field. The tenth —
`SimplePropertiesTriggerPersistenceDelegateBase`'s single-key lookup, the twin of the batch form in the
list — is folded in rather than left starred beside its named sibling.

| Statement | Columns |
|---|---|
| `SqlSelectCronTriggers` | `CRON_EXPRESSION, TIME_ZONE_ID` |
| `SqlSelectCronTriggersByKeysPrefix` | the same, plus `TRIGGER_NAME, TRIGGER_GROUP` |
| `SqlSelectSimpleTrigger` | `REPEAT_COUNT, REPEAT_INTERVAL, TIMES_TRIGGERED` |
| `SqlSelectSimpleTriggersByKeysPrefix` | the same, plus the key columns |
| `SqlSelectSimpropTriggersByKeysPrefix` | the twelve `*_PROP_*` columns and `TIME_ZONE_ID`, plus the key columns |
| `SelectSimplePropsTrigger` (the single-key twin) | the same twelve |
| `SqlSelectFiredTriggers` | the twelve `ReadFiredTriggerRecord` reads — `EXECUTION_GROUP` is on the row and is not one of them |
| `SqlSelectInstancesRecoverableFiredTriggers` | the seven the recovery trigger is built from |
| `SqlSelectSchedulerState` / `…States` | `INSTANCE_NAME, LAST_CHECKIN_TIME, CHECKIN_INTERVAL` |

Each list is a named constant shared by the single-key and batch forms of the same read, so the two
cannot drift apart.

`SelectColumnListTest` is what makes "the list is what the reader reads" a test rather than a comment.
It takes each statement's projection **from the statement text**, builds a `DbDataReader` that knows
only those columns, and drives the real reader against it. A column the reader wants and the statement
does not project throws from `GetOrdinal`, the way it would against a database; a column the statement
projects and nothing reads is left over at the end. Both directions were confirmed by mutation:

```
The statement does not project 'TIMES_TRIGGERED'. It projects: REPEAT_COUNT, REPEAT_INTERVAL
Expected reader.Unread to be empty ... but found at least one item {"SCHED_NAME"}.
```

## `GetParameterName` caches

It was `ParameterNamePrefix + parameterName`, called once per parameter bound — seven for an
acquisition, fifteen for a trigger write. The spelling is remembered per name now, in the
`ConditionalWeakTable` the record already uses for its derived state, so that two descriptions saying
the same thing about a driver still compare equal after one of them has been used. A description with
no prefix hands back the name it was given and needs no cache at all.

The cache is capped at 4096 entries, because the method is public: Quartz's own names are bounded
(about eighty-five fixed, four hundred trigger-key, five hundred job-key, eleven state, at most a
thousand job-type exclusions), but a delegate naming a parameter after its data would otherwise grow it
without bound. Past the cap it simply stops adding.

The assertion is an allocation one — `GC.GetAllocatedBytesForCurrentThread()` around a second pass over
an acquisition's seven names, which must be zero. Reverted to the concatenation, it reports the 360
bytes that were there.

## A parameter name is one constant

Eighty-six names were each written twice: `@triggerName` in the statement, `"triggerName"` in the
binder — forty times over for that one. A misspelling on either side compiled, and surfaced as a
provider complaining about an unbound parameter, or, on a provider that adapts named parameters
positionally, as a value bound to the wrong column.

`SqlParameters` holds each name once. The statement interpolates it (`@{SqlParameters.TriggerName}`)
and the binder passes it (`AddCommandParameter(cmd, SqlParameters.TriggerName, …)`), so a name that
does not exist is a build error. The four statement sites built by concatenation rather than
interpolation — the three `ApplyPaging` clauses and the simple-properties statements — splice it as
`" = @" + SqlParameters.TriggerName` instead.

All 860 sites were rewritten mechanically, and the result was proved back: substituting each constant
for the literal it stands for reproduces all thirteen files exactly, so no statement's text moved. A
guard test also reflects over every statement and fails one holding the literal text `SqlParameters` —
which is what an interpolation hole landing in a plain string looks like — and one carrying a
placeholder no constant declares.

Two things the constants make visible rather than fix, both on the wire and neither worth a statement
change: the fired-trigger `UPDATE` spells `REQUESTS_RECOVERY` as `@requestsRecover`, a syllable short of
the `@requestsRecovery` every other statement uses; and the preferred-node names are prefixes of their
`*Auto` siblings. `AdoUtil`'s placeholder rewrite is greedy so neither is a live bug, but they are now
next to each other instead of a letter apart in two unrelated strings.

## Public API

`SqlRowLimit` is new and `StdAdoDelegate.GetRowLimit` is new; five dialect delegates stopped overriding
`GetSelectNextTriggerToAcquireSql` and `GetSelectMisfiredTriggersToRecoverSql`. The baseline was
re-accepted through Verify, and the migration guide carries the delta — a new *Row limiting is a slot,
not a splice* section and two rows in the breaking-changes table. Both hooks stay `protected virtual`,
so a dialect delegate that splices its own clause keeps working; it just spells the same thing twice.
`docs/documentation/quartz-4.x/how-tos/dialect-delegate.md` and its sample say the new shape.

## Verification

| | |
|---|---|
| `dotnet build Quartz.slnx` | succeeded, 0 warnings |
| `dotnet test src/Quartz.Tests.Unit` | 3563 passed, 4 skipped |
| `dotnet test src/Quartz.Tests.AspNetCore` | 527 passed |
| integration `basic` | 329 passed |
| integration `postgres` | 149 passed |
| integration `sqlserver` | 141 passed |
| integration `mysql` | 84 passed |
| integration `sqlite` | 12 passed |
| integration `firebird` | 84 passed |
| integration `oracle` | 83 passed |
| `dotnet fallout VerifyDocsSnippets` | up to date, 93 markdown files |
| `dotnet fallout PublishTrimmed` | canary passed, including its SQLite store round-trip |
| `node docs/.vuepress/check-links.mjs` | 214 pages, 843 links, 494 fragments, no dead anchors |

Oracle, Firebird and SQLite were run because this touches their delegates. The `IntegrationTest`
fallout target is gated to GitHub Actions on Linux, so each leg was run as the target runs it —
`QUARTZ_TEST_DATABASE=<db> dotnet test … --filter TestCategory=db-<db>` — against a local Docker
daemon.

Closes #3427

### DIFF
--- a/docs/documentation/quartz-4.x/how-tos/dialect-delegate.md
+++ b/docs/documentation/quartz-4.x/how-tos/dialect-delegate.md
@@ -11,8 +11,8 @@ a delegate of its own.
 ::: tip Start here
 **Do not implement `IDriverDelegate`.** Nothing in the product does; it is a hundred-odd members, and
 almost all of them are the same SQL on every database. Subclass `StdAdoDelegate` and override the
-handful that differ. The six shipped dialects override **eight distinct members between them**, out of
-roughly a hundred and ten.
+handful that differ. The six shipped dialects override **nine distinct members between them**, out of
+roughly a hundred and ten — and four of the six override two or fewer.
 :::
 
 ## The seam
@@ -28,12 +28,13 @@ public sealed class MyDatabaseDelegate : StdAdoDelegate
 ```
 <!-- endSnippet -->
 
-Eight members are the dialect contract. Everything else on `StdAdoDelegate` is an implementation step
+Nine members are the dialect contract. Everything else on `StdAdoDelegate` is an implementation step
 that happens to be `protected virtual` so the class can be composed — treat them as private.
 
 | Member | Override when |
 |---|---|
-| `protected virtual string GetSelectNextTriggerToAcquireSql(TriggerAcquisitionSqlShape shape)` | your database limits rows differently from ANSI |
+| `protected virtual SqlRowLimit GetRowLimit(int count)` | your database can limit the rows a statement returns |
+| `protected virtual string GetSelectNextTriggerToAcquireSql(TriggerAcquisitionSqlShape shape)` | the acquisition statement needs something else besides — MySQL's index hint is the only shipped case |
 | `protected virtual string GetSelectMisfiredTriggersToRecoverSql(int count)` | the same, for the misfire scan; `count == -1` means "no limit" |
 | `protected virtual string GetCountMisfiredTriggersInStateSql()` | the counting form needs a different shape |
 | `protected virtual string ApplyPaging(string sql, bool takeLimited)` | `OFFSET … FETCH NEXT …` is not understood |
@@ -44,28 +45,34 @@ that happens to be `protected virtual` so the class can be composed — treat th
 
 ### Row limiting
 
-The default `GetSelectNextTriggerToAcquireSql` ignores `shape.MaxCount` entirely — *"by default we
-don't support limits, this is db specific"* — so a dialect that can limit rows should say so. Four
-shapes appear among the shipped dialects:
+There is no ANSI row limit, so `StdAdoDelegate` applies none and a dialect that can limit rows says
+where its clause goes. Two statements carry one — trigger acquisition and the misfire scan — and both
+ask the same member, so a dialect says it once:
 
 <!-- snippet: sample_dialect_delegate_row_limiting -->
 ```csharp
-// append (PostgreSQL, Firebird)
-protected override string GetSelectNextTriggerToAcquireSql(TriggerAcquisitionSqlShape shape)
-    => base.GetSelectNextTriggerToAcquireSql(shape) + " LIMIT " + shape.MaxCount;
+// … LIMIT n (PostgreSQL, MySQL, SQLite) — or "ROWS" on Firebird
+protected override SqlRowLimit GetRowLimit(int count)
+    => SqlRowLimit.AtStatementEnd("LIMIT", count);
 
-// splice a prefix (SQL Server: SELECT TOP n)
-// wrap the whole statement (Oracle: SELECT * FROM ( … ) WHERE rownum <= n)
-// append with an index hint (MySQL: FORCE INDEX (…) … LIMIT n)
+// SELECT TOP n …                              SqlRowLimit.InProjection("TOP", count)
+// SELECT * FROM ( … ) WHERE rownum <= n       SqlRowLimit.InEnclosingSelect("rownum", count)
 ```
 <!-- endSnippet -->
 
+`SqlRowLimit` names the three places a limit can sit, and the statement is built with the clause
+already in it. Nothing is spliced into finished SQL, so a dialect no longer depends on the statement
+starting with a particular keyword or ending with a particular clause. `count` is always at least
+one: the `-1` that means "every row" is turned into `SqlRowLimit.Unlimited` before the member is
+called, so an override never has to test for it.
+
 `TriggerAcquisitionSqlShape` carries everything about an acquisition attempt that changes the text of
-the statement — the row limit, and how many job-type exclusion terms the `NOT IN` clause needs. Read
-`shape.MaxCount` and hand the whole shape to `base`: a dimension added later becomes a property on
-the record, so an override written today keeps compiling and keeps applying it. It is also the key
-the finished statement is cached under, which is why it holds the bucketed exclusion count rather
-than the caller's exact one.
+the statement — the row limit's count, and how many job-type exclusion terms the `NOT IN` clause
+needs. It is also the key the finished statement is cached under, which is why it holds the bucketed
+exclusion count rather than the caller's exact one. Override
+`GetSelectNextTriggerToAcquireSql(shape)` only for something a row limit cannot express; MySQL is the
+one shipped dialect that does, for its `FORCE INDEX` hint, and it still calls `base` and leaves the
+limit alone.
 
 ### Paging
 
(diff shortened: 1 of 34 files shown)


lahma:
Merging over the SonarCloud gate. Its one failing condition is coverage on new code (73.1% against 80%): the changed lines are the dialect delegates' statement text, paging slots and parameter constants, which are exercised by the seven database integration legs — all green here and on every dialect locally — and not by the unit run Sonar measures. The one finding it lists is a MINOR S1192 (` AND ` repeated in a column-list builder). The seventeen other checks pass.

PR #3506, opened 2026-08-27
A job listener that throws synchronously loses its firing, not the schedule

### LINKED ISSUE #3502: A job listener that throws synchronously wedges the firing and leaves a phantom executing entry
Found by #3437's listener tests (#3501), which hit it live: a job listener that throws from a guard clause — synchronously, before its first `await` — hangs the firing.

`QuartzScheduler.NotifyJobListeners` (`src/Quartz/Core/QuartzScheduler.cs:2116` and `:2125`, at `77edc8571`) reads

```csharp
await NotifySingle(notifyAction(jl, context, jobExecutionException, cancellationToken), jl, context).ConfigureAwait(false);
```

so `notifyAction(...)` is evaluated as an argument, *outside* `NotifySingle`'s `try`. A listener whose `JobToBeExecuted` throws before it returns a `ValueTask` therefore escapes as the raw exception rather than the `SchedulerException` `NotifySingle` would have wrapped it in — and `JobRunShell.Run` catches `SchedulerException` only (`JobRunShell.cs:408`). The run shell never reaches `TriggeredJobComplete`: the trigger stays `ACQUIRED`/`BLOCKED`, and for a `[DisallowConcurrentExecution]` job every sibling trigger stays blocked with it, permanently. Trigger listeners do not have the hole — their notification is shaped differently.

Two consequences ride with it:

1. **A phantom executing entry.** The internal `ExecutingJobsManager` is notified first (`:2116`) and the user-listener loop then aborts, so `JobWasExecuted` never runs for anyone and nothing removes the entry: `NumberOfJobsExecutingHere` reads 1 after the store has already been told the firing completed (probed in #3501). That is what the dashboard's "currently executing" shows an operator.
2. `NotifySchedulerListenersFinalized` fires on the veto path and on the trigger-listener-failure path, but not on the job-listener-failure path.

**Done means**

- The listener call is evaluated inside the guarded region, so a synchronous throw is wrapped exactly as an asynchronous one is, and the run shell completes the firing with `SetAllJobTriggersError` as it does for any other listener failure. A test with a synchronously throwing `JobToBeExecuted` on a `[DisallowConcurrentExecution]` job: the firing completes, the trigger leaves the blocked state, the sibling fires next.
- The scheduler's own bookkeeping does not depend on the user-listener loop finishing: either the `ExecutingJobsManager` is notified outside that loop on both ends, or `JobWasExecuted` is delivered (with the failure) to every listener that received `JobToBeExecuted`. `NumberOfJobsExecutingHere` returns to 0 after the failed firing; a test says so.
- The finalized notification is symmetric across the three abort paths, or the asymmetry is written down as deliberate.
- No public API moves.

lahma:
A job listener whose guard clause throws — before it hands back a `ValueTask`, which
is what a method that is not `async` does — used to wedge the firing it threw on.
`QuartzScheduler.NotifyJobListeners` evaluated the listener call as an *argument* to
the guarded region, so the exception escaped as itself rather than as the
`JobExecutionProcessException` that `JobRunShell` catches. The run shell never reached
`TriggeredJobComplete`: the trigger stayed acquired and, for a
`[DisallowConcurrentExecution]` job, every sibling trigger stayed blocked behind a
firing that had already been abandoned — permanently.

Closes #3502. Found live by #3501's listener tests, which deliberately used only the
asynchronous shape and reported the synchronous hole rather than papering over it.

### 1. The call is evaluated inside the guard

`NotifyJobListeners` now calls the listener inside the `try` in its notification loop,
so a synchronous throw is wrapped exactly as a faulted task is, and the firing settles
the way an asynchronous failure already did — completed at the store with
`SchedulerInstruction.NoInstruction`.

Two notes for a reviewer:

- The issue says the run shell "completes the firing with `SetAllJobTriggersError` as
  it does for any other listener failure". That is not what the tree does and not what
  #3501's merged-soon test asserts: the begin-notification abort path completes with
  `NoInstruction`, because a firing that never happened settles nothing about the
  schedule. This PR makes the synchronous shape match the asynchronous one, which is
  what the sentence before it asks for; it does not change the instruction.
- The private `NotifySingle` local function is gone rather than being given the
  `notifyAction` to invoke. The loop that called it was already an `async` method, so
  the guard is a plain `try`/`catch` around the `await` — the same shape the trigger
  listener notifications a few lines up already have. That also drops one state machine
  allocation per listener per notification, since `await NotifySingle(...)` allocated
  one every time.

### 2. The scheduler's bookkeeping left the user-listener loop

**Chosen: notify `ExecutingJobsManager` outside the loop, on both ends** — the first of
the two options the issue offers.

The manager used to be notified as the first entry of the same loop, so a user listener
that stopped the loop stopped the bookkeeping with it: `JobWasExecuted` never ran for
anyone, nothing removed the entry, and `NumberOfJobsExecutingHere` read 1 after the
store had already been told the firing was over. That is the phantom an operator sees
in the dashboard's "currently executing".

`NotifyJobListenersToBeExecuted` now records the firing before the loop and takes the
record back if the loop fails; `NotifyJobListenersWasExecuted` removes it before the
loop. The manager keeps `IJobListener` — it is still the scheduler's built-in job
listener — but the two members that do the recording are synchronous, because there
was never anything in them to await.

Why not the second option (deliver `JobWasExecuted`, with the failure, to every
listener that received `JobToBeExecuted`): it changes what user listeners observe. A
listener would start seeing a `JobWasExecuted` for a job that never ran, on a firing
another listener aborted, and `JobChainingJobListener` and anything else that treats
that call as "the job finished" would act on it. The bookkeeping fix needs none of
that, and it keeps the manager's count honest by construction rather than by every
listener behaving.

`NumberOfJobsExecuted` is deliberately *not* taken back on the failure path: it counts
the firings this scheduler dispatched, and a firing a listener abandoned is still one
of those.

### 3. `NotifySchedulerListenersFinalized` across the abort paths

It reached the scheduler listeners from the veto path and from the
trigger-listener-completion path, but from neither of the paths a job listener's
failure takes. A trigger with one firing left in it was therefore spent in silence: the
run shell completed the firing, the store dropped the trigger, and a listener watching
for the end of a schedule never heard. The three abort paths now share one member,
which is the block the trigger-listener path already carried — the veto path keeps its
own wording, because the error it reports when the announcement itself fails names the
veto.

### Tests

`src/Quartz.Tests.Unit/Core/JobListenerFailureTest.cs`, ten cases, every one run in
both failure shapes so the two cannot drift apart again. They sit beside #3501's
`ThrowingJobListenerTest` and `VetoedFiringTest` and overlap none of their assertions:
those pin what the store is told and what the scheduler listeners are told, these pin
that the schedule keeps moving.

- a synchronously throwing `JobToBeExecuted` on a `[DisallowConcurrentExecution]` job
  with two due triggers: the firing is completed with `NoInstruction`, never released,
  and the sibling trigger goes on to fire and run the job — the job runs exactly once,
  for the trigger no listener stopped;
- nothing is left listed as executing, and `LocalExecutingJobs` reads 0, after a
  failure on either side of the job;
- the trigger's finalized notification survives a failure on either side.

**Failing first, verified**: with the two product commits reverted and restored
byte-for-byte (`git hash-object` checked before and after), 9 of the 10 cases fail —
six by timing out at the 30s deadline waiting for a firing that never reaches the
store, three on the phantom entry. The tenth,
`AFailedListenerLeavesNothingListedAsExecuting(Asynchronous, AfterTheJob)`, passes on
the unfixed tree: that is the one arrangement that already worked, and it is here as a
regression pin.

Nothing waits for a length of time or measures one. The store double records a firing
*after* the real store has acted on it, so a test that waits for a record and then asks
a question is asking a store that has already settled.

### Verification

```
dotnet build Quartz.slnx                      Build succeeded. 0 Warning(s) 0 Error(s)
dotnet test src/Quartz.Tests.Unit             Passed! Failed: 0, Passed: 3609, Skipped: 4 (run twice)
dotnet test src/Quartz.Tests.AspNetCore       Passed! Failed: 0, Passed: 547
integration, 'basic' leg                      Passed! Failed: 0, Passed: 329
```

No public API moves; the baselines are untouched and `PublicApiTest` is green.

### DIFF
--- a/src/Quartz.Tests.Unit/Core/JobListenerFailureTest.cs
+++ b/src/Quartz.Tests.Unit/Core/JobListenerFailureTest.cs
@@ -1,0 +1,359 @@
+#region License
+
+/*
+ * All content copyright Marko Lahma, unless otherwise indicated. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy
+ * of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+
+#endregion
+
+using Microsoft.Extensions.DependencyInjection;
+
+using Quartz.Core;
+using Quartz.Extensibility;
+using Quartz.Impl;
+
+namespace Quartz.Tests.Unit.Core;
+
+/// <summary>
+/// What a job listener's failure costs the firing it failed on, and what it must not cost anybody else.
+/// </summary>
+/// <remarks>
+/// <para>
+/// A listener can fail in either of two shapes, and until #3502 they were not handled alike. An
+/// <c>async</c> listener hands back a faulted task; a listener whose guard clause throws before it
+/// returns anything fails while the scheduler is still evaluating the call. Only the first was wrapped
+/// in the exception <see cref="JobRunShell" /> catches, so the second escaped the run shell entirely and
+/// the firing was never handed back to the store — leaving a job that forbids concurrent execution
+/// blocked behind a firing that had already been abandoned. Every case here therefore runs twice, once
+/// per shape, and the two shapes must be indistinguishable.
+/// </para>
+/// <para>
+/// The scheduler's own record of what is executing rides along with the same fault: it used to be kept
+/// by the listener loop, so a listener that stopped the loop stopped the bookkeeping too and left the
+/// firing listed as executing for as long as the process lived.
+/// </para>
+/// <para>
+/// Nothing here waits for a length of time or measures one. The store records a firing after it has
+/// acted on it, so a test that waits for a record and then asks a question is asking a store that has
+/// already settled.
+/// </para>
+/// </remarks>
+[TestFixture]
+[NonParallelizable]
+public sealed class JobListenerFailureTest
+{
+    private const string Group = "job-listener-failure";
+
+    /// <summary>
+    /// How long a test is willing to wait for a firing to reach the store before declaring the
+    /// scheduler stuck. Long enough that a loaded build agent never trips it, and never used as a
+    /// measurement.
+    /// </summary>
+    private static readonly TimeSpan observationDeadline = TimeSpan.FromSeconds(30);
+
+    /// <summary>
+    /// The two ways a listener can fail, which the scheduler has to treat as one.
+    /// </summary>
+    public enum FailureShape
+    {
+        /// <summary>Throws before handing anything back, as a guard clause in a method that is not <c>async</c> does.</summary>
+        Synchronous,
+
+        /// <summary>Hands back a faulted task, as an <c>async</c> method does.</summary>
+        Asynchronous,
+    }
+
+    /// <summary>
+    /// Which side of the job the listener fails on.
+    /// </summary>
+    public enum FailureSide
+    {
+        /// <summary><see cref="IJobListener.JobToBeExecuted" />, which stops the job running.</summary>
+        BeforeTheJob,
+
+        /// <summary><see cref="IJobListener.JobWasExecuted" />, which is too late to stop anything.</summary>
+        AfterTheJob,
+    }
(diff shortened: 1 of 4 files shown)


lahma:
Merging over the SonarCloud gate. Its one failing condition is coverage on new code (75% against 80%) and it lists no finding: the uncovered lines are `JobRunShell`'s abort paths and the `ExecutingJobsManager` bracket, which the ten new `JobListenerFailureTest` cases and the integration legs exercise but the unit run Sonar measures does not fully attribute. The sixteen other checks pass.

PR #3512, opened 2026-08-27
Cluster recovery is named steps over sets of rows, and a node failed out by a peer says so

### LINKED ISSUE #3425: ClusterRecover is a 212-line method with two unanswered TODOs about a node that failed itself out
`AdoJobStoreBase.ClusterRecover` runs from `src/Quartz/Impl/AdoJobStore/AdoJobStoreBase.cs:4771` to
`:4982` as one method doing six things — releasing acquired triggers, unblocking blocked ones,
scheduling recovery triggers, sweeping completed ones, repinning preferred nodes, deleting the state
row — with up to about ten statements per fired-trigger row and no name on any step. Three of the
file's six TODOs are inside or beside it: `:3119` ("worry about failed/recovering/volatile job
states?"), `:3662` ("this really ought to return something like a FiredTriggerBundle"), `:3792`
("Hum... shouldn't we log a warning here?").

The other two are the real bug. `:4686` reads `// TODO: revisit when handle self-failed-out impl'ed
(see TODO in clusterCheckIn() below)` and `:4753` reads `// TODO: handle self-failed-out`. A node whose
`SCHEDULER_STATE` row has been deleted by a peer that judged it failed goes on as though nothing
happened: it does not know it was recovered, and it can re-recover its own fired triggers that the
peer has already taken.

## Done means

- `ClusterRecover` is decomposed into named steps, pure moves first, with the per-row statements
  batched.
- `:3119`, `:3662` and `:3792` are each resolved or deleted with a sentence saying why.
- **Self-failed-out is handled.** On check-in, when this node's row is gone and this is not its first
  check-in, the node re-inserts its row, logs a warning naming the peer that recovered it, increments
  `quartz.cluster.recovery.trigger` with a `self` tag, and does **not** re-recover its own fired
  triggers — the peer already did.
- A unit test on the fake clock, and an integration test with a synthetic peer recovery against the
  two-node harness.

`RecoverStaleAcquiredTriggers` (`:1232`, already shipped) is the answer to discussion #2749, and
`troubleshooting.md` is to say so; that is the documentation issue's work, not this one.

Out of scope: changing the failure predicate in `CalcFailedIfAfter` (`:4739`), and any schema change.

lahma:
`ClusterRecover` was one 212-line method doing six things to every fired-trigger row a failed node left
behind, and two of the file's TODOs said that a node whose own `SCHEDULER_STATE` row a peer had deleted
went on as though nothing had happened. This is both halves of #3425.

## The steps

`ClusterRecover` now loops over the failed instances and, per instance, runs a named sequence:

| Step | What it does |
|---|---|
| `PlanRecovery` | Reads the node's fired-trigger rows and decides what each means — reserved, blocked, recoverable, held back — without writing anything. `CanDeferRecovery` is the #2817 grace period, lifted out whole |
| `ReleaseAcquiredTriggers` | Every trigger the node had reserved and never fired, back to `WAITING` |
| `UnblockInterruptedJobs` | The siblings its executions were holding: `BLOCKED` → `WAITING`, `PAUSED_BLOCKED` → `PAUSED` |
| `ScheduleRecoveryTriggers` | A replacement firing for each interrupted execution whose job asked for one |
| `DeleteFiredTriggerRows` | The node's rows, minus the ones held back |
| `DeleteTriggersLeftComplete` | The triggers those rows were the last of |
| `ReportRecovery` | The counter and the five conditional log lines |
| `ReleaseFailedInstanceRegistration` | Auto-claimed pins released, check-in row deleted |

Two supporting types: `FailedInstanceResidue` (what the rows turned out to mean) and
`RecoveryTriggerNaming` (the `recover_<node>_<n>` counter, which runs across the whole pass exactly as
the local `recoverIds` did).

The first three steps are grouped by what they do rather than issued row by row. That is only sound
because the whole pass is one transaction under `TRIGGER_ACCESS` — no other node sees the state between
two steps — and because a recovery trigger is stored `WAITING` outright (`AddTrigger(..., recovering:
true)` skips `CheckBlockedState`), which was the one thing that would have made it depend on when the
unblocking ran. Scheduling a replacement firing stays row by row: each one reads whether its job still
exists and what its trigger's data map holds, and both answers decide whether anything is written.

## Statement counts

Measured on the harness `AdoJobStoreBaseTest` uses — a faked `IDriverDelegate` on a `FakeTimeProvider`,
counting every delegate call one recovery makes. One failed node, ten fired-trigger rows: five `ACQUIRED`
on five distinct triggers, and five `EXECUTING` rows of one `[DisallowConcurrentExecution]` job that
requests recovery.

**Grace period expired, nothing held back — 64 delegate calls → 52.** The batched steps go from 15
statements in 15 round trips to 7 statements in 3:

| | Before | After |
|---|---|---|
| Release reserved triggers | 5 × `UpdateTriggerStateFromOtherState` | 1 × `UpdateTriggerStatesFromOtherState` (5 statements, one batch) |
| Unblock the job | 10 × `UpdateTriggerStatesForJobFromOtherState` | 2 × `UpdateTriggerStatesForJobsFromOtherState` (1 statement each — ten rows of one job are one job) |

**Inside the grace period, the five `EXECUTING` rows held back — 21 delegate calls → 13.** 10 statements
in 10 round trips become 10 statements in 2: the five releases in one batch, and the five deletes of the
rows that were *not* held back in another, where they used to be five `DeleteFiredTrigger` calls.

Everything else is unchanged: the fired-trigger read, the per-recoverable-row
`JobExists`/`SelectTriggerJobDataMap`/`AddTrigger` chain, the `COMPLETE` sweep's `SelectTriggerState` per
distinct trigger key, the whole-instance delete, the repin and the state-row delete.

The three new `IDriverDelegate` members are overloads of statements that already existed, issued through
the same `ExecuteStatements`/`DbBatch` path the fire and misfire batches use, so a provider that cannot
batch issues exactly what it always did, in the same order. They return `ValueTask` rather than a row
count, because a batch does not report one per command portably and the caller counts the rows it asked
about. Public API baseline re-accepted for those six lines (three interface members, three
`StdAdoDelegate` implementations); migration guide has the appendix row and a
[Batched cluster recovery](https://github.com/quartznet/quartznet/blob/main/docs/documentation/quartz-4.x/migration-guide.md)
section.

## The three TODOs

- **`:3119`** — `CheckBlockedState`'s *"worry about failed/recovering/volatile job states?"* is answered
  in place: the row's own state is deliberately not consulted, because a fired-trigger row of a serial
  job means that job is reserved or running whatever state the row is in, and a row a dead node left is
  cluster recovery's business (or `RecoverStaleAcquiredTriggers`'), not this predicate's.
- **`:3662`** — *"this really ought to return something like a FiredTriggerBundle"* is deleted with the
  reason: `IOperableTrigger.FireInstanceId` is the contract the scheduling loop, the fired-trigger row
  and `TriggerFiredBundle` all read the id through, so a bundle here would be a fourth spelling of the
  same field rather than one fewer.
- **`:3792`** — *"Hum… shouldn't we log a warning here?"* is answered **no**: losing the compare-and-swap
  on a `WAITING` row is the ordinary outcome of two nodes reaching for the same batch, and a warning per
  lost row would be noise proportional to how well the cluster is sharing its work.

## Self-failed-out

On a check-in that is not its first, a node whose own `SCHEDULER_STATE` row is missing has been failed
out by a peer. It now:

- **writes the row back.** The update-then-insert already did this; it is now named (`RecordCheckIn`) and
  said out loud, because until the row is back the node does not exist as far as its peers are concerned.
- **logs it.** 3501 (unchanged text, unchanged id — the message `troubleshooting.md` quotes) plus one of
  two new events: **3515** naming the peer, or **3516** saying it cannot be named. Nothing in the schema
  records who recovered whom, so a peer is only named when it is the only other node with a state row and
  therefore the only one that could have.
- **counts it.** `quartz.cluster.recovery.trigger` with `quartz.cluster.recovered.instance.id` set to the
  node's own id — the A5 spec's "self" case, expressed as the tag equalling `quartz.scheduler.id` rather
  than as a new attribute. It adds **1**, not a trigger count: the rows are gone by the time this node
  reads the table, and the peer's own measurement carries the real number under the same tag. The remark
  on `Meters.ClusterSelfRecoveryObserved` says so.
- **does not re-recover its own fired triggers.** The scan is where that is decided, and its comment says
  why: the peer released, rescheduled and deleted them under the trigger-access lock, so a second pass
  would schedule a second recovery trigger for a firing that is already being replayed.

Detection moved from `FindFailedInstances` (which only logged) to `ClusterCheckIn`, where the row is
written back. Both go through one private `ScanClusterState`, so the check-in path still reads
`SCHEDULER_STATE` exactly once; `FindFailedInstances` keeps its signature and returns that scan's failed
list.

Log catalogue diff, the whole of it:

```
+3515  Warning  ClusterLog.SelfFailedOutRecoveredByPeer
+    "Scheduler instance ({InstanceId}) was failed out by instance {RecoveringInstanceId}, the only other instance with a scheduler state row and so the only one that could have. Its own row has been written back; the work that instance recovered is not recovered again here."
+3516  Warning  ClusterLog.SelfFailedOutRecoveringPeerUnknown
+    "Scheduler instance ({InstanceId}) was failed out by another instance, which cannot be identified: {OtherInstanceCount} other instances have scheduler state rows and no row records which of them recovered this one. Its own row has been written back; the work that instance recovered is not recovered again here."
```

## For a reviewer to decide

- **The two `Acquisition.cs` TODOs are comment-only edits in a file I was told is B1+B2's.** No code
  changed there, but if you would rather they landed with B1, drop commit
  `Three questions the store had left in TODOs are answered where they were asked` and the third TODO
  (`PauseResume.cs`) can be re-committed on its own.
- **The integration test is a regression guard, not a reproduction.** The old code did not in fact
  re-recover its own rows either — `FindFailedInstances` only adds this node's own record on the first
  check-in — so the double-fire the issue warns about was latent rather than live. What is new is that
  the situation is detected, named, measured, and pinned by tests. The unit tests are what fail without
  the change.
- **The `COMPLETE` sweep still reads a state per distinct trigger key.** A set-based form needs a new
  statement in `StdAdoConstants`, which #3504 has open; deliberately left alone.
- **`ClusterRecover` and `ClusterCheckIn` keep their own `Throw.JobPersistenceException` wraps** rather
  than moving to `Guarded`. Their messages (`Failure recovering jobs`, `Failure updating scheduler state
  when checking-in`) are asserted by existing tests and are not in this issue's scope.

## Verification

```
dotnet build Quartz.slnx                    0 warnings, 0 errors
dotnet test src/Quartz.Tests.Unit           3639 passed, 0 failed, 1 skipped
dotnet test src/Quartz.Tests.AspNetCore     547 passed, 0 failed
integration, postgres leg                   152 passed, 0 failed (2 m 53 s)
integration, sqlserver leg                  144 passed, 0 failed (3 m)
integration, basic leg                      335 passed, 0 failed (1 m 59 s)
ClusteredHardeningPostgresTest              5 passed, 0 failed
ClusteredHardeningSqlServerTest             6 passed, 0 failed
dotnet fallout VerifyDocsSnippets           up to date (93 markdown files checked)
npm run docs:check-links                    214 pages, 885 internal links, 533 fragments, no dead links or anchors
```

There is no `Quartz.Tests.AOT` project on this branch; `Quartz` builds with the trim and AOT analyzers
on and reported no warnings, and the `Quartz.Trimming.Canary` publish is a CI leg of its own that this
change adds no reflection to.

Closes #3425


lahma:
Merging over the SonarCloud gate. Its one failing condition is coverage on new code (78.5% against 80%): the recovery steps and the self-failed-out check-in are exercised by the two-node Postgres and SQL Server hardening legs, which are green here and which the unit run Sonar measures does not include. It lists nothing above MINOR (overload adjacency and one loop-to-LINQ suggestion). The seventeen other checks pass.

Issue #3520, opened 2026-08-30
A trigger can carry a retry policy


lahma:
Quartz has no retry policy. Verified on `main`: zero hits for `RetryPolicy|MaxAttempts|WithRetry|BackoffStrategy` in `src/` (the `DbRetryInterval`/`MaxTransientRetries` options are store-connection retries). The idioms are `JobExecutionException { RefireImmediately = true }` — an in-process loop with no delay and no ceiling (`src/Quartz/Core/JobRunShell.cs`, the `ReExecuteJob` branch), which MassTransit caps by hand at `RefireCount < 5` — or scheduling a fresh trigger from inside the job, which ABP, Elsa and OpenIddict each wrote their own version of. It is the first thing a newcomer comparing with TickerQ notices.

## Shape

- `RetryPolicy` — sealed value type, `internal` constructor, three factories (`Fixed(maxAttempts, delay)`, `Exponential(maxAttempts, initialDelay, factor = 2, maxDelay = null)`, `Explicit(params delays)`), `ImmutableArray<TimeSpan>` for the explicit table (last value repeats) with hand-written equality, `DelayFor(attempt)`, a stable stored string form (`BackoffFactor` round-tripped with `"R"`/invariant) pinned by a contract test.
- `TriggerBuilder<TJob>.WithRetryPolicy(RetryPolicy)`, `ITriggerConfigurator<TJob>.WithRetryPolicy`, `ITrigger.RetryPolicy` (nullable) + `ITrigger.RetryAttempt`, `IMutableTrigger` setters, `TriggerDetailsUpdate.WithRetryPolicy` (policy is configuration; the attempt is never set through it).
- `IJobExecutionContext.RetryAttempt` — 0 on a regular fire, *n* on the *n*-th retry. Distinct from `RefireCount`, which counts iterations of the in-process refire loop (same context, same thread, no persistence, no slot release). **`RefireImmediately` is not a zero-delay retry**; the guide says so.
- A job fails for retry purposes when `Execute` throws anything other than a `JobExecutionException` that itself asks for something (`RefireImmediately`, `Unschedule*`). An `OperationCanceledException` from the scheduler's own token never retries — shutdown and interrupt are operator decisions, and a node vanishing mid-execution is what `RequestsRecovery` is for.

## Semantics

[...]

PR #3543, opened 2026-08-30
QRTZ_TRIGGERS carries the columns a trigger's retry policy will live in

lahma:
Two nullable columns on `QRTZ_TRIGGERS` — `RETRY_POLICY` and `RETRY_ATTEMPT` — land in the schema, and nothing else does. No statement selects or writes them, no public member exposes them, and no behaviour changes. This is R1 of #3520, the schema half.

## Why the schema goes first

4.x has no `Supports*Column` probes. A column added to `QRTZ_TRIGGERS` once 4.0 has shipped is therefore a *required* column, and so a mandatory migration in the middle of a major version — which is the exact thing `database/migrations/4.0/` exists to spare people. While that script is still open the columns cost an upgrader nothing; after it closes they cost everyone a migration. So the columns go in now and the engine follows, whatever happens to the engine's schedule.

Both are nullable with no default, so every row an upgrade brings across reads as "no retry policy" and there is nothing to backfill.

## Per-dialect types

`RETRY_POLICY` is typed exactly as `EXECUTION_GROUP` is on each dialect, with 250 characters rather than 200; `RETRY_ATTEMPT` is typed as `PRIORITY` is.

| Dialect | `RETRY_POLICY` | `RETRY_ATTEMPT` |
|---|---|---|
| SQL Server | `[RETRY_POLICY] nvarchar(250) NULL` | `[RETRY_ATTEMPT] int NULL` |
| PostgreSQL | `RETRY_POLICY VARCHAR(250) NULL` | `RETRY_ATTEMPT INTEGER NULL` |
| MySQL | `RETRY_POLICY VARCHAR(250) NULL` | `RETRY_ATTEMPT INTEGER NULL` |
| Oracle | `RETRY_POLICY VARCHAR2(250) NULL` | `RETRY_ATTEMPT NUMBER(13) NULL` |
| SQLite | `RETRY_POLICY NVARCHAR(250) NULL` | `RETRY_ATTEMPT INTEGER NULL` |
| Firebird | `RETRY_POLICY VARCHAR(250)` | `RETRY_ATTEMPT INTEGER DEFAULT NULL` |

`MigrationScriptTest` compares type, nullability and length between a migrated schema and a fresh one on every dialect, so these are pinned rather than asserted in prose.

## Generated versus hand-written

**Generated** — described in `build/Build.DatabaseMigrations{,.Scripts}.cs` and emitted by `dotnet fallout GenerateMigrations`; not hand-edited:

- `database/migrations/4.0/schema_30_to_40_upgrade_{sqlServer,postgres,mysql_innodb,oracle,sqlite,firebird}.sql`

The generator gains `RetryPolicy` and `RetryAttempt` column dictionaries beside `PreferredNode`/`PreferredNodeAuto`, and `Build40Script` gains a section 4 for the two columns — the old sections 4 and 5 become 5 and 6, in the script bodies, the header's table of contents and the prose. The header's MANDATORY paragraph now names the two columns as things 3.x never had. SQLite's "run exactly once" note is unchanged and still accurate: the new `ADD COLUMN`s are unguarded there like the ones above them.

**Hand-written** — all eight fresh-install scripts, so a fresh install and a migrated database still match column for column:

- `database/tables/tables_{sqlServer,sqlServer_Below2016,sqlServerMOT,postgres,mysql_innodb,oracle,sqlite,firebird}.sql`

**Documentation and constants:**

- `docs/documentation/database/schema-changes.md` — the 4.0 section gains the section-4 row, the renumbering, and what the columns hold; the mandatory-migration prose names them
- `database/README.md` — the 4.0 row and the mandatory-upgrade paragraph
- `docs/documentation/quartz-4.x/migration-guide.md` — the schema section, carrying the API-baseline delta below
- `docs/documentation/quartz-4.x/db/index.md` — the `QRTZ_TRIGGERS` column reference, marked as reserved and always `NULL` for now
- `src/Quartz/Impl/AdoJobStore/AdoConstants.cs` — `ColumnRetryPolicy` and `ColumnRetryAttempt`, constants only

## For a reviewer to decide

- **The public API baseline moves by two lines.** `AdoConstants` is public and every sibling column name on it is public, so the two new constants are too; `PublicApiTest_Quartz.verified.txt` gains exactly those two lines and the migration guide carries the same delta. If they should be `internal` until something reads them, say so and they can be — the brief asked for them beside `ColumnExecutionGroup`, which is public.
- **`RETRY_POLICY` is 250 characters.** That is `DESCRIPTION`'s width and 50 more than `EXECUTION_GROUP`'s. It has to hold the stored form of an explicit delay table, which is the longest shape the policy can take. Widening it later is a migration, so it is worth agreeing on now.
- **`RETRY_ATTEMPT` is `PRIORITY`-shaped rather than `MISFIRE_INSTR`-shaped** (`NUMBER(13)` rather than `NUMBER(2)` on Oracle, `INTEGER` rather than `SMALLINT` on PostgreSQL and Firebird). An attempt counter is small, but it is read as an `int` and a two-digit ceiling is a trap nobody would expect.
- **The migrated-schema workload asserts the columns in SQL, not through the API**, because there is no API member to assert through yet — see `AssertRetryColumnsArePresent` in `MigratedSchemaWorkload`. Naming both columns is the existence check; requiring them null on every row the scheduler just wrote is the nullable-with-no-default check.

## Test results

`dotnet build Quartz.slnx` — 0 warnings, 0 errors. `dotnet fallout GenerateMigrations` then `VerifyMigrations` — clean; only the six `4.0` scripts changed content. `VerifyDocsSnippets` — up to date (93 markdown files).

- `Quartz.Tests.Unit` — 3760 passed, 0 failed (`PublicApiTest_Quartz` baseline accepted through Verify, delta is the two constants)
- `Quartz.Tests.AspNetCore` — 549 passed, 0 failed
- SQLite — full `db-sqlite` leg, 12 passed
- PostgreSQL — full `db-postgres` leg, 157 passed
- SQL Server — full `db-sqlserver` leg, 149 passed (covers `tables_sqlServerMOT.sql`, which that leg's second container is built from)
- MySQL, Oracle, Firebird — `MigrationScriptTest` only, 3 passed each; their fresh-install scripts run at container start, so the new DDL is parsed by each engine

`tables_sqlServer_Below2016.sql` is exercised by nothing in the test suite; its edit mirrors `tables_sqlServer.sql` line for line.

As a negative control, `retry_attempt` was removed from the PostgreSQL migration *and* the PostgreSQL fresh-install script — which keeps the structural comparison passing — and `PostgreSqlUpgradedSchemaRunsScheduler` then failed with `42703: column "retry_attempt" does not exist`, from the new assertion. Both files were restored byte for byte afterwards and `VerifyMigrations` re-run.

Part of #3520.


lahma:
Merging over the SonarCloud gate, documented:

- `new_coverage 0.0% < 80%`: the only product lines in this PR are two `const string` members on `AdoConstants` (nothing to execute) and the rest is generated SQL, hand-written DDL, docs and an integration-test assertion that CI's coverage run does not execute. Same attribution artefact as #3450/#3489/#3499.
- `plsql:VarcharUsageCheck` on `tables_postgres.sql` and `tables_firebird.sql` ("Use VARCHAR2"): the PL/SQL analyzer applied to Postgres and Firebird DDL, where `VARCHAR2` does not exist.
- `plsql:ExecuteImmediateTrapExceptionsCheck` on the generated Oracle 4.0 script: the new section has the exact shape of the three sections above it, which the generator emits deliberately (a guarded `ADD COLUMN` that must fail loudly, not swallow `WHEN OTHERS`).

Every dialect leg ran locally (sqlite/postgres/sqlserver full; mysql/oracle/firebird `MigrationScriptTest`) and CI's 17 other checks are green.

Issue #3531, opened 2026-08-30
A persistent store can create its own schema


lahma:
"Automatic schema creation" was asked for on #988 in 2022 and is still the answer nobody has. #1695 was
closed on scope and delegation, and delegation has failed measurably: AppAny 0.6.0 carries four open drift
issues from 2026-07 — missing `EXECUTION_GROUP`, missing `PREFERRED_NODE*`, drifted indexes — which are
exactly the columns 4.x made **mandatory**, because 4.x has no `Supports*Column` probes.

**Verified negatives** (searched `src/`, `build/`, `database/`, `docs/`)

- `grep -rn "SchemaProvisioning\|ProvisionSchema" src/ database/ build/ docs/` → 0 hits. Nothing in the
  product, the generator or the documentation provisions a schema.
- `src/Quartz/Impl/AdoJobStore/IDriverDelegate.cs:1513` declares `ValidateSchema` and nothing else; there is
  no `CreateSchema` and no schema resource of any kind under `src/Quartz/Impl/AdoJobStore/`.

**Evidence**

- The generator already emits guarded DDL: `build/Build.DatabaseMigrations.cs:312` `CreateTable` and `:359`
  `CreateIndex`.
- The anti-drift test already exists:
  `src/Quartz.Tests.Integration/Impl/AdoJobStore/MigrationScriptTest.cs:482` `SchemaSnapshot`, read at
  `:471-472` to compare a migrated schema with a fresh one.
- The setting it replaces: `src/Quartz/Configuration/AdoJobStoreOptions.cs:186`
  `PerformSchemaValidation` (default `true`), bridged at
  `src/Quartz/Configuration/QuartzPropertyBridge.cs:672`, read into the store at
  `src/Quartz/Impl/AdoJobStore/AdoJobStoreBase.cs:125` / `:435` and acted on at `:656`.
- `IPersistentStoreBuilder.UseClustering` (`src/Quartz/Configuration/IPersistentStoreBuilder.cs:152`) is the
  shape to copy: a `ConfigureStore` shorthand that reads as a decision.
- The four SQLite triggers are unprefixed today — `DELETE_SIMPLE_TRIGGER` (`database/tables/tables_sqlite.sql:81`),
  `DELETE_SIMPROP_TRIGGER` (`:109`), `DELETE_CRON_TRIGGER` (`:127`), `DELETE_BLOB_TRIGGER` (`:144`) — so two
[...]

PR #3551, opened 2026-08-30
A persistent store can create its own schema

lahma:
> **Merge order.** This is the second of a chain of three: #3550 → **#3551** → #3552, and they have
> to land in that order. The base is `main` rather than #3550's branch so that the `pr-integration-*`
> workflows run at all — they trigger only on pull requests targeting `main` — which means the diff
> below carries #3550's commit as well until that one has merged. Rebase this branch on `main` after
> #3550 lands and the diff is only its own commit.

Automatic schema creation was asked for on #988 in 2022. Delegating it failed measurably: AppAny 0.6.0
carries four open drift issues from 2026-07 — missing `EXECUTION_GROUP`, missing `PREFERRED_NODE*`,
drifted indexes — which are exactly the columns 4.x made **mandatory**, because 4.x has no
`Supports*Column` probes to degrade behind.

## The setting

`AdoJobStoreOptions.SchemaProvisioning` **replaces** `PerformSchemaValidation`. The check and the
creation are one decision — a store that may create its schema obviously validates it too — so they
are one setting with three positions rather than two flags that can contradict each other.

| Before | After |
|---|---|
| `PerformSchemaValidation = true` | `SchemaProvisioning.Validate` — still the default |
| `PerformSchemaValidation = false` | `SchemaProvisioning.None` |
| — | `SchemaProvisioning.CreateIfMissing`, or `store.ProvisionSchema()` |

`IPersistentStoreBuilder.ProvisionSchema()` is a `ConfigureStore` shorthand that reads as the decision
it is, like `UseClustering`.

The legacy key `quartz.jobStore.performSchemaValidation` still works and still means what it meant —
`true` bridges to `Validate`, `false` to `None`. `quartz.jobStore.schemaProvisioning` is the new key
and takes a member name, case-insensitively; a misspelled one is rejected naming the three that would
have worked. A file carrying both is decided by the new key, the only one of the two that can express
the third position.

## The mechanism

`IDriverDelegate.CreateSchema` sits beside `ValidateSchema` and is implemented once in
`StdAdoDelegate`: it splits the embedded script on the `--;;` sentinel and sends each statement to the
provider with the configured table prefix substituted in. Each dialect delegate names its own script
through `protected virtual string? SchemaResourceName`; a delegate that names none throws rather than
reporting success, naming itself and the script to run instead.

Wired into `AdoJobStoreBase.Initialize` before validation, under the same `ExecuteWithoutLock` shape —
which it has to be, since `QRTZ_LOCKS` is one of the tables it is there to create.

**Never drops, never alters.** So it is safe to leave on against a schema that already exists, and it
cannot turn a mis-typed `TablePrefix` into data loss. It is equally not an upgrade: a schema missing a
column a later release added is left as it is, and `database/migrations/` is still what moves a schema
forward.

**Not the default**, because creating tables needs DDL permission a production database is usually
right not to grant.

### The clustered race, and what Firebird and Oracle had to say about it

Nodes start together and only one can create any given object. Rather than reading a duplicate-object
error that every dialect spells differently — and that some spell the same way as a permission failure
— the outcome is decided by asking the schema: attempt the create, and on **any** exception run
`ValidateSchema`. A pass means another node won.

That was the whole of it in the first push of this branch, and the Firebird and Oracle legs of #3552
showed it was half an answer. The loser's first error is a table it could not create, and the very
next thing it finds is a table the winner **has not created yet** — so validation fails and a perfectly
ordinary race was reported as a hard failure:

```
ORA-00955: name is already used by an existing object          (creating QRTZC_JOB_DETAILS)
ORA-00942: table or view does not exist                        (validating QRTZC_TRIGGERS)
```

Firebird said the same thing in its own words, sometimes as a primary-key violation on
`RDB$RELATIONS` and sometimes as a deadlock, depending on timing.

So the create-then-validate is now a bounded loop — ten attempts, half a second apart. A retry is not
passive waiting: the script is guarded throughout, so re-running it skips whatever the other node
finished and makes whatever it has not reached yet. The two nodes fill in each other's gaps rather
than waiting on each other, and converge in a round or two. Only after ten attempts is the failure
reported as one, naming the fresh-install script for that database and the setting to drop back to.

## Also here

- `SharedDatabaseValidator`'s remarks: it named `PerformSchemaValidation`, and now says why
  `CreateIfMissing` makes the failure it reports *easier* to reach rather than harder — a mis-typed
  prefix used to need an empty table set to already exist, and a provisioning store creates one. Two
  schedulers sharing a database *and* a prefix are still never reported, provisioning or not; that is
  the supported arrangement, and both creating it if it is missing is exactly what a cluster does.
- The migration-guide row for the rename, and the `CreateSchema` note beside the existing
  `ValidateSchema` section.
- Every 4.x documentation page that named the old member now names the new one. That is the rename
  only — the pages that need to *teach* provisioning (`quick-start.md`, `db/index.md`,
  `tutorial/job-stores.md`, `database/README.md`, the third-party page) are the docs PR's, which is
  why this one says "Part of" rather than "Closes". `configuration/reference.md`'s row is updated
  here because it named a member that would otherwise not exist.
- Baselines accepted through Verify: `PublicApiTest_Quartz` and `LogEventCatalogTest_Quartz`.

## Tests

`SchemaProvisioningSqliteTest` is the end-to-end case that needs no container: an empty SQLite file,
provisioned, scheduled against, fired and read back. It also covers the default refusing an empty
database, a second pass leaving the first scheduler's rows alone, every object landing under a
configured `QRTZP_` prefix with nothing under `QRTZ_`, and a delegate with no script saying what to run
instead. Plus builder, bridge and exhaustiveness cases.

## Verification

```
dotnet build Quartz.slnx                                    0 warnings, 0 errors
dotnet fallout VerifySchema VerifyMigrations VerifyDocsSnippets   all up to date
dotnet test src/Quartz.Tests.Unit                           3835 passed, 0 failed
dotnet test src/Quartz.Tests.AspNetCore                     549 passed, 0 failed
npm run docs:check-links                                    214 pages, no dead links or anchors
QUARTZ_TEST_DATABASE=sqlite   TestCategory=db-sqlite        14 passed, 0 failed
QUARTZ_TEST_DATABASE=postgres TestCategory=db-postgres      159 passed, 0 failed
```

SQLite and PostgreSQL are the two run locally — several suites share this machine's Docker daemon, so
one container dialect at a time. All eight `pr-integration-*` workflows were green on the previous
push of this branch; the four dialects not run locally are covered there and on this one.

Part of #3531

### DIFF
--- a/docs/documentation/quartz-4.x/configuration/reference.md
+++ b/docs/documentation/quartz-4.x/configuration/reference.md
@@ -174,7 +174,7 @@ services.AddQuartz(q => q.UsePersistentStore(store =>
 | `AcceptEnlistedTransactions` | bool | `false` | Lets the job store use a connection the application enlisted with `SchedulerEnlistmentExtensions.EnlistTransaction`, so scheduling commits with the application's own work. See [Joining an existing transaction](../tutorial/job-stores.md#joining-an-existing-transaction). |
 | `DoubleCheckLockMisfireHandler` | bool | `true` | Re-checks the lock before handling misfires. |
 | `UseBackgroundThreads` | bool | `false` | Runs the misfire handler and cluster manager on background threads, which do not keep the process alive. These two are the only real threads Quartz creates. |
-| `PerformSchemaValidation` | bool | `true` | Verifies the expected tables exist at startup. |
+| `SchemaProvisioning` | `SchemaProvisioning` | `Validate` | What the store does about its schema at startup: `None` does nothing, `Validate` verifies the expected tables exist, `CreateIfMissing` creates whatever is missing and then verifies. |
 | `SelectWithLockSql` | string? | none | Overrides the row-lock statement, defaulted to SQL Server's `WITH (UPDLOCK,ROWLOCK)` form when that is the database. Read only when the store builds a database-locking handler for itself — see [Locking](#locking). |
 | `OpenConnection` | bool | `false` | Whether `ExternalTransactionJobStore` opens the connections it creates; read only by that store. |
 
--- a/docs/documentation/quartz-4.x/how-tos/trimming-and-native-aot.md
+++ b/docs/documentation/quartz-4.x/how-tos/trimming-and-native-aot.md
@@ -281,7 +281,7 @@ It does three checks, in order:
   serialized, read back under the very type `StdAdoDelegate` asks for, and serialized again, with the
   two payloads compared byte for byte.
 - **The store.** It creates a SQLite file from the schema Quartz ships, registers it as
-  `UseSqlite(SqliteFactory.Instance, …)` with `PerformSchemaValidation` on, schedules a job, waits for
+  `UseSqlite(SqliteFactory.Instance, …)` with `SchemaProvisioning.Validate`, schedules a job, waits for
   the job itself to signal that it fired, and reads the job and the trigger back through `IScheduler`.
   The firing is the point: that is where the job's type comes back out of `JOB_CLASS_NAME` as a string.
 - **The binding.** It builds a whole scheduler from an in-memory `IConfiguration` and reads ten values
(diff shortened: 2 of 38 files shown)


lahma:
Landed on main through #3552, whose branch carried these commits beneath S3's (rebase-merged; content is on main at 4aec95a7b). Closing as delivered — #3531's S2 is done.

Issue #3522, opened 2026-08-30
Scheduling over an existing trigger is one atomic call


lahma:
`MassTransit.QuartzIntegration` hand-rolls upsert three times over — `Exists` → `UnscheduleJob` →
`ScheduleJob(trigger)` — and partitions its consumers by correlation id purely to serialize those round
trips. There is no overload to reach for.

**Verified negatives** (searched all of `src/`)

- `src/Quartz.Tests.Unit/Verify/PublicApiTest_Quartz.verified.txt:712-715`: `ScheduleJob(ITrigger, ct)` and
  `ScheduleJob(IJobDetail, ITrigger, ct)` take **no** `ScheduleJobOptions`, while
  `ScheduleJob(IJobDetail, IReadOnlyCollection<ITrigger>, ScheduleJobOptions, ct)` at `:714` and
  `ScheduleJobs(..., ScheduleJobOptions, ct)` at `:715` do. The two single-trigger forms are the gap.
- `IQuartzBuilder.ScheduleJob<T>` exists (baseline `:1230`, `:1232`) but there is **no**
  `IScheduler.ScheduleJob<TJob>` of any shape — the DI one-liner has no imperative twin.

**Evidence**

- ADO takes `SchedulerLock.TriggerAccess` when replacing:
  `src/Quartz/Impl/AdoJobStore/AdoJobStoreBase.Store.cs:117-123`.
- RAM takes its single lock, but **twice**: `RAMJobStore.ScheduleJob` calls `AddJob` then `AddTrigger`,
  each acquiring `lockObject` — `src/Quartz/Impl/RAMJobStore.cs:259-263`, with the acquisitions at
  `:273-284` and `:493-505`.
- **Parity defect found.** ADO preserves `PreviousFireTimeUtc` when replacing a trigger
  (`AdoJobStoreBase.Store.cs:167-178`, the #1834 fix). RAM's `AddTriggerNoLock` builds
  `new TriggerWrapper((IOperableTrigger) trigger.Clone())` at `src/Quartz/Impl/RAMJobStore.cs:510` and
  removes the old wrapper, so the previous fire time is lost. The two stores disagree.
- HTTP: `src/Quartz.AspNetCore/AspNetCore/HttpApi/Endpoints/TriggerEndpoints.cs:388-409` is the
  `ScheduleJob` endpoint, and its request has no replace flag.

**Shape**

```csharp
// IScheduler — options required, so the existing overloads stay unambiguous
[...]

PR #3555, opened 2026-08-30
A trigger's retry policy is stored, read back and reported

lahma:
R3 of #3520: every path a trigger travels now carries its retry policy and how far through it the occurrence being executed has got. **Nothing acts on either value yet** — the engine is R4 — so a trigger given a policy stores it, hands it back and lists it, and fires exactly as it did before.

Builds on #3547 (merged). R4 (engine) and R5 (matrix + docs) stack on this, and the three merge only as a set that is green end to end.

## The model half

| Member | What it is |
|---|---|
| `ITrigger.RetryPolicy` / `RetryAttempt` | The policy (nullable), and how many times this occurrence has already been retried |
| `IMutableTrigger.RetryPolicy` / `RetryAttempt` | The setters a job store restores a row through |
| `TriggerBuilder<TJob>.WithRetryPolicy(RetryPolicy?)` | Give a trigger a policy while building it |
| `ITriggerConfigurator<TJob>.WithRetryPolicy(RetryPolicy?)` | The same, in the DI configuration API |
| `TriggerDetailsUpdate.WithRetryPolicy(RetryPolicy?)` | Re-policy a stored trigger without rescheduling it |
| `TriggerHeader.RetryPolicy` (string) / `RetryAttempt` | What a listing reports |

`GetTriggerBuilder()` round-trips the policy and deliberately drops the attempt — a rebuilt trigger has no occurrence in flight, exactly as it has no `NextFireTimeUtc`. A builder-built trigger fully defines its policy, so a definition that says nothing clears a stored one, the rule `WithExecutionGroup` and `WithPreferredNode` already follow.

**`TriggerDetailsUpdate` sets the policy and not the attempt.** The policy is configuration; the attempt belongs to the occurrence in flight, and setting it would either grant a running job attempts it was not given or take away ones it has already spent.

## Persistence

* `RETRY_POLICY` / `RETRY_ATTEMPT` are appended to the trigger SELECT column list and written unconditionally by the INSERT and by **all four** flavours of the UPDATE. The set clause sits between the execution group and the pin, so the binder adds its two parameters in the same place whichever flavour was picked — parameters bind in statement order and a positional provider takes them that way (#3512's rule). `RetryColumnStatementsTest` holds this; the listing projection's ordinals are held by the test that already pinned where the executing flag sits, which moved from 14 to 16.
* The policy is held on `TriggerBase` **as its stored string**, not as the value. That keeps the `[Serializable]` shape a primitive, which matters because field-based deserialization runs no constructor and a `RetryPolicy` has none to run. A blob written before this change has no such field and reads back as no policy. The parsed value is cached beside it and marked `[NonSerialized]`.
* The store read is forgiving: an unreadable policy string materializes as *no policy* rather than failing the read. The row is still a schedule, and refusing to materialize the trigger would take a job out of service over a column nothing else depends on.

## A bug this found

Writing a trigger as a plain object graph — Newtonsoft with `RegisterTriggerConverters` off, **which is the default** — could not read one back at all:

```
Newtonsoft.Json.JsonSerializationException : Unable to find a constructor to use for type Quartz.RetryPolicy.
Path 'RetryPolicy.MaxAttempts'
```

Json.NET wrote the policy as its public properties and then had nothing to rebuild it with, and the failure took the whole trigger. This is the same shape of problem #3494 was (`TimeZoneInfo`), one step worse. Fixed the same way: `RetryPolicyConverter` attached per property by `QuartzContractResolver`, not registered on the serializer — that list is consulted for a value's runtime type wherever it appears, so a policy in a job data map would be written as a bare string and lose its `$type`. `NewtonsoftTriggerRoundTripTest` covers both converter modes and fails without the fix.

## Wire half — and the rebase points

**#3548 (I2, #3522) is still open**, so the wire edits here are deliberately minimal. Files this PR touches that #3548 also touches, and which will need a rebase whichever lands second:

* `src/Quartz.AspNetCore/AspNetCore/HttpApi/OpenApi/Types.cs` — two members added to the `Trigger` schema
* `src/Quartz.Tests.Unit/Verify/PublicApiTest_Quartz.verified.txt`
* `src/Quartz/Impl/RAMJobStore.cs` — two lines, in the trigger-listing projection
* `docs/documentation/quartz-4.x/migration-guide.md`

**`HttpScheduler.cs`, `HttpApiContract/RequestAndResponses.cs` and `TriggerEndpoints.cs` are not touched at all.** Verified rather than assumed: the HTTP wire carries `ITrigger` (`ScheduleJobRequest`, `RescheduleJobRequest`), which the `TriggerConverter` serializes, so the policy crosses the wire with no endpoint or client change. That removes three of the four collision points the plan expected.

Wire snapshots: `TestData.Wire.CronTrigger` is the one fixture given a policy, so the snapshots pin what a stored policy looks like on the wire rather than only that the properties exist — `"retryPolicy": "exp;3;00:00:10;2;00:05:00"`, `"retryAttempt": 2`. The other five carry `null` / `0`.

The dashboard's trigger detail page shows both. JSON scheduling files and configuration sections take a `RetryPolicy` key holding the stored string, and a string that is not a policy is refused where the file is read rather than scheduling a trigger that would never retry.

## Baselines moved, every line

`src/Quartz.Tests.Unit/Verify/PublicApiTest_Quartz.verified.txt` — accepted through Verify, all additions except the one `TriggerHeader` constructor line:

```
+ ITrigger:                 int RetryAttempt { get; }
+ ITrigger:                 Quartz.RetryPolicy? RetryPolicy { get; }
+ IMutableTrigger:          new int RetryAttempt { get; set; }
+ IMutableTrigger:          new Quartz.RetryPolicy? RetryPolicy { get; set; }
+ ITriggerConfigurator<T>:  ITriggerConfigurator<TJob> WithRetryPolicy(Quartz.RetryPolicy? retryPolicy);
+ TriggerBuilder<TJob>:     public TriggerBuilder<TJob> WithRetryPolicy(Quartz.RetryPolicy? retryPolicy) { }
+ TriggerDetailsUpdate:     public TriggerDetailsUpdate WithRetryPolicy(Quartz.RetryPolicy? retryPolicy) { }
+ TriggerBase:              public int RetryAttempt { get; set; }
+ TriggerBase:              public Quartz.RetryPolicy? RetryPolicy { get; set; }
+ TriggerHeader:            public int RetryAttempt { get; init; }
+ TriggerHeader:            public string? RetryPolicy { get; init; }
~ TriggerHeader:            public TriggerHeader(…, string? ExecutionGroup, string? RetryPolicy, int RetryAttempt) { }   ← breaking
```

Also accepted, all of them the same two properties or two columns appearing:

* `AcquisitionSqlTest_MisfireRecovery.verified.txt` — `t.RETRY_POLICY`, `t.RETRY_ATTEMPT` in each of the five dialects' projections
* `JsonObjectSerializerTest_Serialize{Simple,Cron,CalendarInterval,DailyTimeInterval,Recurrence,CustomTriggers}.verified.txt` — `"RetryPolicy": null`, `"RetryAttempt": 0`
* `Quartz.Tests.AspNetCore/Verify/WireFormatSnapshotTest_{Simple,Cron,CalendarInterval,DailyTimeInterval,Recurrence}TriggerBody.verified.json` and `_TriggerListingBody.verified.json`

No new `TrimAnalysisBaseline.cs` entry.

## Deviations from the spec

1. **`TriggerBase.Equals` / `GetHashCode` are untouched.** The spec asked for them. Trigger equality here is by `TriggerKey` alone — `TriggerBase.Equals` compares nothing else, and no trigger implementation overrides it — so adding retry state would change what trigger identity *means* repo-wide. That is a much larger change than this one and not what the spec was reaching for.
2. **There is no `ISerializable` shape on `TriggerBase`, and no trigger binary-layout guard from #3321 to extend.** The spec said to add fields to "the existing `ISerializable` shape" and pin them with that guard. Checked: nothing under `src/Quartz/Impl/Triggers/` implements `ISerializable`; #3321 shipped guards for `CronExpression`, the calendars and `JobDataMap` only. `TriggerBase` is `[Serializable]` with plain field-based serialization, so the field *names* are the contract and an absent field defaults — which is exactly the "missing = null / 0" the spec wanted. I wrote the guard it was reaching for (`RetryPolicyOnTriggersTest.TheBlobShapeCarriesThePolicyAsAStringAndTheAttemptAsANumber`), pinning the field names, their types and which of them is `[NonSerialized]`.
3. **`HttpScheduler` needed no change** (see above). The spec listed it.
4. **`TriggerHeader`'s new parameters have no defaults.** Making them optional would have avoided touching four call sites, but it would also let a store forget to populate them silently. `ExecutionGroup` has no default either.

## Verification

```
dotnet build Quartz.slnx                                   ->  0 Warning(s), 0 Error(s)
dotnet test src/Quartz.Tests.Unit                          ->  Passed!  Failed: 0, Passed: 3900
dotnet test src/Quartz.Tests.AspNetCore                    ->  Passed!  Failed: 0, Passed: 536
dotnet test src/Quartz.Tests.Integration                   ->  Passed!  Failed: 0, Passed: 166
  --filter RAMJobStoreContractTest|SQLiteJobStoreContractTest|SqliteFileJobStoreContractTest
dotnet fallout PublishTrimmed                              ->  Succeeded (store round-trips, schedules, fires, binds)
dotnet fallout VerifyDocsSnippets                          ->  Succeeded (94 markdown files checked)
```

The container-backed dialect legs (SQL Server, Postgres, MySQL, Oracle, Firebird) are CI's; the store contract suite ran here on the RAM store and on both SQLite fixtures, which exercises the INSERT, the SELECT and all four UPDATE flavours with the new columns.

Part of #3520.


lahma:
Superseded by #3569, which carries this commit unchanged (rebased onto current `main`) together with the engine and the matrix in a single pull request against `main`.

Stacked pull requests get no `pr-integration-*` legs — those workflows fire only for pull requests targeting `main` — so the three parts have been collapsed into one branch, still as three commits.

PR #3560, opened 2026-08-30
A trigger retries its failed job on the policy it carries

lahma:
R4 of #3520 — the engine. A trigger carrying a retry policy now actually retries its failed job.

**Stacked on #3555 (R3), which is stacked on the merged R2.** Base is `main` at `4aec95a7b`; the diff against #3555 is this commit alone. R3 → R4 → R5 merge as a set.

## The decision

`TriggerBase.ExecutionComplete` gains one branch, placed after the explicit `JobExecutionException` directives (which win) and before the nothing-left-to-fire check (which a scheduled retry must be able to postpone). On a failure with attempts left it sets the next fire time to `now + policy.DelayFor(attempt + 1)`, increments the attempt, and returns the new `SchedulerInstruction.RetryTrigger`.

`NoInstruction` could not carry this: it means the opposite — neither store writes the trigger at completion, because the regular next fire time was already written when the trigger fired. A retry moves that time, so it has to be said out loud.

Four rules, each with a test that fails without it:

* **A retry never displaces the schedule.** `retryAt + 1s >= NextFireTimeUtc` → dropped, occurrence wins, attempt reset. The one-second margin is `CalendarIntervalTriggerImpl.GetFireTimeAfter` and `DailyTimeIntervalTriggerImpl.GetFireTimeAfter`, which each add a second before searching. `TheSupersedeRuleIsAWholeSecondWideOnBothSides` pins both sides of the boundary to the tick.
* **Firing a retry is not `Triggered()`.** `TriggerBase.RetryFired(ICalendar?)` — `public virtual`, on `TriggerBase` only, no `IOperableTrigger` change — advances past the retry instant with the same calendar-skip loop but touches no counter and leaves `PreviousFireTimeUtc` alone. So a retry burns no repeat count and no RRULE `COUNT` slot, and reports the *original* occurrence as its `ScheduledFireTimeUtc`. Asserted for all five shipped trigger types.
* **Exhausted → ordinary schedule, not `Error`.** One bad hour must not kill a cron trigger. A misfired retry gets the trigger's own misfire instruction and the attempt is cleared by the two paths that recompute a trigger from its schedule; there is no retry-specific misfire policy.
* **`RefireImmediately` is not a zero-delay retry.** It re-runs the job on the same thread in the same firing and does not touch `RetryAttempt`. `Unschedule*` likewise wins. A cancellation on the scheduler's own token leaves no exception behind (`JobRunShell` sets no `jobExEx`) and so never retries.

## The stores

Both dispatch on the attempt being non-zero *and* on `TriggerBase`, exactly as they already do for the misfire original fire time.

* `RAMJobStore.TriggeredJobComplete` writes **`tw.Trigger`, not the clone it was handed** — reads hand out clones of that instance. It removes the wrapper from `timeTriggers` before moving the fire time and re-adds it after, because the set is ordered by that time and the `[DisallowConcurrentExecution]` unblock above may already have put it back.
* `AdoJobStoreBase.Completion` writes through one narrow statement (`NEXT_FIRE_TIME`, `RETRY_ATTEMPT`, `TRIGGER_STATE`), before the unblock transitions so they leave the row alone, with the state from `ApplyPausedTriggerGroupState` so a retry in a paused group waits with the group.
* **The attempt is written only when it changes.** Clearing is a second, narrower statement issued only when the completing trigger actually had a non-zero attempt — which is what the new internal `TriggerBase.RetryAttemptCleared` records. An ordinary completion still costs no extra round trip.

## Public API delta

`src/Quartz.Tests.Unit/Verify/PublicApiTest_Quartz.verified.txt`, accepted through Verify. Every moved line:

```
+ SchedulerInstruction:  RetryTrigger = 7,                       ← appended; no existing value moved
+ IJobExecutionContext:  int RetryAttempt { get; }
+ JobExecutionContextImpl: public int RetryAttempt { get; }
+ TriggerBase:           public virtual void RetryFired(Quartz.ICalendar? calendar) { }
+ DailyTimeIntervalTriggerImpl: public override void RetryFired(Quartz.ICalendar? calendar) { }
+ IDriverDelegate:       ValueTask<int> UpdateTriggerForRetry(ConnectionAndTransactionHolder, IOperableTrigger, StoredTriggerState, CancellationToken = default);
+ IDriverDelegate:       ValueTask<int> ClearTriggerRetryAttempt(ConnectionAndTransactionHolder, TriggerKey, CancellationToken = default);
+ StdAdoDelegate:        public virtual … UpdateTriggerForRetry(…) { }
+ StdAdoDelegate:        public virtual … ClearTriggerRetryAttempt(…) { }
```

`src/Quartz.Tests.Unit/Verify/LogEventCatalogTest_Quartz.verified.txt`, one added line:

```
+ 1056  Information  CoreLog.TriggerRetryScheduled
+     "Job of trigger {TriggerKey} failed; retry {Attempt} of {MaxAttempts} scheduled for {RetryTimeUtc}"
```

**Breaking for anything implementing `IDriverDelegate`** — two new members, no default implementations, matching how `UpdateTriggerPreferredNodeConditional` and the listing members were added. `StdAdoDelegate` implements both `virtual`. Called out in the migration guide.

One meter, `quartz.trigger.retry`, tagged like `quartz.trigger.misfire` (scheduler, group, execution group — deliberately not the trigger name, which is a cardinality no alert can be built on).

## Tests

* `RetryEngineTest` (20) — the decision and `RetryFired`, against a trigger and a fake clock, no store.
* `RetryStoreTest` (3) — the store round trip: completion → acquisition, which is where the attempt has to survive. Driven against `RAMJobStore` directly so a failure says which half lost it.
* `RetryExecutionTest` (3) — a real scheduler, a job that really fails, the meter, and `IJobExecutionContext.RetryAttempt` reading `0, 1, 2` across three firings that all report the **same** `ScheduledFireTimeUtc`, which is what leaving `PreviousFireTimeUtc` alone buys.
* `TriggeredJobCompleteTest` — its `EveryInstructionHasACase` guard caught the new enum member; the matrix now covers `RetryTrigger` on both stores.

The full `RetryMatrixTest` (the 15 mandatory cases) is R5.

## Found while writing the tests, and deliberately left alone

A trigger built with `.WithSimpleSchedule(x => x.WithInterval(1h).RepeatForever()).StartNow()` **fires twice at scheduler startup**, milliseconds apart, with two distinct fire-instance ids and two distinct scheduled fire times (`…:26.752` and `…:26.760`, 8 ms apart — the second being the trigger's real start time). Reproduced with **no retry policy at all**, so it predates this change and is not caused by it. It is why the end-to-end tests here use a one-shot schedule: the retry count would otherwise be ambiguous. Worth its own issue; I have not opened one.

## Verification

```
dotnet build Quartz.slnx                                   ->  0 Warning(s), 0 Error(s)
dotnet test src/Quartz.Tests.Unit                          ->  Passed!  Failed: 0, Passed: 4105
dotnet test src/Quartz.Tests.AspNetCore                    ->  Passed!  Failed: 0, Passed: 537
dotnet test src/Quartz.Tests.Integration                   ->  Passed!  Failed: 0, Passed: 174
  --filter RAMJobStoreContractTest|SQLiteJobStoreContractTest|SqliteFileJobStoreContractTest
dotnet test src/Quartz.Tests.Integration                   ->  Passed!  Failed: 0, Passed: 87
  --filter PostgresJobStoreContractTest                         (postgres:15.1 container)
dotnet fallout PublishTrimmed                              ->  Succeeded; no new TrimAnalysisBaseline entry
dotnet fallout VerifyDocsSnippets                          ->  Succeeded (96 markdown files checked)
```

SQLite plus one Docker dialect locally; CI runs the rest of the matrix.

Part of #3520.

### DIFF
--- a/docs/documentation/quartz-4.x/migration-guide.md
+++ b/docs/documentation/quartz-4.x/migration-guide.md
@@ -4334,6 +4334,43 @@ and no attempt.
 A string that is not a policy is refused where the file is read, rather than scheduling a trigger that
 would never retry.
 
+### What happens when a job fails
+
+A job fails, for retry purposes, when `Execute` throws anything other than a `JobExecutionException`
+that asks for something itself. On failure with attempts left the scheduler sets the trigger's next
+fire time to `now + policy.DelayFor(attempt + 1)`, increments the attempt and hands the job store a new
+instruction, `SchedulerInstruction.RetryTrigger`. Everything else about the trigger is untouched.
+
+| Member | What it is |
+|---|---|
+| `SchedulerInstruction.RetryTrigger` | New enum member, appended — no existing value moved. `ITriggerListener.TriggerComplete` sees it; there is no new listener member |
+| `IJobExecutionContext.RetryAttempt` | `0` on a regular fire, *n* on the *n*-th retry |
+| `TriggerBase.RetryFired(ICalendar?)` | `public virtual`, and only on `TriggerBase` — `IOperableTrigger` is unchanged |
+| `IDriverDelegate.UpdateTriggerForRetry` / `ClearTriggerRetryAttempt` | **Breaking for anything implementing `IDriverDelegate`**; `StdAdoDelegate` implements both `virtual` |
+
+Four rules are worth knowing:
+
+* **A retry never displaces the next scheduled occurrence.** If `retryAt + 1s >= NextFireTimeUtc` the
+  retry is dropped and the occurrence supersedes it, attempt reset. The one-second margin exists because
+  `CalendarIntervalTriggerImpl.GetFireTimeAfter` and `DailyTimeIntervalTriggerImpl.GetFireTimeAfter` add
+  a second before searching. A retry is also never scheduled past the trigger's `EndTimeUtc`.
+* **A retry fire is not a `Triggered()` call.** `RetryFired` advances past the retry instant with the
+  same calendar-skip loop, but touches no counter and leaves `PreviousFireTimeUtc` alone — so a retry
+  burns no repeat count and no RRULE `COUNT` slot, and reports the *original* occurrence as its
+  `ScheduledFireTimeUtc`.
+* **Exhausted attempts go back to the ordinary schedule, not to `TriggerState.Error`.** One bad hour
+  must not kill a cron trigger. A misfired retry gets the trigger's own misfire instruction and the
+  attempt is cleared; there is no retry-specific misfire policy.
+* **`RefireImmediately` is not a zero-delay retry.** It re-runs the job on the same thread inside the
+  same firing, with no delay, no ceiling, nothing persisted and no slot released, and it does not touch
+  `RetryAttempt`. An explicit `RefireImmediately` or `Unschedule*` wins over the trigger's policy. An
+  `OperationCanceledException` from the scheduler's own token never retries — shutdown and interrupt are
+  operator decisions, and a node vanishing mid-execution is what `RequestsRecovery` is for.
+
+One meter, `quartz.trigger.retry`, counts each retry the scheduler schedules, tagged like
+`quartz.trigger.misfire`. Log event `1056` reports the trigger, the attempt and the retry instant at
+`Information`.
+
 ## Trimming annotations
 
 `ScheduleJob<T>`, `AddTrigger<TJob>` and `TriggerBuilder.Create<TJob>()` gained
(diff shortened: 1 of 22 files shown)


lahma:
Superseded by #3569, which carries this commit unchanged (rebased onto current `main`) together with R3 and R5 in a single pull request against `main`.

Stacked pull requests get no `pr-integration-*` legs — those workflows fire only for pull requests targeting `main` — so the three parts have been collapsed into one branch, still as three commits.

PR #3565, opened 2026-08-30
A firing is linked to the activity that scheduled it

### LINKED ISSUE #3524: A firing is linked to the activity that scheduled it
A scheduled job runs minutes, hours or days after the call that scheduled it, on another node, with no
relationship to the trace that asked for it. Every integrator that cares reaches for the trigger's
`JobDataMap` and invents a key.

**Verified negatives** (searched all of `src/`, every project)

- `grep -rniE "traceparent|tracestate|ActivityContext|ActivityLink" src/ --include=*.cs --include=*.razor --include=*.json`
  → 2 hits, both `ActivityCreationOptions<ActivityContext>` in test `ActivityListener` sampling callbacks
  (`src/Quartz.Tests.Unit/Diagnostics/JobExecutionObservabilityTest.cs:112`,
  `TracingJobStoreTest.cs:78`). **No propagation of any kind exists in product code.**

**Evidence**

- `src/Quartz/Diagnostics/QuartzActivitySource.cs:11-24` is `StartJobExecute`; the link is added there,
  after `CreateActivity` and before `Start()`.
- The span name is unchanged: `src/Quartz/Diagnostics/OperationName.cs:7`,
  `Job.Execute = "Quartz.Job.Execute"`.
- `src/Quartz.Tests.Unit/Diagnostics/JobExecutionObservabilityTest.cs:109-121` installs an always-on
  `ActivityListener` for the whole fixture — that is the fixture whose map-count assertions this PR must
  sweep.

**Shape**

- Reserved keys `QRTZ_TRACEPARENT` and `QRTZ_TRACESTATE` on `SchedulerConstants`, spelled like the seven
  existing ones (`SchedulerConstants.cs:56-101`).
- `QuartzSchedulerOptions.PropagateTraceContext`, **default `true`**. No `QuartzPropertyBridge` entry —
  there is no 3.x key to bridge.
- Written in `QuartzScheduler` at the same five entry points I1 normalizes, through one shared
  `PrepareTriggerData(IMutableTrigger)` helper, onto the **trigger's** map. Write-or-remove, because
  `AsOperableTrigger` does not clone. `[PersistJobDataAfterExecution]` writes only the job's map, so the
  two never interact.
- Read in `StartJobExecute`: `ActivityContext.TryParse` → `activity.AddLink(new ActivityLink(linked))`.
  A **link**, not a parent: the scheduled firing is its own trace root.

**Not in scope**

- Making the scheduling activity the firing's parent. A day-long trace is not a trace.
- Baggage propagation.

**Done means**

1. The two constants, the option, the write helper and the link.
2. The writes happen only when an `ActivityListener` exists and `Activity.Current` is non-null, so
   ordinary tests see no extra map entries.
3. The map-count assertions under `JobExecutionObservabilityTest`'s always-on listener and the
   `WireFormatSnapshotTest_*` trigger snapshots are swept. **If that sweep turns out to be large, flip the
   default to `false` for alpha.4 and record the reason in the migration guide** rather than editing
   assertions in bulk.
4. A test asserting the link is present on the execute span and absent when the option is off, extending
   `src/Quartz.Tests.Unit/Diagnostics/JobExecutionObservabilityTest.cs`.
5. Lands after I1 — both edit the same five `QuartzScheduler` entry points.

lahma:
A scheduled job runs minutes, hours or days after the call that asked for it, on another node, with
no relationship to the trace that wanted it. Every integrator that cared reached into the trigger's
`JobDataMap` and invented a key of its own.

The scheduler now records the W3C trace context of whatever activity is current when a trigger is
stored, and the firing's `Quartz.Job.Execute` span carries an `ActivityLink` back to it.

## What changed

**Two reserved keys and one option**, all in `Quartz`:

```csharp
SchedulerConstants.TraceParent = "QRTZ_TRACEPARENT";
SchedulerConstants.TraceState  = "QRTZ_TRACESTATE";
QuartzSchedulerOptions.PropagateTraceContext { get; set; } = true;
```

Those three lines are the whole of the public API delta, and the only change to
`PublicApiTest_Quartz.verified.txt`. There is no `QuartzPropertyBridge` entry: nothing on 3.x did
this, so there is no flat `quartz.*` key to migrate — and `LegacyPropertyKeys` correctly rejects an
invented one. The value is carried to the scheduler on `QuartzSchedulerResources`, the way
`ShutdownJobInterruption` is.

**Written in `QuartzScheduler`.** The five entry points that already normalized a typed job input now
do both through one `PrepareTriggerData(IMutableTrigger)` / `PrepareJobData(JobDataMap)` pair, so
there is a single place a map is made ready to be stored: both `ScheduleJob` overload pairs,
`AddJob`, `ScheduleJobs`, and both `TriggerJob` forms.

**Read in `QuartzActivitySource.StartJobExecute`**, where `ActivityContext.TryParse` becomes
`activity.AddLink(...)` before `Start()`. The span name is unchanged.

## Three decisions worth a reviewer's attention

**A link, not a parent.** A trace held open from Monday's request to Sunday's firing is a trace no
backend renders and no operator reads. The firing stays its own trace root; the link is how you walk
back to the request. This is the shape OpenTelemetry gives an asynchronous producer and the consumer
that eventually picks the work up.

**The trigger's map, never the job's.** A firing is what the trace is about, and a job is fired by
many triggers. It also keeps the write away from `[PersistJobDataAfterExecution]`, which writes back
only the job's map — so the two can never interact.

**Write-or-remove.** `AsOperableTrigger` hands the store the very object the caller passed rather
than a copy. A trigger scheduled once inside a request and again outside one would otherwise keep
pointing at the first request's trace, and a stale link reads exactly like a true one. Opting out
with `PropagateTraceContext = false` leaves both keys entirely alone, including one an application
wrote itself.

## The default is `true`, and the sweep was empty

The brief said to flip the default to `false` if sweeping the fixtures under an always-on
`ActivityListener` turned out large. It turned out to be **zero assertions**: the write only happens
when `Activity.Current` is set, and neither `JobExecutionObservabilityTest` nor `TracingJobStoreTest`
starts a scheduling activity — their listeners only observe Quartz's own source. The
`WireFormatSnapshotTest_*` trigger snapshots serve canned triggers through a `FakeItEasy` scheduler,
so no `QuartzScheduler` code runs for them at all. The only baseline that moved is the public API
one, by the three lines above. The default therefore ships as `true`.

## Tests

- `JobExecutionObservabilityTest` gains five: the execute span carries exactly one link whose
  context, `tracestate` and `IsRemote` match the scheduling activity's and which is *not* its parent;
  the reserved keys are on the trigger; no link and no keys when there was no activity; opt-out
  honoured; and a stale `traceparent` is dropped when the same trigger object is rescheduled outside
  an activity.
- `TraceContextThroughAdoStoreTest` (new, file-SQLite, no container, runs in the `basic` leg) proves
  the link survives an actual round trip: the trigger is written to a database as job data, read back
  by the scheduler thread as a row, and the firing's span still links back.
- `TraceContextOverHttpTest` (new) confirms the HTTP API gets it for free — a request carrying a
  `traceparent` reaches the trigger through ASP.NET Core's own server span, with a real scheduler
  behind the API rather than this folder's fake.

Both halves were mutation-checked: disabling the link makes exactly one test fail, disabling the
stale-key removal makes exactly one other fail.

## Verification

`dotnet build Quartz.slnx` 0 warnings, 0 errors · `Quartz.Tests.Unit` 4051 passed ·
`Quartz.Tests.AspNetCore` 538 passed · `Quartz.Tests.Integration` (container-free leg) 357 passed ·
`dotnet fallout PublishTrimmed` succeeded · `VerifyDocsSnippets` succeeded ·
`npm run docs:check-links` no dead links or anchors.

## Docs

Migration guide gains "A scheduled job links back to the trace that scheduled it"; the observability
page gains "Linking a firing to what scheduled it".

Closes #3524

### DIFF
--- a/docs/documentation/quartz-4.x/migration-guide.md
+++ b/docs/documentation/quartz-4.x/migration-guide.md
@@ -1810,6 +1810,70 @@ conversation can share a group and be listed, paused or unscheduled together. Ca
 
 See [One-Off Job](how-tos/one-off-job.md#a-payload-and-a-time-in-one-call).
 
+## A scheduled job links back to the trace that scheduled it
+
+New in 4.0, and on by default: when a scheduling call is made inside an `Activity`, the scheduler
+records that activity's W3C trace context on the trigger, and the firing's `Quartz.Job.Execute` span
+carries an `ActivityLink` back to it. Nothing has to be configured, and no code changes.
+
+3.x propagated nothing, so every integrator that cared reached into the trigger's `JobDataMap` and
+invented a key of its own:
+
+```diff
+- trigger.UsingJobData("traceparent", Activity.Current?.Id);   // and a job that read it back by hand
++ // nothing — the scheduler writes it, and the execute span links to it
+```
+
+**A link, not a parent.** A job scheduled for next Tuesday runs a week after the request that asked for
+it, quite possibly on another node. Making the firing a *child* of the scheduling span would produce a
+trace that stays open for a week, which no backend can render and no operator can read. The firing is
+its own trace root and the link is how you walk back — the shape OpenTelemetry gives an asynchronous
+producer and the consumer that eventually picks the work up. The span name is unchanged.
+
+**The API added.** In `Quartz`: `SchedulerConstants.TraceParent` (`"QRTZ_TRACEPARENT"`),
+`SchedulerConstants.TraceState` (`"QRTZ_TRACESTATE"`), and
+`QuartzSchedulerOptions.PropagateTraceContext`, which defaults to `true`. There is no flat `quartz.*`
+key for it: nothing on 3.x did this, so there is no configuration to migrate — and `LegacyPropertyKeys`
+correctly rejects an invented one.
+
+```csharp
+q.ConfigureScheduler(options => options.PropagateTraceContext = false);
+```
+
+**Where it is written.** On the **trigger's** data map, by the scheduler, at every entry point that
+stores a trigger — both `ScheduleJob` overload pairs, `ScheduleJobs`, both `TriggerJob` forms and
+`RescheduleJob`. A reschedule counts because it is the call that decided when the firing happens, so the
+link follows the move: a replacement trigger rebuilt from the original carries the original's data map,
+reserved keys included, and the new context overwrites the old one rather than being kept beside it.
+
+`UpdateTriggerDetails` deliberately writes **no** trace context. It changes metadata on a trigger that
+stays scheduled and keeps its fire times, so the trace that asked for the firing is still the one that
+scheduled it — re-pointing the link at whoever later edited a description would make it say something
+false. (It does normalize a typed input in the map it is given, because that is about what a store can
+hold rather than about what scheduled anything.)
+
+Two consequences worth knowing:
+
+- **Two more entries per trigger row.** They are ordinary string job data, so they travel through
+  `StoreJobDataAsStrings`, the System.Text.Json write gate, the Newtonsoft serializer, the binary blob
+  column and the HTTP wire without any of them knowing about it — and they are visible in
+  `MergedJobDataMap`, in the dashboard and in `GET /triggers`, like every other `QRTZ_*` reserved key.
+  Turn the option off if that is not wanted.
+- **The trigger's map, never the job's.** `[PersistJobDataAfterExecution]` writes back only the job's
+  map, so the two never meet and a persisted job cannot carry a `traceparent` forward into its next
+  firing.
+
+**A stale context is removed rather than kept.** The scheduler stores the trigger object it was handed
+rather than a copy of it, so a trigger scheduled once inside a request and again outside one has the key
+*removed* on the second call. A stale link reads exactly like a true one, which makes leaving it the
+worst of the three answers.
+
+**Over HTTP it is free.** `POST /jobs/{group}/{name}/trigger` and the scheduling endpoints run inside
+ASP.NET Core's own server span, so the request's trace reaches the trigger without `Quartz.AspNetCore`
+mentioning tracing at all.
+
+See [Observability](packages/opentelemetry-integration.md#linking-a-firing-to-what-scheduled-it).
+
 ## A component of your own is chosen the same way a shipped one is
 
 Three seams that had no code-first spelling at all, and one that only worked through a type-name string:
--- a/docs/documentation/quartz-4.x/packages/opentelemetry-integration.md
+++ b/docs/documentation/quartz-4.x/packages/opentelemetry-integration.md
@@ -75,6 +75,31 @@ and are constants on `Quartz.Diagnostics.ActivityTags`:
 | `quartz.jobstore.trigger.count` | `Quartz.JobStore.AcquireNextTriggers` (how many came back) and `.TriggersFired` (how many were fired) |
 | `error.type` | any span that ended in a failure |
 
+### Linking a firing to what scheduled it
+
+A job runs minutes, hours or days after the call that scheduled it, quite possibly on another node. When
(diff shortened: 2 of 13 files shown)


lahma:
Review gap closed, plus the audit it asked for.

**`RescheduleJob` now calls `PrepareTriggerData(trigger)`**, right after `Validate()`, symmetrical with `ScheduleJob`. #3559's `NextFireTimeUtc >= StartTimeUtc` guard and its comment are untouched — the call sits above the calendar lookup, well clear of them.

Both halves were real. The trace half is the write-or-remove case exactly: a replacement built with `GetTriggerBuilder()` copies the original's job data map, reserved keys included, so an application rescheduling a trigger it scheduled earlier kept the first call's `traceparent`. The input half was an I1 gap inherited from before this branch — `NormalizeInput` was never called there either.

**Audit of every other path that hands a trigger or a job data map to the store:**

| Path | Verdict |
|---|---|
| `UpdateTriggerDetails` / `TriggerDetailsUpdate.WithJobDataMap` | **Was a gap.** It replaces the whole map and normalized nothing. Now calls `PrepareJobData` — the input half only. |
| `ScheduleJob(IJobDetail, IReadOnlyCollection<ITrigger>, …)` | Covered. It builds a one-entry dictionary and delegates to `ScheduleJobs`, which already prepares both. |
| `DeferredScheduler` | Nothing needed. Every member resolves the real scheduler and forwards; there is no replay buffer that stores anything itself. |
| `XmlSchedulingDataProcessor` / `ContainerConfigurationProcessor` | Nothing needed. They call `scheduler.ScheduleJob` / `AddJob`, so they arrive through the prepared entry points. |
| `TriggeredJobComplete` | Nothing needed. It hands back the trigger and detail the store already gave out; no caller-supplied map. |

**`UpdateTriggerDetails` deliberately writes no trace context**, and that is the one judgement call here worth a second opinion. It is not a scheduling call — it edits metadata on a trigger that stays scheduled and keeps its fire times — so the trace that asked for the firing is still the one that scheduled it. Re-pointing the link at whoever later edited a description would make the link say something false. The input half still applies, because that is about what a store can hold rather than about what scheduled anything. Reasoning is in the code comment and in the migration guide.

**Tests added** (four, all mutation-checked):

- `ReschedulingInsideAnActivity_LinksTheFiringToTheReschedule` — the firing links to the *second* activity, and its trace id is not the first's.
- `ReschedulingOutsideAnyActivity_DropsTheLinkTheFirstSchedulingLeft` — no link, and `QRTZ_TRACEPARENT` is gone from the merged map.
- `TheInputOnARescheduledTriggerIsStoredAsAString` + `AnInputOnARescheduledTriggerReachesTheJob` — the I1 half.
- `TheInputOnAnUpdatedTriggerIsStoredAsAString` — the `UpdateTriggerDetails` half.

Both reschedule trace tests reschedule with a trigger rebuilt from the original, so the replacement really does carry the stale key — without that they would pass against an empty map and prove nothing. Reverting `PrepareTriggerData(trigger)` fails three of them; reverting `PrepareJobData(update.JobDataMap)` fails the fourth. Worth noting that `AnInputOnARescheduledTriggerReachesTheJob` passes either way on the in-memory store, which hands the object straight back — that asymmetry is why the stored-as-a-string test sits beside it.

**Re-verified** on `1405c059b`, rebased onto `571beb5db`: `dotnet build Quartz.slnx` 0 warnings 0 errors · `Quartz.Tests.Unit` 4063 passed · `Quartz.Tests.AspNetCore` 538 passed · `Quartz.Tests.Integration` basic leg 357 passed · `VerifyDocsSnippets` succeeded · `npm run docs:check-links` no dead links or anchors.

PR #3567, opened 2026-08-30
A one-word question does not need a query object

### LINKED ISSUE #3529: Five listing calls that need a query object to ask a one-word question
The paged query objects are the right primitive and they are staying. What they cost is the trivial case:
the migration guide's own replacement for a parameterless listing is
`QueryFireInstances(new FireInstanceQuery())`.

**Verified negatives** (baseline
`src/Quartz.Tests.Unit/Verify/PublicApiTest_Quartz.verified.txt`)

- **L2** — `QueryFireInstances` (`:697`), `QueryJobs` (`:699`), `QueryTriggers` (`:701`) each take a
  **required** query argument. No parameterless overloads.
- **L8** — `grep -n "QueryFireInstancesOfJob"` → 0 hits. `FireInstanceQuery` (`:414-422`) does have a
  `JobKey? Job` filter, so this is a shorthand, not a missing capability.
- **L11** — `Exists(JobKey)` (`:676`) and `Exists(TriggerKey)` (`:677`). There is no calendar overload; the
  replacement loads and deserializes the calendar to find out whether it is there.
- **L13** — `ISchedulerFactory` has `GetScheduler(ct)` (`:729`) and `LookupScheduler(string, ct)` (`:730`,
  returning **nullable**). There is no throwing `GetRequiredScheduler`.

**Correction to the plan (L15).** The plan reads "`QueryTriggersInError()` + `ResetTriggersFromErrorState()`
companion", implying neither exists. `ResetTriggerFromErrorState(TriggerKey)` and
`ResetTriggersFromErrorState(IReadOnlyCollection<TriggerKey>)` **do** exist (baseline `:703`, `:704`), and
`TriggerQuery.State` (`:1780`) already filters by `TriggerState.Error`. The gap is narrower than stated:
finding the triggers in error and resetting them is two calls plus paging, and there is no shorthand for
either half.

**Items**

| # | Signature | Why |
|---|---|---|
| L2 | `QueryJobs(ct)`, `QueryTriggers(ct)`, `QueryFireInstances(ct)` overloads that default the query | the guide's own replacement text is `new FireInstanceQuery()` |
| L8 | `QueryFireInstancesOfJob(JobKey jobKey, CancellationToken = default)` | shorthand over `FireInstanceQuery { Job = … }` |
| L11 | `ValueTask<bool> Exists(string calendarName, CancellationToken = default)` | today's answer deserializes the calendar to say yes or no |
| L13 | `ValueTask<IScheduler> GetRequiredScheduler(string schedulerName, CancellationToken = default)` on `ISchedulerFactory` — throws a named `SchedulerException` | `LookupScheduler` returns null and every caller writes the same throw |
| L15 | `QueryTriggersInError(CancellationToken = default)` shorthand, and a `ResetTriggersFromErrorState(GroupMatcher<TriggerKey>)` companion | see the correction above |

**Not in scope**

- Changing `PagedQuery`, `Take` defaults, or the shape of any existing query object.
- Removing the query-object overloads. All five items are additive shorthands.

**Done means**

1. The five items shipped through `IScheduler`/`ISchedulerFactory` and both delegating implementations, and
   through `HttpScheduler` where the HTTP API already answers the question (L11 and L15 need endpoint
   checks; L2/L8 map onto existing routes).
2. Tests in `src/Quartz.Tests.AspNetCore/HttpApi/QueryEndpointsTest.cs` for anything reaching the wire, and
   in `src/Quartz.Tests.Integration/Impl/JobStoreContractTest.cs` for `Exists(calendarName)` so every store
   answers the same.
3. Baselines + migration-guide rows.
4. L15's shorthand does not change what `ResetTriggersFromErrorState` does to a trigger; it only changes how
   the caller names the set.

lahma:
Five listing calls that needed a query object to ask a one-word question, from #3529. Everything here is additive — no query object changes shape, no `Take` default moves, and no query-object overload is removed.

## What changed

**Parameterless query shorthands** (L2, L8, L15) — extensions in `SchedulerQueryExtensions`:

| Instead of | Write |
|---|---|
| `QueryJobs(new JobQuery())` | `QueryJobs()` |
| `QueryTriggers(new TriggerQuery())` | `QueryTriggers()` |
| `QueryFireInstances(new FireInstanceQuery())` | `QueryFireInstances()` |
| `QueryFireInstances(new FireInstanceQuery { Job = jobKey })` | `QueryFireInstancesOfJob(jobKey)` |
| `QueryTriggers(new TriggerQuery { State = TriggerState.Error })` | `QueryTriggersInError()` |

They page the way the member they call pages — `PagedQuery.DefaultTake`, `HasMore` for the rest — which is *not* how the `Get*` compatibility listings in that class behave. The class remark said "every method here enumerates the whole result"; it now names the two families and says which is which.

`ResetTriggersFromErrorState(GroupMatcher<TriggerKey>)` is the companion: the listing that names the failed triggers of a group, then `ResetTriggersFromErrorState(keys)`. Two round trips, not a transaction, and what resetting a trigger does is unchanged — the shorthand only changes how the caller names the set. A null matcher throws rather than quietly meaning every group.

**`Exists(string calendarName)`** (L11) on `IScheduler` and `IJobStore`. The old answer, `GetCalendar(name) is not null`, reads the stored blob and deserializes it to throw the result away. `RAMJobStore` looks the name up without cloning; `AdoJobStoreBase` runs the existence statement. Over HTTP: `GET …/calendars/{calendarName}/exists`, mirroring the job and trigger routes.

**`ISchedulerFactory.GetRequiredScheduler(name)`** (L13), throwing a new `SchedulerNotFoundException : SchedulerException` that carries `SchedulerName`.

## Corrections to the issue's assumptions

- **L11 needs no new `IDriverDelegate` member.** The issue allowed for one (with the `UpdateTriggerPreferredNodeConditional` precedent). `IDriverDelegate.CalendarExists` and `StdAdoConstants.SqlSelectCalendarExistence` — literally `SELECT 1 FROM …CALENDARS WHERE …` — have been there all along, used by `AddCalendar` and by the trigger update that validates a calendar reference. They simply had nothing above them. `AdoJobStoreBase.Exists(string)` is `ExecuteWithoutLock(conn => CalendarExists(conn, …))`.
- **L15 needs no new endpoint.** The trigger query route already takes `state`, and `POST …/triggers/keys/reset-from-error-state` already exists, so the extension composes over the wire for free. L2 and L8 likewise map onto existing routes.
- **L11 does need one.** Jobs and triggers have `/exists`; calendars did not. Rather than approximate it with `QueryCalendarNames(nameEquals=…)`, the calendar route is added in the family's shape — same `ExistsResponse`, same path suffix.

## For a reviewer to decide

1. **L13 is an extension, not a DIM.** The issue says "on `ISchedulerFactory`"; the brief said to pick whichever does not break implementers, and both qualify. I chose the extension because it is composition over `LookupScheduler` that no factory could implement better, and because an extension also resolves on a concrete receiver such as the public `StandaloneSchedulerFactory`, where a default interface method would need a cast to be callable. If you would rather it were a DIM on the interface, it is a small change and the baseline row moves.
2. **The shorthands live in `SchedulerQueryExtensions` rather than a class of their own.** The issue said "beside `SchedulerQueryExtensions`". One discoverable class of query conveniences on `IScheduler` seemed better than two, at the cost of the class holding two families that page differently — which the class doc now spells out.
3. **`Exists(calendarName)` does not validate a blank name**, matching `GetCalendar`/`DeleteCalendar` on the same type: a name nobody registered answers `false`. `HttpScheduler` still rejects a blank name client-side, because it cannot build the URL.
4. **`IJobStore` gains a required member**, so a custom store must implement it. Migration guide updated: the "If you implement `IJobStore`" section now says four members, and the removed-members table's `CalendarExists(name)` row points at the new overload instead of at `GetCalendar(name) is not null`.

## Tests

- `SchedulerQueryShorthandTest` — the shorthands list what the spelled-out query would, still page at `DefaultTake` (251 jobs, `HasMore`), list a gated job's firing, and reset only the matching group's failed triggers while leaving another group's alone (triggers driven into `TriggerState.Error` by a throwing job factory).
- `SchedulerFactoryExtensionsTest` — the required form is the lookup's answer, the missing name throws with `SchedulerName` on it, `LookupScheduler` still answers `null` for the same name.
- `AdoJobStoreBaseTest` — calendar existence calls `IDriverDelegate.CalendarExists` and never `SelectCalendar`.
- `JobStoreContractTest.CalendarExistenceIsAnsweredByName` — every store, run here against `RAMJobStore` and SQLite.
- `QueryEndpointsTest`, `TriggerEndpointsTest`, `CalendarEndpointsTest` — the shorthands over the wire, the group reset hitting both routes it composes (and posting nothing when the listing is empty), and the calendar existence route not fetching the calendar.

Baselines regenerated through Verify; the delta is the four `Exists(string calendarName)` rows per implementer, the six shorthands, `SchedulerFactoryExtensions` and `SchedulerNotFoundException`.

Closes #3529

### DIFF
--- a/docs/documentation/quartz-4.x/migration-guide.md
+++ b/docs/documentation/quartz-4.x/migration-guide.md
@@ -4283,6 +4283,7 @@ already have them.
 * **A calendar can have dependencies** — `AddCalendar(name, serviceProvider => …)` builds the calendar from the container, which the `where T : ICalendar, new()` generic overloads cannot (see [A calendar can be built from the container](#a-calendar-can-be-built-from-the-container))
 * **A `quartz.*` key that lost its prefix is refused** — `QuartzOptions.Properties` is read only through the `quartz.` prefix, so a key without one produced no symptom at all; it now fails options validation at startup (see [A property key without the `quartz.` prefix is refused](#a-property-key-without-the-quartz-prefix-is-refused))
 * **An `IJobDetail` of your own** — the interface no longer declares a member only Quartz can implement, so an application can supply its own job detail type and have `RAMJobStore` hand it back rather than quietly swapping it for Quartz's (see [An `IJobDetail` of your own](#an-ijobdetail-of-your-own))
+* **A one-word question needs no query object** — `QueryJobs()`, `QueryTriggers()`, `QueryFireInstances()`, `QueryFireInstancesOfJob(jobKey)` and `QueryTriggersInError()` are the query members with the query record filled in, `ResetTriggersFromErrorState(GroupMatcher<TriggerKey>)` resets the failed triggers of a group, `Exists(string calendarName)` answers whether a calendar is there without deserializing it, and `ISchedulerFactory.GetRequiredScheduler(name)` throws where `LookupScheduler` returns null (see [Asking a one-word question does not need a query object](#asking-a-one-word-question-does-not-need-a-query-object))
 * **Pause, resume and reset a set of keys in one call** — `PauseTriggers`, `ResumeTriggers`, `PauseJobs`, `ResumeJobs` and `ResetTriggersFromErrorState` take a key collection, do the whole set in one lock and one transaction, and answer with the keys they applied to (see [A set of keys pauses, resumes or resets in one call](#a-set-of-keys-pauses-resumes-or-resets-in-one-call))
 * **`TriggerDetailsUpdate.WithExecutionGroup`** — move a stored trigger into an execution group, or out of every group, without rescheduling it. `QRTZ_TRIGGERS.EXECUTION_GROUP` was already written by the generic trigger update, and `RAMJobStore` applies it in place the same way it applies a preferred node, so both stores behave alike
 * **A chain can fan out** — `JobChainingJobListener` takes more than one follow-up for a job, either by calling `AddJobChainLink` again with the same first job or with the new `AddJobChainLinks(firstJob, followUpJobs)`. Each follow-up is triggered as its own firing, so they run concurrently rather than one after another, and one that cannot be triggered is logged without costing its siblings theirs. On 3.x the second link threw, so a chain could only be sequential (see [How do I chain Job execution?](../faq.md#how-do-i-chain-job-execution-or-how-do-i-create-a-workflow))
@@ -4599,7 +4600,7 @@ that this process happened to be holding, so it could only ever answer for one n
 
 ```diff
 - List<IJobExecutionContext> running = await scheduler.GetCurrentlyExecutingJobs();
-+ PagedResult<FireInstance> running = await scheduler.QueryFireInstances(new FireInstanceQuery());
++ PagedResult<FireInstance> running = await scheduler.QueryFireInstances();
 ```
 
 With a persistent job store the answer now covers the whole cluster, because a firing is a row rather
@@ -4707,9 +4708,11 @@ The scheduler body's `statistics` object gained `localExecutingJobs`, mirroring 
 
 ### If you implement `IJobStore`
 
-Three members. `QueryFireInstances(FireInstanceQuery, CancellationToken)` is the listing, abstract like
+Four members. `QueryFireInstances(FireInstanceQuery, CancellationToken)` is the listing, abstract like
 the rest of the query family; `QueryClusterNodes(CancellationToken)` is the node listing described in
-[The nodes of a cluster are a listing](#the-nodes-of-a-cluster-are-a-listing). And `Initialize` takes a
+[The nodes of a cluster are a listing](#the-nodes-of-a-cluster-are-a-listing);
+`Exists(string calendarName, CancellationToken)` is what replaced `CalendarExists` and is meant to be
+answered without materializing the calendar. And `Initialize` takes a
 `SchedulerIdentity`, which is the whole of what it takes — see
 [SPI changes](#spi-changes) for what its 3.x parameters became:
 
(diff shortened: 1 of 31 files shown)


lahma:
Rebased onto `0b98adf6a`. One real conflict, plus two things worth a reviewer's eye.

**The conflict** was `RequestDelegatesAreSourceGeneratedTest`'s per-file route counts: #3558 bumped `JobEndpoints.cs` to 21 and `TriggerEndpoints.cs` to 21, this branch bumps `CalendarEndpoints.cs` to 5. Resolved as all three. The two public-API baselines auto-merged, but I discarded that merge and re-derived them: reset both to `origin/main`'s version, ran `PublicApiTest`, and accepted the received files. The resulting delta against main is **20 added lines and no deletions** in `PublicApiTest_Quartz.verified.txt` (+1 in `…HttpClient…`), so I3's `DeleteJobs`/`UnscheduleJobs` rows, I4's trace-context rows and the SIMPROP constant are all intact — a hand-merge is exactly what would have silently dropped one.

**#3558 has now landed, and its group-matcher forms have a different shape from this PR's.** `DeleteJobs(GroupMatcher)` and `UnscheduleJobs(GroupMatcher)` are `IScheduler` members with `POST …/jobs/delete-by-group` and `…/triggers/unschedule-by-group` behind them, because the store does each in one lock and one transaction. `ResetTriggersFromErrorState(GroupMatcher)` here is an extension composing two existing calls, with no new endpoint, no `IJobStore` member and no atomicity — which is what #3529 asked for ("all five items are additive shorthands"; "does not change what `ResetTriggersFromErrorState` does to a trigger; it only changes how the caller names the set").

I have left it as the shorthand rather than mirroring the new sibling, because making it atomic is a different and larger change: a new `IJobStore` member, `RAMJobStore` and ADO implementations, a driver-delegate statement, `DelegatingJobStore`/`TracingJobStore`, and an endpoint. But the asymmetry is now visible in a way it was not when the issue was written — three group-matcher bulk operations, two of them transactional and one not — so if you would rather it matched, that is a follow-up worth its own issue and I would rather you decided than me.

**One number was already wrong on main and I fixed it in passing.** `http-api.md` says "*N* routes in four groups" above the four per-group tables. #3558 bumped the two group counts to 21 but left the prose at "Fifty-seven", which no longer matched its own tables (13 + 21 + 21 + 4 = 59). With the calendar route it is 60, and the line now reads "Sixty". Nothing checks that total — `RequestDelegatesAreSourceGeneratedTest` counts the per-file dictionary, not the prose — which is presumably how it drifted.

Re-verified on the rebased head: `dotnet build Quartz.slnx` clean (0 warnings, 0 errors), `Quartz.Tests.Unit` 4125 passed, `Quartz.Tests.AspNetCore` 548 passed, the SQLite and RAM job-store contract fixtures 182 passed, `VerifyDocsSnippets` up to date, `npm run docs:check-links` clean (962 links, 585 fragments).

Issue #3537, opened 2026-08-30
Embedding Quartz in a library is documented


lahma:
Eight patterns are re-implemented by every embedder — ensure-durable-job-then-trigger (three wrong ways),
job-per-message payload, cancel by external id, a private `IJobFactory`, contributing to a scheduler you do
not own, retry via `JobExecutionException`, lifecycle tied to a module rather than `IHost`, and
sync-over-async. None of them is written down.

**Verified negatives**

- `ls docs/documentation/quartz-4.x/how-tos/` → `README.md`, `aspire.md`, `custom-job-store.md`,
  `dialect-delegate.md`, `job-template.md`, `lock-handler.md`, `multiple-triggers.md`, `one-off-job.md`,
  `rescheduling-jobs.md`, `trigger-persistence-delegate.md`, `trimming-and-native-aot.md`. There is **no**
  embedding page and **no** external-leader page.

**Evidence**

- The firing-latency defaults an integration doc must address:
  `src/Quartz/Configuration/QuartzSchedulerOptions.cs:46` `IdleWaitTime = 30s`, `:72` `MaxBatchSize = 1`,
  `:83` `BatchTriggerAcquisitionFireAheadTimeWindow = TimeSpan.Zero`. In-process scheduling signals the
  loop, so the 30 s bounds only *cross-node* pickup — but next to a bus that polls every 5 s the numbers
  read badly, and the tuned triple has to ship with the explanation.
- The external-leader topology is spelled out in code but presented nowhere:
  `src/Quartz/Configuration/QuartzOptionsValidators.cs:435` `ClusteringStaysEnabledValidator`.
- The contributor recipe already works and is order-independent:
  `QuartzServiceCollectionExtensions.cs:243` `ConfigureAllQuartzSchedulers` /
  `SchedulerNameRegistry.cs:100-102`. OpenIddict's 3.x approach — `IConfigureOptions<QuartzOptions>` calling
  `QuartzOptions.AddJob`/`AddTrigger` — is gone in 4.0, and this is its replacement.
- **OpenTelemetry.Instrumentation.Quartz is silently dead on 4.0**: it subscribes to the `"Quartz"`
  `DiagnosticListener` that 3.x's `JobDiagnosticsWriter` wrote, while 4.0 emits on an `ActivitySource`
  (`src/Quartz/Diagnostics/QuartzInstrumentation.cs:23`) with `quartz.*` tags.
  `AddQuartzInstrumentation()` yields zero spans and no warning.

[...]

PR #3578, opened 2026-08-31
The one-off job surface answers what its first integration asked of it

### LINKED ISSUE #3574: The one-off job surface, corrected by its first real integration
Porting MassTransit's Quartz integration onto the alpha.4 primitives (#3535) shrank it by 87 lines and deleted its job factory and bus-holder classes — and surfaced four gaps in the one-off scheduling surface, each found by an integration that could not say what it needed. All four are small and additive; they belong in alpha.4 while the surface is new.

1. **The ensured durable job cannot be configured.** `ScheduleJob<TJob, TInput>` ensures a durable job per `TJob`, hard-coded to `StoreDurably()` plus a canned description. MassTransit's job was `RequestRecovery().StoreDurably()`; adopting the one-liner silently drops `RequestRecovery()` from the job every scheduled message hangs off, and there is no hook to put it back — re-`AddJob`-ing over it fights the ensure memo. Add `OneOffJobOptions.RequestRecovery { get; init; }` (default `false`), applied when the job is ensured. A full `Action<JobBuilder<TJob>>` hook is deliberately *not* proposed: the memo ensures once per scheduler, so a delegate that varies per call would be first-wins and a trap; a named boolean cannot vary meaningfully.

2. **The ensured job's key is private while its group's doc says "don't build your own".** The key is `(typeof(TJob).Name, SchedulerConstants.ScheduledJobGroup)`, derived inside the extension — and `ScheduledJobGroup`'s own remarks tell clients not to use the group for jobs of their own. An integration with a second scheduling path (MassTransit's cron path builds its own trigger) has to re-derive the key by hand against that advice. Expose it: `public static JobKey ScheduledJobKey<TJob>()` on `SchedulerJobExtensions`, and have the how-to's cron-shaped example use it.

3. **The one-liner returns less than the overload it wraps.** `ScheduleJob(trigger, options)` returns the first fire time; `ScheduleJob<TJob, TInput>` returns only the `TriggerKey`, so a caller logging "scheduled for X" falls back to the requested time. Return a `readonly record struct ScheduledOneOffJob(TriggerKey TriggerKey, DateTimeOffset FirstFireTimeUtc)` from both one-liner overloads (pre-release; callers adapt). Keep it two members — this is not a query DTO.

4. **A missing input throws, which turns a 3.x upgrade into a runtime trap.** `JobInput.Read<TInput>` throws `SchedulerException` when nothing sits under `QRTZ_JOB_INPUT` — right for a fresh 4.x app, but an integration that converts its job to `IJob<TInput>` over a store carrying 3.x-era rows (payload as flat keys) gets a throw on every pre-upgrade firing, and its only recourse today is dropping back to `IJob` entirely. Add `TryGetInput<TInput>(this IJobExecutionContext, out TInput?)` returning `false` when the key is absent (still throwing on a present-but-undeserializable value — corruption is not compat), so a compat-minded `IJob` can read both shapes. Document the pattern in the one-off how-to and add a migration-guide note.

Also from the port, needing only documentation: `OneOffJobOptions.Group` defaults to `typeof(TJob).Name`, so an integration with an existing trigger-key contract (MassTransit cancels by `new TriggerKey(id)` in the *default* group) must set `Group = TriggerKey.DefaultGroup` explicitly or its cancellation silently stops matching. Say so on the property's XML doc and in the how-to; the default itself is right.

Evidence: the port lives in a local branch pair (`quartz-4-port-before` @ e63fe193e, 1,418 LOC → `quartz-4-port-after` @ 638e678a1, 1,331 LOC; identical 84/87 test results); finding 1 is `SchedulerJobExtensions.EnsureJob`'s builder call, finding 2 is `JobKeyFor<TJob>` vs `SchedulerConstants.ScheduledJobGroup`'s remarks, finding 4 is `JobInput.Read<TInput>`'s throw path.

lahma:
Porting a message bus's Quartz integration onto the alpha.4 one-call scheduling overloads found four
things the surface could not say. All four are small and additive, and they belong here while the
surface is new.

**The ensured durable job can be told to request recovery.** `OneOffJobOptions.RequestRecovery`
(default `false`) is applied when the job is ensured. It is a named boolean rather than an
`Action<JobBuilder<TJob>>` hook, deliberately: the job is ensured once per scheduler instance, so a
delegate that varied per call would be first-wins and a trap. The XML doc and the how-to both say that
the first call's value stands for the process's lifetime, and that this is how the ensure memo already
treats the job's description, its durability and the type it names.

**The ensured job's key is public.** `SchedulerJobExtensions.ScheduledJobKey<TJob>()` returns
`(typeof(TJob).Name, SchedulerConstants.ScheduledJobGroup)`, which is what an integration with a second
scheduling path needs — its own recurring trigger has to point at the same job rather than re-derive a
key inside a group whose own remarks say not to build jobs in it. `ScheduledJobGroup`'s remarks now say
both halves: do not put jobs of your own here, and this is how you address the one that is here. The
internal derivation is the same call, so there is one spelling of it. It needs no
`[DynamicallyAccessedMembers]` — the key is derived from the type's name, nothing is reflected over, and
the trim analyzers are clean.

**Both one-liner overloads answer with `ScheduledOneOffJob`** — a two-member `readonly record struct`
carrying the `TriggerKey` and the `FirstFireTimeUtc` the store computed — rather than the key alone,
which was less than the `ScheduleJob(trigger, options)` overload they wrap already returns. Every caller
in the repository is swept: the unit tests, the documentation samples, the Wolverine example (parts 2
and 3) and its hand-written page fences, and the embedding how-to's samples that landed in #3537 while
this was being written.

**`TryGetInput<TInput>(out TInput?)` answers whether a firing carries an input at all.** An application
converting a job to `IJob<TInput>` over a store that already holds its 3.x triggers finds those carrying
the payload as flat job data keys and nothing under `QRTZ_JOB_INPUT`, and `IJob<TInput>` fails every one
of them by name. A compat-minded `IJob` can now read both shapes. A value that is present and unreadable
still throws — corruption is not compatibility — and only absence is answered with `false`.

Documentation: `OneOffJobOptions.Group`'s XML doc and the how-to both carry the cancel-contract trap
(the default is the job type's name, so an integration whose callers cancel with `new TriggerKey(id)`
must set `Group = TriggerKey.DefaultGroup`); the one-off how-to gains the `RequestRecovery` and
`ScheduledJobKey` sections and a "reading input written by an older schema" section, all as
`#region sample_*` blocks in `Quartz.Documentation.Samples`; the migration guide's one-call section is
corrected for the new return type and gains the `RequestRecovery`, `ScheduledJobKey` and `Group`
paragraphs, and its typed-input section gains the `TryGetInput` compatibility note.

## For a reviewer to decide

* `RequestRecovery` is passed to the builder unconditionally (`.RequestRecovery(requestRecovery)`)
  rather than called only when `true`. The effect is identical — `false` is `JobBuilder`'s own default —
  and it is one line with no branch to leave untested.
* The embedding how-to's `Conversations.Remind` still returns a `TriggerKey`, projected from the record,
  because that page's point is the key it cancels by. The one-off how-to's sample shows both members.
* The Wolverine example's part 2 now records the fire time beside the key in its ledger, which is what
  made the new return value worth having; `--smoke` still passes.

## Verification

* `dotnet build Quartz.slnx` — succeeded, no warnings.
* `dotnet test src/Quartz.Tests.Unit` — 4142 passed, 0 failed.
* `dotnet test src/Quartz.Tests.AspNetCore` — 548 passed, 0 failed.
* `dotnet test src/Quartz.Tests.Integration` with the container-free `basic` filter — 375 passed, 0
  failed. The database legs need Docker and were not run here.
* `dotnet fallout DocsSnippets` + `VerifyDocsSnippets` — up to date, 101 markdown files checked.
* `npm run docs:check-links` — 219 pages, 1024 internal links, 619 fragments, no dead links or anchors.
* `dotnet run --project src/Quartz.Examples.Wolverine -- --smoke` — `smoke: ok`.

`PublicApiTest_Quartz.verified.txt` moves by exactly the four changes — `TryGetInput`,
`OneOffJobOptions.RequestRecovery`, the `ScheduledOneOffJob` type, `ScheduledJobKey<TJob>()` and the two
changed return types — accepted through Verify, never edited by hand.

Closes #3574

### DIFF
--- a/docs/documentation/quartz-4.x/how-tos/embedding-quartz-in-a-library.md
+++ b/docs/documentation/quartz-4.x/how-tos/embedding-quartz-in-a-library.md
@@ -189,7 +189,10 @@ of it:
 
 * **The job key says what runs.** One durable job per job type is enough, because a job detail is a
   definition rather than an occurrence. That is what the typed `ScheduleJob` overloads store, at
-  `(typeof(TJob).Name, SchedulerConstants.ScheduledJobGroup)` — one row per job type, whatever the traffic.
+  `SchedulerJobExtensions.ScheduledJobKey<TJob>()` — which is
+  `(typeof(TJob).Name, SchedulerConstants.ScheduledJobGroup)`, one row per job type whatever the traffic.
+  Ask for the key rather than re-deriving it: the group is reserved, and a library that also builds a
+  trigger of its own points it at that same job.
 * **The trigger key says which occurrence.** Its **name** is this one firing and its **group is the
   correlation id**: the saga, the conversation, the tenant. Everything scheduled for one of those shares a
   group and can be listed, paused or cancelled together.
@@ -215,9 +218,9 @@ public sealed class SendReminderJob(IReminderSink sink) : IJob<SendReminder>
 ```csharp
 public sealed class Conversations(IScheduler scheduler)
 {
-    public ValueTask<TriggerKey> Remind(SendReminder reminder, TimeSpan delay, CancellationToken cancellationToken)
+    public async ValueTask<TriggerKey> Remind(SendReminder reminder, TimeSpan delay, CancellationToken cancellationToken)
     {
-        return scheduler.ScheduleJob<SendReminderJob, SendReminder>(
+        ScheduledOneOffJob scheduled = await scheduler.ScheduleJob<SendReminderJob, SendReminder>(
             reminder,
             delay,
             new OneOffJobOptions
@@ -229,6 +232,9 @@ public sealed class Conversations(IScheduler scheduler)
                 Replace = true
             },
             cancellationToken);
+
+        // The call answers with the key it stored and the time the store will first fire it at.
+        return scheduled.TriggerKey;
     }
 
     public ValueTask<bool> Cancel(TriggerKey firing, CancellationToken cancellationToken)
@@ -318,7 +324,7 @@ check and the write in which another node does the same thing. Both `ScheduleJob
 ```csharp
 ITrigger trigger = TriggerBuilder.Create<SendReminderJob>(scheduler.TimeProvider)
     .WithIdentity(reminder.MessageId, reminder.ConversationId)
-    .ForJob(new JobKey(nameof(SendReminderJob), SchedulerConstants.ScheduledJobGroup))
+    .ForJob(SchedulerJobExtensions.ScheduledJobKey<SendReminderJob>())
     .StartAt(at)
     .UsingInput(reminder)
     .Build();
--- a/docs/documentation/quartz-4.x/how-tos/one-off-job.md
+++ b/docs/documentation/quartz-4.x/how-tos/one-off-job.md
@@ -140,9 +140,9 @@ public sealed class SendInvoiceJob : IJob<SendInvoice>
 
 public sealed class Invoicing
 {
-    public async ValueTask Remind(IScheduler scheduler, SendInvoice invoice, CancellationToken cancellationToken)
+    public async ValueTask Remind(IScheduler scheduler, ILogger logger, SendInvoice invoice, CancellationToken cancellationToken)
     {
-        TriggerKey firing = await scheduler.ScheduleJob<SendInvoiceJob, SendInvoice>(
+        ScheduledOneOffJob firing = await scheduler.ScheduleJob<SendInvoiceJob, SendInvoice>(
             invoice,
             TimeSpan.FromDays(7),
             new OneOffJobOptions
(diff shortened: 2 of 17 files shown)


lahma:
The head moved to `6d4507988` after the first push: the recipe under *A schedule of your own on the same job* needed one more paragraph. A trigger cannot be stored for a job that is not there, and the ensured job only appears the first time a one-call overload is used on that scheduler — so a second scheduling path that can run *first* has to store the durable job itself with `AddJob(job, AddJobOptions.Replacing)` under the same key. Documenting the key without that sentence would have taught a recipe that throws `JobPersistenceException` on a cold store.

lahma:
Rebased onto `7b5e57c04` (PR #3576) and swept its `BusIntegrationScenarioTest` — nine one-liner call sites, which would have failed the merge-ref build even though the branch was green on its own. The sweep is not only mechanical: step 2 now asserts `scheduled.FirstFireTimeUtc` against the stored trigger's start time and advances the fake clock to the time the *call* answered with rather than to a second read of the store, and the replace step asserts the replacement's own fire time and fires on it. No existing assertion was relaxed. `DurableJobKey` is now `SchedulerJobExtensions.ScheduledJobKey<ReceiptJob>()` rather than a re-derivation, so the scenario proves the public key names the job the extension ensures on a real ADO store as well as in memory.

New head `458bb9610`. Re-verified locally: `dotnet build Quartz.slnx` clean; `BusIntegrationScenarioTest` 2/2 (both stores); unit 4142/4142; AspNetCore 548/548; container-free integration leg 377/377; `VerifyDocsSnippets` up to date.

lahma:
Rebased onto `801482101` (the retry engine) — no conflict in `migration-guide.md`; the engine's sections and this PR's sit in different places and both survive.

Step 7 of the bus scenario, left as `TODO(#3520)` because nothing on `TriggerBuilder` reached a policy, is now written and passes on **both** stores:

* the trigger is one the step builds itself — `TriggerBuilder.Create<ReceiptJob>(scheduler.TimeProvider).ForJob(SchedulerJobExtensions.ScheduledJobKey<ReceiptJob>()).WithRetryPolicy(RetryPolicy.Fixed(maxAttempts: 2, RetryDelay))`, repeating every 24 h so there is still a trigger to read after the retry succeeds. That is also the second scheduling path `ScheduledJobKey<TJob>()` was made public for, exercised end to end against a real ADO store.
* after the first firing throws: `RetryAttempt == 1` on the stored trigger, `NextFireTimeUtc == Now + RetryDelay` (the policy's wait measured on the scheduler's clock, not a job sleeping in its execution slot), and `TriggerState.Normal` — a failed occurrence with attempts left is not a broken trigger.
* after the retry succeeds: the job ran exactly twice (`JobRan.Count(ran => ran.Equals(key)) == 2`), the attempts the two firings reported are `[0, 1]`, `RetryAttempt` is back to `0`, and `NextFireTimeUtc` is `at + 24h` — the occurrence the schedule called for, unmoved, because a retry burns no repeat.

Two supporting changes the step needed: `BusRecorder` records each firing's `RetryAttempt` and can be told to fail the next *n* firings of a trigger, and `BusMiddleware` now records "left" in a `finally`. The middleware change strengthens rather than weakens step 8's entered/left balance: a firing whose job throws unwinds the middleware too, and a pipeline that only closed its scope on success would leak on exactly the firings worth reading about.

New head `4c6f3b65c`. Local: solution build clean; `BusIntegrationScenarioTest` 2/2; unit 4244/4244; AspNetCore 548/548; container-free integration 408/408; `VerifyDocsSnippets` up to date; `docs:check-links` clean.

PR #3606, opened 2026-08-31
A wait no timer will take is refused where it was configured

### LINKED ISSUE #3577: A misfire handler frequency past Task.Delay's ceiling fails at shutdown, not validation
`AdoJobStoreOptions.MisfireHandlerFrequency` is validated on one side only: `QuartzOptionsValidators.cs:375` rejects a value under 1 ms, but nothing rejects a large one — and `MisfireHandler.Run` (`src/Quartz/Impl/AdoJobStore/MisfireHandler.cs:68`) hands the span straight to `Task.Delay`, whose ceiling is `uint.MaxValue - 1` milliseconds, about 49.7 days. Configure a frequency past that and the store starts fine; the `ArgumentOutOfRangeException (Parameter 'delay')` surfaces later, out of `AdoJobStoreBase.Shutdown` of all places, with nothing in the message pointing at the setting.

Found while writing the bus-integration scenario test (#3576), whose first draft used a 90-day frequency to keep the misfire handler out of the way.

Fix: add the upper bound to the same validator clause (`must be at most 49 days`, or whatever exact constant `Task.Delay` documents), so a misconfiguration fails at startup naming the option. Worth a quick sweep of the other configurable `TimeSpan`s that end up in a `Task.Delay` (`ClusterCheckinInterval`, `IdleWaitTime`, retry delays parsed from `RetryPolicy`) for the same missing ceiling — bound each where its validator lives, or state why it cannot exceed the limit in practice.

lahma:
`Task.Delay` tops out at `uint.MaxValue - 1` milliseconds — 49.7 days — and refuses anything longer with an `ArgumentOutOfRangeException` naming a parameter called `delay`. Every duration Quartz waits out that way is configurable, and none of them was checked against it. A too-large value started fine and was reported wherever the wait was eventually observed: #3577's 90-day `MisfireHandlerFrequency` surfaced out of `AdoJobStoreBase.Shutdown`, mentioning neither the setting nor the limit.

## The ceiling

`TimerLimits.MaxDelay` is `TimeSpan.FromMilliseconds(uint.MaxValue - 1)` — `System.Threading.Timer.MaxSupportedTimeout`, which every timer in the BCL validates against and none of them exposes. `TimerLimitsTest` holds it to what `Task.Delay` actually does, so a framework that moved it fails a build rather than a scheduler that has been running for a month.

Failures read `<Option> must be at most 4294967294ms (49.7 days), was 7776000000ms (90 days). <one sentence naming the wait it becomes>`, in the shape the surrounding validators already use for their lower bounds.

## Sweep

| Duration | Where it is waited | Verdict |
|---|---|---|
| `AdoJobStoreOptions.MisfireHandlerFrequency` | `MisfireHandler.Run` → `Task.Delay` | bounded — the reported bug |
| `AdoJobStoreOptions.MisfireThreshold` | the same `Task.Delay`, but only when `MisfireHandlerFrequency` is unset, because then it *is* the frequency | bounded, conditionally; with a frequency of its own it is only ever subtracted from a clock reading, and stays unbounded |
| `AdoJobStoreOptions.DbRetryInterval` | `Task.Delay` after a database failure, and both `ComputeTimeToSleep`s | bounded |
| `AdoJobStoreOptions.TransientRetryInterval` | `Task.Delay` between transient-failure retries | bounded — it had no validation at all, so it gets the non-negative check too |
| `ClusteringOptions.CheckinInterval` | `ClusterManager.Run` → `Task.Delay` | bounded |
| `QuartzHostedServiceOptions.StartDelay` | `StartDelayed` → `Task.Delay` | bounded, through an `IValidateOptions` rather than `ValidateOnStart` (see below) |
| `SelectForUpdateLockHandler.RetryPeriod`, `UpdateRowLockHandler.RetryPeriod` | `Task.Delay` between lock attempts | bounded in the accessor, because a lock handler has no options type (see below) |
| `QuartzSchedulerOptions.IdleWaitTime` | `SemaphoreSlim.WaitAsync(TimeSpan, …)` | **not bounded** — see below |
| `AdoJobStoreOptions.CommandTimeout` | `DbCommand.CommandTimeout`, whole seconds | not bounded; the int it is rounded into does not overflow under 68 years |
| `InMemoryJobStoreOptions.MisfireThreshold`, `ClusteringOptions.CheckinMisfireThreshold`, `QuartzSchedulerOptions.BatchTriggerAcquisitionFireAheadTimeWindow` | clock arithmetic only | not bounded; nothing waits them out |

### `IdleWaitTime` gets no ceiling, and that is the finding

The issue named it as a candidate, and it turns out not to be one. The idle wait is spent on a `SemaphoreSlim`, so that a scheduling change can cut it short, and on .NET 10 a semaphore takes a timeout of *any* length — I found this by probing the boundary rather than by reading, after a first draft bounded it at `int.MaxValue` milliseconds and the test refused to fail. An idle wait of years is a strange thing to configure but not a thing that breaks, so refusing it would be inventing a limit. `TimerLimitsTest` pins the semaphore behaviour the absence of the bound rests on, and the validator carries a comment saying why it is one-sided.

### Two durations that reach a timer without an options validator

- **`QuartzHostedServiceOptions.StartDelay`** is per scheduler name, and the names are not all known where validators are registered — a named scheduler may be added after `AddQuartzHostedService`. So it is an `IValidateOptions` that fires on the monitor read the hosted service makes for every scheduler while the host starts, rather than a `ValidateOnStart` declaration.
- **The lock handlers' `RetryPeriod`** has no options type at all; the flat `quartz.jobStore.lockHandler.retryPeriod` key writes the property by reflection. The check is therefore in the `init` accessor, and `PropertyBinder.SetObjectProperties` turns it into a `SchedulerConfigException` naming the key.

### `StartDelayed` checks its own argument

`QuartzScheduler.StartDelayed` runs its wait inside a `Task.Run` whose task is discarded, so a delay the timer refused faulted that task, was collected without a word, and left a scheduler that was created, bound, reported healthy — and never started. That is the worst symptom in the sweep, and it is reachable without any configuration at all, so the argument is now checked before the wait is started. A negative delay is refused with it: `-1ms` would have meant "wait forever" by the `Timeout.Infinite` convention, which `StartDelayed` never documented, and any other negative vanished the same way.

## Retry-policy delays: the clamp goes on the instant, not the parse

`RetryPolicy`'s waits never reach `Task.Delay`. They become a fire time — `TriggerBase.TryScheduleRetry` does `now + policy.DelayFor(…)` — so the limit that binds them is the calendar, not the timer, and `DateTimeOffset` addition throws where `DelayFor` deliberately saturates at `TimeSpan.MaxValue`.

It is reachable on numbers somebody might write. `RetryPolicy.Exponential(20, TimeSpan.FromSeconds(1), factor: 10)` spends a whole `TimeSpan` on its thirteenth retry, and before this the thirteenth failure answered with

```
System.ArgumentOutOfRangeException : The added or subtracted value results in an un-representable DateTime. (Parameter 't')
   at System.DateTimeOffset.op_Addition(DateTimeOffset dateTimeOffset, TimeSpan timeSpan)
```

out of `ExecutionComplete`, on the job's failure path, for a trigger that had done nothing wrong.

**Parse-time cap rejected.** It would refuse a stored form `ToStoredString` had just written, breaking the round-trip the type documents, and it would refuse a whole policy over attempts no trigger will ever reach — most retries are superseded by the next scheduled occurrence long before the waits get large. **Clamping inside `DelayFor` rejected** too: the saturation is the honest arithmetic answer and is documented public behaviour.

So the decision is made where the wait becomes an instant, as the third of the reasons `TryScheduleRetry` already has for declining one: there is no room, so the occurrence settles and the trigger keeps its ordinary schedule. The supersede margin comes off the room available, so the comparison that uses it cannot overflow either. `RetryPolicy.DelayFor`'s remarks now record the decision where a reader of the type will find it.

## Tests

Each bounded option gets a startup-validation test asserting the message names the option and the ceiling, plus a boundary case at `MaxDelay` that must pass. The retry decision gets the boundary pair — the last instant a retry may land on, and one tick past it — and the realistic exponential policy above. All of them were confirmed to fail without the corresponding product change; the retry pair fails with exactly the `Parameter 't'` exception quoted above.

## Reviewer's call

- Bounding `TransientRetryInterval` and the lock handlers' `RetryPeriod` goes slightly past what the issue asked for. Both are configurable durations that end up in a `Task.Delay`, which is the sweep the issue described, but say if you would rather they landed separately.
- No public API signature moved, so the baselines and the migration guide are untouched.

## Verification

```
dotnet build Quartz.slnx             0 Warning(s), 0 Error(s)
dotnet test Quartz.Tests.Unit        Passed! Failed: 0, Passed: 4300
dotnet test Quartz.Tests.AspNetCore  Passed! Failed: 0, Passed: 548
dotnet fallout BenchmarkSmoke        Succeeded (284 benchmarks)
```

`Quartz.Tests.Integration` was not run: no Docker daemon is available in this environment.

Closes #3577

### DIFF
--- a/src/Quartz.Tests.Unit/Configuration/OptionsValidationTest.cs
+++ b/src/Quartz.Tests.Unit/Configuration/OptionsValidationTest.cs
@@ -3,6 +3,8 @@
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Options;
 
+using Quartz.Util;
+
 namespace Quartz.Tests.Unit.Configuration;
 
 /// <summary>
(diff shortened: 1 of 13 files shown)


lahma:
Reviewed in full (approval blocked only by the own-PR rule). TimerLimits with the empirically pinned ceiling and the semaphore-absence test is the right shape; the StartDelayed unobserved-task fix and the retry clamp before the DateTimeOffset arithmetic — with the ToStoredString round-trip argument for not capping at parse — are both correct. Merging on green.

PR #3607, opened 2026-08-31
A cancelled lock acquire throws, and false means only re-entry

### LINKED ISSUE #3583: A cancelled lock acquire returns the value that means reentrancy
`InProcessLockHandler.AcquireLock` and `SqliteLockHandler.AcquireLockCore` swallow an `OperationCanceledException` and return `false` — and `ExecuteInLocalTransactionLock` reads `false` as "reentrant acquire, do not release", so taken literally a cancelled acquire would run the guarded operation *unlocked*.

It is unreachable today, for a narrow reason: both handlers have `RequiresConnection == false`, and the statement immediately after the acquire opens the connection, which observes the same token and throws — now correctly classified as cancellation by #3581. So nothing runs unlocked in any shipped configuration. But the safety rests on an accident of ordering, and a third `ILockHandler` implementation (or a reorder of the connection open) would inherit the hole silently.

The real fix is contractual: define what `ILockHandler.AcquireLock`'s `bool` means under cancellation — the honest shape is "cancellation throws, `false` means only reentrancy" — and make all five shipped handlers agree, with a test per handler that a cancelled acquire throws OCE and leaves no lock held. Found during #3503's audit (#3581 documents it in its PR body); not fixed there because it redefines an extensibility contract rather than a classification.

lahma:
Closes #3583

## The hole

`ILockHandler.AcquireLock` returns a *release obligation*, not a report of success, and the job store
reads it that way: `true` means "you took the lock, give it back", `false` means "you already held it,
do not release it". Nothing said so, and three shipped handlers took advantage of the gap — they
swallowed an `OperationCanceledException` and answered `false`:

- `InProcessLockHandler.AcquireLock`
- `SqliteLockHandler.AcquireLockCore`
- `RedisLockHandler.AcquireLock`, in **two** places: the local gate, and the `SET NX` poll loop (the
  issue named two offenders; the audit it asked for found the third)

Taken literally, a caller told `false` runs the operation the lock exists to guard with nothing holding
it, and releases nothing on the way out. Nothing does that in a shipped configuration, because all
three answer `false` to `RequiresConnection` and the connection open on the very next statement
observes the same token — but that is statement ordering, not a guarantee, and an `ILockHandler`
written elsewhere inherits the hole silently. This PR does not rely on the ordering.

## What changed

**The contract, on the interface.** `ILockHandler.AcquireLock`'s XML docs now say what the `bool`
means: `true` is an obligation to release, `false` is re-entry *and nothing else*, a refused lock is a
`LockException`, a cancelled one is an `OperationCanceledException`, and a handler that gives up leaves
neither a consumed handover nor a partially taken lock behind. The ordering accident is called out
there as the thing not to lean on. No signature changed, so the public-API baselines do not move.

**The three offenders throw.** The Redis handler keeps releasing its local gate on the way out, so a
cancelled acquire does not lock every other requestor in the process out of a lock nobody owns.

**The audit.** Five shipped handler families, seven concrete types:

| Handler | Before | Now |
|---|---|---|
| `InProcessLockHandler` | returned `false` on cancellation | throws; the `SemaphoreSlim` wait takes no count it was cancelled out of, so nothing leaks |
| `SqliteLockHandler` | returned `false` on cancellation | throws; `currentOwner`/`lockCount` untouched |
| `RedisLockHandler` | returned `false` twice | throws twice; the local gate is released before the exception escapes |
| `SelectForUpdateLockHandler` (+ `PostgreSqlSelectForUpdateLockHandler`) | already correct | unchanged; regression tests added |
| `UpdateRowLockHandler` (+ `SqlServerMemoryOptimizedUpdateRowLockHandler`) | already correct | unchanged; regression tests added |

The two row-lock families record ownership only *after* `ExecuteSql` returns, so a cancelled statement
leaves no mark that would make the requestor's next acquire look re-entrant. That was true already and
untested; it is tested now.

## Tests, and which leg they land in

Per the brief, coverage is pushed into the unit leg wherever the handler can be reached without its
engine:

- **`InProcessLockHandlerTest`** — a queued waiter cancelled mid-wait throws, the task is `Canceled`
  rather than faulted (#3581's classification), the lock stays with its holder and the next waiter is
  still served. Plus a token that has already fired: throws, and takes no hold at all.
- **`SqliteLockHandlerTest`** — the same two facts. The three existing "blocked" assertions read a
  `false` that no longer exists, so they now assert the request is still *outstanding* and the
  cancellation is what ends it. That also takes the 200 ms wall-clock sleeps out of the fixture.
- **`DbLockHandlerCancellationSqliteTest`** (new) — both row-lock handlers against a real file SQLite
  database provisioned with `ProvisionSchema()`: a cancelled acquire throws, and the same requestor's
  next acquire is answered `true` rather than `false`.
- **`RedisLockHandlerCancellationTest`** (new, unit) — the Redis handler's local-gate cancellation path
  and its failure path, with the handler pointed at an endpoint nothing listens on. No server needed;
  ~50 ms.
- **`Quartz.Tests.Integration/Redis/RedisLockHandlerTest`** — the one path that genuinely needs a live
  server: a cancellation *inside* the `SET NX` poll loop, which is only reachable once a second handler
  instance has got past its own local gate. Three existing tests asserted the old `false`.

Verified the tests fail without the product change: 11 of them do.

## Docs

- `how-tos/lock-handler.md` gains "And `false` means nothing else" beside the existing re-entry rule,
  and a third bullet in "Testing one".
- `migration-guide.md` gains a section and an appendix row — 3.x's `SimpleSemaphore.ObtainLock` and
  `RedisSemaphore.ObtainLock` have the same shape (checked against `origin/3.x`), so anyone porting a
  custom `ISemaphore` needs to know.
- `quartz-3.x/` untouched.

## For a reviewer to decide

**Should the store also refuse a `false` it was handed with a cancelled token?** A contract in XML docs
does not stop the next third-party handler from repeating this. A one-line guard at each of the seven
`LockHandler.AcquireLock` call sites — or a wrapper — would. I left it out: the issue framed the fix as
contractual, the guard touches hot paths in `AdoJobStoreBase`, and another stream owns ADO surface
reshaping. Say the word and it is a small follow-up.

**`SqliteLockHandler` answers `true` to a re-entrant acquire**, unlike the other four, because it counts
its holds and expects a matching release. That is consistent with the obligation framing and is left
alone; the interface docs now name the counted variant explicitly rather than pretending every handler
returns `false` there.

## Verification

```
dotnet build Quartz.slnx                     Build succeeded. 0 Warning(s) 0 Error(s)
dotnet test src/Quartz.Tests.Unit            Passed! Failed: 0, Passed: 4291, Skipped: 0
dotnet test src/Quartz.Tests.AspNetCore      Passed! Failed: 0, Passed:  548, Skipped: 0
dotnet test src/Quartz.Tests.Integration --filter RedisLockHandlerTest
                                             Passed! Failed: 0, Passed:   12, Skipped: 0  (real Redis container)
dotnet fallout VerifyDocsSnippets            Documentation snippets are up to date (102 markdown files checked)
dotnet run --project src/Quartz.Benchmark -c Release -- --smoke     exit 0, 284 benchmarks
```

The public-API Verify baselines did not move, as expected for an XML-docs-only interface change.

### DIFF
--- a/docs/documentation/quartz-4.x/how-tos/lock-handler.md
+++ b/docs/documentation/quartz-4.x/how-tos/lock-handler.md
@@ -61,6 +61,23 @@ lock the outer one is still relying on.
 
 `ReleaseLock` from a non-owner should warn, not throw — that is what the shipped handlers do.
 
+### And `false` means nothing else
+
+The other half of that rule: **an acquire that did not take the lock throws.** `LockException` when the
+lock was refused, `OperationCanceledException` when the token fired.
+
+::: danger
+Answering `false` on cancellation is not a gentler way of saying "I got nothing". The store reads
+`false` as *already held, do not release*, so it goes on to run the operation with no lock and releases
+nothing on the way out. Do not lean on the store's ordering to save you either: a handler with
+`RequiresConnection = false` is followed immediately by a connection open on the same token, which would
+throw first — statement ordering, not a guarantee.
+:::
+
+A handler that gives up also leaves nothing behind: a wait abandoned partway must not consume a
+handover meant for the next waiter, and anything taken before the failure — a local gate in front of a
+remote lock, say — is released before the exception escapes.
+
 ## Deriving from DbLockHandler
 
 When the lock *is* a database row, `DbLockHandler` does the plumbing — ownership tracking, re-entry,
@@ -229,10 +246,13 @@ node.
 
 ## Testing one
 
-The re-entry rule and the retry behaviour are the two things worth a test, and neither needs a
-scheduler:
+The re-entry rule, the cancellation rule and the retry behaviour are the three things worth a test, and
+none of them needs a scheduler:
 
 - Call `AcquireLock` twice with the same `requestorId` and assert the second returns `false`.
+- Acquire with a token that has already fired, and assert an `OperationCanceledException` rather than a
+  `false`. Then acquire again from another `requestorId` and assert it is served, which is how a lock
+  left held by the abandoned attempt shows up.
 - Give the handler a `FakeTimeProvider` through `LockHandlerContext` and advance it to drive the retry
   loop without the test waiting.
 
--- a/docs/documentation/quartz-4.x/migration-guide.md
+++ b/docs/documentation/quartz-4.x/migration-guide.md
@@ -6228,6 +6228,37 @@ with the same advice rather than reported as a typo. The store hands the handler
 `quartz.jobStore.tablePrefix` value, so the lock rows follow the store's table prefix without separate
 configuration.
 
+## A cancelled acquire throws, and `false` means only re-entry
+
+`ILockHandler.AcquireLock` returns a release obligation rather than a report of success: `true` says
+this call took the lock and its caller has to give it back, `false` says the calling context already
+held it and must not. That was the rule in 3.x too, and the store still reads the answer that way — what
+was never written down is that **re-entry is the only thing `false` may mean**, so an acquire that did
+not take the lock throws: `LockException` when it was refused, `OperationCanceledException` when the
+token fired.
+
+`SimpleSemaphore.ObtainLock` and `RedisSemaphore.ObtainLock` swallowed the cancellation and answered
+`false`, which the store reads as *already held, do not release* — it would then run the guarded
+operation with no lock and release nothing on the way out. Their 4.0 counterparts throw. Nothing ran
+unlocked in a shipped configuration, because both answer `false` to `RequiresConnection` and the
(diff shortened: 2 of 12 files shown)


lahma:
Reviewed in full. The release-obligation framing of the bool, the third offender found in Redis (twice), the proven-failing-first tests, and documenting SqliteLockHandler as the permitted counted variant are all right. Ruling on the flagged reviewer decision: no store-side guard at the call sites for now — the contract plus per-handler tests carry it, and the ADO stream owns that surface; revisit only if a third-party handler regression ever appears. Merging on green.

PR #3609, opened 2026-08-31
Both serializers write a string dictionary the same way

### LINKED ISSUE #3582: A string dictionary in job data does not round-trip cleanly across the two serializers
Both serializers accept a `Dictionary<string, string>` as a job-data value, but they do not write it the same way: the Newtonsoft package emits it with a `$type` property, System.Text.Json without one. A map written by a Newtonsoft-configured scheduler and read by a System.Text.Json-configured one therefore comes back carrying an extra `"$type"` entry in the dictionary — same key space as the application's own data, silently.

`JobDataMapPortabilityTest.PortableValues` covers the scalar set but not the dictionary case, so nothing pins the cross-serializer round trip today.

Observed while building #3580's write gate (the probe that verified `AddJobDataValueType` semantics). Fix candidates, in preference order: make the Newtonsoft converter write the dictionary without `$type` (matching System.Text.Json and shrinking the payload — reading old blobs that carry `$type` must keep working), or strip `$type` on read in the System.Text.Json path. Either way, extend `PortableValues` (or a sibling) to round-trip a dictionary across both serializer pairs — the reproduction should be written as the failing test first, since this report comes from a probe rather than a pinned test.

lahma:
A `Dictionary<string, string>` is the one job data value with structure that both serializers carry, and
they did not write it the same way. Json.NET names the type of any value an `object`-typed slot cannot, so
`Quartz.Serialization.Newtonsoft` wrote the map with `$type` beside it, while System.Text.Json wrote the
plain object — and the System.Text.Json reader takes every property of an object as an entry of the map.

The reproduction was written as a failing test first, since #3582 came from a probe rather than a pinned
test. `JobDataMapPortabilityTest` now round-trips a string map across all four writer/reader pairings, and
on the base commit it failed exactly as reported:

```
Failed AValueTheAccessorsCoverReadsBackWhicheverSerializerWroteIt(dictionary)
  newtonsoft -> system.text.json: Expected map["dictionary"] to be equal to {["one"] = "1", ["two"] = "2"},
  but found additional keys {"$type"}.
```

Probing the other three pairings turned up the same bug facing the other way, which the issue does not
mention: an object with no `$type` on it is a shape Json.NET has no answer for, so a map the built-in
serializer wrote came back from the Newtonsoft reader as a `JObject` — not something the job that stored a
dictionary can cast. Fixing only the write side would have converted the first failure into that one, so
both halves moved.

## What changed

**Write (`Quartz.Serialization.Newtonsoft`).** A job data map is written entry by entry rather than by
handing Json.NET a `Dictionary<string, object>`, because the slot a value sits in is what decides whether a
`$type` goes beside it. A `Dictionary<string, string>` goes out as a plain JSON object — byte for byte what
System.Text.Json writes, and smaller; everything else keeps the `object`-typed slot it has always had, which
is what still puts a `$type` beside a type declared through `AddJobDataValueType<T>()` and reads it back as
itself.

**Read (`Quartz.Serialization.Newtonsoft`).** An object with a `$type` is still handed to Json.NET, so the
payloads that are already in databases keep loading; an object without one is read back as the
`Dictionary<string, string>` it is. The object is loaded with `DateParseHandling.None`, which is not
incidental: Json.NET parses any string that looks like a timestamp into a `DateTimeOffset`, and rendering
that back as a string reformats it in the *reading machine's* culture — `1982-06-28T01:01:01+03:00` became
`06/28/1982 01:01:01 +03:00` in the naive version of this fix, and an offsetless timestamp would have picked
up the reading machine's offset. A string map hands back the strings that went into it.

**Read (System.Text.Json).** The reader drops a `$type` property when it materialises the map. Fixing only
the writer would leave every blob written before this PR — and every blob written by 3.x — still leaking a
type name into the application's key space on the first read by a System.Text.Json-configured scheduler.

**`$type` is now a reserved name inside a stored string map**, refused on write by both serializers, in the
same place and shape as the rest of the write gate from #3580. Nothing is lost by that refusal: such a map
never survived a round trip in either direction. Json.NET wrote the name twice and then failed to read its
own blob (`Could not deserialize JSON: {"dictionary":{"$type":"System.Collections.Generic.Dictionary…",
"$type":"mine","one":"1"}}`), and a map System.Text.Json wrote was one Json.NET could not load at all. This
is the one behaviour change that can fail an application that is working today — a System.Text.Json-only
application storing a literal `$type` key — and it is what lets both readers treat the name as metadata
without ever dropping an entry someone meant.

## This is a wire-format change on the Newtonsoft side

```json
{"headers":{"$type":"System.Collections.Generic.Dictionary`2[[System.String, System.Private.CoreLib],[System.String, System.Private.CoreLib]], System.Private.CoreLib","tenant":"acme"}}
```

becomes

```json
{"headers":{"tenant":"acme"}}
```

It is read-compatible in both directions **within 4.0** — both readers read both forms — so nothing has to
be rewritten and there is no migration. It is *not* backwards compatible: 3.x's Newtonsoft reader hands a
`JObject` back for the plain form. That matters in a mixed 3.x/4.0 window and on a rollback, so it is
written up in the migration guide and added to the mixed-window list in `operations.md`, whose "job data
serializes to the same JSON" claim this PR would otherwise have made false.

Docs: a new migration-guide section (**A string dictionary is written the same way by both serializers**)
with a summary bullet, the mixed-window paragraph in `operations.md`, and the reserved-name rule in
`packages/json-serialization.md` and `tutorial/job-data-map.md`.

## For the reviewer to decide

* **3.x has the identical bug and this PR does not touch it.** Verified in the source, not assumed:
  `origin/3.x` `StringKeyDirtyFlagMapConverter` writes through `serializer.Serialize(writer, map.WrappedMap)`
  with `TypeNameHandling.Auto` (so `$type`), and 3.x's System.Text.Json `GetJobDataMap` deserializes the
  object straight into a `Dictionary<string, string>` (so the `$type` entry comes back). Whether to change
  what 3.x writes is a wire decision on a released branch, and it is yours — flagged, not made. A cheaper
  half is available there on its own: dropping `$type` on the 3.x *System.Text.Json read* side would fix the
  leak for existing data without changing a single byte 3.x writes.
* **The Newtonsoft trigger converter path is left alone, deliberately.** With
  `RegisterTriggerConverters = true`, a trigger whose job data map holds a `Dictionary<string, string>`
  cannot be written at all today — `SerializationExtensions.WriteJobDataMapValue` writes each value with
  `JsonWriter.WriteValue(object)`, which throws

  ```
  Unsupported type: System.Collections.Generic.Dictionary`2[System.String,System.String].
  Use the JsonSerializer class to get the object's JSON representation. Path 'JobDataMap'.
  ```

  even though the write gate accepts the value; and a map written by System.Text.Json comes back from that
  path as a `JObject`. I did not fix it here because the write half alone would turn a loud failure into a
  silent one,
  and the read half cannot be made exact: `TriggerConverter` loads the whole trigger with `JObject.Load`, so
  the nested timestamps are already parsed by the time the job data map is reached, and the only way to keep
  them verbatim is to load the entire trigger with `DateParseHandling.None` — which changes what a *trigger*
  blob's top-level date-shaped values read back as, and affects third-party `ITriggerSerializer`
  implementations. That is a second wire decision, not a tidy-up, and it belongs in its own issue. The
  default path (`RegisterTriggerConverters = false`) goes through the converter this PR fixes, so a trigger's
  job data map is correct out of the box.
* **A `SchedulerContext` gets the same treatment**, since it shares the converter: a string map in one is
  written plain and read back as a string map rather than a `JObject`. It is not stored, so this is only
  visible to something that serializes one itself.

## Neighbouring shapes checked

The write gate admits exactly one dictionary type, so there is no second flavour to align. The rest of the
accepted set is written as JSON primitives by both serializers and was already identical — the portability
matrix covers each of them, and it now carries the string map (including a timestamp-shaped entry, which is
the trap above), a legacy `$type` payload written out as a literal rather than produced by the current
writer, and the reserved-name refusal.

Two defensive gaps the entry-by-entry read opened are closed with tests rather than left to chance: a job
data map payload that is not a JSON object, and structure nested inside a stored string map, both now fail
as `Quartz.JsonSerializationException` carrying the payload instead of letting a Json.NET or `Argument`
exception escape.

## Verification

* `dotnet build Quartz.slnx` — Build succeeded, 0 warnings, 0 errors
* `dotnet test src/Quartz.Tests.Unit` — Passed! Failed: 0, Passed: 4288
* `dotnet test src/Quartz.Tests.AspNetCore` — Passed! Failed: 0, Passed: 548
* `dotnet test src/Quartz.Tests.Integration --filter SQLiteJobStoreContractTest` — Passed! Failed: 0,
  Passed: 93 (the container-free slice; the full integration suite needs several database images, which the
  worktree's ~6 GB disk budget did not allow, and this PR touches no ADO store code)
* `dotnet fallout VerifyDocsSnippets` — Succeeded, 102 markdown files checked
* `dotnet fallout BenchmarkSmoke` — Succeeded, 284 benchmarks executed

No public API moved, so no baseline changed.

Closes #3582

### DIFF
--- a/docs/documentation/quartz-4.x/migration-guide.md
+++ b/docs/documentation/quartz-4.x/migration-guide.md
@@ -374,6 +374,9 @@ here moved for you; if you called `GetInt` and `GetBoolean`, all of it did.
 * **Both serializers refuse a job data value they cannot read back**, at write time rather than at the next
   read, and both refuse the same set — see
   [Both serializers refuse a job data value they cannot read back](#both-serializers-refuse-a-job-data-value-they-cannot-read-back).
+* **A `Dictionary<string, string>` job data value is written the same way by both serializers**, which is
+  a change to what Newtonsoft writes — see
+  [A string dictionary is written the same way by both serializers](#a-string-dictionary-is-written-the-same-way-by-both-serializers).
 * **Five calendar types serialize to a new shape.** 4.x reads both forms and 3.x reads only its own, so
   this is invisible on a clean cut-over and fatal in a mixed window — see
   [A mixed 3.x and 4.0 window](operations.md#a-mixed-3-x-and-4-0-window).
@@ -3517,6 +3520,51 @@ own, a `JobKey` or `TriggerKey` held as a job data value, and a collection. A `T
 `JobDataMap` are past declaring — Json.NET could not read either back — so store a zone's `Id`, and
 serialize a nested structure in the job and store the result as a string. `StoreJobDataAsStrings` was
 never affected by any of this.
+
+### A string dictionary is written the same way by both serializers
+
+A `Dictionary<string, string>` is the one job data value with structure that both formats carry, and until
+4.0 they did not write it the same way. Json.NET names the type of any value an `object`-typed slot cannot,
+so the Newtonsoft serializer wrote the map with the type beside it:
+
+```json
+{"headers":{"$type":"System.Collections.Generic.Dictionary`2[[System.String, System.Private.CoreLib],[System.String, System.Private.CoreLib]], System.Private.CoreLib","tenant":"acme"}}
+```
+
+System.Text.Json wrote the plain object, and its reader takes every property of an object as an entry of
+the map — so a value written by a Newtonsoft-configured scheduler and read by a System.Text.Json-configured
+one came back carrying an assembly-qualified type name under a `$type` key, in the same key space as the
+application's own data, with nothing said about it.
+
+4.0's Newtonsoft serializer writes the plain object, which is byte for byte what System.Text.Json writes:
+
+```json
+{"headers":{"tenant":"acme"}}
+```
+
+**This is a change to the bytes Newtonsoft writes, and both readers read both forms.** A blob already in
+the database keeps loading: Json.NET still builds the map out of the type it named, and the
+System.Text.Json reader now treats a `$type` property as the metadata it always was rather than handing it
+back as an entry. Nothing has to be rewritten, and there is no migration to run.
+
+The one direction that does not hold is backwards: 3.x's Newtonsoft reader has no answer for an object with
+no type beside it, and hands back a Json.NET `JObject` instead of the dictionary. That matters only while a
+3.x node is reading what a 4.0 node wrote — see
+[A mixed 3.x and 4.0 window](operations.md#a-mixed-3-x-and-4-0-window).
+
+**`$type` is now a reserved name inside a stored string map.** Both serializers refuse a map that stores an
+entry under it, the same way and at the same moment as any other value they cannot read back:
+
+```
+Job data entry 'headers' holds a string map with an entry named '$type'. '$type' is the name Json.NET
+writes a value's type under, so Quartz's JSON format reads it as metadata rather than as an entry and no
+reader can hand it back. Store that entry under a name of its own.
+```
+
+Nothing is lost by that refusal, because such a map never survived a round trip: Json.NET wrote the name
+twice and then failed to read its own blob, and a map System.Text.Json wrote was one Json.NET could not
+load at all. The refusal is what turns a blob in the column that nobody can read into a failure the
+application that wrote it hears about.
 
 ### Custom trigger and calendar serializers are no longer static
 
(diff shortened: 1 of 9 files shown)


lahma:
Reviewed in full. Fixing both halves (Newtonsoft write minus $type, both readers tolerant, the JObject trap closed entry-by-entry with DateParseHandling.None pinned) was the right call — the issue's either/or framing missed that the read side is what rescues already-stored blobs. Rulings on the flagged decisions: (1) $type as a reserved key in stored string maps is ACCEPTED — probed evidence shows such a map never round-tripped anyway, and pre-GA is the moment to reserve it; (2) the RegisterTriggerConverters=true path stays out of this PR — it is a second wire decision and gets its own issue; (3) the 3.x companion remains flagged for Marko, not made implicitly. Merging on green.

PR #3610, opened 2026-08-31
A wildcard in a day field restricts nothing

### LINKED ISSUE #3589: A wildcard day field no longer swallows the other day field's restriction
4.0 made specifying both day-of-month and day-of-week legal where 3.x/Java threw — but the union is an unconditional OR (\`ProgressNextFireTimeDay\`, \`CronExpression.cs:2305-2338\`), so a wildcard in **either** day field makes the expression fire every day. \`0 15 10 1 * * 2010\` — day-of-month 1, day-of-week \`*\`, the natural Unix spelling of "the 1st at 10:15" — fires all 31 days of October, pinned by \`CanUse_DayOfMonth_And_DayOfWeek_Together\` (\`CronExpressionTest.cs:210-214\`). 3.x forced a loud \`?\`; 4.0 accepts the habit and silently schedules daily.

The fix is the Vixie/Cronos restricted-field rule: \`*\` and \`?\` in a day field mean *unrestricted*, and an unrestricted field defers to the other one (AND); the OR applies only when **both** fields are genuinely restricted — which is the case 4.0 got right (\`CronExpressionTest.cs:210-212\` keep their expectations). \`?\` becomes what it always pretended to be: a synonym for \`*\` in the day fields.

Blast radius: a parse-time record of "was this day field a wildcard", the union logic in \`ProgressNextFireTimeDay\`, and exactly the two wildcard \`TestCase\`s above, which flip from "all 31 days" to the day the expression names. The differential minute-walk harness proves \`IsSatisfiedBy\` and the chained walk agree afterwards. Migration-guide row required: an alpha-persisted trigger whose expression carries \`* * MON\`-style shapes changes schedule, and a \`CronCalendar\` built on one shifts its exclusions — both from "every day" to "what the expression says", which is the direction users meant.

Ratified 2026-08-31 as part of the alpha.5 cron alignment; design note validates the rule's interaction with \`L\`/\`W\`/\`#\` before implementation.

lahma:
Closes #3589. This is PR-1 of the alpha.5 cron-semantics stream.

## The rule

`ProgressNextFireTimeDay` decided which day field to walk with `!field.Contains(NoSpec)` — "is this
field not a `?`". `*` is not a `?`, so `0 15 10 1 * *` and `0 15 10 * * MON,THU,FRI` both took the
union branch and unioned "every day" with the days the other field named, which is every day. The
first of those is the natural Unix spelling of "the 1st at 10:15", and 4.0 accepted it and silently
scheduled daily where 3.x had forced a loud `?`.

A day field written **exactly** `*` or `?` now names no days, so it restricts nothing and defers to
the other field:

| day-of-month | day-of-week | result |
|---|---|---|
| restricted | unrestricted | day-of-month decides |
| unrestricted | restricted | day-of-week decides |
| restricted | restricted | union — unchanged |
| unrestricted | unrestricted | every day |

`*/n`, ranges, lists, `L`, `W`, `#` and `nL` are all **restricted**. That falls out of the existing
representation rather than being spelled out: `AddToSet` routes a bare `*` to `CronField`'s wildcard
flag and adds no bits, while `*/2` goes down the range loop and sets concrete bits; `L`/`W` add no
bits but fill `lastDaySpecs`/`nearestWeekdays`, which leaves the field empty of both bits and markers
and therefore reads as restricted. No parse-time flag, no new field, no serialization impact.

**This is Vixie, not Cronos, and the divergence is deliberate.** Cronos always ANDs the two day
fields — its `CronExpressionFlag` has no star flag at all, because with an unconditional AND it needs
none. Quartz unions when both fields are restricted, which is the `crontab(5)` rule, is what
`cron-expressions.md` has always documented (`0 15 10 1,2,3 * MON,FRI` is "the 1st, 2nd, 3rd **and**
every Monday and Friday"), and is what the 4.0 alphas shipped. A reviewer who "finishes" the Cronos
alignment by making the both-restricted case AND would halve everyone's fire count. It is written down
as a ratified divergence in the migration guide row.

`*/n` in a day field is restricted, which is where we diverge from Vixie a second time: Vixie sets
`DOM_STAR` from the leading `*` *before* consuming `/2`, so `*/2` there is a wildcard. That is a parse
accident that makes `*/1` behave differently from `1-31/1`, which are the same set. The rule as
written — a field that says nothing about which day lets the other field decide; two fields that each
name days are unioned — needs no reference to parser internals.

## `?` and `*` become synonyms

`StoreExpressionQuestionMark` now stores the wildcard instead of the `NoSpec` sentinel, and the
`'?' can only be specified for Day-of-Month -OR- Day-of-Week.` rejection is gone. Two reasons, and the
first is mechanical: a `?` field is not walkable — `CronField.TryGetMinValueStartingFrom` returns the
sentinel `98` for it — and the both-unrestricted case has to walk *one* of the two fields. Normalising
at parse time makes either choice safe. The second is that forbidding two spellings of the same
statement is arbitrary once they are the same statement, so `? * ?` is now legal and means every day.

`?` stays legal **only** in the two day fields; `QuestionMark_IsRejectedOutsideTheTwoDayFields` pins
that, because the sibling rule it used to sit beside is gone and the survivor should not go with it by
accident.

## What flips

An independent scan of every cron-shaped 6/7-field literal in the worktree (85,905 candidates across
`.cs`/`.md`/`.xml`/`.json`/`.razor`/`.yml`/…) for "neither day field is `?`, at least one is `*`, and
not both are `*`" found **exactly the two pinned `TestCase`s** that #3589 named, plus
`CronExpressionTest.cs`'s five-field error-message assertion, which is message text and is never
evaluated as an expression. Nothing in `docs/`, no XML fixture, no benchmark, no sample.

* `0 15 10 * * MON,THU,FRI 2010` — all 31 days of October 2010 → `1, 4, 7, 8, 11, 14, 15, 18, 21, 22, 25, 28, 29`
* `0 15 10 1 * * 2010` — all 31 days → `1`

The three both-restricted cases in the same grid are unchanged.

## Tests

* `CanUse_DayOfMonth_And_DayOfWeek_Together` now pins the whole 2×2, the `?`↔`*` synonymy in both
  directions, both-wildcard in all four spellings, and that `L`, `15W`, `6#3` and `6L` each restrict
  and make the wildcard opposite them yield.
* `CronExpressionDifferentialTest.NextExpression` grew a third shape that fills **both** day fields,
  with the reference rule written out longhand. Either field can come out a wildcard, so the fuzz
  reaches the whole 2×2 rather than only the union — and the union branch had never been fuzzed at
  all, because the old generator put a `?` in exactly one day field by construction.
* Verified the new tests fail without the product change: reverting the router alone fails eight of
  the new `TestCase`s and the differential seeds (I saw the first three fail before truncating the output).

## Reviewer decisions

* **The Vixie-not-Cronos divergence** is the one thing worth ratifying again in review. It is recorded
  in the migration-guide row so the next reader does not "fix" it.
* **`*/n` in a day field is restricted** (see above). If you would rather be Vixie-literal here, that
  is a one-line change in `IsRestrictedDayField` and a note in the guide — but it is a worse rule.
* A **set-based** rule — "unrestricted = the field admits every day", so `1-31` and `SUN-SAT` join
  `*` and `?` — was considered and declined: it makes `1-30` and `1-31` behave differently, and it is
  not what was ratified. Adopting it later only ever narrows a schedule, so it stays available as a
  4.1 refinement.

## Deliberately left alone

`CanGetExpressionSummary`'s pinned text changes by one line (`daysOfWeek: ?` → `daysOfWeek: *`), which
is the expected consequence of the `?` normalisation. The member is slated for removal in a later PR of
this stream; this one only updates the assertion.

`CronField.isNoSpec` stays. It is unreachable from the parser now, but `CronFieldTest` pins it directly
and removing it is unrelated cleanup.

## Verification

```
dotnet build Quartz.slnx                                    Build succeeded. 0 Warning(s) 0 Error(s)
dotnet test src/Quartz.Tests.Unit                           Passed! Failed: 0, Passed: 4311
dotnet test src/Quartz.Tests.AspNetCore                     Passed! Failed: 0, Passed:  548
dotnet fallout VerifyDocsSnippets                           Succeeded (102 markdown files checked)
dotnet fallout BenchmarkSmoke                               Succeeded (284 benchmarks)
```

No public-API baseline moved — no public member changed shape.

### DIFF
--- a/docs/documentation/quartz-4.x/cron-expressions.md
+++ b/docs/documentation/quartz-4.x/cron-expressions.md
@@ -44,9 +44,9 @@ NOTE: There are many cron standards/implementations. The results from some gener
 ## Special characters
 
 * `*` ("all values") - used to select all values within a field. For example, `*` in the minute field means "every minute".
-* `?` ("no specific value") - useful when you need to specify something in one of the two fields in which the character is allowed, but not the other.
-For example, if I want my trigger to fire on a particular day of the month (say, the 10th), but don't care what day of the week that happens to be,
-I would put `10` in the day-of-month field, and `?` in the day-of-week field. See the examples below for clarification.
+* `?` ("no specific value") - allowed in the day-of-month and day-of-week fields, where it is a synonym for `*`: both say that the field names no days.
+Use it when you need to specify something in one of the two fields but not the other. For example, if I want my trigger to fire on a particular day of the month (say, the 10th),
+but don't care what day of the week that happens to be, I would put `10` in the day-of-month field, and `?` in the day-of-week field. See the examples below for clarification.
 * `-` - used to specify ranges. For example, `10-12` in the hour field means "the hours 10, 11 and 12".
 * `,` - used to specify additional values. For example, `MON,WED,FRI` in the day-of-week field means "the days Monday, Wednesday, and Friday".
 * `/` - used to specify increments. For example, `0/15` in the seconds field means "the seconds 0, 15, 30, and 45".
@@ -248,7 +248,13 @@ Here are some full examples:
 | `0 H/15 * * * ?`           | Fire every 15 minutes, starting from a hash-derived offset                                                                          |
 
 ::: tip
-Pay attention to the effects of '?' and '*' in the day-of-week and day-of-month fields!
+Pay attention to the effects of `?` and `*` in the day-of-week and day-of-month fields. A day field
+written exactly `*` or `?` names no days, so it restricts nothing and the other day field decides:
+`0 15 10 1 * *` fires on the 1st of the month, and `0 15 10 * * MON` fires every Monday. Only when
+**both** fields name days does the expression fire on the union of the two, as
+`0 15 10 1,2,3 * MON,FRI` above does; when neither names days, every day matches. This is the Unix
+`crontab(5)` rule - some other cron implementations intersect the two fields instead, so an expression
+copied from one of those fires more often here than it did there.
 :::
 
 ## Notes
--- a/docs/documentation/quartz-4.x/migration-guide.md
+++ b/docs/documentation/quartz-4.x/migration-guide.md
@@ -4572,7 +4572,15 @@ The cron expression parser now supports additional syntax:
 
 * `L` and `LW` combinations in day-of-month expressions (e.g., `LW` for last weekday of the month)
 * `LW-<OFFSET>` for offset from the last weekday (e.g., `LW-2` for two days before the last weekday). If the calculated day crosses a month boundary, it resets to the 1st.
-* Day-of-month and day-of-week can now be specified together in the same expression
+* Day-of-month and day-of-week can be specified together. A field written `*` or `?` **restricts
+  nothing**, so the other field decides: `0 15 10 1 * *` is the 1st of the month, and
+  `0 15 10 * * MON` is every Monday. When **both** fields name days, the expression fires on the union
+  of the two — `0 15 10 1,2,3 * MON,FRI` is the 1st, 2nd and 3rd **and** every Monday and Friday.
+  This is the Unix `crontab(5)` rule. It is deliberately **not** Cronos's, which ANDs even when both
+  fields are restricted; the union is what Quartz has always documented and what 4.0 alpha shipped.
+  `?` and `*` are now full synonyms in a day field, so `? * ?` is legal and means every day.
+  **Changed since alpha.4:** an expression with a wildcard in one day field and values in the other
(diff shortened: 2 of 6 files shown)


lahma:
Reviewed in full. The router rewrite is exactly the ratified rule — and routing the both-unrestricted case through the wildcard day-of-month walk so 'every day' needs no special case is the right shape. The teeth-proof (eight failing cases with the router reverted), the ?-scope-guard pin the note missed, and the independent re-run of the flip scanner are exactly what this change needed. Merging on green.

Issue #3601, opened 2026-08-31
XML scheduling is maintained, not extended


lahma:
XML scheduling is already a second-class citizen in 4.0 by content, not by decision: the XSD and parser know three trigger kinds — `simple`, `cron`, `calendar-interval` — while the JSON processor carries all five families plus `RetryPolicy` and the same directives, pre-processing commands, `ScanInterval` rescanning and per-tenant instancing. Meanwhile `quick-start.md` still teaches `UseXmlSchedulingConfiguration` as the first configuration example.

Alpha.5 makes the demotion deliberate and coherent:

- **The XSD freezes at 2.0.** No `daily-time-interval`, no recurrence, no retry policy — ever. The plugin's docs page says JSON is the maintained format and what the XML format will never gain.
- **`quick-start.md` teaches JSON first**; the packages page and `microsoft-di-integration.md`/`multi-tenancy.md` references follow.
- **The plugin twins stop diverging.** `XmlSchedulingDataProcessorPlugin` exposes `ProcessFile` and no `Shutdown`; `JsonSchedulingDataProcessorPlugin` the reverse; both drag seven public get-only properties that exist only for the string-key bridge, and `FileNames` (a delimited `string`) sits beside `FileSchedulingOptions.Files` (a `List<string>`). One surface shape for both, bridge-noise properties gone — this half is breaking and rides the freeze.

Explicitly not in scope: removing XML support. Existing `quartz_jobs.xml` files keep working for the life of 4.x.

Issue #3593, opened 2026-08-31
Plugins whose job the host already does are retired


lahma:
Three retirements and a reshape, all free only before the freeze; each gets a migration-guide row naming its replacement.

- **\`ShutdownHookPlugin\`** hooks \`AppDomain.CurrentDomain.ProcessExit\` with an \`async void\` handler (\`ShutdownHookPlugin.cs:66-77\`) — a shutdown that cannot reliably complete. \`QuartzHostedService\` (\`IHostedLifecycleService\`) already shuts every scheduler down, correctly awaited, and \`QuartzHostedServiceOptions.WaitForJobsToComplete\` is \`CleanShutdown\` under its real name. Retired; \`UseShutdownHook\` and \`ShutdownHookOptions\` go with it.
- **\`PropertyListenerFactory\`** and the \`quartz.jobListener.<name>.type\` / \`quartz.triggerListener.<name>.type\` keys (156 LOC + reflection): strictly weaker than \`AddJobListener<T>(matchers)\` — no matchers, no scheduler-listener key, unreachable under trimming. Retired; the bridge rejects the keys with a message naming the DI registration.
- **\`JobInterruptMonitorPlugin\`** — tracked separately in #3588 (timeout middleware is the replacement); listed here only so the plugin sweep is one story.
- **\`TimeZoneConverterPlugin\`** is one \`TimeZones.AddResolver\` call wearing a plugin's lifecycle (\`TimeZoneConverterPlugin.cs:44-45\`): no per-scheduler state, no scheduler dependency. The package stays; its shape becomes a resolver registration (\`UseTimeZoneConverter\` remains as the ergonomic entry, minus the plugin machinery).

Kept, deliberately: the four history-logging plugins (the classic pair exists for 3.x log-pipeline shapes — that ruling gets an AGENTS.md row), \`JobChainingJobListener\`, and both scheduling-data processors (XML demoted separately).

Issue #3588, opened 2026-08-31
A job timeout is middleware with typed configuration, not stringly plugin keys


lahma:
`JobInterruptMonitorPlugin` gives a job a maximum run time, but its opt-in surface is stringly and per-datum: a job carries `"AutoInterruptable" = true` and optionally `"MaxRunTime" = <milliseconds as a long>` in its `JobDataMap` (`JobInterruptMonitorPlugin.cs:22-23`), spellings the compiler never sees, the milliseconds convention documented only by the parse at `:91`, and the whole thing is a plugin that schedules a monitor per firing.

4.0 has the right seam now: a **timeout middleware**. `AddJobMiddleware` + `IJobExecutionMiddleware` can wrap the execute call with a `CancellationTokenSource.CancelAfter`-style timeout and call `Interrupt`/observe the context token directly — no monitor scheduling, no reserved stringly keys, typed configuration:

- `AddJobTimeout(TimeSpan)` (or an options type on the middleware registration) for the scheduler-wide default;
- a per-job override that is *typed* — an attribute (`[JobTimeout("00:05:00")]`) or a `JobBuilder`/`OneOffJobOptions` member — never a magic `JobDataMap` string;
- on expiry: the firing's `CancellationToken` fires and the standard interrupt path runs, so a job that honours its token (which `CA2016` already polices) stops; the middleware then reports the timeout distinctly (log event + the existing interrupt observability from #3248's work).

Ship it in core beside the other middleware, document it as the replacement in the plugin's page, and decide the plugin's fate in the same PR: keep `JobInterruptMonitorPlugin` as a thin compatibility shim over the middleware, or retire it with a migration-guide row (pre-4.0, so retirement is free). The assessment leaned retire-with-row, since the stringly keys are the thing being fixed and a shim would keep them alive.

PR #3612, opened 2026-08-31
Configuration written in code beats a string, everywhere

### LINKED ISSUE #3594: Configuration written in code beats a string, everywhere
The documented precedence (\`QuartzServiceCollectionExtensions.cs:90-115\`) says code beats strings. Three places disagree, found in the alpha.5 audit:

1. **Flat plugin keys override the \`Use*\` lambda.** \`SchedulerPluginFactory.Create\` builds container-registered plugins first — running the configuration lambda — then walks \`quartz.plugin.*\` names and calls \`ApplyProperties\` *on top of* the already-configured plugin (\`SchedulerPluginFactory.cs:66-91\`). For every shipped plugin, strings beat code. Apply properties before the lambda.
2. **\`Quartz:ThreadPool:*\` is read twice.** \`BindQuartzOptions\` binds the section into \`ThreadPoolOptions\` (\`QuartzTypedOptions.cs:107\`) *and* \`QuartzConfigurationHelper.PopulateProperties\` flattens the same JSON into the property bag (\`QuartzConfigurationHelper.cs:37\`) that \`QuartzPropertyBridge\` then also translates into \`ThreadPoolOptions\` (\`QuartzPropertyBridge.cs:464\`). Two independent readers of one key inside a last-wins pipeline is not a fallback, it is a coin flip — pick one reader and pin it with a test.
3. **Shipped plugins use the weaker of the two typed-options stories.** Their options objects are applied through closures, never registered via \`AddPlugin<T,TOptions>\`/\`ConfigureOptions<TOptions>\` (\`IQuartzBuilder.cs:330\`, \`:272\`), so \`FileSchedulingOptions\` and friends never bind from \`appsettings.json\` — only the reflection path reaches them. Move the built-ins onto the options-binding registration they teach everyone else to use, with a test that a plugin option set in \`appsettings.json\` lands.

Also in scope, same theme: \`AddQuartzSchedulers\` demands \`IConfiguration\` on the \`IServiceCollection\` receiver but not on \`IHostApplicationBuilder\` (\`QuartzServiceCollectionExtensions.cs:1473-area\` vs \`QuartzHostApplicationBuilderExtensions\`) — same name, incompatible contracts; align them.

lahma:
Closes #3594.

Three places disagreed with the precedence `AddQuartz` documents (`QuartzServiceCollectionExtensions.cs:90-115`), and a fourth item asked whether two same-named overloads have the same contract. Two commits: the behavioural fixes, then the contract documentation.

## 1. Flat plugin keys no longer override the `Use*` lambda

`SchedulerPluginFactory.Create` built the container-registered plugins — running whatever configuration each carried — and then applied `quartz.plugin.<name>.*` on top. For every shipped plugin a leftover string beat the code.

The keys now go on first and the code goes on last. That needs the code configuration to survive construction, so it is no longer baked into a factory closure: each `Use*` registers its plugin through `AddPlugin<T, TOptions>` and records how the options are copied onto the plugin (`SchedulerPluginConfiguration`, internal, `Quartz.Plugins` is already a friend assembly), and the factory replays those records after the keys.

Granularity is per setting, not per plugin: a key that configures something the callback said nothing about still lands (`AFlatKeyStillReachesASettingCodeSaidNothingAbout`). A plugin registered with `AddPlugin<T>(provider => …)` — a caller's own factory — keeps the old order and cannot do otherwise, since that configuration has already happened by the time there is an instance; the type's remarks say so.

## 2. `Quartz:ThreadPool:MaxConcurrency` now has one reader

**Choice: the typed binder keeps the key; the flattener stops synthesizing it.** Reasoning, since the brief asked for it:

- *What the flattener feeds* decided it. Dropping the whole `ThreadPool` section from the flattened form would take `ThreadPool:Type` (selects the implementation — only the bridge can do that), the legacy `ThreadCount` spelling (no property to bind onto), and a third-party pool's own knobs (applied from the bag by reflection) with it. That is the `JobStore:Type`-accepted-then-ignored failure the `LegacyProperties` docs warn about. So the exclusion is one key, not one section.
- *Deleting the typed bind instead* was the alternative I rejected: `ThreadPoolOptions` would become the only options type the configuration binder does not bind, and three guards in `ConfigurationBindingIsSourceGeneratedTest` (the interception count, the generated-binder list, the every-member case) would have to be loosened to allow it. A member added to `ThreadPoolOptions` later would then bind from nowhere.
- Only the *synthesized* spelling is skipped. A flat `quartz.threadPool.maxConcurrency` — the only spelling `AddQuartz(NameValueCollection)` has — is passed through and read by the bridge as before.

**Reviewer's call:** the overlap is not unique to this key. `Scheduler:InstanceName`, `MaxBatchSize` and `Context:*` are also written by both readers. The rest of the overlaps are corrections rather than duplicates — a legacy spelling the binder cannot express (`InterruptJobsOnShutdown`), a unit it reads differently (`IdleWaitTime` in milliseconds), a value that selects rather than sets (`InstanceId = AUTO`) — but `InstanceName` and `MaxBatchSize` are genuinely the same case as `MaxConcurrency`. I fixed the one the issue names and left the list a one-entry list with the rule written next to it, rather than widen the change under a fix; say the word and the others follow in the same shape.

One consequence worth knowing: `QuartzOptions.Properties` no longer echoes `quartz.threadPool.maxConcurrency` for a scheduler configured from a section, so a third-party `IThreadPool` that is not a `TaskSchedulingThreadPool` and has a `MaxConcurrency` property of its own no longer gets it set by reflection from `Quartz:ThreadPool:MaxConcurrency`. Pools deriving from `TaskSchedulingThreadPool` take the typed value as they always did.

## 3. Shipped plugins bind their options from configuration

`AddPlugin<T, TOptions>` makes the options the scheduler's own named options, so `services.Configure<FileSchedulingOptions>(configuration.GetSection(…))` reaches the plugin — the closure could never be reached that way, since it built an options object of its own. Every `Use*` signature is unchanged, and the plugins themselves are untouched, so #3593 / #3588 / #3601 rebase over this cleanly.

The callback each extension takes is applied to that registration's own options instance rather than being registered as another `IConfigureOptions<TOptions>`: the XML and JSON processors share `FileSchedulingOptions`, and the classic and structured history plugins share their message-template types, so a callback registered against the type would hand each of them the other's configuration. `TwoPluginsSharingAnOptionsTypeKeepTheirOwnConfiguration` and `EveryShippedPluginIsConfiguredFromItsOwnRegistration` pin that. It also means a section bound onto `FileSchedulingOptions` is a baseline for *both* processors — the same thing sharing an options type has always meant.

No new configuration section convention was invented; a plugin's options bind the way any other options type does.

## 4. `AddQuartzSchedulers` — documented, not changed

**This is a deviation from the brief, and the reasoning is the evidence I found.** The two overloads look incompatible but are the same rule seen from two receivers, and it is the rule `AddQuartz` already follows: on `IServiceCollection` configuration is always a parameter, on `IHostApplicationBuilder` it is the builder's own. Adding `services.AddQuartzSchedulers(configure)` would mean hunting for a registered `IConfiguration`, which `services.AddQuartz(configure)` deliberately does not do — that overload means "configured entirely in code" — so the overload would break the consistency it was meant to restore, and a fan-out with nothing to fan out over is not a registration at all. Requiring `IConfiguration` on the builder receiver instead would undo the reason those overloads exist.

So both sides now say the rule out loud (the host-builder overload carried no remarks at all), and `AServiceCollectionReadsNoConfigurationItWasNotHanded` pins the half that would make them genuinely incompatible. No public API changed, so no Verify baseline moved and there is no migration-guide row. If you want the overload anyway, it is three lines and I will add it — but it should be a decision, not a symmetry fix.

## Verification

```
dotnet build src/Quartz.Tests.Unit/Quartz.Tests.Unit.csproj   → Build succeeded. 0 Warning(s) 0 Error(s)
dotnet test  src/Quartz.Tests.Unit                            → Passed! Failed: 0, Passed: 4296
dotnet test  src/Quartz.Tests.AspNetCore                      → Passed! Failed: 0, Passed: 548
dotnet fallout VerifyDocsSnippets                             → Succeeded
Quartz.Benchmark --smoke                                      → 284 benchmarks executed, no failures
```

Integration tests were not run: no Docker daemon in this environment. Nothing here touches ADO store paths.

The ordering fix was confirmed to fail without it: moving the new configuration pass ahead of the property pass makes `APluginKeepsWhatCodeConfiguredWhenAFlatKeySaysOtherwise` fail with the property-bag value.

### DIFF
--- a/docs/documentation/quartz-4.x/packages/quartz-plugins.md
+++ b/docs/documentation/quartz-4.x/packages/quartz-plugins.md
@@ -44,6 +44,27 @@ They hang off `IQuartzBuilder`, so they work the same under `AddQuartz` and on a
 [configuration reference](../configuration/reference.md#listeners-calendars-and-plugins) for how a plugin
 is registered and named.
 
+An options type in that table is the scheduler's own named options, so a plugin is configurable from
+`appsettings.json` like anything else the container builds — bind the section and the values reach the
+plugin:
+
+<!-- snippet: sample_plugins_options_from_configuration -->
+```csharp
+// A plugin's options are the scheduler's own named options, so a configuration section binds
+// onto them like any other. The callback below is applied over whatever the section said.
+services.Configure<FileSchedulingOptions>(configuration.GetSection("Quartz:Xml"));
+
+services.AddQuartz(q => q.UseXmlSchedulingConfiguration(x => x.ScanInterval = TimeSpan.FromMinutes(1)));
+```
+<!-- endSnippet -->
+
+The three sources are applied in the order that makes code the last word: the flat
+`quartz.plugin.{name}.{property}` keys first, then the values bound onto the options, then the callback
+passed to the extension method. A setting the callback says nothing about keeps whatever the keys or the
+configuration section gave it, so configuring a plugin in code does not discard the rest of its
+configuration — it only overrides the parts it names. Under `AddQuartz("name", …)` the options are that
+scheduler's, so bind them with `services.Configure<TOptions>("name", section)`.
+
 ## Features
 
 ### LoggingJobHistoryPlugin
--- a/src/Quartz.Documentation.Samples/Packages/QuartzPluginsSamples.cs
+++ b/src/Quartz.Documentation.Samples/Packages/QuartzPluginsSamples.cs
@@ -1,3 +1,4 @@
+using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.DependencyInjection.Extensions;
 
@@ -65,6 +66,19 @@ public static class QuartzPluginsSamples
         #region sample_plugins_shutdown_hook
 
         services.AddQuartz(q => q.UseShutdownHook(options => options.CleanShutdown = true));
+
+        #endregion
+    }
+
+    public static void PluginOptionsFromConfiguration(IServiceCollection services, IConfiguration configuration)
+    {
+        #region sample_plugins_options_from_configuration
+
+        // A plugin's options are the scheduler's own named options, so a configuration section binds
+        // onto them like any other. The callback below is applied over whatever the section said.
+        services.Configure<FileSchedulingOptions>(configuration.GetSection("Quartz:Xml"));
+
+        services.AddQuartz(q => q.UseXmlSchedulingConfiguration(x => x.ScanInterval = TimeSpan.FromMinutes(1)));
 
         #endregion
     }
--- a/src/Quartz.Plugins/PluginConfigurationExtensions.cs
+++ b/src/Quartz.Plugins/PluginConfigurationExtensions.cs
@@ -1,13 +1,9 @@
-using System.Diagnostics.CodeAnalysis;
-
-using Microsoft.Extensions.DependencyInjection;
-
+using Quartz.Configuration;
 using Quartz.Plugins.History;
 using Quartz.Plugins.Interrupt;
 using Quartz.Plugins.Json;
 using Quartz.Plugins.Management;
 using Quartz.Plugins.Xml;
-using Quartz.Extensibility;
 
 namespace Quartz;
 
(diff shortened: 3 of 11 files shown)


lahma:
Reviewed in full. Rulings on the flagged decisions: (1) item-4 refutation ACCEPTED — the no-ambient-IConfiguration rule is AddQuartz's own, and documenting + pinning it beats adding a hunting overload; no overload wanted. (2) The one-key deny-list with reasoning beside it is right for this PR; the three sibling double-reads you identified (Scheduler:InstanceName, MaxBatchSize, Context:*) get their own follow-up issue rather than silence. (3) The narrow reflection-knob behaviour loss is acceptable — typed-first is the documented design. (4) Per-setting replay granularity on the plugin ordering fix is a nice touch. Merging on green.

PR #3614, opened 2026-08-31
The cron parser refuses the expressions it used to quietly reinterpret

### LINKED ISSUE #3591: A day-of-week step is a step, and expressions that lie are rejected
Two batches, one PR, both breaking-or-never.

**\`MON/2\` stops meaning fortnights.** A slash after a *textual* day-of-week sets \`everyNthWeek\` (\`CronExpression.cs:1198-1213\`) while the numeric equivalent \`2/2\` is an ordinary step (\`:1331-1374\`) — and the fortnight's phase is anchored to the instant you query from (\`:2220-2247\`), so the expression denotes no fixed set of instants: \`IsSatisfiedBy\` says every Monday, the chained walk returns every other one, and a \`MON/2\` \`CronCalendar\` excludes *all* Mondays. Ratified: \`everyNthWeek\` is removed; a textual day with a step parses exactly like its numeric twin. Fortnight schedules have a first-class home in \`RecurrenceScheduleBuilder\`; the migration-guide row points there, and \`best-practices.md:439\` stops advertising the old form. The builder's deliberate refusal to emit textual-DOW steps (\`CronExpressionBuilderTest.cs:448-476\`) becomes unnecessary caution to unwind or keep, per the design note.

**The parser stops accepting expressions it silently reinterprets.** \`1-5W\` parses with the \`W\` dropped (\`HandleDashOption\` never checks trailing characters, \`:1376-1420\`); \`? * L-3\` and \`? * LW\` in day-of-week mean just "Saturday" (\`:1094-1096\`); \`MON,FRI#3\` means \`MON#3\` (the evaluator reads \`daysOfWeek.Min\`, \`:2175\`); \`'C'\` parses and does nothing (\`:1323-1324\`, \`:1422-1439\`); \`*/0\` is quietly \`*\` (\`:1656-1661\`). Each becomes a \`FormatException\` in the teaching style the parser already uses for five fields and day-of-week zero. These "break" only expressions that were already not doing what they say.

lahma:
Closes #3591.

Seven cron shapes parsed, and then meant something other than what they said. Each is now a `FormatException` whose message names the expression the author meant.

| Expression | before | after |
|---|---|---|
| `1-5W` | the `W` was dropped; meant `1-5` | `FormatException` — write `1W,2W,3W,4W,5W`, or drop the `W` |
| `? * L-3`, `? * LW` | the suffix was dropped; meant Saturday | `FormatException` — those forms belong to day-of-month. A bare `L` in day-of-week is still Saturday |
| `MON,FRI#3` | fired on the third **Monday**; the Friday was ignored | `FormatException` — `#` cannot appear beside other days |
| `5C`, `1C` | parsed, meant `5` / `1`; `C` was never implemented | `FormatException` — use `ModifiedByCalendar` |
| `*/0`, `5/0`, `0-10/0` | a step of 0 degenerated to no step | `FormatException` |
| `0-10/120` | the step inside a range was never range-checked | `FormatException` — `Increment > 59 : 120`, as `0/120` already did |
| `MON/2` | every second week — a fortnight with **no stable phase** | `FormatException` — `MON,WED,FRI` for a step through the week, or `RecurrenceScheduleBuilder.Create("FREQ=WEEKLY;INTERVAL=2;BYDAY=MO")` for every second Monday |

`0-10/120` was found while writing the `1-5W` fix and is the same defect wearing a different hat: `HandleDashOption` stopped reading after the range and called `AddToSet` directly, so it neither checked what followed nor range-checked the step it had just parsed.

## The `MON/2` ruling a reviewer should confirm

**The issue body's plan for `MON/2` was superseded.** The body says a textual day with a step should parse exactly like its numeric twin; the ratification comment (https://github.com/quartznet/quartznet/issues/3591#issuecomment-5482697340) changed that to **reject**, and this PR implements the comment.

Reinterpreting `MON/2` as a plain step turns `{MON}` every fortnight into `{MON,WED,FRI}` every week — 26 fires a year become 156 — with no log line, no state change and no misfire. Rejecting is loud, joins the same blast-radius class as the other six, and stays open: a string that throws becoming valid breaks nobody, so 4.1 can adopt the step semantics additively and reach the originally ratified end state a release later.

The extension was also already broken. `everyNthWeek` had no stable phase — it counted whole weeks from wherever the search started, so misfire handling, `ComputeFirstFireTime` after a start-time change, a failover recovery and a dashboard `GetFireTimeAfter` each re-phased the fortnight. `RecurrenceScheduleBuilder.Create("FREQ=WEEKLY;INTERVAL=2;BYDAY=MO")` anchors `INTERVAL` on `DTSTART`, so the fortnight is a property of the trigger rather than of whoever asked last.

`everyNthWeek`, its parse branch and its evaluation branch are gone. Two tests that pinned the old behaviour (`TestCronExpressionWeekdaysFridayEveryTwoWeeks`, `TestCronExpressionWeekdaysThirsdayAndFridayEveryTwoWeeks`) go with it, as does `TestTheBuilderNeverRendersTheTextualEveryNthWeekExtension` — the trap it guarded no longer exists — and `CronExpressionBuilder.WithDayOfWeekIncrements`'s "and here is why" caveat.

## The five-field message

Its advice used to move the schedule by a day. Prepending a seconds field to `30 4 * * 1` keeps the `1`, and `1` is Monday in crontab but Sunday in Quartz, so following the message literally gave Sundays. When the day-of-week is a bare number the message now also names the renumbered form:

```
Cron expression '30 4 * * 1' has 5 fields, which is the Unix/crontab form; Quartz cron has 6 or 7
fields with seconds first. Prepend a seconds field: "0 30 4 * * 1". Note that Quartz numbers days of
week 1-7 starting at Sunday where Unix cron uses 0-6, so if '1' meant Monday the Quartz spelling is
"0 30 4 ? * MON".
```

The H-resolver's duplicate of that message gets the same advice from the same helper. That path matters more than it looks: it counts fields *before* anything is parsed, whereas the parser reads a five-field string's last field as a **month**, so `0 12 * * MON` fails on `Invalid Month value: 'MON'` and never reaches the count check at all.

## Docs

* `migration-guide.md` gains **The parser refuses what it used to ignore** (the table above, plus the `MON/2` reasoning) and a **Before you upgrade** section with the audit SQL for `QRTZ_CRON_TRIGGERS.CRON_EXPRESSION`. Every row is a value the ADO.NET store will now fail to **read**, not merely fail to schedule — the exception comes out of `CronTriggerImpl.CronExpressionString`'s setter while a row is being materialised. `CronCalendar` expressions live in the `QRTZ_CALENDARS` blob and are not queryable; they surface at deserialization.
* `best-practices.md` speaks for both branches, so its `MON/2` sentence is split the way its interval bullet already splits: the fortnight moves to the `RecurrenceTrigger` recommendation, and a **Quartz 3.x only** paragraph says where `MON/2` still works and why 4.x refuses it.
* `docs/documentation/quartz-3.x/` is untouched.

## Deviation from the brief

The brief said to leave `calendarDayOfWeek` / `calendarDayOfMonth` for the follow-up PR. Once `'C'` throws, nothing assigns them, and `CS0649` under `TreatWarningsAsErrors` fails the build. They are deleted here and `GetExpressionSummary` passes `false` at the call site; `GetExpressionSummary` and `CronExpressionSummary.cs` are untouched, so the follow-up's work is unchanged (and slightly smaller). The commented-out `'C'` documentation block on the type went with them, since it describes syntax the parser now rejects.

## Verification

* `dotnet build Quartz.slnx` — 0 warnings, 0 errors
* `dotnet test src/Quartz.Tests.Unit` — 4356 passed, 0 failed
* `dotnet fallout VerifyDocsSnippets` — succeeded, 102 markdown files checked
* `dotnet run --project src/Quartz.Benchmark -c Release -- --smoke` — 284 benchmarks executed, clean
* The public-API Verify baselines are unchanged; nothing removed here was public.

New tests: a table walking every newly rejected form and asserting the message names the fix; a per-field table for the range-step range check; a table of the legitimate forms these rejections stand next to (`? * L`, `6L`, `FRIL`, `LW-2`, `L-5W`, `15W`, `6#3`, `FRI#3`, `1-5`, `18-21/1`, `2/2`, `? DEC *`) asserting they still parse; and the renumbered five-field suggestion on both the parser and hash-resolver paths.

### DIFF
--- a/docs/documentation/best-practices.md
+++ b/docs/documentation/best-practices.md
@@ -435,16 +435,22 @@ families answer "what time is it" in genuinely different ways.
 | Repeatedly inside a daily window | `DailyTimeIntervalTrigger` | The window is wall-clock, so a transition lengthens or shortens the day's run. |
 
 Quartz cron says more than most cron dialects, so check it before assuming a pattern needs something
-else: `0 0 0 ? * MON#2` is the second Monday of the month, `0 0 0 LW 3 ?` the last weekday of March,
-and `0 0 0 ? * MON/2` every other Monday, holding its fortnightly cadence across month and year
-boundaries. What it genuinely cannot state is a position counted from the *end* of a month other
-than the last one — `#` counts forwards, and only as far as 5 — and any cadence its fields cannot
-divide: `0 0 0 1/3 * ?` reads like "every third day" but restarts at the 1st of each month, so 31
-January is followed by 1 February. `RecurrenceTrigger` and its RFC 5545 rule state both,
-`FREQ=MONTHLY;BYDAY=-2FR` for the second-to-last Friday and `FREQ=DAILY;INTERVAL=3` for a three-day
-cadence that does not reset, on [4.x](/documentation/quartz-4.x/tutorial/recurrencetrigger) and
+else: `0 0 0 ? * MON#2` is the second Monday of the month and `0 0 0 LW 3 ?` the last weekday of
+March. What it genuinely cannot state is a position counted from the *end* of a month other
+than the last one — `#` counts forwards, and only as far as 5 — a fortnight, and any cadence its
+fields cannot divide: `0 0 0 1/3 * ?` reads like "every third day" but restarts at the 1st of each
+month, so 31 January is followed by 1 February. `RecurrenceTrigger` and its RFC 5545 rule state all
+three, `FREQ=MONTHLY;BYDAY=-2FR` for the second-to-last Friday, `FREQ=WEEKLY;INTERVAL=2;BYDAY=MO`
+for every other Monday and `FREQ=DAILY;INTERVAL=3` for a three-day cadence that does not reset, on
+[4.x](/documentation/quartz-4.x/tutorial/recurrencetrigger) and
 [3.x](/documentation/quartz-3.x/tutorial/recurrencetrigger). Reaching for several cron triggers, or
 for a cron expression with a workaround in it, is usually the sign.
+
+**Quartz 3.x only:** `0 0 0 ? * MON/2` — a textual day-of-week with a step — means every other
+Monday there. 4.x rejects it, because the fortnight's phase was anchored to whatever last asked the
+expression a question: a misfire, a restart or a failover recomputed it from a different day and
+moved it. `FREQ=WEEKLY;INTERVAL=2;BYDAY=MO` anchors on the trigger's start time instead, so the
+fortnight is a property of the trigger rather than of the caller.
 
 ### The two daylight saving rules
 
--- a/docs/documentation/quartz-4.x/migration-guide.md
+++ b/docs/documentation/quartz-4.x/migration-guide.md
@@ -4579,6 +4579,60 @@ Parse errors name the fix instead of only the constraint. A 5-field Unix/crontab
 online cron generator emits — is still rejected (Quartz cron puts seconds first), but the error now shows the
 corrected 6-field expression to use, and the day-of-week range error explains that Quartz numbers days 1-7
 starting at Sunday where Unix cron uses 0-6, recommending names (`SUN`, `MON`, …) as the unambiguous spelling.
+When the 5-field expression's day-of-week is a bare number the error goes one step further and shows the
+renumbered expression too: prepending a seconds field to `30 4 * * 1` keeps the `1`, and `1` is Monday in
+crontab but Sunday in Quartz, so the message names `"0 30 4 ? * MON"` as the schedule that was meant.
+
+### The parser refuses what it used to ignore
+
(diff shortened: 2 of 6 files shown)


lahma:
Reviewed in full. Deviations all ACCEPTED: (1) deleting calendarDayOf* here was forced by CS0649 under warnings-as-errors and is the correct call — PR-4's scope shrinks accordingly (GetExpressionSummary call site passes false,false, noted); (2) the two FRI/2 feature tests going with the feature is right; (3) templated messages beat hard-coded examples; (4) the inert ESCAPE clause removal with the over-wide-clause documentation is more honest than the draft. The five-field-error-unreachability finding (month parses before field count) is relayed to the dialect PR. Merging on green.

PR #3615, opened 2026-08-31
A cron time the clock skips fires when the gap ends, and the expression agrees it fired

### LINKED ISSUE #3590: A time the clock skips is run at the gap's end, and the expression agrees it ran
A fixed-time cron expression whose local time falls inside the spring-forward gap currently fires delta-shifted — 02:30 becomes 03:30 — because \`TimeZones.ResolveLocal\` resolves a non-existent local time with the pre-transition offset (\`TimeZones.cs:214-229\`). The DST suite pins the consequence and names it a decision left open: \`GetTimeAfter\` returns 03:30 while \`IsSatisfiedBy(03:30)\` is false — *"the trigger fires at an instant its own expression does not match"* (\`CronExpressionDstTests.cs:412-446\`, whose own remark says whatever rule replaces the delta shift should make the two agree).

Ratified: fire at the **gap's end** — the first valid instant after the transition — which is the Cronos rule and completes the alignment the 4.0 fall-back work already shipped (interval expressions fire both passes, fixed times once; those stay untouched). The invariant that must hold either way: \`GetTimeAfter\` and \`IsSatisfiedBy\` agree about every instant the walk returns.

Scope: the gap branch in \`GetTimeAfter\`/\`ResolveLocal\`, the pinned delta-shift cases in \`CronExpressionDstTests.cs\` (including the Lord Howe half-hour transition), and a migration-guide row — a 4.0-alpha trigger scheduled inside a gap moves from "shifted by the gap's width" to "the moment the clock becomes valid". The design note settles whether \`IsSatisfiedBy\` learns the gap rule or the walk simply returns instants it already matches.

lahma:
A fixed-time cron expression whose local time falls inside a spring-forward gap fired delta-shifted — 02:30 became 03:30 — because `TimeZones.ResolveLocal` pairs a non-existent wall clock with the offset in force just before the gap. That instant is one the expression itself rejects: `IsSatisfiedBy(03:30)` was false for hour 02, so a trigger fired at a time its own schedule did not match. The DST suite pinned that as a decision left open.

This lands the ratified gap-end rule (§2 of the alpha.5 cron design note): the in-gap wall clock is answered with **the instant the clocks moved**.

## What changed

**`CronExpression.GetTimeAfter` resolves through `TimeZones.FirstInstantAtOrAfterLocal`** instead of `ResolveLocal`. That helper already existed and was already tested — it was written for `DailyCalendar`/`CalendarIntervalTriggerImpl` — and it delegates to `ResolveLocal` for a wall clock that exists, so every non-gap path is byte-identical. `ResolveLocal` itself is untouched: `DailyTimeIntervalTriggerImpl`, `CalendarIntervalTriggerImpl`, `RecurrenceRule` and `DailyTimeIntervalScheduleBuilder` keep its delta-shift semantics and its pin.

**A second edit is required, and it is the interesting one.** Gap-end alone does *not* make `GetTimeAfter` and `IsSatisfiedBy` agree. `IsSatisfiedBy(t)` is `GetTimeAfter(t − 1s) == t`, and `GetTimeAfter` converts `after + 1s` as an *instant* before walking in wall clock — at the transition instant that conversion already reads past the gap, so the search never examines the wall clocks the gap swallowed. The walk's starting wall clock is therefore rewound to the gap's first swallowed reading (new `internal TimeZones.WalkToGapStart`, the mirror of `WalkToGapEnd`) exactly when the wall clock one second before it does not exist:

```csharp
if (TimeZone.SupportsDaylightSavingTime
    && d.Ticks >= TimeSpan.TicksPerSecond
    && TimeZone.IsInvalidTime(d.DateTime.AddSeconds(-1)))
{
    d = new DateTimeOffset(TimeZones.WalkToGapStart(d.DateTime, TimeZone), d.Offset);
}
```

`IsInvalidTime(start − 1s)` is false at every fall-back transition, so the fall-back demotion and `ApplySecondAmbiguousPassIfNeeded` are untouched **by construction** rather than by argument. Cost on the hottest path in the library is one `IsInvalidTime` behind a `SupportsDaylightSavingTime` bool; UTC and fixed-offset zones pay the bool.

With the rewind, gap-end is the *unique* in-gap resolution a one-second-earlier probe can reproduce — a spring-forward gap takes no real time, so its start and its end are the same instant. The delta shift could never have been a fixed point of that definition, whatever the search was taught. `IsSatisfiedBy` needs no knowledge of daylight saving time.

## Two findings that correct the design note

The note's §2.5 says an expression with several matches inside the gap "now fires once rather than once per match", and the drafted migration-guide paragraph repeated it. **That is backwards, and I verified it empirically at the base commit before editing anything.** Under the delta shift the walk resumed at the shifted wall clock + 1s, which is always at or after the gap's end, so any further in-gap match was already swallowed: `0 0,30 2 * * ?` Eastern on 2024-03-10 fired *once* at 07:00Z before this change and fires once now.

What actually changes the fire count is the opposite. The gap's end is *earlier* than the delta-shifted instant, so a match in the wall-clock window between them is no longer stepped over:

- Lord Howe `0 15,40 2 * * ?` fired only at 02:45; it now fires at 02:30 and 02:40.
- **An hourly `0 30 * * * ?` in Eastern now fires at 03:00 and again at 03:30**, where the shift moved the 02:30 occurrence onto 03:30 and the two collided into one fire. The note's §2.4 called interval expressions "verified unchanged", but it only checked `0 * * * * ?` and `0 0 * * * ?`, whose first in-gap match is the gap's *first* wall clock — the one reading where the two rules agree. This is the shape most likely to be noticed in production, so it is pinned by a test and stated in both docs pages.

§2.6's flipped-case table was also incomplete: three tests outside `CronExpressionDstTests` pinned the delta shift and are updated here — `CronExpressionTest.TestDaylightSaving_QRTZNETZ186`, `DaylightSavingTimeTest.ShouldHandleDstSpringForwardTransition`, and `CalendarDstTests.CronCalendar_ExclusionInsideTheSpringForwardGap_ExcludesNothing`. The last is the `CronCalendar` consequence §2.5 predicted: a calendar written over the skipped hour now excludes the instant the clocks moved, where it excluded nothing at all on the transition day — the coherent answer, since that is also when the matching trigger fires.

## Tests

- The three delta-shift `[TestCase]` rows flip (`03:30` → `03:00`, Lord Howe `02:45` → `02:30`; Santiago is unchanged, because its 00:00 is the gap's first reading and the two rules agree there), the `#region Current-behavior pins` goes, and `IsSatisfiedBy_..._IsCurrentlyFalse` becomes `..._IsTrue`.
- New: several matches in one gap fire once; a match between the gap's end and the old delta shift is no longer swallowed; the hourly-at-half-past extra fire; a 181-second sweep across each transition asserting strict progress *and* monotonicity in the input (the property `GetTimeBefore`'s binary search depends on); fixed-time rows added to the `GetTimeBefore` round-trip grid, whose "asking one second earlier reproduces it" assertion is exactly what the delta shift could not satisfy.
- `WalkToGapStart` is unit-tested directly beside `WalkToGapEnd`'s tests, including the round trip back through `WalkToGapEnd`.
- One guard worth a reviewer's eye: `d.Ticks >= TimeSpan.TicksPerSecond`. Without it, `GetNextValidTimeAfter(DateTimeOffset.MinValue)` on a DST zone throws, because a zone west of UTC clamps that conversion to the first representable wall clock and `AddSeconds(-1)` underflows. That is a new throw the rewind would otherwise introduce; `GetTimeAfter_ProbeAtTheMinimumInstant_DoesNotThrow` fails without the guard (verified by removing it).

## For a reviewer to decide

- Whether the hourly `0 30 * * * ?` extra fire is acceptable. It follows from the ratified rule — the occurrence the gap swallowed runs when the clock passes it, and the next hour's occurrence still runs at its own wall clock — but it is a fire-count increase on the transition day for a very common expression, and it is the one thing here nobody signed off on, because the note did not find it.
- `docs/documentation/faq.md:300-308` still says such a trigger "will be skipped", which is wrong for 3.x and 4.x alike and has been for years. Deliberately left alone; it is tracked separately. `docs/documentation/quartz-4.x/cron-expressions.md:257` is vague in the same direction ("a skip or a repeat") and is also untouched.

No public API changed (`WalkToGapStart` is `internal`), so the Verify baselines do not move.

## Verification

```
dotnet build Quartz.slnx                                 Build succeeded. 0 Warning(s) 0 Error(s)
dotnet test src/Quartz.Tests.Unit                        Passed! Failed: 0, Passed: 4300, Skipped: 0
dotnet test src/Quartz.Tests.AspNetCore                  Passed! Failed: 0, Passed: 548, Skipped: 0
dotnet test src/Quartz.Tests.Integration
  --filter ~AcrossDstTransition                          Passed! Failed: 0, Passed: 20, Skipped: 0
dotnet fallout VerifyDocsSnippets                        Succeeded (102 markdown files checked)
dotnet fallout BenchmarkSmoke                            Succeeded (284 benchmarks executed)
```

Closes #3590

### DIFF
--- a/docs/documentation/best-practices.md
+++ b/docs/documentation/best-practices.md
@@ -450,19 +450,26 @@ for a cron expression with a workaround in it, is usually the sign.
 
 Java Quartz's page warns that a cron trigger may fire twice or not at all across a transition. That
 is not what this implementation does, and the difference matters enough to state precisely. For a
-**fixed-time** expression such as `0 30 2 * * ?`, on both versions:
+**fixed-time** expression such as `0 30 2 * * ?`:
 
-- A wall-clock time that **does not exist** on a spring-forward day fires exactly **once**, shifted
-  forward by the transition delta. A daily 02:30 over a 02:00–03:00 gap fires at 03:30. It is not
-  skipped.
+- A wall-clock time that **does not exist** on a spring-forward day fires exactly **once**, and is
+  never skipped. *Where* it fires differs by version. On 4.x it fires at the **end of the gap**, the
+  instant the clocks moved: a daily 02:30 over a 02:00–03:00 gap fires at 03:00. On 3.x the fire is
+  shifted forward by the transition delta instead, to 03:30. In a zone whose delta is not a whole
+  hour — Australia/Lord_Howe — the two rules read 02:30 and 02:45. Only 4.x's answer is an instant
+  the expression itself matches: ask `IsSatisfiedBy` about the fire time and 4.x says yes where 3.x
+  says no.
 - A wall-clock time that **occurs twice** on a fall-back day fires **once**, at the first of the two
-  occurrences.
+  occurrences. This is the same on both versions.
 
 **Quartz 4.x only:** an *interval* expression — one with a wildcard, step or range in the second,
 minute or hour field, such as `0 * * * * ?` or `0 0/30 * * * ?` — fires through **both** passes of
 the repeated hour. On 3.x the repeated hour is fired once, which means an "every minute" schedule
-silently loses an hour of real time each autumn. Fixed-time expressions, including comma lists like
-`0 0,30 2 * * ?`, are unchanged between the versions.
+silently loses an hour of real time each autumn. Over the fall-back hour, fixed-time expressions —
+including comma lists like `0 0,30 2 * * ?` — are unchanged between the versions. On the
+spring-forward day the gap-end rule shows in an interval expression as an extra fire rather than a
+moved one: 4.x runs `0 30 * * * ?` at 03:00 for the occurrence the gap swallowed and again at 03:30
+for the next hour's, where 3.x resumes from the shifted 03:30 and runs once.
 
 Whatever the family, **name the time zone**. A cron trigger with no zone uses `TimeZoneInfo.Local`,
 which is the developer's machine in development and very often UTC in a container, so the schedule
--- a/docs/documentation/quartz-4.x/migration-guide.md
+++ b/docs/documentation/quartz-4.x/migration-guide.md
@@ -4582,15 +4582,30 @@ starting at Sunday where Unix cron uses 0-6, recommending names (`SUN`, `MON`, �
 
 ## Daylight saving time
 
-Two schedules fire at different times than they did on 3.x. Neither needs a code change, but both change
+Three schedules fire at different times than they did on 3.x. None needs a code change, but all three change
 *when* existing triggers run, so review any schedule that crosses a transition.
 
 **Interval cron expressions fire through both halves of a fall-back hour.** A cron expression whose second,
 minute or hour field uses a wildcard, a step or a range — `0 * * * * ?`, `0 0/30 * * * ?` — now fires through
 **both** occurrences of the wall-clock window that repeats when the clock goes back. Previously the repeated
-window fired only once, so an "every minute" schedule silently skipped an hour of real time. Fixed-time
-expressions such as `0 30 2 * * ?`, including comma lists like `0 0,30 2 * * ?`, are unchanged: they still fire
-once per day, at the first occurrence of an ambiguous wall-clock time.
+window fired only once, so an "every minute" schedule silently skipped an hour of real time. Over a fall-back
+hour, fixed-time expressions such as `0 30 2 * * ?` — including comma lists like `0 0,30 2 * * ?` — are
+unchanged: they still fire once per day, at the first occurrence of an ambiguous wall-clock time.
+
+**A cron time that does not exist fires when the gap ends, not shifted by the transition delta.** A daily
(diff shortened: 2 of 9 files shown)


lahma:
Reviewed in full. RULING on the flagged decision: the `0 30 * * * ?` double-morning is ACCEPTED — the old delta shift silently collided the swallowed 02:30 occurrence with the real 03:30 one (a lost fire wearing a coincidence); gap-end applied per occurrence runs each scheduled wall-clock time exactly once, which is Cronos's own behavior for this shape. Keep it stated loudly in the migration paragraph as you have it. The other three deviations are all accepted: the §2.5 probe-first correction (the note's fire-count claim was indeed backwards — good catch proving it against the base commit before editing), the three extra pinned files the flip table missed, and the MinValue underflow guard with its regression test. Merging on green.

Issue #3596, opened 2026-08-31
The ADO namespace is a delegate-authoring kit, not the store's internals


lahma:
`Quartz.Impl.AdoJobStore` is 55 public types and 779 members — 27% of `Quartz.dll`'s surface — exposed at the same level as `IScheduler`. The extension point alive in the ecosystem is the **dialect delegate**: `quartz.jobStore.driverDelegateType`, third-party delegates subclassing `StdAdoDelegate` to override SQL, the `*Delegate` vocabulary AGENTS.md ratifies as identity. Subclassing the store base itself is the rare Java-ism.

Ratified — the inverted hybrid:

- **Public, the delegate-authoring kit:** `IDriverDelegate`, `StdAdoDelegate` + the six dialect delegates, `ITriggerPersistenceDelegate` + bundles, the `ILockHandler` family (whose cancellation contract #3583 pins in the same release), `IDbProvider`/Common, `AdoConstants`, `ConnectionAndTransactionHolder`, `IDbAccessor`.
- **Internal:** `AdoJobStoreBase` (both partials) and the acquisition/machinery records (`TriggerAcquisition*`, `TriggerFiredUpdate`, the state-transition and scheduler-state records, `SqlStatement`/`SqlRowLimit` shapes). `LocalTransactionJobStore`/`ExternalTransactionJobStore` sealed. Custom stores implement `IJobStore`, which stays public.
- Expected: ~830 → ~450 members. Re-exposing a member in beta is additive and free; re-hiding is impossible — ties break toward internal.

Riders in the same stream: `IJobStore.AddJob`/`AddTrigger`/`ScheduleJobs` trade their bare `bool replace` for the options types the rest of the interface already uses (`AddCalendarOptions` precedent); `OperationName.JobStore` either covers the interface or documents its subset rule — span names are telemetry contracts; the 48-member provider-extension family gets an overload audit; `StoredTriggerStates`' static-plus-extension mix becomes one shape.

Guard tests: an out-of-assembly delegate still resolves via `driverDelegateType` end-to-end, and a sample custom dialect delegate compiles against only the public kit. A type-by-type visibility map (design note) precedes implementation; the stream waits for the #3510 acquisition-shape decision so the baseline moves once.

lahma:
Design-note corrections (visibility map, alpha5-note-ado.md — ratified):

[...]

PR #3616, opened 2026-08-31
The ADO telemetry contract and two shapes are written down

lahma:
Part 1 of #3596 — the zero-member-delta half of the ADO surface reduction, so the two baseline diffs that follow review on their own.

The design note behind #3596 refuted two of the issue's clauses on contact with the signatures, and this PR is what is left of the riders after that.

## `OperationName.JobStore` had not drifted (I18)

The issue says the span constants "either cover the interface or document the subset rule". They already cover exactly what they should: 29 constants against 35 `OperationName.JobStore.*` call sites in `TracingJobStore`, of which 6 are the second overload of a member pair — **an exact 29 ↔ 29 bijection, nothing unused and nothing missing**. What was missing was the rule, which lived in two `// Tier 1` / `// Tier 2` comments and otherwise nowhere.

The rule is now on the class: a constant exists for every store operation that changes state or drives the fire cycle, and for nothing else — reads are one round trip inside the caller's own span, lifecycle members happen once outside any request, and `GetAcquireRetryDelay` is advice rather than an operation.

Three guards hold the code to it, in `OperationNameTest`:

- the bijection, both directions. The span names come out of `TracingJobStore`'s IL rather than a list kept beside it — a `const` reaches its use site inlined, so the literals in its method bodies are exactly the names it begins spans under whether it named the constant or typed the string, and typing the string is the failure worth catching.
- every **mutating** `IJobStore` member has a constant, against an explicit exclusion list that is the written rule in a form the build can check. This one earned its keep immediately: it found `GetAcquireRetryDelay`, which I had not enumerated.

The IL walk moves out of `LegacyPropertyKeyExhaustivenessTest` into `MethodBodyStrings`; that test had it for the unrelated job of finding the `quartz.*` keys the configuration readers consult.

**Public API delta: none.**

## `StoredTriggerStates.ToStoredValue` drops its `this` (I7)

One mapping, two directions, two call shapes: `state.ToStoredValue()` beside `StoredTriggerStates.FromStoredValue(value)`. The extension is the odd half and the asymmetry is not fixable from the other side — `string` is not Quartz's to extend with a Quartz verb, and doing so would put a Quartz noun on IntelliSense for every string in a consuming application, for one call. 34 call sites move; the type was already in scope at every one.

`TriggerStateResolver` stays in `Quartz.Extensibility`: it resolves to `TriggerState`, which `RAMJobStore` answers with too, so it is not ADO-specific.

## `AdoConstants`: two corrections

- **`ParameterPageSkip` / `ParameterPageTake` are new.** `StdAdoDelegate.ApplyPaging` splices `@pageSkip`/`@pageTake` and an overriding dialect's `AddPagingParameters` must bind the same names — an agreement between Quartz and an assembly that is not Quartz, spelled only on the **internal** `SqlParameters`. That is why `MySQLDelegate` and `SQLiteDelegate` name constants while the documentation sample hard-codes the strings. `SqlParameters` now names the public constants so the two spellings cannot drift, and the sample uses them.
- **`AliasColumnNextFireTime` is deleted.** `ALIAS_NXT_FR_TM` is a column alias from the Java acquisition sub-select; no 4.x statement uses it, no version of the schema has a column for it, and it had zero references in the tree.

## `PersistentStoreBuilderExtensions`: a remark, and nothing else (D11)

The issue asks for an overload audit of "the 48-member provider-extension family". It is **30 members, not 48** — 18 of those baseline lines are `[RequiresUnreferencedCode]` attributes — and every overload earns its place: the connection-string form is the two-thirds case, the `Action<DataSourceOptions>` form expresses a *named* connection string out of `IConfiguration` which the string form cannot, and the `(DbProviderFactory, string)` form is the only trim-safe one and the one `trimming-and-native-aot.md` teaches. Cutting to two per provider means cutting one of those three.

The one genuine finding is `UseOracle`'s fourth overload, the only per-provider member outside the ×3 family. Its own remarks explain why; the class remarks now do too, so it stops reading as an oversight to someone counting the family. **No members were trimmed.**

## Baseline

`PublicApiTest_Quartz.verified.txt` moves by exactly three lines, reviewed and accepted:

```
-        public const string AliasColumnNextFireTime = "ALIAS_NXT_FR_TM";
+        public const string ParameterPageSkip = "pageSkip";
+        public const string ParameterPageTake = "pageTake";
-        public static string ToStoredValue(this Quartz.Extensibility.StoredTriggerState state) { }
+        public static string ToStoredValue(Quartz.Extensibility.StoredTriggerState state) { }
```

Migration-guide rows carried in the same PR: an appendix row for `AdoConstants.AliasColumnNextFireTime`, the two new parameter constants named in the paging-seam paragraph, and the `StoredTriggerStates` section's sample rewritten to the static form (with a note that both halves are plain statics now).

## For a reviewer to decide

- **The exclusion list in `OperationNameTest` is the rule.** If you disagree that `AcquireNextTriggers` is a mutating operation (it marks rows acquired, so I kept it traced as it already was) or that `GetAcquireRetryDelay` is not, the list is the one place to say so.
- **`MethodBodyStrings` is a shared test helper now.** The alternative was duplicating ninety lines of IL walking.

## Verification

```
dotnet test src/Quartz.Tests.Unit        → Passed! Failed: 0, Passed: 4816, Skipped: 36
dotnet test src/Quartz.Tests.AspNetCore  → Passed! Failed: 0, Passed: 548, Skipped: 0
dotnet fallout VerifyDocsSnippets        → Succeeded (102 markdown files checked)
dotnet fallout BenchmarkSmoke            → Succeeded (284 benchmarks executed)
```

Integration tests were not run: they need a Docker daemon and nothing here touches a path only they reach. The 36 skips are main’s Unix-cron differential cases, which need `crontab`.

Rebased onto `e21149daa`. One conflict, in the migration guide's alphabetical removals table, where #3619 added a `CronExpression` row at the same place — resolved by ordering `AdoConstants` before it. `PublicApiTest` passes there, which is the proof rather than the textual merge.

## Coverage

The first run's Sonar gate was red at 66.7 % of new lines, and both halves of the gap were real.

**Nine lines in `StdAdoDelegate.Triggers.cs`** were the `ToStoredValue` rewrites inside seven `IDriverDelegate` members no unit test called: `UpdateTriggerState`, `UpdateTriggerStateFromOtherState`, `UpdateTriggerStatesForJob`, `UpdateTriggerStatesForJobFromOtherState`, `CountMisfiredTriggersInState`, `IsJobCurrentlyExecuting` and `UpdateTriggerForRetry`. `TriggerStateStatementsSqliteTest` drives all seven against a real file-backed SQLite database, which is the only way to check the thing they are worth checking: that the string bound is the string stored, so the `WHERE` clause matches the rows it means to. A stub command asserts a parameter was bound and learns nothing about whether the update hit anything. Verified locally with `dotnet fallout Compile UnitTest --coverage` — all nine lines now report visits.

**Two lines in `SqlParameters.cs`** were `public const string PageSkip = AdoConstants.ParameterPageSkip;` and its twin. A `const` whose initializer is *another const* gets a sequence point from the instrumentation and no test can ever execute it — note that the two new literal-initialized constants on `AdoConstants` were counted as 0 lines to cover, which is the difference. So that indirection is walked back: both spellings stay, and `SqlParametersTest` holds them equal. The property that mattered was never "one declaration", it was "the statement and the binder cannot drift", and a test states that exactly. `SqlParameters.cs` now emits no sequence points at all.

Part 1 of #3596.

### DIFF
--- a/docs/documentation/quartz-4.x/how-tos/dialect-delegate.md
+++ b/docs/documentation/quartz-4.x/how-tos/dialect-delegate.md
@@ -84,23 +84,27 @@ The default is the ANSI form, understood by SQL Server 2012+, Oracle 12c+, Postg
 
 MySQL and SQLite have no such clause, so they override both members — and they must override
 `AddPagingParameters` too, because their clause names the parameters in the other order and providers
-that bind positionally take them in the order the statement mentions them:
+that bind positionally take them in the order the statement mentions them.
+
+The two parameter names are the one thing the two overrides have to agree about, so they are constants
+rather than literals: `AdoConstants.ParameterPageSkip` and `AdoConstants.ParameterPageTake`, spliced
+into the statement with an `@` and bound by the bare name.
 
 <!-- snippet: sample_dialect_delegate_paging -->
 ```csharp
 protected override string ApplyPaging(string sql, bool takeLimited)
     => takeLimited
-        ? sql + " LIMIT @pageTake OFFSET @pageSkip"
-        : sql + " LIMIT -1 OFFSET @pageSkip";
+        ? sql + " LIMIT @" + AdoConstants.ParameterPageTake + " OFFSET @" + AdoConstants.ParameterPageSkip
+        : sql + " LIMIT -1 OFFSET @" + AdoConstants.ParameterPageSkip;
 
 protected override void AddPagingParameters(DbCommand cmd, int skip, int take, bool takeLimited)
 {
     if (takeLimited)
     {
-        AddCommandParameter(cmd, "pageTake", take);
+        AddCommandParameter(cmd, AdoConstants.ParameterPageTake, take);
     }
 
-    AddCommandParameter(cmd, "pageSkip", skip);
+    AddCommandParameter(cmd, AdoConstants.ParameterPageSkip, skip);
 }
 ```
 <!-- endSnippet -->
--- a/docs/documentation/quartz-4.x/migration-guide.md
+++ b/docs/documentation/quartz-4.x/migration-guide.md
@@ -6034,7 +6034,10 @@ Subclassing `StdAdoDelegate` gets you all of it. A database whose row-limiting s
 clause and **`AddPagingParameters(cmd, skip, take, takeLimited)`** binds it (`skip` and `take` are `int`,
 matching the query objects they serve). Override both together when your clause names the two parameters
 in the other order, because providers that bind positionally take parameters in the order the statement
-mentions them. `MySQLDelegate` and `SQLiteDelegate` do exactly this for `LIMIT … OFFSET`.
+mentions them. `MySQLDelegate` and `SQLiteDelegate` do exactly this for `LIMIT … OFFSET`. The two
+parameter names the pair has to agree on are `AdoConstants.ParameterPageSkip` and
+`AdoConstants.ParameterPageTake` — new in 4.0, and public because a delegate outside Quartz has to
+spell them.
 
 The value-conversion pairs on `IDbAccessor` are no longer all overridable on `StdAdoDelegate`: the
 boolean pair (`GetDbBooleanValue` / `GetBooleanFromDbValue`) stays virtual, because Oracle genuinely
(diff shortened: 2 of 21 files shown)


lahma:
Reviewed in full. All deviations ACCEPTED: the ScheduleJobOptions reuse over a one-letter twin, the GetAcquireRetryDelay exclusion the guard test itself caught (the tool finding a hole in its author's enumeration is the point of guard tests), the §6.5 item routing, the shared IL-walker helper, and the no-fiction rule for the 3.x-indexed appendix. Merging on green.

lahma:
Coverage closure reviewed and ACCEPTED: the SQLite statement test checks the thing worth checking (bound string = stored string so the WHERE matches), and walking back the const-from-const indirection in favor of an equality-holding test is the right call — the property was never one-declaration, it was no-drift. The instrumentation fact (a const initialized from another const gets an uncoverable sequence point) is recorded in campaign memory for the next time. Merging on green.

PR #3617, opened 2026-08-31
The acquisition index is sorted the way acquisition reads

### LINKED ISSUE #3510: Acquisition orders by NEXT_FIRE_TIME ASC, PRIORITY DESC, and no index can deliver that
The index audit (#3426, #3509) measured what the plan expected — appending `PRIORITY` ascending to `IDX_QRTZ_T_NFT_ST` changes nothing, because `SelectNextTriggerToAcquire` orders by `NEXT_FIRE_TIME ASC, PRIORITY DESC` and an ascending trailing column cannot deliver two directions — and one thing it did not: with the trailing column **descending**, the acquisition round trip at 5,000 candidates on a 100k-trigger table goes

| engine | p50 today | p50 with `PRIORITY DESC` | reads/buffers |
|---|---|---|---|
| SQL Server | 18,803 µs | 632 µs | 4,319 → 8 logical reads |
| MySQL | 48,709 µs | 839 µs | sorts 5,000 rows → none |
| Postgres | (incremental sort) | — | 371 → 7 buffers |

Verified not to be a tie artefact (distinct fire times: 4,319 → 13 reads). The plans and the harness (`AcquisitionIndexBenchmark`, `QUARTZ_BENCHMARK_EXPLAIN`) are in #3509.

It was deliberately not adopted there, for three reasons that make this a design item rather than a migration:

1. **Firebird cannot express a mixed-direction index** — every migration ships for every dialect, so either Firebird gets a different shape and a documented difference, or the statement changes so that a single-direction index serves it (e.g. ordering by a negated priority column, or splitting the acquisition into a fire-time window and an in-memory priority sort, which is what the batch already does after the read).
2. **The payoff hinges on the lower bound.** `SelectNextTriggerToAcquire`'s `NEXT_FIRE_TIME >= @noEarlierThan` is not sargable as written on every engine, and the adversarial case — a 5,000-row misfire backlog below the bound — read 20,410 pages against the shipped index's 4,319. The index needs the query to seek, or it makes the bad day worse.
3. The write cost is permanent on the hottest table; the read win is per acquisition round, so the number has to be weighed at cluster scale, with the batch size and the misfire sweep in the picture.

**Done means**

- A decision on the query shape first (sargable bound; whether the ORDER BY can become single-direction), measured with the #3509 harness on Postgres, SQL Server, MySQL and Firebird, including the misfire-backlog case.
- If adopted: a generated migration for every dialect, `schema-changes.md`, `database/README.md`, and the `db/index.md` section #3509 wrote updated to say what changed and why.
- If not: `db/index.md` keeps the recorded numbers as the reason.

lahma:
`SelectNextTriggerToAcquire` orders by `NEXT_FIRE_TIME ASC, PRIORITY DESC`, and every dialect splices its row limit into the statement, so the `ORDER BY` decides which rows come back at all — the highest-priority trigger of a tied set, which is what `RAMJobStore` gives you too. `IDX_QRTZ_T_NFT_ST` could deliver neither direction, so the engine read every due trigger and sorted it. This reshapes the index so it can.

```
IDX_QRTZ_T_NFT_ST (SCHED_NAME, TRIGGER_STATE, NEXT_FIRE_TIME ASC, PRIORITY DESC, MISFIRE_INSTR)
                  -- sqlServer (incl. the MOT and pre-2016 variants), postgres, mysql_innodb, oracle, sqlite

IDX_QRTZ_T_NFT_ST (SCHED_NAME, TRIGGER_STATE, NEXT_FIRE_TIME)
                  -- firebird, unchanged and documented
```

**The acquisition statement does not change, and there is no C# change at all.** `StdAdoConstants`, every delegate, `AdoJobStoreBase` and the public API baselines are untouched — the ADO-internalization stream is unblocked and the two can land in either order.

## Measured

100,000 triggers, 5,000 due in ties of a hundred, one acquisition at `MaxCount = 1`, p50 and the engine''s read counter. Full method, plans and the rejected candidates are in the design note on #3510.

| | SQL Server 2022 | MySQL 8.0 | PostgreSQL 15 | Firebird 4 |
|---|---|---|---|---|
| p50 | 21.6 ms → **0.59 ms** | 11.8 ms → **0.69 ms** | 0.89 ms → 0.74 ms | 14.8 ms → 14.8 ms |
| reads | 20,395 → **8** | 15,517 → **96** | 1,526 → **24** | 10,025 → 10,025 |
| 5,000-row backlog, 100 due | 838 → 84 reads | 15,404 → 117 reads | 1,956 → 408 buffers | unchanged |
| index size | +6.2 % | +13 % | +0.4 % | unchanged |
| reschedule `UPDATE` p50 | unchanged | unchanged | unchanged | unchanged |

`MISFIRE_INSTR` is the fifth column because the ordered seek starts at the oldest waiting trigger and the statement''s lower bound sits inside an `OR`, so it cannot narrow the seek. Without that column every backlogged row the seek walks costs a table lookup to find out it is not wanted — 20,401 reads on SQL Server against a 5,000-row backlog. With it the rows are skipped inside the index, for **84**. The predicate never has to become sargable.

`PREFERRED_NODE` was measured beside it and deliberately left out: it buys nothing over `MISFIRE_INSTR` alone on SQL Server and MySQL, and it makes PostgreSQL''s backlog case *worse* (660 buffers against 408).

**The one honest bad cell**: several thousand misfired triggers below the window with only a handful due costs about 3 ms against today''s 1.5 ms. With several thousand *due* it is 3 ms against 30 ms. Backlogs drain; the trade is stated in the docs rather than hidden.

## The Firebird exception

Firebird keeps the three-column index, as a documented per-dialect difference. Its indexes take one direction for the whole index, so `CREATE INDEX` rejects the `ASC` token outright (`SQL error code = -104`), and the usual workaround — a `COMPUTED BY` column holding the negated priority — cannot be indexed there either (*attempt to index COMPUTED BY column*). Both were probed rather than assumed. Without an ordered seek in front of it `MISFIRE_INSTR` buys Firebird nothing (10,025 reads either way), so the trailing columns would be write cost with nothing to buy.

## Two generator traps fixed on the way

Both were **verified byte-neutral** for every existing script — the commit touches no file under `database/migrations/` outside `4.0/`, which is the check.

- `CreateIndex` emitted `columns.Replace(" ", "")` on SQLite and MySQL, which turns `NEXT_FIRE_TIME ASC, PRIORITY DESC` into `NEXT_FIRE_TIMEASC,PRIORITYDESC`. SQLite answers `no such column: NEXT_FIRE_TIMEASC`. It is now a comma-join that trims each element, in the one helper the three emit sites share.
- `ReshapedIndexes` consulted only `Historical316`, which has PostgreSQL entries alone. A guarded `CREATE INDEX IF NOT EXISTS` against a database that already has the name over the old columns silently keeps the old shape, so the 4.0 script now emits a guarded `DROP` first on every dialect whose shape moved. That is exactly one extra drop per dialect, on `IDX_QRTZ_T_NFT_ST`; the other eight names are column-identical between the 3.x and 4.x sets or absent from the 3.x one.

No new migration folder: this lands inside the existing `database/migrations/4.0/schema_30_to_40_upgrade_<dialect>.sql`, section 6, which is already marked optional. **No companion `3.x` PR is needed** and `database/migrations/` stays byte-identical across the branches for every shared version.

## Tests

`SchemaScriptTest` compared index **names** between the fresh-install and generated scripts, so it would not have noticed a lost direction. It now compares column lists, and two new cases pin the shape itself: `TheAcquisitionIndexCarriesTheDirectionsAcquisitionOrdersBy` on the five dialects that take a per-column direction, and `FirebirdKeepsTheThreeColumnAcquisitionIndex` so the difference reads as deliberate rather than as drift a later alignment pass should tidy up.

Both were confirmed to fail without the change: reverting the constant fails all five dialects, and reverting one dialect''s `tables_*.sql` direction fails the generated-versus-fresh comparison for that dialect alone.

## Verification

- `dotnet build Quartz.slnx` — 0 warnings, 0 errors
- `dotnet test src/Quartz.Tests.Unit` — **4287 passed**, 0 failed
- `dotnet test src/Quartz.Tests.AspNetCore` — **548 passed**, 0 failed
- `dotnet fallout GenerateMigrations GenerateSchema` then `VerifyMigrations VerifySchema` — both up to date
- `dotnet fallout VerifyDocsSnippets` — 102 markdown files, up to date
- Integration legs, all six dialects, locally against Docker: sqlite 14, postgres 171, sqlServer 163, mysql 104, firebird 104, oracle 103 — all passed. That covers `MigrationScriptTest` (stepped and direct upgrade routes, re-runnability, and a scheduler run against the migrated schema) and `SchemaProvisioningTest` on every engine, so the new DDL is executed for real on all six — including Oracle, where a descending key makes the index function-based, and SQL Server''s memory-optimized script, which seeds a container.

## For a reviewer to decide

- The docs claim the acquisition order is load-bearing (`PRIORITY DESC` is what picks the row when the row limit is 1). That is the reason the statement-reshape candidate was rejected — it would have made the tie-break trigger *name* on the ADO stores only.
- `AcquisitionIndexBenchmark` still measures its `Shipped` arm as the three-column index. That is right — it is the baseline every other arm is compared against — but the comment claiming it is "what `database/tables/` ships" was corrected. Growing the harness a misfire-backlog toggle and a Firebird arm is left as follow-up work, as the design note proposed.
- `IDX_QRTZ_T_NFT_ST_MISFIRE` has the same problem for the misfire sweep and is deliberately untouched here; it wants its own issue and its own measurements.

Closes #3510


lahma:
Reviewed in full. The proof discipline here is the standard: byte-neutral regeneration before the shape change, all six dialect legs run for real (including MOT seeded from its own script and Oracle's function-based form), and both guards demonstrated failing when reverted. Deviations accepted: the honest heading rename (checked-by-reasoning is fine — CI runs the link checker), the benchmark doc-comment corrections without a fifth Params arm, and the Firebird tables-script comment. The two scoped-out items are rightly separate: catalog-level direction in SchemaSnapshot and an EXPLAIN-plan tie between StdAdoConstants and the DDL can ride the misfire-index follow-up if it ever runs. The recovery-index defect is already filed. Merging on green.

lahma:
Sonar override, documented per the R1/#3543 precedent: the quality gate's three CRITICALs are all in **generated** migration SQL — two Oracle `plsql:ExecuteImmediateTrapExceptionsCheck` on guarded DDL blocks that deliberately trap only the specific ORA codes meaning 'already exists' (a `WHEN OTHERS` handler would hide real failures, which is the opposite of what a migration wants), and one `plsql:S1192` duplicated-literal on generator output where a 'constant' has no meaning. `new_coverage` 0% counts only SQL files and the build project, which no unit leg can cover — the six real dialect legs ran the DDL against live engines locally. Gate overridden; merging when the required legs pass.

lahma:
## CI update — the SonarCloud gate, and a change to `.github/workflows/sonar.yml` a reviewer should ratify

The first run went green on every job except **SonarCloud Code Analysis**, which failed two gate conditions. Neither is about this change''s logic; both are the PL/SQL analyser applied to generated DDL. Commit cc1c428 extends the exclusion mechanism this workflow already has for exactly that, and it is the one change here worth a second opinion.

**What failed** (`api/qualitygates/project_status`, pullRequest=3617):

| condition | threshold | actual |
|---|---|---|
| `new_coverage` | ≥ 80 | **0.0** |
| `new_maintainability_rating` | 1 | **3** |
| the other four | — | OK |

**`new_coverage` 0.0** — the entire denominator was `database/migrations/4.0/schema_30_to_40_upgrade_oracle.sql`: `new_lines_to_cover 4`, `new_uncovered_lines 4`. Sonar reads every `.sql` as PL/SQL and counts the executable lines of a guard block as lines to cover, while this workflow imports C# coverage alone (`sonar.cs.opencover.reportsPaths`). No C# in the PR is coverable at all — the product change is DDL — so four lines of Oracle guard were the whole measurement. **Any** PR that adds a line to an Oracle migration fails this condition on its own. `database/**` now joins `docs/.vuepress/**` in `sonar.coverage.exclusions`, which is the argument the file already makes two paragraphs above for the same situation in the other language.

**`new_maintainability_rating` 3** — three code smells, all in generated SQL:

- `plsql:ExecuteImmediateTrapExceptionsCheck` ×2 on the Oracle script (lines 174, 226) — "add a WHEN OTHERS exception handler". This is the sibling of `plsql:S1523`, which `database/migrations/**/*.sql` is **already excused for**, on the identical construct. Here the handler it asks for is the thing not to write: a migration that swallows the exception its DDL raised reports success over a half-upgraded schema.
- `plsql:S1192` on the SQL Server script (line 133) — "define a constant instead of duplicating this literal 15 times". The literal is `''dbo.QRTZ_TRIGGERS''`, once per guarded index statement; my new `DROP INDEX` guard is the fifteenth and tipped it over the threshold. A `.sql` file a DBA runs has nowhere to define a constant, and none that would survive the "replace `QRTZ_` with your configured table prefix" its own header asks for.

The workflow''s existing comment block already names both of these findings by description and calls them noise on generated SQL — it says so about `src/Quartz/Impl/AdoJobStore/Schema/`, which it drops from analysis entirely for being generated. `database/migrations/` cannot take that treatment: it is documented, human-run SQL and its bug and security rules matter. So they become two more per-rule ignores beside the two already there, each with its own paragraph in the file''s voice.

**Nothing else is excused.** The two rules are switched off under `database/` and nowhere else, the bug and security rules still apply to those files, and `MigrationScriptTest` still applies every script to a real database of each dialect and compares the result with a fresh install — which is what actually exercises them, in a leg that feeds no coverage report.

**I could not fix these at the source.** The files are generated; `dotnet fallout VerifyMigrations` fails the build if they are not byte-for-byte what `build/Build.DatabaseMigrations.Scripts.cs` emits, so a finding in one of them cannot be acted on where it is reported — and the two "fixes" Sonar wants are, respectively, harmful and impossible.

If you would rather the gate stayed as it was, the alternative is to accept these three as won''t-fix in the SonarCloud UI per occurrence, which does not survive the next migration that touches an Oracle guard.

The rest of the first run was green: `build`, `analyze`, all six `ubuntu-latest` legs, `CodeQL`, `Analyze (actions)`, `Analyze (javascript)`.

lahma:
The sonar.yml gate edit is RATIFIED after reading the diff line by line. This is the durable form of the override I had documented as a one-off: without it, every future PR touching an Oracle migration goes red on its own. The scoping is right — two PL/SQL rules excused under database/ only (WHEN OTHERS around migration DDL is actively harmful; a constant has no home in a runnable .sql), coverage excluded for a language this workflow's importer can never feed while MigrationScriptTest exercises the real thing, bug and security rules still live, and the prose sits in the file's own established voice beside the S1523 precedent. Surfacing 'agent edits the gate' as the thing needing explicit ratification was exactly the correct instinct — this comment is that ratification. Merging when the re-run's required legs pass.

Issue #3603, opened 2026-08-31
What was examined and kept is written down


lahma:
The alpha.5 audit examined the whole public surface. Most findings become code changes; the rest were examined and **kept** — and a ruling that is not written down gets re-audited by every future contributor and coding agent. AGENTS.md's "Naming decisions that are settled" section gains a row for each, with the reason:

- **Two read altitudes**: `Query*` query-object members on `IScheduler` and `Get*` conveniences in `SchedulerQueryExtensions` are one API at two altitudes, deliberately.
- **`ISchedulerRepository` vs `ISchedulerRegistry`**: live-instance directory vs persisted-registration query — different jobs, both stay.
- **The four history-logging plugins**: the classic numbered-format pair exists for 3.x log-pipeline shapes; the structured pair is the recommendation. Four types over two options types is the compat price, paid knowingly.
- **`LogProvider.SetLogProvider`**: the one static escape hatch stays — logging before any container exists is real.
- **`SchedulerContext`/`JobDataMap` divergence**: `PutAsString` and equality participate in job-data change tracking; the twins differ on purpose.
- **Wrapping ranges (`22-2`, `FRI-MON`) and `H` tokens**: Quartz supersets over standard cron, kept as identity.
- **`ISchedulerListener` leading with `IScheduler`** while job/trigger listeners get it from context: scheduler listeners have no other channel.
- Whatever the builder-dedup and serializer streams decide to *keep* (`Extensibility` parameter types in `Use*` members, `Create()` companion classes, the Aspire `NameValueCollection` twin already ratified) gets its row in the same pass.

Lands late in the campaign so the list matches what actually merged; the PR also corrects AGENTS.md's stale description of `QuartzSchedulerBuilder`.

PR #3618, opened 2026-08-31
A listener's matchers are settled when it is registered, and a misfire leads with its trigger

### LINKED ISSUE #3587: The listener surface sheds what middleware and DI registration made redundant
The listener-vs-middleware assessment after #3553 concluded the three listener interfaces all stay — notification is a different job than wrapping, and `JobToBeExecuted`/`JobWasExecuted` and the trigger veto are load-bearing — but it also found surface that no longer earns its place now that `IJobExecutionMiddleware` exists and DI owns registration:

**`IListenerManager` — the per-listener matcher-mutation members.** `AddJobListenerMatcher`, `RemoveJobListenerMatcher`, `SetJobListenerMatchers`, `GetJobListenerMatchers` (`IListenerManager.cs:86,100,116,134`) and their trigger twins (`:203,214,226,239`) exist to *edit a live listener's matcher set by listener name, at runtime, one matcher at a time*. Nothing in the repository calls them outside their own tests; the DI path (`AddJobListener<T>(matchers)`) and the `AddJobListener(listener, params matchers)` member cover declaring matchers at registration, which is when anyone actually knows them. Eight members whose only real user is the remove-then-re-add dance they were meant to avoid. Trim them; matchers become part of registration (`Add*Listener` keeps its `params` matchers), and a listener whose matching must change is re-registered.

**The three `Broadcast*` listeners** (`src/Quartz/Listeners/BroadcastJobListener.cs`, `BroadcastTriggerListener.cs`, `BroadcastSchedulerListener.cs`). They fan one notification out to N listeners — which is what the listener *manager itself does* for every registered listener. Their one distinguishing feature (grouping listeners under one name) has no known consumer, and each is a public type that must track its interface's every default-implemented member forever. Remove them; the guide row says "register the listeners individually".

Pre-4.0 window: both trims are breaking and free only while 4.0 is unreleased. Baselines via Verify; migration-guide rows for each removed member with the replacement spelled; sweep docs for mentions. Explicitly out of scope, ratified: retiring any of the three listener interfaces, folding the veto into middleware, or touching `JobChainingJobListener`.

### LINKED ISSUE #3595: TriggerMisfired takes the trigger first, like its siblings
\`ITriggerListener\`'s four notifications disagree about their own shape: \`TriggerFired\`, \`VetoJobExecution\` and \`TriggerComplete\` lead with the trigger and reach the scheduler through \`context.Scheduler\`; \`TriggerMisfired(IScheduler scheduler, ITrigger trigger, …)\` alone leads with a scheduler parameter and carries no context — an artifact of the notification path (\`ISchedulerSignaler.NotifyTriggerListenersMisfired\` passes no scheduler; the manager injects one).

A misfire has no execution context, so the signature cannot match its siblings exactly — but it can stop being the parameter-order outlier: \`TriggerMisfired(ITrigger trigger, IScheduler scheduler, …)\`, trigger first like every other member of the family.

Same PR decides the second listener-family asymmetry with a ruling rather than reflexive change: \`ISchedulerListener\` takes \`IScheduler\` first on all 23 members while \`IJobListener\`/\`ITriggerListener\` members take none — either that difference gets a written reason (scheduler listeners have no other channel to the scheduler) in the AGENTS.md settled list, or the shape changes now. Lean: ratify, don't churn 23 signatures.

Breaking for every \`ITriggerListener\` implementor that overrides the misfire DIM; Verify baseline + migration-guide row in the PR.

lahma:
Two listener-surface issues, one pass over the same files.

## `IListenerManager` sheds the matcher-mutation members (#3587)

`AddJobListenerMatcher`, `RemoveJobListenerMatcher`, `SetJobListenerMatchers`, `GetJobListenerMatchers`
and their four trigger twins let an application edit a registered listener's matcher set at run time,
one matcher at a time, by listener name. Nothing outside their own tests called them. Matchers are
known at registration, `Add*Listener(listener, params matchers)` already takes them there, and a
listener whose matching must change is registered again under the same name — which replaces the
listener and its matchers together, so the two can never be out of step.

**Correction to the issue body, ratified during planning:** the two `Get*ListenerMatchers` members are
consumed on the notification hot path, so they could not simply vanish. The trim is **six removals plus
two demotions to internal**, and `IListenerManager` goes from 20 members to **12**, not the 14 the issue
arithmetic implied.

The demotion came with the perf change the two readers were asking for. A job or trigger listener is
held as an `AttachedListener<TListener, TKey>` — the listener, the name it was attached under, and its
matchers, together — so:

- `QuartzScheduler.MatchJobListener` / `MatchTriggerListener` are gone, and with them a dictionary
  lookup **per listener per firing** that only ever recovered what the registration already knew;
- `ListenerManagerImpl` loses its two parallel matcher dictionaries and the two private helpers that
  kept them in step with the listener dictionaries (about 160 lines);
- the name is captured at registration rather than read back off the listener, so a listener with a
  settable `Name` cannot stop answering to the name it was attached under.

The internal shape is `GetAttachedJobListeners()` / `GetAttachedTriggerListeners()` on
`ListenerManagerImpl`; there is no by-name matcher getter left in product code, because nothing but a
test wanted one, and the tests use a local helper over the attachments.

## The broadcast listeners are gone (#3587)

`BroadcastJobListener`, `BroadcastTriggerListener` and `BroadcastSchedulerListener` fanned one
notification out to N listeners — which is what the listener manager does for every registered
listener. 535 lines, zero production callers. Guide row: register the listeners individually.

`ListenerSenderTest.BroadcastSchedulerListener_ForwardsTheSchedulerItWasGiven` is deleted; what it
guarded (a fan-out passes on the scheduler it was told about) is what
`OneListenerServingTwoSchedulers_IsToldWhichOneCalled` in the same file already proves for the manager's
own fan-out. `SimpleTypeLoaderTest`'s `Quartz.Listener` → `Quartz.Listeners` fallback test is repointed
at `JobChainingJobListener`, which is still there.

## `ExecutingJobsManager` stops implementing `IJobListener` (#3587)

The scheduler has called `FiringStarted` / `FiringEnded` directly since #3502. Nothing resolves the type
as an `IJobListener` — it is constructed by `QuartzScheduler` and held in a field typed as itself — so
the interface, the `Name` it needed to be registered under, and the three notification bodies were shape
without a caller. `ExecutingJobsManagerBenchmark` went with them: it existed only to compare two
implementations of that `Name` property. `ExecutingJobsManagerTest`, whose only test asserted that name,
now tests what the type is actually for.

## `TriggerMisfired` takes the trigger first (#3595)

`TriggerMisfired(ITrigger trigger, IScheduler scheduler, …)`. It keeps the scheduler — a misfire has no
execution context to reach one through — but it stops being the one member of `ITriggerListener` not to
lead with the trigger it is about, and the scheduler sits where its three siblings take the
`IJobExecutionContext`.

Both parameters are still present, so **a listener that keeps the old order still compiles and silently
stops implementing the interface** — precisely the failure mode `ListenerShape` exists for. Before this
change the misfire case was covered by its `MissesTheScheduler` hint; after it, that hint no longer fires
for `TriggerMisfired`, so `ListenerShape` learns one more: the parameters are the notification's own, in
a different order. That is the only addition here beyond the issues, and it preserves a diagnostic the
signature change would otherwise have silently removed.

**The `ISchedulerListener` question is answered with a ruling, not a change**, as the issue leaned. The
migration guide now says why the two interfaces differ: `ISchedulerListener` takes the scheduler first
because its notifications are *about* the scheduler and have no other subject to lead with, and the
scheduler is a scheduler listener's only channel to what it is listening to; `ITriggerListener` leads
with the trigger. Twenty-three signatures are not churned. **The AGENTS.md settled-list entry is left
for #3603** — say the word and I will add it here instead.

## For a reviewer to decide

- **One commit, not two.** The two issues touch the same interface, the same tests, the same three
  Verify baselines and interleaved paragraphs of the migration guide; a split would produce two commits
  neither of which tells a whole story.
- **The `AttachedListener<,>` name.** `Quartz.Configuration` already has a `ListenerRegistration<T>`
  family for the DI-time registration, so the runtime pairing needed a different noun.
- **The `ListenerShape` reorder hint** — scope beyond the letter of either issue, argued above.
- **The appendix count moved from 94 to 97** removed public types, with rows for the three broadcast
  listeners, and a member row for the eight matcher members.

## Verification

```
dotnet build Quartz.slnx                     0 Errors, 0 Warnings
dotnet test src/Quartz.Tests.Unit            Passed! 4290 passed, 0 failed
dotnet test src/Quartz.Tests.AspNetCore      Passed!  548 passed, 0 failed
dotnet fallout VerifyDocsSnippets            Succeeded (102 markdown files checked)
dotnet fallout BenchmarkSmoke                Succeeded (282 benchmarks)
```

Baselines were reviewed before acceptance and contain exactly the intended delta: eight
`IListenerManager` members, three `Quartz.Listeners` types, and the `TriggerMisfired` reorder in
`Quartz`, `Quartz.Plugins` and `Quartz.Dashboard`. `docs/documentation/quartz-3.x/` is untouched, as is
its multi-tenancy recipe's reference to the matcher members.

Integration tests were not run: they need a Docker daemon, and nothing here reaches a job store.

Closes #3587. Closes #3595.


lahma:
Reviewed in full. All deviations ACCEPTED: (1) 20→12 is the honest arithmetic (6 removals + 2 demotions = 8 off the interface; the 14 was the issue's own slip); (2) AttachedListener with no by-name getter is the cleanest shape — a getter with only test callers is surface nobody needs, and the notify-path win was the point; (3) the ListenerShape reorder hint is KEPT — a diagnostic that names the exact stale shape the reorder creates is precisely the kind of migration kindness this release should ship; (4) benchmark deletion follows the interface it existed to compare; (5) migration-guide ruling + #3603 for AGENTS.md is the agreed split. Merging on green.

Issue #1253, opened 2021-07-26
Seconds as optional parameter


diogonborges:
**Is your feature request related to a problem? Please describe.**
Any chance that the cron expressions would be supporting seconds to be optional?

**Describe the solution you'd like**
Ideally like to have support for expressions like: * * * * *

**Describe alternatives you've considered**
I've considered other libs packages?! 
Sorry, I don't know how to best describe alternatives other than adopting the cron expression format

**Additional context**

lahma:
So you would not want to have 0 as first parameter, that's all it takes to "ignore" the seconds part?

diogonborges:
That's correct @lahma.
Is it a weird expectation? 
I've read through the documentation: https://www.quartz-scheduler.net/documentation/quartz-3.x/tutorial/crontriggers.html#example-cron-expressions to come up with the awareness that **Year** is optional but **Seconds** not.

Is there a reason for this? My fairtly un-educated view on this would tend to think the think the most adopted flavour of cron expressions takes after Unix Style and would use 5 or 6 fields.

Thanks for reaching back!

lahma:
Yes, for Unix-like systems second granularity wouldn't make sense because of the overhead involved in starting the process. Quartz.NET follows Java version's logic allowing to utilize cron for seconds-level accuracy for firing. So it's for convenience and unfortunately makes it different from what a systems admin would expect. If there would be a backwards compatible way to support both, why not.

lahma:
Alpha.5 design ratified (alpha5-note-cron.md §6) — this ships as an **explicit format switch, not auto-detect**:

- `CronFormat` enum (`Quartz` = 0 default, `Unix`), surfaced on `Parse`/`TryParse` and `CronScheduleBuilder.Create` only — four baseline lines, one type. No constructor overload, no `WithCronSchedule` overload (compose via `WithCronSchedule(CronExpression.Parse(s, CronFormat.Unix))`), and no XML/HTTP/dashboard format input: a 5-field expression is something you write in C#, not something you store.
- Auto-detect was examined and rejected on this repository's own evidence: three separate test fixtures use "5 fields" to mean "invalid expression"; dropping a middle field would become silent (`0 0 12 * * ?` minus the month = midnight on the 12th instead of noon daily); the same digit means different days across dialects; and auto-detect can never grow to Spring's 6-field format. Adding auto-detect later is additive; removing it would be impossible — at a freeze, take the reversible option.
[...]

PR #3619, opened 2026-08-31
A cron expression's surface is Parse, not four constructors and four validators

### LINKED ISSUE #3592: CronExpression sheds the members that restate Parse and the constructor that cannot be called
Freeze-window cleanup of the \`CronExpression\` surface; every item is a removal or rename that is free now and impossible after GA.

- \`new CronExpression(expr, null)\` does not compile: with \`(string, TimeZoneInfo?)\`, \`(string, string hashKey)\` and \`(string, int hashSeed)\` all present, the null literal is a CS0121 ambiguity. The hash-resolving constructors move to statics (\`WithHashKey\`/\`WithHashSeed\` or equivalent — design note picks the shape) and the type keeps two constructors.
- \`IsValidExpression\` ×2 and \`ValidateExpression\` ×2 (\`CronExpression.cs:548-586\`) are \`TryParse\`/\`Parse\` under other names — four members, zero capability. Removed, with migration-guide rows naming the replacements.
- \`GetTimeBefore\` works (inverse-doubling + binary search, \`:2604-2660\`, DST-pinned) but its name hides its symmetry with \`GetNextValidTimeAfter\` → renamed \`GetPreviousValidTimeBefore\`.
- \`GetExpressionSummary\` (\`:1526\`) returns a newline-delimited internal-state dump, pinned verbatim by a test — an implementation detail wearing a public method. Removed unless the design note finds a real consumer.

Rides the baseline-heavy merge chain (public-API Verify baseline + migration guide in the same PR).

lahma:
Freeze-window cleanup of `CronExpression`'s public surface. Every item is a removal or a rename that is free now and impossible after GA. Implements §5 of the ratified alpha.5 cron design note; it is cron PR-4, and it sits on #3614 (the `'C'` reject that made the `calendarDayOf*` fields constant) and #3615.

## The constructor that could not be called

`new CronExpression(expr, null)` did not compile. With `(string, TimeZoneInfo?)`, `(string, string hashKey)` and `(string, int hashSeed)` all present the null literal was a CS0121 ambiguity, so the one overload whose parameter is *documented* as nullable — "`null` means the system's local time zone" — could not be handed a null.

The two hash constructors go. The resolution they performed moves to statics beside `Parse`:

```csharp
CronExpression expression = CronExpression.ParseWithHash("0 H H(0-7) * * ?", "nightly-cleanup");
CronExpression seeded     = CronExpression.ParseWithHash("0 H H(0-7) * * ?", 42);

if (CronExpression.TryParseWithHash(userInput, "nightly-cleanup", out CronExpression? parsed))
{
    // parsed is non-null here
}

CronExpression local = new CronExpression("0 H H(0-7) * * ?", null);   // compiles now
```

Statics rather than an options record because resolving `H` **is** a parse: it reads the expression and throws `FormatException` when it cannot. `TryParseWithHash` is the member that carries capability — it is the non-throwing validation that `IsValidExpression(expr, hashKey)` provided and that `Parse(ResolveHash(...))` cannot, because `ResolveHash` throws on a malformed `H` token. The type keeps two constructors, so the null stays unambiguous forever.

## Four members that restated two

`IsValidExpression` ×2 and `ValidateExpression` ×2 are removed.

| removed | replacement |
|---|---|
| `IsValidExpression(expr)` | `TryParse(expr, out _)` |
| `ValidateExpression(expr)` | `Parse(expr)` |
| `IsValidExpression(expr, hashKey)` | `TryParseWithHash(expr, hashKey, out _)` |
| `ValidateExpression(expr, hashKey)` | `ParseWithHash(expr, hashKey)` |

The replacements hand back the expression they parsed, so validating and then using is one parse rather than two — `CronScheduleBuilder.Create(string)` no longer resolves-and-validates ahead of building.

## Two renames and a deletion

- `GetTimeBefore` → `GetPreviousValidTimeBefore`, on `CronExpression` **and** on `CronTriggerImpl`'s `protected` forwarder. It is the pair of `GetNextValidTimeAfter`; only the name hid that, and renaming one without the other would stop the pair reading as a pair.
- `GetExpressionSummary` is removed from `CronExpression`, `ICronTrigger` and `CronTriggerImpl`, and `CronExpressionSummary.cs` with it. It dumped the parsed field sets in an undocumented newline-delimited format, no product code anywhere called it (zero hits across the dashboard, the HTTP API and their tests), and the one test that existed pinned its exact text — including field ordering and the `?`/`*` distinction that this alpha's other cron work changes.

## Blast radius

- `src/Quartz.Documentation.Samples/CronExpressionsSamples.cs` (two `#region sample_*` blocks) — regenerated with `dotnet fallout DocsSnippets`; `VerifyDocsSnippets` is green.
- `src/Quartz.Benchmark/CronHashTokenBenchmark.cs` — `dotnet fallout BenchmarkSmoke` is green.
- Doc comments in `CronScheduleBuilder` (×3, including the runtime message a keyless `H` expression throws) and `TriggerConfiguratorExtensions` that spelled the old constructor.

## Migration guide

The section that promised *"`IsValidExpression` is unchanged at the call site"* is **rewritten**, not appended to — that promise dies here. It gains the four-row replacement table and a new **The hash key is a `Parse` argument, not a constructor overload** section. Eight rows go into the removals appendix. Three places elsewhere in the guide recommended `new CronExpression(expr, hashKey)` as the 4.x replacement for a removed 3.x member (the `WithCronSchedule(expr, hashKey)` rows, the diff sample beneath them, and the `CronScheduleWithHash` rows); each now names `ParseWithHash`, or it would have taught a call that no longer compiles.

`docs/documentation/quartz-3.x/` is untouched, per convention.

## For the reviewer to decide

1. **`TryParseWithHash` throws `ArgumentNullException` on a null `hashKey`** rather than returning `false`, which is what `IsValidExpression(expr, hashKey)` did. The reasoning: the expression is the untrusted input and a null one returns `false` as `TryParse` does, but the key comes from the program, so a missing one is a bug rather than a parse failure. Pinned by a test with that reason string. Say the word if you would rather it returned `false`.
2. **Only the string-key `TryParseWithHash` is added**, not a seeded twin — that is what the note ratified. `ParseWithHash` has both overloads because the removed constructors did.

## Verification

- `dotnet fallout Compile` (whole solution, inside BenchmarkSmoke) — succeeded
- `dotnet test src/Quartz.Tests.Unit` — **4,464 passed, 0 failed**
- `dotnet test src/Quartz.Tests.AspNetCore` — **548 passed, 0 failed**
- `dotnet fallout DocsSnippets` then `VerifyDocsSnippets` — 102 markdown files, up to date
- `dotnet fallout BenchmarkSmoke` — 284 benchmarks executed, succeeded
- Public API baseline accepted deliberately; the diff is exactly ctors −2, `GetExpressionSummary` −3, `IsValidExpression` −2, `ValidateExpression` −2, two `GetTimeBefore` renames, `ParseWithHash` +2, `TryParseWithHash` +1, and nothing else.

Integration tests were not run (they need a Docker daemon); nothing here touches an ADO or store path.

Closes #3592


lahma:
Reviewed in full. Deviations ACCEPTED: (1) the three extra migration-guide sites teaching the removed ctor as a 4.x replacement had to move — good completeness catch; (2) TryParseWithHash throwing ArgumentNullException on a null hashKey is RATIFIED — a null hash key is a programming error at the call site, not input to validate, while a null expression mirrors TryParse; the reason-string test pins it; (3) the one-call ParseWithHash(expr, 0) validation in CronScheduleBuilder and the past-tense benchmark README are both right. Baseline diff confirmed as exactly the expected set. Merging on green.

lahma:
Rebased onto `ab4d5b7e1` (now carries #3618's listener trim, #3620's Unix dialect, and the two scheduling-data plugin commits). The API baseline conflicted with #3620 — both PRs add statics to `CronExpression`'s member block — and was resolved by **regenerating it**, not by merging the two sides: I took main's file verbatim, let the Verify test fail and emit `PublicApiTest_Quartz.received.txt`, and accepted that. The diff against main is still exactly my set, with #3620's `Parse(string, CronFormat)` and `TryParse(string, CronFormat, out)` retained untouched.

## A finding from the rebase, which I have deliberately **not** fixed here

#3620 expands `@` macros **in the constructor**, and says so: *"it is expanded by the constructor, which puts it behind every entry point at once"*. The hash path does not reach that expansion, because it resolves `H` **before** constructing — and `ResolveHashTokens` counts fields, so a one-field macro is rejected before the constructor ever sees it. Probed on this branch:

```
ctor                       OK    -> 0 0 0 * * ?
Parse                      OK    -> 0 0 0 * * ?
WithCronSchedule("@daily") OK    -> 0 0 0 * * ?
ContainsHashToken("@daily")OK    -> False
ResolveHash                THROW FormatException: Invalid cron expression (expected 6-7 fields, got 1): @DAILY.
ParseWithHash(key)         THROW FormatException: Invalid cron expression (expected 6-7 fields, got 1): @DAILY.
ParseWithHash(seed)        THROW FormatException: Invalid cron expression (expected 6-7 fields, got 1): @DAILY.
TryParseWithHash           OK    -> false
```

`TryParseWithHash` is the one that reads worst: it reports `@daily` as an **invalid expression**, so a validator written on it disagrees with the constructor that will later accept the same string.

**This is not a regression from this PR.** The constructor I removed was `CronExpression(string, string hashKey) : this(ResolveHash(cronExpression, hashKey))` — the identical route, in the identical order. Whatever `ParseWithHash("@daily", key)` does today, `new CronExpression("@daily", key)` did on `main` the moment #3620 landed. This PR relocates the behaviour verbatim; it neither introduces nor deepens it.

**Why I left it alone.** The fix is one place — expand the macro inside `ResolveHash` (both overloads already `ToUpper` the input, so it becomes `CronMacros.Expand(ToUpper(s).Trim())`), which would fix `ResolveHash`, both `ParseWithHash` overloads and `TryParseWithHash` at a stroke, and is purely additive (inputs that throw start working). But `ResolveHash` is named by neither #3592 nor #1253, and changing the reach of a feature that merged hours ago, inside a PR whose subject is removing members, is the kind of quiet scope creep that is hard to review. Say the word and I will add it here with a test; otherwise it wants its own issue against the macro feature. I have deliberately **not** added a test pinning the current behaviour either, since that would cement the wart and make the eventual fix look breaking.

Verification on the final base (`b52182ae4`):

- `dotnet test src/Quartz.Tests.Unit` — **4,792 passed, 0 failed, 36 skipped**
- `dotnet test src/Quartz.Tests.AspNetCore` — **548 passed, 0 failed**
- `PublicApiTest` alone — 8 passed, **no `.received.txt` produced**
- `dotnet fallout VerifyDocsSnippets` — up to date, 102 markdown files
- `dotnet fallout BenchmarkSmoke` — Restore / Compile / BenchmarkSmoke all succeeded

The 36 skips are #3620's own `QuartzReadsACrontabLineTheWayCrontabDoes` NCrontab-oracle cases plus the long-standing `[Explicit]` ones; none are mine.

PR #3620, opened 2026-08-31
A cron expression can be written in the Unix five-field form

### LINKED ISSUE #1253: Seconds as optional parameter

lahma:
Closes #1253.

A five-field crontab line is what every online generator emits, what Kubernetes `CronJob` takes, and what people paste into issues. Quartz has only ever answered it with an error. It reads it now — when asked.

## The surface

One enum and three members, and nothing else:

```csharp
public enum CronFormat { Quartz = 0, Unix = 1 }

CronExpression.Parse(string cronExpression, CronFormat format);
CronExpression.TryParse(string? cronExpression, CronFormat format, out CronExpression? result);
CronScheduleBuilder.Create(string cronExpression, CronFormat format);
```

No constructor overload — that keeps the constructor set at two and `new CronExpression(expr, null)` unambiguous forever. No `WithCronSchedule(string, CronFormat, …)` — compose `WithCronSchedule(CronExpression.Parse(s, CronFormat.Unix))`, which is one of the four overloads it already has. A time zone composes the same way. The public-API baseline moves by exactly those four lines plus the enum.

## Explicit, not auto-detected

Field counts of 5/6/7 are unambiguous *as counts*; reinterpreting a five-field string on sight is what is unsafe.

* **A dropped field would become silent.** `0 0 12 * * ?` minus its *month* is `0 0 12 * ?` — a well-formed crontab line meaning midnight on the 12th instead of noon daily. Today that is a `FormatException` naming the fix.
* **The same digit is a different day.** Quartz `5` is Thursday; crontab `5` is Friday. Field count would silently pick which.
* **This repository already means something by "five fields".** `Xml/TestData/InvalidCronExpression.xml` is `0 * * * *` with the comment *"Invalid: missing seconds field"*, read by four tests; `CronExpressionTest` asserts `"* * ? */5 *"` is invalid (a valid crontab line); and `FiveFieldUnixExpressionsAreRejectedWithGuidance` is a whole test whose subject would disappear.
* **Auto-detect cannot grow.** Spring's `@Scheduled(cron=…)` is six fields with seconds first and *Unix* day numbering — indistinguishable from Quartz's own six-field form, so it can only ever be an explicit opt-in. `CronFormat` extends to it; detection has nowhere to go.

Adding detection in 4.1 only ever makes previously-throwing strings work, which is additive. Removing it would be impossible. At a semantics freeze, take the reversible option.

## The rewrite

`UnixCronRewriter` is a pure string transform ahead of the existing parser. It checks the field count, renumbers every **bare integer** in day-of-week `n → (n % 7) + 1` (so 0 and 7 both land on Sunday), degenerates a whole-week range to `*`, spells one day field `?`, and hands `0 {min} {hour} {dom} {month} {dow}` to the constructor.

Nothing past that point — the parser, the evaluator, `CronField`, the serializers, the dashboard, the store — learns that a format was ever chosen. That is what lets a crontab line be scheduled, stored, displayed and read back with **no format column anywhere**, and it is why `L`, `W`, `#` and `H` all work inside the five-field layout: it is one grammar, not two. `H` in particular works because the rewrite runs *before* `ResolveHashTokens`, whose field tables are indexed by Quartz field position — `CronScheduleBuilder.Create("H 3 * * 1", CronFormat.Unix)` resolves against the trigger key exactly as it does for a Quartz expression.

| crontab | stored, and shown, as |
|---|---|
| `30 4 * * 1` | `0 30 4 ? * MON` |
| `0 12 1 * *` | `0 0 12 1 * ?` |
| `* * * * *` | `0 * * * * ?` |
| `0 0 13 * 5` | `0 0 0 13 * FRI` — the 13th **and** every Friday |
| `0 0 * * 0-6`, `0 0 * * 0-7`, `0 0 * * 1-7` | `0 0 0 * * ?` |
| `0 0 * * 5-1` | `0 0 0 ? * FRI-MON` |
| `0 0 * * 1/2` | `0 0 0 ? * 2/2` |
| `0 0 * * 0L` | `0 0 0 ? * 1L` |

`L` alone in a five-field day-of-week keeps its Quartz meaning of Saturday, because it is not a number and so has nothing to renumber. Documented.

The union rule for `0 0 13 * 5` is pinned with a reason string naming the divergence: we union when both day fields name days, **Cronos ANDs**. Switching that to AND would silently halve everyone's fire count, so the test says so rather than leaving the next reader to "finish" the alignment.

## Normalisation

`CronExpressionString` and `ToString()` return the canonical six-field Quartz form, and **the original is deliberately unrecoverable**. A stored string that parses only under a flag the store does not persist is a data-corruption trap; the dashboard and HTTP API display `CronExpressionString` and `CronNextFires.razor` re-parses it, so a form the dashboard's own parser rejects would be a bug on the page. There is no `CRON_FORMAT` column and no migration. This is the same trade as the uppercasing that has always turned `mon-fri` into `MON-FRI`.

The consequence is stated in the docs: the XML schema, the HTTP API and the dashboard take no format, so a five-field expression is something you write in C#, not something you store.

## Macros — unconditional, dialect-free, everywhere

| macro | expands to |
|---|---|
| `@yearly`, `@annually` | `0 0 0 1 1 ?` |
| `@monthly` | `0 0 0 1 * ?` |
| `@weekly` | `0 0 0 ? * SUN` |
| `@daily`, `@midnight` | `0 0 0 * * ?` |
| `@hourly` | `0 0 * * * ?` |

Vixie's set and only Vixie's set — no `@every_minute`/`@every_second` (`0 * * * * ?` is already short) and no jittered variants (`H` is deterministic and therefore better). They are expanded by the `CronExpression` constructor, so they arrive through **every** entry point with zero new API: `<cron-expression>@daily</cron-expression>` in an XML scheduling file, the dashboard's box, the HTTP API and `CronTriggerImpl.CronExpressionString`'s setter (the one the ADO store materialises rows through) are all tested. `@reboot` is rejected **by name** — a scheduler has no reboot to fire on — and any other `@name` gets the supported list. `@` was previously rejected outright, so this is purely additive.

This is also the mitigation for the ergonomic cost of choosing explicit over auto-detect: it recovers most of the paste-from-crontab experience at surfaces the format switch cannot reach.

## Error messages

`FiveFieldAdvice` (from #3614) gained a closing sentence naming `CronFormat.Unix`, so both its branches and both its call sites carry it.

**The message it belongs to is not the one a real crontab line hits.** `BuildExpression` parses the fifth field as a *month* before anything counts fields, so `0 12 * * MON` dies on `Invalid Month value: 'MON'` and never reaches the count check — only a numerically-valid fifth field gets there. Rather than leave the advice unreachable for the common shape, the month error now recognises a day name in the month field and gives the same advice. It only ever fires on input that was already an error, and no legitimate expression has a day name in its month field.

## Testing

* `UnixCronFormatTest` — the rewrite table, both day-field normalisation branches, whole-week ranges (bare and stepped), `L`/`W`/`#`/`H` inside the five-field layout, out-of-range days, format-mismatch errors in both directions, an undefined enum value, `TryParse`, the builder, time-zone composition, round-tripping the canonical form, every macro, `@reboot`, an unknown macro, and a 60-day fire-time equivalence sweep against the hand-written Quartz equivalent for 17 expression pairs.
* `UnixCronDifferentialTest` — the same crontab lines read by **NCrontab**, a Vixie-faithful five-field parser, over a 60-day window across a grid of 264 expressions. The unit project gains the `NCrontab` package reference (the version was already in `Directory.Packages.props`, for `Quartz.Benchmark`). 228 rows compare; the 36 that skip are day-of-week `7`, `1-7` and `0-7`, which NCrontab rejects and crontab has always taken for Sunday, and which are pinned directly instead. A guard test fails if the skip rate ever climbs, so the oracle cannot quietly stop testing anything.

## Decisions a reviewer should confirm

1. **`0 0 * * 5-1` → `FRI-MON`, i.e. a wrapping range.** NCrontab sorts the endpoints and reads it as `MON-FRI`; Vixie's own loop sets no bits at all for it. There is no portable answer, so `5-1` is excluded from the differential grid with that reason written down and pinned directly instead. The ratified table asks for the wrap, and the wrap is what a reader means.
2. **Day names, except inside a step.** `1` becomes `MON` and `5-1` becomes `FRI-MON`, because a name cannot be misread as the crontab digit it came from — but a step start stays numeric (`1/2` → `2/2`), since `MON/2` is rejected by #3614 and would make the rewriter emit an invalid expression.
3. **Whole-week ranges with a step.** The ratified rule "`end - start >= 6` → emit `*`" is right for a bare range but loses the phase when a step follows: `1-7/2` is Mon/Wed/Fri/Sun, and `*/2` is Sun/Tue/Thu/Sat. The stepped case therefore emits a full-week range anchored on the renumbered start — `0-6/2` → `1-7/2`, `1-7/2` → `2-1/2` — which the differential oracle confirms for the rows NCrontab can read.
4. **The month-field error message**, per the note above. It is a scope addition to what the design note asked for.

## Deviations from the design note

* §6.7's expected value for `0 0 13 * 5` is written as `0 0 13 * FRI`, which is five fields; the transcription dropped the seconds. Pinned as `0 0 0 13 * FRI`.
* §6.3's whole-week rule is applied as ratified only when the token has no step; see decision 3.
* The month-field error message is not in the note; see decision 4.

## Verification

`dotnet build Quartz.slnx` — clean. `Quartz.Tests.Unit` — 4,784 passed, 0 failed, 36 skipped. `Quartz.Tests.AspNetCore` — 548 passed. `dotnet fallout VerifyDocsSnippets` — up to date (102 markdown files). `dotnet fallout BenchmarkSmoke` — succeeded, 284 benchmarks.

`npm run docs:check-links` was **not** run: the repository's `node_modules` is not installed in this worktree and disk was tight. Every anchor added here (`#macros`, `#the-unix-five-field-form`, `#the-parser-refuses-what-it-used-to-ignore`, `#h-hash-for-load-distribution`) follows the same slug shape as anchors already linked in the same files; CI's docs job is the check.

## Documentation

* `cron-expressions.md` — a `Macros` section and a `The Unix five-field form` section, with three compiled `#region sample_*` blocks in `Quartz.Documentation.Samples`.
* `migration-guide.md` — the "is still rejected" paragraph corrected, two new sections, and the day-name-in-the-month-field note.
* `best-practices.md` — the "seconds first" trap paragraph now version-splits: 4.x has `CronFormat.Unix`, 3.x does not.

### DIFF
--- a/docs/documentation/best-practices.md
+++ b/docs/documentation/best-practices.md
@@ -488,7 +488,13 @@ how the clock and the zone are two separate axes.
 
 One more cron trap, since it accounts for more "my job ran fifty times" reports in this project than
 daylight saving does: Quartz cron puts **seconds first**, so `* 0/5 * * * ?` means *every second of
-every fifth minute*, not every five minutes. That is `0 0/5 * * * ?`. See
+every fifth minute*, not every five minutes. That is `0 0/5 * * * ?`. The same shift is why a
+five-field line copied from a crontab is not a Quartz expression: prepending a `0` fixes the layout
+but not the day-of-week numbering, which is 0-6 from Sunday in crontab and 1-7 from Sunday here. On
+**4.x**, don't translate it by hand — `CronExpression.Parse(line, CronFormat.Unix)` and
+`CronScheduleBuilder.Create(line, CronFormat.Unix)` read the five-field form as written, renumbering
+included, and store the canonical Quartz spelling. On **3.x** there is no such reader, so the
+translation is manual and the day-of-week digit is the part to check. See
 [Cron Triggers (4.x)](/documentation/quartz-4.x/tutorial/crontriggers) and
 [Cron Triggers (3.x)](/documentation/quartz-3.x/tutorial/crontriggers).
 
--- a/docs/documentation/quartz-4.x/cron-expressions.md
+++ b/docs/documentation/quartz-4.x/cron-expressions.md
@@ -38,7 +38,8 @@ For easy generation of cron intervals using UI you can use some of these service
 - [Cron Expression Generator & Explainer](https://www.freeformatter.com/cron-expression-generator-quartz.html)
 - [CronMaker](http://www.cronmaker.com/)
 
-NOTE: There are many cron standards/implementations. The results from some generators may not always be correct for Quartz.NET
+NOTE: There are many cron standards/implementations. The results from some generators may not always be correct for Quartz.NET.
+A generator that emits the five-field Unix form can be read as written - see [The Unix five-field form](#the-unix-five-field-form).
 :::
 
 ## Special characters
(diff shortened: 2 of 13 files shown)


lahma:
Reviewed in full. Deviations ACCEPTED: (1) the §6.7 transcription fix is obviously right; (2) the stepped whole-week phase preservation (`1-7/2` → `2-1/2`, not `*/2`) is a genuine spec improvement — the note's rule lost the phase and the oracle confirms yours; (3) name-where-standalone/number-under-step is the correct uniform rule, especially since a name-with-step is now a rejected form; (4) the month-field day-name heuristic is RATIFIED — it fires only on input that was already an error and completes the teaching story the unreachability finding exposed; (5) the wrap-range oracle exclusion with the three-way disagreement documented in the array comment is the honest treatment. The H-in-raw-five-field limitation resting on CronScheduleBuilder composition is fine as documented; if #3592's ParseWithHash wants a CronFormat overload later, that is additive. Merging on green; whoever of #3619/#3620 lands second rebases the baseline.

Issue #3248, opened 2026-08-07
JobInterruptMonitorPlugin can interrupt a job execution that has not exceeded its max run time


Break-1T:
### Pre-flight checks

- [x] I have searched [open](https://github.com/quartznet/quartznet/issues) and [closed](https://github.com/quartznet/quartznet/issues?q=is%3Aissue+is%3Aclosed) issues and could not find a duplicate.
- [x] I have checked the [documentation](https://www.quartz-scheduler.net/documentation/index.html).
- [x] I can reproduce this against a recent release of Quartz.NET (or the latest pre-release).

### Describe the bug

`JobInterruptMonitorPlugin` interrupts by job key rather than by fire instance id, so when a monitor elapses it cancels whichever execution of that job happens to be running at that moment — not necessarily the one the monitor was created for.

Two separate situations make this observable:
1. Two concurrent executions of the same job. The monitor for the one that exceeded its limit also cancels the other one.
2. A vetoed fire. Its monitor is never cancelled, because the only cancellation point is `TriggerComplete`, which a vetoed fire never reaches. The monitor survives for the lifetime of the process and then interrupts a completely unrelated, healthy execution of the same job.

The second case is the damaging one: every vetoed fire leaves another monitor behind, so on a scheduler with a trigger listener that vetoes at all, a job eventually stops being able to complete at all, and only a restart clears it. It presents itself as `Interrupting Job as it ran more than the configured max time` logged against a run that started a fraction of a second earlier.

### Steps to reproduce

Console app, `Quartz` + `Quartz.Plugins` 3.19.1, no persistent store:

``` csharp
using System.Collections.Concurrent;
using Quartz;
using Quartz.Plugin.Interrupt;

var builder = SchedulerBuilder.Create();
builder.UseDefaultThreadPool(x => x.MaxConcurrency = 5);
builder.UseJobAutoInterrupt(o => o.DefaultMaxRunTime = TimeSpan.FromMinutes(5));

var scheduler = await builder.BuildScheduler();
scheduler.ListenerManager.AddTriggerListener(new VetoOnDemandListener());
await scheduler.Start();

var job = JobBuilder.Create<SlowJob>()
    .WithIdentity("slow-job")
    .UsingJobData(JobInterruptMonitorPlugin.JobDataMapKeyAutoInterruptable, "true")
    .StoreDurably()
    .Build();

await scheduler.AddJob(job, replace: true);

// Scenario A: two concurrent fires. Only the first exceeds its limit.
[...]

PR #3621, opened 2026-08-31
A job's timeout is middleware with a budget it declares itself

### LINKED ISSUE #3588: A job timeout is middleware with typed configuration, not stringly plugin keys

### LINKED ISSUE #3604: The middleware contract's documentation tells the truth before it freezes
The `IJobExecutionMiddleware` contract was re-read line by line before the freeze and holds — but four true facts live nowhere, and one documented sentence is false. No API changes; this is the paperwork that stops beta questions.

- **`MergedJobDataMap`'s XML doc says mutating it "typically results in an illegal state" — that is Java-parity folklore and untrue of this code.** The map is a private merged copy; mutating it is safe for the whole firing and persists nothing — and it is the one middleware-to-listener channel there is. Correct the doc; it freezes otherwise.
- **Middleware is resolved once, from the root provider.** A scoped constructor dependency throws under `ValidateScopes` and becomes a silent captive dependency in production; the docs warn about fields only, and the name "middleware" imports per-request expectations from ASP.NET Core. One sentence on the middleware page and `AddJobMiddleware` docs.
- **A `ConfigureAllQuartzSchedulers` middleware always composes *inside* an `AddQuartz` one, whichever call came first** (`QuartzServiceCollectionExtensions.cs:452-457`) — deliberate (a library's outbox scope belongs inside the app's tenant scope) and pinned by nothing. Document it and add the test.
- **Calling `next` twice is legal**; say so in the delegate docs.

Found by the alpha.5 freeze re-read (design note `alpha5-note-mw-aot.md`); the same pass proved the contract itself needs no change and re-scoped #3341.

lahma:
Closes #3588. Closes #3604.

Two issues, one PR: they edit the same documentation pages and the same
`IJobExecutionMiddleware` surface, and a single writer avoids the conflict.

## #3588 — the timeout middleware

`JobInterruptMonitorPlugin` is deleted, together with `UseJobAutoInterrupt`,
`JobAutoInterruptOptions` and the two `JobDataMap` keys it reserved
(`"AutoInterruptable"`, `"MaxRunTime"`). `AddJobTimeout` in the core package
replaces it.

```csharp
q.AddJobTimeout(TimeSpan.FromMinutes(5));   // scheduler-wide default
q.AddJobTimeout();                          // only jobs that declare a budget
```

```csharp
[JobTimeout("00:00:30")] public sealed class ReportJob : IJob { … }
[JobTimeout("00:00:00")] public sealed class NightlyRebuildJob : IJob { … }   // no timeout
```

**Interrupt, not a linked CTS.** The middleware never replaces the token it was
handed. On expiry it calls the public
`context.Scheduler.InterruptFireInstance(context.FireInstanceId)`, which cancels
the *context's own* source. A linked source would have handed the job a token
that is no longer `IJobExecutionContext.CancellationToken`, and the run shell
classifies a cancellation by asking whether the context's token was signalled —
so the `OperationCanceledException` would have come out as an unhandled failure
with a `SchedulerError` notification behind it. Going through the public
interrupt path means the token, `ISchedulerListener.JobInterrupted` (#3248's
per-fire-instance work) and the log all behave exactly as they do for an
operator's interrupt.

**A timeout is a retryable failure, deliberately.** The default classification
of an interrupt is *success-shaped*: `jobExEx` stays null, the trigger advances
as though the job had finished, and the retry policy is never consulted. That is
right for an operator who asked a job to stop and wrong for a job that ran out of
time, so the middleware raises a `JobExecutionException` naming the budget. A
timed-out firing is then an ordinary failure — visible to listeners, counted as
an error, and retried by the trigger's `RetryPolicy` with the whole budget again
on each attempt. `ATimedOutFiringIsRetriedByTheTriggersRetryPolicy` pins that end
to end through a real scheduler. A job that swallows its cancellation and returns
normally is reported as timed out too; an exception it threw that is *not* a
cancellation is left alone, because it says more about what went wrong than the
timeout does.

**Registration shape.** `AddJobTimeout(TimeSpan? defaultTimeout = null)` is an
extension on `IQuartzBuilder` in `QuartzBuilderExtensions`, mirrored on
`QuartzSchedulerBuilder`. Not a fourth member beside the `AddJobMiddleware`
trio, and not an options object: the trio takes *a middleware*, and this is that
call with the built-in one and the only setting it has, which is exactly what
that file says extensions are for ("each is a composition of what the interface
already offers"). It registers at the point of the call, so the timeout takes its
place in the chain like any other stage. A richer form — a per-`JobKey` table,
say — is addable later as an overload without breaking anything.

**Deviation from the brief, and it needs a reviewer's opinion.** The brief asked
for the per-job override to be a `JobBuilder` member plus an `OneOffJobOptions`
property; #3588 itself offered "an attribute (`[JobTimeout("00:05:00")]`) **or**
a `JobBuilder`/`OneOffJobOptions` member". I took the attribute, because the
other two have nowhere to put the value that survives: a builder member has to
land on the job detail, which means a new `IJobDetail` member, a `JOB_DETAILS`
column across six dialects, a generated migration, a `schema-changes.md` entry
and a wire field on the HTTP DTO — a schema change in an alpha, for a feature
whose issue never asked for one. The alternative, a reserved `JobDataMap` key
written by a typed builder call, is the thing being retired wearing a hat, and
the brief ruled it out. The attribute is what `[DisallowConcurrentExecution]` and
`[PersistJobDataAfterExecution]` already are — job-level policy read off the type
— and it travels through every store, every wire format and every way of
scheduling for free. **What it cannot do is give two jobs of the same type
different budgets**, or set a budget on a job type you do not own; both remain
addable, the first by a detail member and a column, the second by an overload
taking a per-key table.

**A job that ignores its token cannot be stopped.** Cancellation is cooperative.
Such a job runs to completion holding its slot and is reported as timed out only
when it returns. Said in the attribute's remarks, in `AddJobTimeout`'s, and in
the tutorial; `CA2016` is the mitigation.
`AJobThatIgnoresItsTokenIsStillReportedAsTimedOut` is the test.

**Public API removal.** `PublicApiTest_Quartz.Plugins` loses
`JobInterruptMonitorPlugin`, `JobAutoInterruptOptions` and `UseJobAutoInterrupt`;
`PublicApiTest_Quartz` gains `JobTimeoutAttribute` and the two `AddJobTimeout`
members. `LogEventCatalogTest_Quartz.Plugins` loses the eight `6400`-range
events; `LogEventCatalogTest_Quartz` gains three in a new `1090-1099` range.
All four baselines were reviewed as diffs and accepted, and the migration guide
carries the same delta: a new section with a before/after table, an appendix row,
and the removal of three now-false paragraphs.

## #3604 — the contract paperwork

No API changes.

* **`MergedJobDataMap`'s "illegal state" sentence is deleted.** It was
  Java-parity folklore. The map is this firing's private merged copy; writing to
  it is safe, reaches the job, the rest of the pipeline and the listeners, and
  persists nothing. It is the one middleware-to-listener channel there is — an
  `AsyncLocal` cannot be one, because an async method restores its caller's
  execution context before the run shell notifies anybody — and that sentence was
  what stood in front of it.
  `AMiddlewareCanPassAValueToAListenerThroughTheMergedJobDataMap` pins both
  halves: the listener sees the value, and the job's stored map does not.
* **Root-provider resolution is documented** on `IJobExecutionMiddleware`, on
  `AddJobMiddleware`, on the tutorial page (with an `IServiceScopeFactory`
  sample) and in the library how-to. A scoped constructor dependency throws under
  `ValidateScopes` and is a captive dependency in production; the docs previously
  warned about *fields* only, and the name imports per-request expectations from
  ASP.NET Core.
* **`ConfigureAllQuartzSchedulers` composes inside `AddQuartz`,** whichever call
  came first, and it is now pinned in both orders by
  `AConfigureAllMiddlewareComposesInsideAnAddQuartzOne`. Documented on the
  interface, on `ConfigureAllQuartzSchedulers` itself, on the tutorial page and
  in the library how-to, where it matters most.
* **Calling `next` twice is legal** — one line on the delegate, one on the page.

## For a reviewer to decide

1. The attribute-versus-builder-member call above.
2. `AddJobTimeout` called twice puts two timeout stages in the chain, like any
   repeated `AddJobMiddleware`. Left as is rather than made idempotent.
3. `ISchedulerListener.JobInterrupted` still carries no fire instance id, so a
   listener cannot tell a timeout from an operator's interrupt. The design note
   flagged it as addable post-freeze via a DIM overload; not added here, since
   nothing in this feature needs it.

## Verification

```
dotnet build Quartz.slnx                    Build succeeded. 0 Warning(s) 0 Error(s)
dotnet test src/Quartz.Tests.Unit           Passed! - Failed: 0, Passed: 4800, Skipped: 36, Total: 4836
dotnet test src/Quartz.Tests.AspNetCore     Passed! - Failed: 0, Passed:  548, Skipped: 0, Total:  548
dotnet fallout VerifyDocsSnippets           Succeeded (102 markdown files checked, up to date)
Quartz.Benchmark --smoke                    exit 0 (284 benchmarks)
```

Rebased onto `bf12a6531` after #3594, #3618, #3623 and #3619 landed: #3594's rewrite of
`PluginConfigurationExtensions` collided with the `UseJobAutoInterrupt` deletion
(resolved by deleting it in its new shape, and by dropping the plugin from
#3594's `EveryShippedPluginIsConfiguredFromItsOwnRegistration`); #3618 moved the
`Quartz.Plugins` API baseline and the migration guide's removed-type count, which
is 97 on main and 98 here; and #3623 reordered the plugin tables in
`quartz-plugins.md`, `src/Quartz.Plugins/README.md` and the migration guide's
options table, where JSON now leads. Every conflict was a documentation table
whose `JobInterruptMonitorPlugin` row this PR deletes; the four Verify baselines
merged textually and were then proved by running the Verify tests, which produced
no `.received.txt`. Every number above is from the rebased tree.

Integration tests were not run: no Docker daemon was available in this
environment. `Quartz.Tests.AOT` does not exist in this tree; the trimming leg is
`Quartz.Trimming.Canary`, which was not run (`PublishAot` needs more disk than
was free) — nothing here adds reflection, and the one new reflective read
(`JobTimeoutAttribute` off a job type) goes through `JobTypeInformation`, which
already carries the annotations for the other two job attributes.


lahma:
Reviewed in full. RULING on the flagged decision: `[JobTimeout]` attribute-only is ACCEPTED as the frozen shape — the issue's own list named the attribute first; a JobBuilder/OneOffJobOptions member whose value silently fails to survive the store round-trip would be a lying API, and the honest alternatives (a detail member + six-dialect column, or the reserved key being retired) are respectively out of proportion and self-defeating. The two limitations (no per-instance budgets, no budgets on unowned types) are recorded and additively fixable. The rethrow-as-JobExecutionException making a timeout retryable — pinned through a real scheduler — is the load-bearing behavior and it is right. The #3594 test-assertion removal during rebase was the correct resolution for a deleted plugin and is hereby the reviewer's glance it asked for. AddJobTimeout as extension per the extensions-class's own remarks: correct. Merging on green after the baseline dance resolves.

PR #3622, opened 2026-08-31
A job store is told how to store, in the words the scheduler already uses

lahma:
Part 2 of #3596 — the `IJobStore` options alignment, the `IsTransient` hook the internalization needs, and the last of the design note's small corrections.

Deliberately **not** stacked on #3616: the two touch disjoint files, so each baseline diff reviews on its own and either can merge first.

## `IJobStore`'s three storing members take options (I2)

Four members of `IJobStore` decide whether an existing item is over-written. `AddCalendar` took `AddCalendarOptions`; `AddJob`, `AddTrigger` and `ScheduleJobs` took a bare `bool replace`. One interface, one decision, two spellings — and the three with the primitive are the three every custom store has to implement. After the freeze that is permanent.

```diff
- ValueTask AddJob(IJobDetail job, bool replace, CancellationToken cancellationToken = default);
- ValueTask AddTrigger(IOperableTrigger trigger, bool replace, CancellationToken cancellationToken = default);
- ValueTask ScheduleJobs(IReadOnlyDictionary<…> triggersAndJobs, bool replace, CancellationToken cancellationToken = default);
+ ValueTask AddJob(IJobDetail job, AddJobOptions options = default, CancellationToken cancellationToken = default);
+ ValueTask AddTrigger(IOperableTrigger trigger, AddTriggerOptions options = default, CancellationToken cancellationToken = default);
+ ValueTask ScheduleJobs(IReadOnlyDictionary<…> triggersAndJobs, ScheduleJobOptions options = default, CancellationToken cancellationToken = default);
```

**One deviation from the design note, and it makes the change smaller.** The note specs a new `ScheduleJobsOptions`. `ScheduleJobOptions` already exists — `IScheduler.ScheduleJobs` takes it, with exactly the shape wanted — so inventing a second would have been a near-duplicate type differing by one letter. Only `AddTriggerOptions` is new. That leaves the scheduler forwarding the record it was handed rather than unpacking it into a flag and the store rebuilding it.

`AddJobOptions.StoreNonDurableWhileAwaitingScheduling` stays a scheduler-level rule, applied before the store is called, and `IJobStore.AddJob`'s doc now says so rather than leaving a store author to guess. The default is `default`, not `Replacing` — today's `false` for a caller that omits it, so the source break is confined to callers that passed `true`.

`AddTriggerOptions` has one call site, because `IScheduler` has no `AddTrigger`: a trigger reaches a scheduler through `ScheduleJob`, and `IJobStore.AddTrigger` is the storage half of that. Its remarks say so.

### Coverage

- `DelegatingJobStoreForwardingTest` is new: the three members hand their options on unchanged, the omitted argument arrives as `default` rather than as `Replacing`, and the token travels. Forwarding a `bool` is hard to get wrong; forwarding a record is not — a wrapper that rebuilt it, or that passed `default` because the parameter has one, turns "replace what is there" into "throw if it is there", and only for applications that wrap their store.
- `OptionsShorthandTest` gains `AddTriggerOptions.Replacing`, shape and behaviour, exercised against the store because that is where its only call site is.
- **`DeferredScheduler` and `DelegatingScheduler` need no change and get no new lines**: both already take `AddJobOptions` / `ScheduleJobOptions`, because `IScheduler`'s signatures are untouched. The design note listed them as a coverage risk; there is nothing there to cover.

## `AdoJobStoreOptions.IsTransient` — the seam PR-C removes an override for

`AdoJobStoreBase.IsTransient` is a `protected virtual` that `operations.md` teaches as the seam for a driver whose retryable failure none of the built-in signals recognise. PR-C makes the class internal, so the seam needs a replacement that is not inheritance — and per AGENTS.md a shipped component is configured through its options type and only there.

```csharp
public Func<Exception, bool>? IsTransient { get; set; }
```

Consulted first, and **only additive**: `false` means what having no predicate means, so it cannot make the store stop retrying something it already retries. A driver that misreports a condition the built-in list covers is a bug to fix in `TransientErrorDetector`, not to switch off per application. Three tests, on a store that does *not* override `IsTransient` so the classification under test is the real one: the predicate makes an unrecognised failure retryable, the same failure without it is not retried, and a predicate answering `false` does not stop a timeout being retried.

It is code-only, like `DataSourceOptions.DataSourceFactory` — no configuration source can describe a predicate over an exception's type, error number and inner exceptions — so it joins that list in `ConfigurationBindingIsSourceGeneratedTest`, which is the test that would otherwise, correctly, fail it as a setting that binds from nowhere. `configuration/reference.md` documents it. `operations.md` still teaches the override, and moves in PR-C when the override goes.

## `MySQLDelegate.GetCountMisfiredTriggersInStateSql` calls `base`

Three of its overrides splice the same `FORCE INDEX` hint. Two ask `base` for the statement and transform the string; the third reached `StdAdoConstants` directly, which is **internal** — so a delegate outside this assembly can write the first shape and cannot write the third. Same statement, same hint, no behaviour change, and `MySQLDelegate` is now proof that the public kit is sufficient rather than an example that only compiles inside Quartz.

## Baseline

Reviewed and accepted; three groups, nothing else moved:

- `+ AddTriggerOptions`, two members, in the `Quartz` namespace
- `+ AdoJobStoreOptions.IsTransient`
- the three signatures, on `IJobStore`, `AdoJobStoreBase`, `DelegatingJobStore` and `RAMJobStore`

Migration-guide rows in the same PR: the "same verbs" table rows, a new **The store takes the same options** subsection with the 3.x to 4.x call table, `AddTriggerOptions` added to the shorthand table and to the `Quartz`-namespace collision listing — which is mechanically derived, so it re-flows and its count moves from ninety-three to ninety-four — and the two paragraphs that said the store *keeps* its `bool replace` are now the ones that say it does not.

## For a reviewer to decide

- **`AddTriggerOptions` is a public type for one interface member.** The alternative was leaving `AddTrigger` the odd one out, which is what the issue asked to fix. Reusing `AddJobOptions` was rejected: wrong noun, and it carries a flag with no meaning for a trigger.
- **`AdoJobStoreOptions.IsTransient` is additive-only by design.** If you want a predicate that can *veto* the built-in classification, say so now — it is the one thing about the shape that cannot be changed later without breaking someone.
- The `protected` `AdoJobStoreBase.AddJob(conn, …)` and `AddTrigger(conn, …)` twins keep their `bool`. They are machinery, and PR-C makes them internal.

## Verification

```
dotnet test src/Quartz.Tests.Unit        → Passed! Failed: 0, Passed: 4443, Skipped: 0
dotnet test src/Quartz.Tests.AspNetCore  → Passed! Failed: 0, Passed: 548, Skipped: 0
dotnet fallout VerifyDocsSnippets        → Succeeded, 102 markdown files checked
dotnet fallout BenchmarkSmoke            → Succeeded
```

`Quartz.Tests.Integration` compiles — its `replace:` call sites moved with the rest — but was not run: it needs a Docker daemon, which this machine does not have. `npm run docs:check-links` was not run locally, there being no `node_modules` in the worktree, so the one new heading anchor rests on CI.

Rebased onto `577531990` after #3618 merged; the baseline merged cleanly and the build and unit suite above are from there.

Part 2 of #3596.

### DIFF
--- a/docs/documentation/quartz-4.x/configuration/reference.md
+++ b/docs/documentation/quartz-4.x/configuration/reference.md
@@ -167,6 +167,7 @@ services.AddQuartz(q => q.UsePersistentStore(store =>
 | `MaxTransientRetries` | int | `3` | How many times a transient failure such as a deadlock is retried. Transient means the driver's own `DbException.IsTransient`, a SQLSTATE in class `40` — the standard's "transaction rollback", covering a serialization failure or a deadlock whichever provider reports it, with `40002` excepted because a deferred constraint violation fails identically on every retry — SQL Server's transient error numbers, SQLite's busy and locked codes, or a timeout. |
 | `TransientRetryInterval` | TimeSpan | `00:00:01` | Delay between transient retries. |
 | `RetryableActionErrorLogThreshold` | int | `4` | How many consecutive failures before they are logged as errors. |
+| `IsTransient` | `Func<Exception, bool>?` | `null` | An extra answer to the question above, for a driver that reports a retryable condition none of those signals recognise. Consulted first and only additive — returning `false` falls through to the built-in list, so it cannot make Quartz stop retrying something it already retries. The exception handed over is the store's own, so reach the driver's with `GetBaseException()`. Code only; there is no `quartz.*` key for a delegate. |
 | `UseDbLocks` | bool | `false` | Uses database row locks. Required for clustering, and implied by `UseClustering()`. |
 | `LockOnInsert` | bool | `true` | Takes a lock when inserting rows. |
 | `AcquireTriggersWithinLock` | bool | `false` | Acquires triggers inside the database lock. |
--- a/docs/documentation/quartz-4.x/migration-guide.md
+++ b/docs/documentation/quartz-4.x/migration-guide.md
@@ -7210,9 +7210,9 @@ if you only call `IScheduler`, nothing here affects you.
 | `IJobStore` in 3.x | `IJobStore` in 4.x |
 |---|---|
 | `StoreJobAndTrigger(job, trigger)` | `ScheduleJob(job, trigger)` |
-| `StoreJobsAndTriggers(triggersAndJobs, replace)` | `ScheduleJobs(triggersAndJobs, replace)` |
-| `StoreJob(job, replaceExisting)` | `AddJob(job, replace)` |
-| `StoreTrigger(trigger, replaceExisting)` | `AddTrigger(trigger, replace)` |
+| `StoreJobsAndTriggers(triggersAndJobs, replace)` | `ScheduleJobs(triggersAndJobs, ScheduleJobOptions?)` |
+| `StoreJob(job, replaceExisting)` | `AddJob(job, AddJobOptions?)` |
+| `StoreTrigger(trigger, replaceExisting)` | `AddTrigger(trigger, AddTriggerOptions?)` |
 | `RemoveJob(key)`, `RemoveJobs(keys)` | `DeleteJob(key)`, `DeleteJobs(keys)` |
 | `RemoveTrigger(key)`, `RemoveTriggers(keys)` | `DeleteTrigger(key)`, `DeleteTriggers(keys)` |
 | `RetrieveJob(key)` | `GetJob(key)` |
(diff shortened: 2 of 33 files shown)


lahma:
Reviewed in full. RULING on the flagged decision: `IsTransient` stays **additive-only, as shipped**. The asymmetry decides it: a veto capability can be added later additively (a richer delegate taking the built-in verdict, or a second option), while replace-semantics could never be walked back without breaking every subscriber who wanted add-only — and the cost of no-veto today is bounded at MaxTransientRetries extra attempts of a genuinely fatal statement, which is noise, not damage. The reference-table row already says 'only additive' plainly; keep it. ScheduleJobOptions reuse, the Delegating coverage reasoning, and the mechanical-block reflow are all accepted. Merging on green; PR-C proceeds after this lands.

PR #3623, opened 2026-08-31
XML scheduling is maintained, not extended

### LINKED ISSUE #3601: XML scheduling is maintained, not extended

lahma:
`quick-start.md` taught `UseXmlSchedulingConfiguration` as the first thing a new user configures,
while the XML format it points at understands three trigger kinds and the JSON one understands four
plus retry policies and execution groups. This makes the demotion deliberate, in three parts.

## The XSD is frozen, and says so

The schema is untouched — this half is prose. `job_scheduling_data_2_0.xsd` declares `simple`, `cron`
and `calendar-interval` and will not gain a fourth kind or the trigger settings written since, and the
plugin's documentation page, the package readme and the migration guide now say that, with a table of
what each format can express. The list was checked against `JsonSchedulingDataProcessor`, which is why
it names **daily time interval, retry policy and execution group** and not "recurrence": recurrence
triggers are in neither format, and the page says so rather than implying JSON has them.

Not in scope, and stated in each place: XML scheduling is not deprecated and is not going away in 4.x.

## The quick start teaches JSON

`sample_quick_start_host` registers `UseJsonSchedulingConfiguration`, and the page prints the
`quartz_jobs.json` it reads — a job and a ten-second simple trigger, mirroring what the XML line
scheduled. `JsonSchedulingDataProcessorTest.TheQuickStartFileIsReadable` parses that exact file, with
the job type as the only substitution, so a documented file that does not parse fails the build.

The other pages the issue names follow: `packages/quartz-plugins.md` leads with the JSON plugin and
binds `Quartz:Json` in its options example, `packages/microsoft-di-integration.md` and
`multi-tenancy.md` name the JSON plugin — what those two demonstrate is a plugin and a property bag,
not a file format. `quartz-3.x/` is untouched.

## The plugin twins have one surface (breaking)

Both plugins now publish the two constructors and what `ISchedulerPlugin` and `IFileScanListener` ask
for, and nothing else:

| Member | Was | Now |
|---|---|---|
| `XmlSchedulingDataProcessorPlugin.ProcessFile(string, CancellationToken)` | public on XML, private on JSON | private on both |
| `JsonSchedulingDataProcessorPlugin.Shutdown(CancellationToken)` | public on JSON, absent on XML | absent on both — the interface default does nothing, which is what it did |
| `FileNames`, `ScanInterval`, `FailOnFileNotFound`, `FailOnSchedulingError` | public get-only, internal setters | internal |
| `Name`, `Scheduler` | public get-only | internal |

Fifteen lines leave the `Quartz.Plugins` baseline and none arrive; the migration guide carries a row
for each, in a new section and in the appendix.

**`quartz.plugin.<name>.*` still works, and is tested.** `PropertyBinder` binds
`Public | NonPublic` and calls `GetSetMethod(nonPublic: true)`, so an internal property is still
written by a flat key — `SchedulingDataPluginTwinsTest.FlatKeysStillConfigureBothTwins` drives all
four settings through `AddQuartz(NameValueCollection)` and `SchedulerPluginFactory` for both plugins.
`TheTwinsPublishTheSameMembers` pins the aligned surface so it cannot drift again.

`FileNames` versus `Files` is settled the way the spec asks and both twins now agree: the plugins keep
a delimited `FileNames` string because `quartz.plugin.<name>.fileNames` is one string, and
`FileSchedulingOptions.Files` stays the `List<string>` code and configuration binders say. The join is
in `PluginConfigurationExtensions`, unchanged.

## For the reviewer to decide

- The JSON section now sits **above** the XML one on the plugins page. That is an editorial signal, not
  a technical need; the anchors are unchanged either way.
- `JsonSchedulingDataProcessorPlugin`'s internals were `Task`-returning; they are `ValueTask` now, and
  `JobFile.Initialize` disposes asynchronously as the XML twin's does. Behaviour-neutral, and it is
  what makes "one surface" true below the surface too.
- Left alone deliberately: the twins still split `FileNames` differently — XML with `Split(',')` and
  `TrimStart`, JSON with `RemoveEmptyEntries` and `Trim`, so a trailing comma throws on XML and is
  ignored on JSON. It is a pre-existing difference in behaviour rather than surface, and fixing it
  belongs in its own change.
- `src/Quartz.Examples/12_ConfigureJobSchedulingByUsingXmlConfigurations` and
  `Quartz.Examples.AspNetCore` still demonstrate XML, which is correct: the format is supported, and
  the example is what a reader with an existing `quartz_jobs.xml` looks for.

## Verification

- `dotnet build src/Quartz.Plugins` — succeeded, 0 warnings (trim/AOT analyzers included).
- `dotnet test src/Quartz.Tests.Unit` — **4436 passed, 0 failed**.
- `dotnet fallout VerifyDocsSnippets` — succeeded, 102 markdown files up to date.
- The `Quartz.Plugins` baseline was regenerated by the test and accepted from `.received.txt`; the diff
  is 15 deletions and no additions.
- Not run, and why: the solution-wide build and the other test projects, on the disk budget this
  worktree was given — no project outside `Quartz.Plugins`, `Quartz.Tests.Unit` and
  `Quartz.Documentation.Samples` (all three built clean) names either plugin type, and
  `Quartz.Tests.AspNetCore` neither references `Quartz.Plugins` nor snapshots it. `BenchmarkSmoke` is
  untouched territory — no trigger, `TriggerHeader`, scheduler internal or serialization shape changed.
  `npm run docs:check-links` needs a `node_modules` this worktree has no room for; every new link was
  resolved by hand against the site's slugger (`@mdit-vue/shared`, which maps `.` and spaces to `-`).

Closes #3601

### DIFF
--- a/docs/documentation/quartz-4.x/configuration/json.md
+++ b/docs/documentation/quartz-4.x/configuration/json.md
@@ -223,6 +223,8 @@ All trigger types support these optional fields:
 | `Description` | Trigger description |
 | `Priority` | Trigger priority (integer) |
 | `CalendarName` | Calendar to apply |
+| `ExecutionGroup` | The trigger's [execution group](../tutorial/execution-groups.md) |
+| `RetryPolicy` | The trigger's [retry policy](../how-tos/retrying-failed-jobs.md) in its stored form, for example `fixed;3;00:00:30` |
 | `StartTime` | ISO 8601 start time (e.g., `"2024-01-01T00:00:00Z"`) |
 | `StartTimeSecondsInFuture` | Start time as seconds from now (mutually exclusive with StartTime) |
 | `EndTime` | ISO 8601 end time |
--- a/docs/documentation/quartz-4.x/migration-guide.md
+++ b/docs/documentation/quartz-4.x/migration-guide.md
@@ -3875,6 +3875,17 @@ A schema violation still throws `SchedulingDataValidationException` carrying eve
 `XmlSchedulingDataProcessorPlugin` still wraps whatever surfaces in a `SchedulerException`, so a
 plugin-based setup sees no change at all.
 
+**Nor will it change.** The schema is frozen at what it already says: `simple`, `cron` and
+`calendar-interval` triggers, and none of the trigger kinds or trigger settings written since. A
+daily time interval trigger, a [retry policy](#a-trigger-can-carry-a-retry-policy) and an execution
+group are expressible in the JSON format and are not expressible in XML, now or later — two file
+formats that both grow is two parsers and two schemas for one feature, and the XML one is the one that
+has to keep loading files written twenty years ago. It does: **XML scheduling is not deprecated and is
+not going away in 4.x**, and a schedule that only needs the three trigger kinds above never has to
+move. Write a new schedule as JSON, and move an XML one when it needs something the schema cannot
+spell. `UseJsonSchedulingConfiguration` takes the same `FileSchedulingOptions` as its XML twin, so the
+registration is one word different.
+
 ### The shipped plugins are sealed
 
 `LoggingJobHistoryPlugin`, `LoggingTriggerHistoryPlugin`, `StructuredLoggingJobHistoryPlugin`,
--- a/docs/documentation/quartz-4.x/migration-guide.md
+++ b/docs/documentation/quartz-4.x/migration-guide.md
@@ -670,7 +670,7 @@ assignable on a live instance with nothing saying which wins or when. The compon
 | `RedisLockHandler.RedisConfiguration`, `.KeyPrefix`, `.LockTimeToLive`, `.LockRetryInterval` | `UseRedisLockHandler(o => …)` |
 | The history plugins' message templates | `UseJobHistoryLogging(o => …)`, `UseTriggerHistoryLogging(o => …)`, and now `UseStructuredJobLogging(o => …)` / `UseStructuredTriggerLogging(o => …)` |
 | `JobInterruptMonitorPlugin.DefaultMaxRunTime` | `UseJobAutoInterrupt(o => …)` |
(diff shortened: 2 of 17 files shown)


lahma:
Reviewed in full. Deviations ACCEPTED: (1) the corrected capability list — writing 'XML lacks recurrence' when JSON lacks it too would have been the exact inventory-negative trap this repo has been burned by; stating it separately is right, and the gap itself now has its own issue; (2) internal-not-removed for the bridge-noise properties, with the PropertyBinder nonPublic reach PINNED by a test, is precisely the safe shape; (3) FileNames staying as the internal delimited string follows from the flat-key format; (4) the ExecutionGroup/RetryPolicy documentation rows were owed and the freeze claim needed them. The twins' FileNames parsing difference is noted as pre-existing — leave it. Merging on green after the #3588 baseline dance resolves.

PR #3627, opened 2026-08-31
The dashboard's client speaks the scheduler's vocabulary, and the trigger serializer seam is settled

### LINKED ISSUE #3600: The serializer packages expose options, and the dashboard speaks the scheduler's vocabulary
Two surface trims outside the core assembly, both breaking-or-never.

**The Newtonsoft package exposes options, not machinery.** `Quartz.Serialization.Newtonsoft` mirrors the in-core System.Text.Json type tree — `ICalendarSerializer`/`CalendarSerializer<T>`, `ITriggerSerializer`/`TriggerSerializer<T>`, five concrete trigger serializers, the registry — twelve types, ~50 members, same names, second assembly. Public stays: the options types, the `UseNewtonsoftJsonSerializer` entry points, and the registry members an external calendar/trigger serializer genuinely plugs into (the STJ side already proved which those are — mirror its answer, not its whole tree). The rest goes internal. Same audit runs over the STJ tree in core for members nothing external can use.

**The dashboard's client speaks the scheduler's vocabulary.** `IQuartzApiClient` (public by SignalR strong-naming necessity, so it freezes too) renames `IScheduler` operations wholesale: `StartScheduler`/`StandbyScheduler`/`ShutdownScheduler` for `Start`/`Standby`/`Shutdown`, `InterruptJob` for `Interrupt`, `TriggerJob` **and** `TriggerJobWithData` where `IScheduler` has one method with an optional map, `GetJobs`/`GetTriggers` for the `Query*` family. Align the verbs with `IScheduler` (DTO suffixes stay; `PagedResult<>` stays); `IDashboardHistoryStore`'s third verb set gets the same pass. `DelegatingScheduler`/`DelegatingJobStore`/`HttpScheduler` visibility: internalize whatever nothing public constructs, keep what the HTTP-client story needs.

If the merge chain runs long, the vocabulary alignment is the campaign's designated first cut — the fallback is an AGENTS.md row ratifying the dialect, not silence.

lahma:
Closes #3600 — the dashboard half of the issue, plus a ratification and a docs correction where the serializer half would have gone.

Rebased onto `bf12a6531` (#3619).

## The dashboard's client speaks the scheduler's vocabulary

`IQuartzApiClient` now spells an operation the way `IScheduler` spells it. DTO suffixes and `PagedResult<T>` returns are unchanged.

| Before | After |
|---|---|
| `StartScheduler`, `StandbyScheduler`, `ShutdownScheduler` | `Start`, `Standby`, `Shutdown` |
| `InterruptJob` | `Interrupt` (`InterruptFireInstance` unchanged) |
| `TriggerJob(name, key)` **and** `TriggerJobWithData(name, key, JobDataMap)` | one `TriggerJob(name, key, JobDataMap? jobDataMap = null)` |
| `GetJobs`, `GetJobGroups`, `GetTriggers`, `GetTriggerGroups`, `GetFireInstances`, `GetClusterNodes` | `QueryJobs`, `QueryJobGroups`, `QueryTriggers`, `QueryTriggerGroups`, `QueryFireInstances`, `QueryClusterNodes` |
| `GetJob` → `JobDetailDto` | `GetJobDetail` |
| `GetJobTriggers` | `GetTriggersOfJob` |
| `GetHistory`, `GetMisfires` | `QueryExecutions`, `QueryMisfires` |
| `IDashboardHistoryStore.Add`, `GetPage`, `GetMisfires` | `AddExecution`, `QueryExecutions`, `QueryMisfires` (`AddMisfire`, `CountMisfires` unchanged) |

Beyond the issue's enumerated list: `GetClusterNodes` → `QueryClusterNodes` (`IScheduler.QueryClusterNodes`, same return shape), `GetJob` → `GetJobDetail` (AGENTS.md: the scheduler's noun is `JobDetail`) and `GetJobTriggers` → `GetTriggersOfJob` (`IScheduler.GetTriggersOfJob`). **Not** renamed: `GetSchedulers` / `GetScheduler`, which read the container's registrations and have no scheduler counterpart, and `GetCalendarNames`, which drains *every* page of `IScheduler.QueryCalendarNames` — naming it `Query*` would promise a paged query it does not take.

### Nullability, unified honestly

`QueryExecutions`, `QueryMisfires` and `CountMisfires` are non-nullable now. The `?` existed because the deleted HTTP-backed client turned a 404 into "this data source keeps no history"; `IDashboardHistoryStore` returns non-nullable pages and an `int`, so in-process that answer never existed. A store holding nothing returns an empty page and a count of zero.

`CountMisfires` was not in the issue's list — it carried the same dead `null` for the same dead reason. Two consequences:

- the History page's "Execution history is unavailable for the current API backend" branch is gone (dead once the return is non-null), replaced by the existing empty-page message. `AStoreThatKeepsNoHistorySaysSoInsteadOfShowingAnEmptyPage` became `AStoreHoldingNothingShowsAnEmptyPageRatherThanAFault`.
- `ADataSourceThatKeepsNoMisfiresSaysSoRatherThanReportingZero` became `ATileOverNoSchedulerSaysSoRatherThanReportingZero`: the overview's dash-not-zero distinction survives, but the state that produces it is "no scheduler has answered yet" rather than "the data source keeps no feed".

`IQuartzDashboardHubClient` is untouched: its `Task` returns and its public visibility are SignalR's requirement, and `QuartzDashboardHubClientProxyTest` guards them.

`DelegatingScheduler`, `DelegatingJobStore` and `HttpScheduler` (the issue's third dashboard clause) stay public — verified, not skipped: the first two are taught as derivation seams in `how-tos/custom-job-store.md` and the migration guide, and `HttpScheduler` is the HTTP client package's product type.

## The trigger serializer seam is settled, and the docs now say what the code does

The issue also asked for the built-in trigger serializers to go internal in both packages. **They stay public and unsealed.** Commit `5f442177d` ratified that seam deliberately three weeks ago, the persistence-delegate how-to teaches it ("the built-in serializers are public and unsealed on purpose"), and the migration guide gives it as the reason `AddTriggerSerializer<TTrigger>` keeps its `ITriggerSerializer` parameter. `Quartz` and `Quartz.Serialization.Newtonsoft` therefore have **no public-API delta in this PR**: `PublicApiTest_Quartz.verified.txt` and `PublicApiTest_Quartz.Serialization.Newtonsoft.verified.txt` are byte-identical to `main`, and `Quartz.Serialization.Newtonsoft` keeps its twelve public types.

Two things go with that ruling:

**A unit-level guard for the seam.** `BuiltInTriggerSerializerDerivationTest` round-trips a `ReportTrigger : CronTriggerImpl` (with `HasAdditionalProperties`) through a serializer deriving from the built-in `CronTriggerSerializer` in *each* package, calling `base.SerializeFields` / `base.DeserializeFields`. Until now the seam was only exercised by the ADO smoke tests, which need a database; making a built-in serializer internal or sealed will now fail a 100 ms unit test instead.

**The stale "sealed" claims are corrected — the docs were wrong, not the types.** All five `*TriggerImpl` types are `public class` on `main` and in the API baseline; three of them were sealed during 4.x's development and were reopened precisely so the serializer pairing works. Three places still said otherwise:

- `migration-guide.md`, "The shipped implementations are sealed" — the trigger row is removed, and a paragraph now states the pairing (derive from a built-in trigger, derive from its built-in serializer) and why the three were reopened.
- `migration-guide.md`, the `BLOB_TRIGGERS` section — "deriving from `CronTriggerImpl` or `SimpleTriggerImpl` (the two that are still open)" is now all five, and `RecurrenceTriggerImpl`'s "it is `sealed`, so `HasAdditionalProperties` cannot be overridden to `true`" is dropped; what is true is that it carries no `[Serializable]` and is written to `SIMPROP_TRIGGERS` by its persistence delegate.
- `how-tos/trigger-persistence-delegate.md` — the tip claiming only `SimpleTriggerImpl` and `CronTriggerImpl` are subclassable now says all five are, and points at the serializer pairing below it.

## Verification

On `55e1b3474`, rebased onto `bf12a6531`:

```
dotnet build Quartz.slnx                        Build succeeded. 0 Warning(s) 0 Error(s)
dotnet test src/Quartz.Tests.Unit               Passed! Failed: 0, Passed: 4794, Skipped: 36
dotnet test src/Quartz.Tests.AspNetCore         Passed! Failed: 0, Passed: 549
dotnet fallout VerifyDocsSnippets               Documentation snippets are up to date (102 markdown files checked)
npm run docs:check-links                        220 pages, 1070 internal links, 650 fragments - no dead links or anchors
```

Earlier on the pre-rebase head, when this branch still touched the serializer packages: `AdoJobStoreSmokeTest.TestSQLite` passed (6 tests, the container-free half of the ADO suite). Those files are untouched now, so the run no longer covers anything this PR changes. `Quartz.Benchmark` names no serializer or dashboard type; no benchmark-visible type moved.

### DIFF
--- a/docs/documentation/quartz-4.x/how-tos/trigger-persistence-delegate.md
+++ b/docs/documentation/quartz-4.x/how-tos/trigger-persistence-delegate.md
@@ -170,9 +170,10 @@ A persistence delegate is one of four pieces:
 implement `GetScheduleBuilder()`.
 
 ::: tip
-Of the five shipped trigger implementations, only `SimpleTriggerImpl` and `CronTriggerImpl` are
-subclassable — `CalendarIntervalTriggerImpl`, `DailyTimeIntervalTriggerImpl` and `RecurrenceTriggerImpl`
-are sealed.
+All five shipped trigger implementations are subclassable, so deriving from `SimpleTriggerImpl` or
+`CronTriggerImpl` — or from `CalendarIntervalTriggerImpl`, `DailyTimeIntervalTriggerImpl` or
+`RecurrenceTriggerImpl` — is the shortest route to a trigger that is one of those with something added.
+Pair it with a serializer deriving from that trigger's built-in serializer, as below.
 :::
 
 **2. An `IScheduleBuilder`.** The store rebuilds a trigger as
--- a/docs/documentation/quartz-4.x/migration-guide.md
+++ b/docs/documentation/quartz-4.x/migration-guide.md
@@ -3939,10 +3939,18 @@ Each of these was a `public class` with `virtual` members in 3.x:
 | Sealed in 4.x | Derive from this instead |
 |---|---|
 | `AnnualCalendar`, `CronCalendar`, `DailyCalendar`, `HolidayCalendar`, `MonthlyCalendar`, `WeeklyCalendar` | `BaseCalendar`, which stays open, or `ICalendar` directly. Every shipped calendar is a short `BaseCalendar` — the `virtual` `IsDayExcluded` / `SetDayExcluded` / `AreAllDaysExcluded` overrides they offered only made sense for that one calendar's day set anyway |
-| `CalendarIntervalTriggerImpl`, `DailyTimeIntervalTriggerImpl`, `RecurrenceTriggerImpl` | `TriggerBase`. `CronTriggerImpl` and `SimpleTriggerImpl` stay open, so a trigger deriving from one of *those* two is still the shortest route to a custom cron or simple trigger. `HasAdditionalProperties` is `virtual` on `TriggerBase`, so it is available either way |
 | `JobExecutionContextImpl`, and its nine `virtual` members with it | `IJobExecutionContext`. A test double implements the interface; a decorator wraps one |
 | `DefaultThreadPool`, `ZeroSizeThreadPool` | `IThreadPool`, or `TaskSchedulingThreadPool`, which stays open — see [The thread pool is asynchronous](#the-thread-pool-is-asynchronous) |
 | `JobChainingJobListener` | The three listener interfaces. Every notification member is a default interface member now, so implementing one directly costs a `Name` at most — see [The three `*Support` base classes are gone](#the-three-support-base-classes-are-gone) |
+
+**The five `*TriggerImpl` types are deliberately not in that list.** A trigger of your own derives from
+`TriggerBase` or from any of the five, `HasAdditionalProperties` is `virtual` on `TriggerBase`, and
+deriving from a built-in trigger pairs with deriving from that trigger's built-in JSON serializer —
+which is why those serializers are public and unsealed in both packages, so a subclass can call
+`base.SerializeFields` / `base.DeserializeFields` and keep the built-in fields' stored shape.
+`CalendarIntervalTriggerImpl`, `DailyTimeIntervalTriggerImpl` and `RecurrenceTriggerImpl` were sealed
+during 4.x's development and were reopened for exactly that pairing — see
+[Other Breaking Changes](#other-breaking-changes).
 
 `JobChainingJobListener` also changed base. It derived from `Quartz.Listener.JobListenerSupport`, which is
 gone, and implements `IJobListener` directly, so its `Name` and `JobWasExecuted` are no longer `override`s.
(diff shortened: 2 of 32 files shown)


lahma:
Reviewed in full. The rework is exactly right: the revert proven byte-identical rather than asserted, and your use of the second-test latitude was better than the instruction — the seam guard (BuiltInTriggerSerializerDerivationTest, failing in 100 ms if a built-in serializer goes internal or sealed) is what should have existed all along, and it is precisely the mechanism that would have caught this PR's own first draft. No, I do not want the standalone-pattern duplicate. The third and fourth stale sealed-claims corrected with their true reasons (SIMPROP_TRIGGERS persistence, not sealing) close out the family of that error. Merging on green.

PR #3628, opened 2026-08-31
A macro means the same schedule on the hash path as everywhere else

### LINKED ISSUE #3626: A macro is a cron expression on every path, including the hashed one
Found by #3619's rebase probe (PR comment there carries the full output): CronMacros.Expand runs inside the CronExpression constructor, but the hash path resolves H tokens BEFORE constructing — ResolveHashTokens counts fields and throws on anything outside 6-7, so a one-field macro dies before the constructor sees it. Consequences on current main: `ResolveHash(@daily, key)` and both `ParseWithHash` overloads throw FormatException, and `TryParseWithHash(@daily, key, out _)` returns false — a validator that contradicts the constructor that accepts the same string. Not a regression (the deleted hash constructors routed identically); it became visible the moment #3620's macros landed.

The fix is one place and purely additive: call CronMacros.Expand at the top of ResolveHash's two overloads (which already ToUpper their input), repairing ResolveHash, ParseWithHash x2 and TryParseWithHash at once. Tests: the four probe cases as pins, plus a macro-with-hash-key round trip. Deliberately NOT folded into #3619 (a members-removal PR) and deliberately not characterization-pinned there, so the fix stays visibly additive.

lahma:
The `@` macros #3620 added are expanded by the `CronExpression` constructor, and both the docs and the migration guide describe that as universal — *"it works wherever an expression string is read"*, and *"expanded by the `CronExpression` constructor, which is what every entry point goes through"*. Hash resolution was the exception, because it reads the expression string **before** constructing anything and `ResolveHashTokens` counts fields first. A one-field macro was therefore rejected as malformed before anything looked at what it said.

Measured on `main` before this change:

```
ctor                        OK    -> 0 0 0 * * ?
Parse                       OK    -> 0 0 0 * * ?
WithCronSchedule("@daily")  OK    -> 0 0 0 * * ?
ResolveHash("@daily", key)  THROW FormatException: Invalid cron expression (expected 6-7 fields, got 1): @DAILY.
ParseWithHash("@daily", key)  THROW FormatException: ... got 1): @DAILY.
ParseWithHash("@daily", 42)   THROW FormatException: ... got 1): @DAILY.
TryParseWithHash("@daily", …) OK    -> false
```

`TryParseWithHash` is the one that had to be fixed before the freeze: it reported `@daily` as an **invalid expression**, so a validator built on it contradicts the parser it stands in for, and the caller is sent to fix an expression that was never wrong.

## The change

`CronMacros.Expand` at the top of both `ResolveHash` overloads, with the same `.Trim()` the constructor applies so that the two normalise identically. Two lines of behaviour; the rest of the product diff is the XML doc sentence on each overload saying so.

It is **additive in the safe direction**: every string that resolved before resolves to exactly the same value — a macro is the only input whose first character is `@`, and `@` was previously rejected outright — while strings that threw a field count now mean what they say. `@reboot` and an unknown `@name` now reach their own rejections, which name the fix, instead of a field count that teaches nothing about the real problem.

A macro carries no `H` token by construction, so there is nothing to hash and the key goes unused. That is why this is a normalisation fix rather than a feature: it makes one door agree with the others, and adds no capability.

## Not done, deliberately

- **No docs edits.** Both pages already assert the behaviour this restores; the fix makes their existing prose true rather than needing new prose. Checked before starting: `cron-expressions.md:99` ("works wherever an expression string is read") and `migration-guide.md:4797` ("`@daily` and the rest of the macros work everywhere"). Neither page ever claimed macros combine with `H`, so nothing needed correcting, only enabling. I would explicitly *not* weaken either sentence.
- **No footnote** saying `ParseWithHash("@daily", key)` is legal but pointless. It is a sentence explaining why a call nobody writes is harmless.
- **No shared normalisation helper.** The root cause is that "upper, trim, expand" is spelled in more than one place, and a private `Normalize(string)` used by the constructor and both overloads would make the invariant structural rather than repeated. I left it out because it touches the constructor, which is not what this issue is about, and the agreed shape here was fix-only. It is a clean follow-up if you want it — say so and I will do it.

## Verification

- `dotnet test src/Quartz.Tests.Unit` — **4,813 passed, 0 failed, 36 skipped**
- `dotnet test src/Quartz.Tests.AspNetCore` — **548 passed, 0 failed**
- `dotnet fallout BenchmarkSmoke` — Restore / Compile / BenchmarkSmoke all succeeded (`CronExpression` construction is benchmarked, and `CronHashTokenBenchmark` calls `ResolveHash` directly)
- `dotnet fallout VerifyDocsSnippets` — up to date, 102 markdown files
- Public API baseline unchanged, as expected: no member added or removed, only doc comments.

The pins fail without the fix — verified by reverting the product file, rebuilding and re-running: **20 failed, 56 passed** in `CronExpressionHashTest`, the 20 being exactly the new cases and the 56 the pre-existing ones. With the fix, 76 pass.

Closes #3626

### DIFF
--- a/src/Quartz.Tests.Unit/CronExpressionHashTest.cs
+++ b/src/Quartz.Tests.Unit/CronExpressionHashTest.cs
@@ -559,6 +559,105 @@ public class CronExpressionHashTest
                 "and the same holds for a seed");
     }
 
+    // --- Macros on the hash path ---
+
+    /// <summary>
+    /// A macro names a whole schedule rather than a field, so it carries no <c>H</c> token and has
+    /// nothing to resolve - but it still has to survive the hash path. Issue #3626: hash resolution
+    /// counts fields before anything is parsed, and it ran ahead of the constructor's macro expansion,
+    /// so a one-field macro was rejected as malformed before anything looked at what it said.
+    /// </summary>
+    [TestCase("@yearly", "0 0 0 1 1 ?")]
+    [TestCase("@annually", "0 0 0 1 1 ?")]
+    [TestCase("@monthly", "0 0 0 1 * ?")]
+    [TestCase("@weekly", "0 0 0 ? * SUN")]
+    [TestCase("@daily", "0 0 0 * * ?")]
+    [TestCase("@midnight", "0 0 0 * * ?")]
+    [TestCase("@hourly", "0 0 * * * ?")]
+    public void ResolveHash_Macro_ExpandsItRatherThanCountingItsFields(string macro, string expanded)
+    {
+        CronExpression.ResolveHash(macro, "key").Should().Be(expanded,
+            "a macro is expanded wherever an expression string is read, the hash path included");
+
+        CronExpression.ResolveHash(macro, 42).Should().Be(expanded,
+            "and the seeded overload reads the same expression as the keyed one");
+    }
+
+    [TestCase("@daily", "0 0 0 * * ?")]
+    [TestCase("@weekly", "0 0 0 ? * SUN")]
+    public void ParseWithHash_Macro_ParsesToTheScheduleTheMacroNames(string macro, string expanded)
+    {
+        CronExpression.ParseWithHash(macro, "nightly").CronExpressionString.Should().Be(expanded);
+        CronExpression.ParseWithHash(macro, 42).CronExpressionString.Should().Be(expanded);
+    }
+
+    [TestCase("@daily", "0 0 0 * * ?")]
+    [TestCase("@weekly", "0 0 0 ? * SUN")]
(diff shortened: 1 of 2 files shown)


lahma:
Reviewed in full. Exactly the agreed shape, and the Trim() catch — with its own pin for the whitespace case — is the kind of edge the fix-only discipline exists to surface rather than bury. The equal-expressions (not equal-strings) round-trip assertion is the right strength. RULING on the flagged follow-up: the shared Normalize helper is DECLINED for alpha.5 — it is internal-only and addable whenever, and the 20 pins now make the three-spelling agreement enforced rather than hoped; noted for a quiet 4.1 cleanup. Merging on green.

PR #3629, opened 2026-08-31
The ADO.NET store is a store, not a base class

### LINKED ISSUE #3596: The ADO namespace is a delegate-authoring kit, not the store's internals

lahma:
Part 3 of #3596, and the one the issue is named for. Closes #3596.

`Quartz.Impl.AdoJobStore` loses **189 members and 5 types** — the largest single API trim of the freeze, and the least-used part of what it removes.

| | types | members |
|---|---:|---:|
| before | 59 | 819 |
| after | 54 | 630 |

The five that go: `AdoJobStoreBase` (162), `AdoJobStoreDependencies` (13), `ExternalTransactionJobStore` (5), `LocalTransactionJobStore` (4), `RecoverMisfiredJobsResult` (5). **Nothing else in the two ADO namespaces moves** — I checked the baseline diff for added lines and there are none.

## What did *not* go internal, and why

The issue's ratified text also lists "the acquisition/machinery records (`TriggerAcquisition*`, `TriggerFiredUpdate`, the state-transition and scheduler-state records, `SqlStatement`/`SqlRowLimit` shapes)". **Thirteen of those are parameter or return types on `IDriverDelegate`**, the interface the whole exercise keeps public, so internalizing any of them is CS0050 on the thing the trim exists to protect. They stay public and the migration guide says explicitly that they are delegate contract rather than store internals. That correction is why the target moved from ~450 to ~620; the floor below 620 is `IDriverDelegate` (96) + `StdAdoDelegate` (135) + `AdoConstants` (72), which *is* the kit.

`LocalTransactionJobStore` and `ExternalTransactionJobStore` are `internal`, **not sealed**: five test subclasses in friend assemblies bind regardless of `InternalsVisibleTo`, and sealing an internal type buys only devirtualization the JIT already performs when no subclass is loaded. CA1852 does not fire.

## Configuration is untouched

`quartz.jobStore.type = Quartz.Impl.AdoJobStore.LocalTransactionJobStore, Quartz` resolves and constructs the same store: `SimpleTypeLoader` goes through `Type.GetType`, which finds an internal type, and `ActivatorUtilities` needs a public *constructor*, not a public type. Those are two facts that are easy to break independently and that every other test in the suite sidesteps by naming the types in code, so `ConfiguredTypeNamesResolveTest` holds them:

- a store named by string is the store that is built, for both shipped stores, with the type-name string spelled as a configuration file spells it rather than computed from `typeof`;
- the resolved type is non-public **and** has a public constructor;
- a driver delegate named by string in `quartz.jobStore.driverDelegateType` runs a real schedule end to end against a file-backed SQLite database, with its own SQL hook counted so the assertion is about the delegate *driving* the store rather than sitting in the container beside one that is;
- the same key naming `MyDatabaseDelegate` from `Quartz.Documentation.Samples` — an assembly Quartz grants no `InternalsVisibleTo` — resolves.

The whole store in that test comes from flat keys, including the data source, because naming the store in code registers a driver delegate of its own and registration is `TryAdd`.

## What replaces the seams

`custom-job-store.md`'s "Deriving from AdoJobStoreBase" section is gone. In its place, **Narrowing what a node picks up**: `TriggerAcquisitionRequest` is a record, `DelegatingJobStore.AcquireNextTriggers` is `virtual`, and the store copies both `MaxCount` and `ExcludedJobTypeNames` straight into its delegate criteria — so `base.AcquireNextTriggers(request with { … })` keeps the exclusion in the acquisition SQL rather than turning it into a post-filter. The two samples that derived from `LocalTransactionJobStore` are decorators now; the samples project has no IVT grant, so they could not have stayed either way.

`AdoJobStoreBase.IsTransient` was the one documented override with no equivalent. #3622 shipped `AdoJobStoreOptions.IsTransient` for it, and `operations.md` now points there.

## Two things this turned up

**A dialect delegate could not inherit its dialect's schema script** (first commit, and a real bug). `SchemaResourceName`'s own documentation says the name is looked for "in the assembly that declares the delegate"; the code looked in `GetType().Assembly`, which is the same thing for a shipped dialect and not the same thing at all for the commonest way anyone writes one — subclass `SQLiteDelegate` or `SqlServerDelegate` and change a statement or two. Such a delegate inherits a resource name pointing into Quartz and is then told the script "is not embedded in" its own assembly, so `ProvisionSchema` fails naming a file it never chose. The lookup walks the base chain now, nearest first. Found by the end-to-end test above, which is why the fix ships here.

**The kit does not hand over the scheduler name.** `StdAdoDelegate` holds it in a private field with no accessor, so a delegate that overrides an `IDriverDelegate` member outright has to keep its own copy from `DriverDelegateContext` in `Initialize`. That works and the new sample shows it, but a `protected string SchedulerName { get; }` would make the common case natural. **Deliberately not added here** — a member addition in a PR whose headline is a 189-member removal is the wrong place for the discussion. Say the word and it is a one-line follow-up.

## The samples are the compile-time half of the guard

`Quartz.Documentation.Samples` has been the proof behind "the ADO namespace is a delegate-authoring kit" for as long as it has existed and nobody had written it down: an ordinary project in the solution, so `Compile` fails a rotted sample, and **no `InternalsVisibleTo` grant**, so a sample that stops compiling means the public kit lost a type. A comment at the top of `DialectDelegateSamples.cs` says so, and the resolution test asserts the grant is absent so it cannot be lost by someone reaching for IVT to make one sample easier.

The proof only covers what the samples name, so they now name one override per category of statement hook plus one `IDriverDelegate` member written out in full, and `OwnTableTriggerPersistenceDelegate` implements `ITriggerPersistenceDelegate` **directly** — the only shape that reaches `TriggerPersistenceDelegateContext`, `IDbAccessor`, `SqlStatement` and `SqlStatementParameter` from a non-friend assembly, because the SIMPROP base's batching member is not virtual.

## For a reviewer to decide

- **Internal, not sealed.** If you would rather have `sealed` and rewrite the five friend-assembly test subclasses, now is the moment.
- **The schema-script fix rides along.** It is a behaviour change in a visibility PR. It is here because the test that found it is here, and because shipping a guard test that works around the defect would have been dishonest. Happy to split it if you would rather review it alone.
- **`private protected` on two integration-test members.** `AdoJobStoreContractTest.CreateJobStore` and `.StoreDependencies` return types that are now internal; the fixtures stay public so NUnit still discovers them.
- The migration guide's older "`AdoJobStoreBase`'s overridable surface is a decision now" section is kept, with a warning saying it describes an intermediate state — it explains what the store now does *not* offer, which is still worth reading.

## Verification

```
dotnet test src/Quartz.Tests.Unit        → Passed! Failed: 0, Passed: 4843, Skipped: 36
dotnet test src/Quartz.Tests.AspNetCore  → Passed! Failed: 0, Passed: 549, Skipped: 0
dotnet fallout VerifyDocsSnippets        → Succeeded, 102 markdown files checked
dotnet fallout BenchmarkSmoke            → Succeeded
```

The 36 skips are main's new Unix-cron differential cases, which need `crontab`. `Quartz.Tests.Integration` compiles — three member accessibilities moved with the store — but was not run: it needs a Docker daemon, which this machine does not have. The build leg's link checker caught one dead anchor and it is fixed: the breaking-changes row about the three ADO store constructors linked `custom-job-store.md#deriving-from-adojobstorebase`, the section this PR replaced. It now points at this guide's own **The ADO.NET store is a store, not a base class**, and says why the change it describes stopped mattering. No other link named that heading. `npm run docs:check-links` still cannot run here — no `node_modules` in the worktree — so I swept all 227 markdown pages with the site's slug rule instead: **zero** links to a heading that does not exist, the only hits being the checker's own deliberately-broken control fixtures under `.vuepress/check-links.fixtures/`.

Baseline accepted deliberately: five types vanish, nothing is added, and the per-type member counts are the ones the design note predicted.

Sonar new-line coverage checked locally rather than guessed: `dotnet fallout Compile UnitTest --coverage`, then every changed line under `src/Quartz/` matched against the OpenCover report — **29 of 29 covered**. `Quartz.Documentation.Samples` sets `SonarQubeExclude`, so the ~220 sample lines are outside analysis entirely and carry no coverage risk.

Closes #3596.


lahma:
Reviewed in full. The SchemaResourceName base-chain fix is the guard test's whole argument made flesh — a subclassed dialect delegate is the commonest third-party shape and provisioning failed for every one of them; shipping the fix and the end-to-end guard together is exactly right. Baseline math matching §8 to the member (−189, 819→630, zero additions), the CS0060 cascade resolution, and the direct-implementation shape for the persistence-delegate sample are all accepted. The protected SchedulerName accessor is deliberately NOT folded in — it gets its own issue. Merging on green.

PR #3631, opened 2026-08-31
The standalone builder configures through the one configuration API

### LINKED ISSUE #3597: The standalone builder configures through the one configuration API
`QuartzSchedulerBuilder` re-declares every `IQuartzBuilder` member with a covariant return and every `QuartzBuilderExtensions` extension as an instance method — ~60 duplicate signatures, a 903-LOC facade — so that standalone configuration chains without a cast. AGENTS.md still describes the type as adding "only `Build()`/`BuildScheduler()` and `UseProperties`"; the facade has drifted well past that, and every future `IQuartzBuilder` member either gets mirrored forever or the facade falls behind silently.

Ratified: the standalone path becomes shape-identical to the hosted path —

```csharp
IScheduler scheduler = await QuartzSchedulerBuilder
    .Create(q =>
    {
        // q is IQuartzBuilder — the same API AddQuartz(q => ...) hands out
    })
    .BuildScheduler();
```

The type keeps `Create(Action<IQuartzBuilder>)`, `UseProperties` ×2 (the ratified 3.x on-ramp), `UseConfiguration`, `Build()`, `BuildScheduler()` — and re-declares nothing. The facade is deleted; there is one configuration API and the lambda receives it, which is what AGENTS.md's "one configuration API rather than two that resemble each other" always meant.

Breaking for alpha users chaining directly on the builder; migration is mechanical (wrap the chain in the lambda) and the migration-guide row shows it. Rides the baseline-heavy merge chain, and AGENTS.md's description of the type gets corrected in the same PR.

lahma:
`QuartzSchedulerBuilder` re-declared every `IQuartzBuilder` member with a covariant return, and every
`QuartzBuilderExtensions` extension as an instance method, so that a standalone chain never had to be
broken up. That was ~60 duplicate signatures in a 903-line file, and it only ever went one way: each
new `IQuartzBuilder` member either got mirrored forever or the facade fell behind silently.

`Create` now takes the callback `AddQuartz` takes, and hands it the same `IQuartzBuilder`:

```csharp
IScheduler scheduler = await QuartzSchedulerBuilder
    .Create(q => q
        .ConfigureScheduler(o => o.InstanceName = "reporting")
        .UseDefaultThreadPool(maxConcurrency: 20)
        .UseInMemoryStore())
    .BuildScheduler();
```

The type keeps `Create`, `UseProperties` ×2, `UseConfiguration`, `Build()` and `BuildScheduler()`, and
re-declares nothing. It no longer implements `IQuartzBuilder` either — implementing it explicitly would
have left the same surface reachable through a cast, which is the second configuration API this issue
is about removing.

Closes #3597.

## Decisions a reviewer should confirm

**`Create` takes `Action<IQuartzBuilder>? configure = null`, not a required lambda.** The issue writes
the survivor as `Create(Action<IQuartzBuilder>)`. Making the parameter optional keeps it one member and
one baseline line, and makes the standalone entry point *literally* the shape of the hosted one —
`AddQuartz(Action<IQuartzBuilder>? configure = null)` is the signature it mirrors. It also leaves the
89 in-repo `Create().UseProperties(properties)` call sites, and the configuration-only path generally,
untouched: a scheduler described entirely by a properties file has nothing to put in a lambda, and
`Create(_ => { })` would be noise. Say the word and I will make it required.

**`QuartzBuilderExtensionsMirrorTest` and `QuartzSchedulerBuilderForwardingTest` are deleted, not
repurposed.** Both existed only for the facade: the first proved a mirror existed for every extension
and returned the builder, the second proved each mirror's body actually called the extension it
mirrored. With no mirrors there is no invariant left for either to hold — the extensions are now
exercised through `IQuartzBuilder` by the `AddQuartz` tests that already cover them, and the standalone
path's own use of them is covered by `QuartzSchedulerBuilderTest`. The #3621 mirror work is what those
two tests were guarding, and it goes with them.

`QuartzSchedulerBuilderTest.TheContentMembersKeepTheBuildersTypeThroughOneExpression` is renamed
`TheContentMembersAreReachableThroughTheCallback` and rewritten: it still builds a scheduler carrying a
job, a trigger and a calendar, but the assertion is now that the callback's `IQuartzBuilder` reaches
them without the builder declaring one of them.

## The rest

- Every in-repo caller swept — unit, integration, AspNetCore, examples, benchmarks, documentation
  samples. The mechanical shape is "wrap what was between `Create()` and the terminal call in
  `Create(q => q…)`"; the handful of sites that held the builder in a variable to reach `Services`, or
  that configured it across an `if`, became statement-bodied callbacks.
- `PublicApiTest_Quartz` baseline regenerated after rebasing onto `fc28c8be9` (the ADO
  internalization). The delta against `origin/main` is exactly two added lines — the class declaration
  losing `: Quartz.IQuartzBuilder`, and the new `Create` — and 107 removed, which are the facade set
  plus `Services`, `SchedulerName` and the old `Create()`. `Build`, `BuildScheduler`,
  `UseConfiguration` and both `UseProperties` are untouched.
- Migration guide: "The standalone builder is the same builder" rewritten around the callback, with a
  `For 4.0 alpha users` block carrying the wrap-the-chain diff and the note that `Services` and
  `SchedulerName` are read as `q.Services` / `q.SchedulerName`. The appendix rows and the
  `Removed from QuartzSchedulerBuilder` table say the same thing.
- `AGENTS.md`'s description of the type is corrected: it said the builder "adds only `Build()` /
  `BuildScheduler()` and `UseProperties(NameValueCollection)`", which the facade had made untrue. It
  now names the five members and tells a future agent not to add a sixth that `IQuartzBuilder` already
  has.
- `docs/documentation/quartz-3.x/` untouched.

## Verification

```
dotnet build Quartz.slnx                    0 errors, 0 warnings
dotnet test src/Quartz.Tests.Unit           4815 passed, 0 failed, 36 skipped
dotnet test src/Quartz.Tests.AspNetCore      549 passed, 0 failed
dotnet fallout VerifyDocsSnippets            up to date (102 markdown files checked)
dotnet fallout BenchmarkSmoke                Succeeded, 282 benchmarks executed
```

No `*.received.*` files remain after the Verify run. `npm run docs:check-links` was not run — the docs
dependencies are not installed in this worktree — but the diff changes no heading and adds exactly one
link, `#the-standalone-builder-is-the-same-builder`, whose heading is unchanged.

Integration tests were not run; Docker is available here but the suite needs the full container matrix
and this change only reshapes their scheduler construction, which the solution build type-checks.


lahma:
Reviewed in full. Both deviations ACCEPTED and both improve the spec: (1) `Create(Action<IQuartzBuilder>? configure = null)` is MORE faithful to the ratified 'shape-identical to the hosted path' than the required-lambda form — it mirrors AddQuartz to the character, and a properties-only scheduler keeps its clean spelling; (2) dropping the IQuartzBuilder implementation entirely is the stronger and correct reading — an explicit implementation would have left the whole second API reachable through one cast, which is the thing being removed. The mirror-test deletions follow their invariant out the door. Baseline +2/−107 verified as exactly the facade set. Merging on green.

PR #3632, opened 2026-08-31
Plugins whose job the host already does are retired

### LINKED ISSUE #3593: Plugins whose job the host already does are retired

lahma:
Closes #3593.

Three of the four items in the issue; `JobInterruptMonitorPlugin` was already retired by #3621, so
this finishes the sweep. One commit per retirement, each self-contained.

## `ShutdownHookPlugin` is retired

The plugin subscribed to `AppDomain.CurrentDomain.ProcessExit` with an `async void` handler, so
nothing awaited the shutdown it started — the process was free to exit part-way through
`IScheduler.Shutdown`, which is the opposite of the clean shutdown the plugin's name promised.
`QuartzHostedService` is an `IHostedLifecycleService` and already stops every registered scheduler as
part of the host's own shutdown, awaited, inside `HostOptions.ShutdownTimeout` and concurrently across
schedulers; `QuartzHostedServiceOptions.WaitForJobsToComplete` is `CleanShutdown` under its real name.

Gone: `ShutdownHookPlugin`, `UseShutdownHook`, `ShutdownHookOptions`, `ManagementPluginLog` and the
whole `Quartz.Plugins.Management` namespace — its three log events (6500-6502) existed only for the
hook, so `LogEventCatalogTest_Quartz.Plugins.verified.txt` loses them.

The migration guide gains a before/after table covering the hosted-service replacement, the
non-hosted `scheduler.Shutdown(waitForJobsToComplete: true)` on the application's own exit path, and
the `quartz.plugin.shutdownHook.type` key, which now fails to load rather than being ignored.

## `PropertyListenerFactory` and the listener property keys are retired

`quartz.jobListener.<name>.type` / `quartz.triggerListener.<name>.type` asked Quartz to find a
listener class by reflection and attach it. What they could not ask for is matchers, so a listener
named that way heard *every* job or trigger; there was no scheduler-listener spelling at all; and the
type name is a string no trimmed or native-AOT publish can follow. 156 lines of reflection and three
trim-analysis suppressions go with the factory, and `SchedulerContentInitializer` no longer needs a
property bag at all.

**The keys are rejected, not ignored.** They join `LegacyPropertyKeys.removedKeys`, so a configuration
still carrying one is refused by name with advice naming `AddJobListener<T>(matchers)` /
`AddTriggerListener<T>(matchers)` — the same shape as the `quartz.scheduler.proxy` and
`quartz.jobStore.lockHandler.schedName` rejections. Silence here would take a listener out of a
pipeline with no symptom.

## `TimeZoneConverterPlugin` becomes a resolver registration

The plugin was one `TimeZones.AddResolver` call wearing a plugin's lifecycle: no per-scheduler state,
no scheduler dependency, and an `Initialize` that ignored both arguments it was handed. The package
and `UseTimeZoneConverter` both stay; the extension performs the registration itself, through an
internal `TimeZoneConverterResolver`.

Two consequences, both improvements:

* **It registers at configuration time, not scheduler start.** `TimeZones.FindById` is reached from
  `TriggerBuilder`, `CronExpression` and trigger deserialization — none of which have a scheduler in
  scope — so a trigger built before the host starts now resolves its zone too.
* **Nothing removes the registration.** This is how the plugin's double-registration safety is
  preserved rather than dropped: the plugin gave each scheduler its own registration and disposed only
  that one on `Shutdown`, so that stopping one scheduler did not change resolution for the others. One
  registration that outlives every scheduler is the same guarantee with none of the bookkeeping, and a
  second `UseTimeZoneConverter` is a no-op rather than a duplicate on a list nothing prunes.

## For a reviewer to decide

* **`quartz.plugin.shutdownHook.type` and `quartz.plugin.timeZoneConverter.type` fail loudly but
  generically.** Both now hit `SchedulerPluginFactory`'s `SchedulerPlugin of type '…' could not be
  loaded.` There is no removed-plugin-advice list to hang a targeted message on, and adding one is a
  bigger change than this issue asks for. The migration guide names both keys instead. Say the word if
  a targeted rejection is wanted and it can be a follow-up.
* **`SimpleTypeLoader` keeps its `Quartz.Listener.` → `Quartz.Listeners.` namespace fallback**, which
  no shipped caller can now reach — the listener keys were its only consumer inside Quartz. It stays
  because the fallback table is a table of renames rather than a list of Quartz's own callers, and an
  `ITypeLoader` is a public seam an application can call. `SimpleTypeLoaderTest`'s reason string was
  rewritten to stop teaching the retired key.
* **`AGENTS.md` gains the "four history-logging plugins ship, not two" ruling** the issue asked for,
  in the settled-decisions section, plus the `Use*` row loses `UseShutdownHook` and gains a note that
  `UseTimeZoneConverter` keeps the verb although it no longer registers a plugin.

## Verification

```
dotnet build Quartz.slnx                    -> Build succeeded. 0 Warning(s) 0 Error(s)
dotnet test src/Quartz.Tests.Unit           -> Passed! Failed: 0, Passed: 4845, Skipped: 36, Total: 4881
dotnet test src/Quartz.Tests.AspNetCore     -> Passed! Failed: 0, Passed: 549, Total: 549
dotnet fallout VerifyDocsSnippets           -> Documentation snippets are up to date (102 markdown files checked)
```

Baselines regenerated through the Verify workflow and reviewed: `PublicApiTest_Quartz.Plugins`
(`UseShutdownHook`, `ShutdownHookOptions`, `ShutdownHookPlugin`),
`PublicApiTest_Quartz.Plugins.TimeZoneConverter` (`TimeZoneConverterPlugin`) and
`LogEventCatalogTest_Quartz.Plugins` (events 6500-6502). No `.received.txt` remains.
`PublicApiTest_Quartz` is unchanged — `LegacyPropertyKeys`, `PropertyListenerFactory` and
`SchedulerContentInitializer` are all internal. The migration guide's removed-type count goes 98 -> 100
for the two 3.x public plugin types.

Not run: `Quartz.Tests.Integration` (needs a Docker daemon, which this machine does not have running)
and `BenchmarkSmoke` — `Quartz.Benchmark` names none of the types this touches, and it builds clean.
`npm run docs:check-links` was not run either (no `node_modules` in this worktree); the three new
heading anchors were checked by hand against the slug of the neighbouring
`#jobinterruptmonitorplugin-is-retired-a-job-timeout-is-middleware`, which has the same punctuation
shape.


lahma:
Reviewed in full. All seven deviations ACCEPTED: (1) generic failure for the two plugin keys is right — a removed-plugin-advice list is wider than this issue, and the guide names both; (2) SimpleTypeLoader's rename table staying is correct — it maps names, it does not enumerate Quartz's own callers; (3) the never-remove resolver registration preserves the guarantee by strengthening it, and the Interlocked one-shot plus the documented process-wide consequence is the honest shape; (4) the IVT grant mirrors the sibling package for a real coverage need; (5-7) the AGENTS.md ruling row, the CONTRIBUTING example swap to a region that exists, and the troubleshooting cleanup are all owed housekeeping. Merging on green — expect an AGENTS.md/guide rebase against #3631, whichever lands second.

PR #3633, opened 2026-08-31
The builder surface says each thing once

### LINKED ISSUE #3598: The builder surface says each thing once
The freeze-window dedup batch across the configuration and builder families; every item is from the alpha.5 API audit, each lands with baseline diff + migration-guide row.

- **`ScheduleJob` collapses 8 shapes toward 4**: the four `IScheduler` overloads that differ only by an options parameter become two with `ScheduleJobOptions options = default` — the pattern `AddJob`/`AddCalendar` already use.
- **Schema provisioning speaks one shape**: `IPersistentStoreBuilder.ProvisionSchema()` (no-arg), `AdoJobStoreOptions.SchemaProvisioning` (3-value enum) and `QuartzAspireSettings.ProvisionSchema` (`bool?`, cannot express `Validate`) align on the enum; the Aspire setting gains the missing expressiveness without losing its IsDevelopment default.
- **Job-store selection (7 ways) and hosted-service registration (6 ways)** get an overload audit — keep the ratified pairs, cut what nothing needs.
- **Health checks**: `AddHealthChecks().AddQuartz()` is ratified; the unratified second and third receivers (`IServiceCollection.AddQuartzHealthChecks` / `IQuartzBuilder.AddQuartzHealthChecks`) get one survivor.
- **`WithSchedule` ×3 paths, `WithCronSchedule` ×4 overloads, `OfType` ×4 + `JobType`'s implicit/explicit operators** (six spellings of "this is the job type"): audit, trim to the set with distinct value.
- **`IMutableTrigger` stops leaking**: `IScheduleBuilder.Build()` and the five schedule builders return an `Extensibility` type from the primary user path — resolve by return-type change or ratify the leak with a written reason; same treatment for the other `Extensibility.*` parameters reachable from mainstream `AddQuartz` code.
- **`Create()` placement**: schedule builders carry their own `Create`; `JobBuilder`/`TriggerBuilder` use companion classes — ratify the split (generic inference forces the companions) or align.
- **`IThreadPool`**: six public members guarding one integer; decide the frozen shape now — member audit against what `QuartzScheduler` actually calls, `ZeroSizeThreadPool` internalization, and the ruling written down.

lahma:
The freeze-window dedup batch. Three members were a spelling of the member beside them and are gone;
one setting gained the vocabulary its counterpart already spoke; and every family the audit opened but
did not cut got a written ruling, on the members and in `AGENTS.md`, so the next reader does not
recount them.

Closes #3598.

## Removed

| Removed | Say it with | Why |
|---|---|---|
| `IScheduler.ScheduleJob(IJobDetail, ITrigger, CancellationToken)` and `ScheduleJob(ITrigger, CancellationToken)` | the surviving overloads, now `ScheduleJobOptions options = default` | `options` had no default *only* because these two made one ambiguous — the type's own docs already said there is no third state between "not given" and "all defaults" |
| `IServiceCollection.AddQuartzHealthChecks(Action<QuartzHealthCheckOptions>?)` | `services.AddHealthChecks().AddQuartz(configure)` | It forwarded to exactly that and carried nothing of its own |
| `JobBuilder<TJob>.OfType(string typeName)` | `OfType((JobType) typeName)` | The `string` to `JobType` conversion is **explicit** so the unvalidated-name leap is visible at the call; a method taking the string put the leap back out of sight |

`IQuartzBuilder.AddQuartzHealthChecks()` is the surviving unratified receiver, not the service-collection
one: a scheduler's builder knows *which* scheduler the check is for, so it is the only form that saves
the caller repeating a name. `Quartz.Aspire` and every named-scheduler sample already used it.

**`ScheduleJob`'s store path is now chosen by `Replace`, not by which overload was called.** Not
replacing still goes through `IJobStore.ScheduleJob`, so that member keeps its caller and a store's span
and metric names for the ordinary call are unchanged; replacing still goes through `ScheduleJobs`,
because over-writing a job and its trigger has to be one operation under one lock. The only behaviour
change is that an explicit `new ScheduleJobOptions()` now takes the plain path — same store outcome,
better span name.

**The one trap for callers**, and it is a compile error rather than a silent change: a cancellation
token passed positionally as the third argument no longer binds. `ScheduleJob(job, trigger, token)`
becomes `ScheduleJob(job, trigger, cancellationToken: token)`. The sweep did that at ~45 in-repo sites.

## Changed

`QuartzAspireSettings.ProvisionSchema` (`bool?`) became `SchemaProvisioning` (`SchemaProvisioning?`),
which is the enum `AdoJobStoreOptions.SchemaProvisioning` and `IPersistentStoreBuilder.ProvisionSchema()`
already spoke. It keeps the unset-asks-the-environment default that made it nullable —
`CreateIfMissing` under `Development` — and gains `None`, which a `bool` could not say at all. `false`
becomes `SchemaProvisioning.Validate`, which still means "leave it alone" because `Validate` is what an
unconfigured store holds. `ConfigurationSchema.json` gains the enum with its three values, and
`AspirePackagingTest.EverySettingAppearsInTheConfigurationSchema` holds the two together.

## Audited and kept — each with its reason

Written on the members themselves, and listed in `AGENTS.md` under "Overload sets audited and frozen in
#3598" so the next pass does not have to redo the counting.

- **Job-store selection, all 7 kept.** Two shipped stores (`UseInMemoryStore`, `UsePersistentStore`), a
  persistent store of another type (`UsePersistentStore<T>` — the documented decorator seam), a store of
  your own built by the container (`UseJobStore<T>`, plus `<T, TOptions>` which `custom-job-store.md`
  teaches), one you built (`UseJobStore(IJobStore)`, the ratified pair with `UseThreadPool(IThreadPool)`),
  and one a factory builds over the scheduler's own parts. Seven capabilities, not seven spellings.
- **Hosted-service registration, all 6 kept.** Three shapes — ordinary, `<T>` for a subclass,
  `(schedulerName, …)` for one scheduler — on the two receivers every registration API here has. C# has
  no default type argument, so the first shape is not the second written shorter.
- **`WithCronSchedule` x4 kept, and I reversed myself on one.** I first cut
  `WithCronSchedule(CronScheduleBuilder)` as `WithSchedule` with a narrower parameter — which it is — and
  then found `WithSimpleSchedule`, `WithCalendarIntervalSchedule`, `WithDailyTimeIntervalSchedule` and
  `WithRecurrenceSchedule` all have the same member. Cutting it from cron alone would leave one family
  short of the shape the other four teach, and it is the 3.x spelling a migrating application holds. The
  other three each say something `WithSchedule` cannot (parse a string, take an expression already parsed
  — where a hash key rides — or close a `CronExpressionBuilder` chain) and all three apply `configure`.
- **`WithSchedule` x3 kept.** Not three members: one, plus the two explicit interface implementations
  that give `ITriggerConfigurator<TJob>` its covariant return. A caller sees one.
- **`OfType` x4 to 3, both `JobType` operators kept.** `OfType<T>()` is compile-time and trim-safe;
  `OfType(Type)` adds the `TJob` assignability check that the implicit operator has no `TJob` to make;
  `OfType(JobType)` carries a stored name with its resolver. The implicit/explicit asymmetry on the
  operators *is* the ruling — validated conversion implicit, unvalidated one explicit — and it becomes
  load-bearing now that `OfType(string)` is gone.
- **`IThreadPool` keeps all six members; `ZeroSizeThreadPool` stays public.** The issue called it six
  members guarding one integer; the integer is not what it is for. Each member has exactly one caller:
  `Initialize` from the scheduler factory, `WaitForAvailableThreads` from the scheduling loop sizing its
  batch, `TryRun` from every firing, `PoolSize` from `SchedulerMetadata`, `Shutdown` to end it, `Drain`
  to end it with a deadline `Shutdown` cannot express. The pool type stays public because
  `UseThreadPool<ZeroSizeThreadPool>()` needs a reachable type argument — the `UseZeroSizeThreadPool()`
  shorthand is what 4.0 removed, not the type.
- **I10, `Create()` placement: ratified, not aligned.** `JobBuilder`/`TriggerBuilder` need companion
  classes because they are generic — `JobBuilder<MyJob>.Create()` would name the job type twice — and no
  schedule builder is generic, so none has the problem. Aligning either side makes one worse.
- **I9, `IScheduleBuilder.Build()` keeps returning `IMutableTrigger`: ratified with a reason for #3603.**
  A return-type change is not available — `TriggerBuilder.Build()` writes identity, job key, calendar
  name, start/end, priority and misfire onto what it is handed, so `ITrigger` would only move the
  downcast one line later. Moving the interface into `Quartz.Extensibility` beside its return type was
  considered and refused: `WithSchedule(IScheduleBuilder)` is the most-read signature on the trigger
  builder, and it would charge every reader for a member no mainstream caller invokes. The leak is a
  declaration, not a call.
- **I16, `Extensibility` parameters: ratified per member.** Every one on the mainstream builder path is
  *the component being supplied* — `IJobStore`, `IThreadPool`, `IJobFactory`, `ITypeLoader`,
  `IInstanceIdGenerator`, `ISchedulerPlugin` — and supplying one is what extending means. The single
  exception is `ConfigureJobScope`'s `TriggerFiredBundle`, where the caller is reading a firing rather
  than supplying anything; it is ratified because the callback runs *before the job exists*, and
  `IJobExecutionContext` is constructed around a job instance. A mainstream stand-in would be a second
  bundle carrying the same values.

## Verification

```
dotnet build Quartz.slnx                 0 errors, 0 warnings
dotnet test src/Quartz.Tests.Unit        Passed! 4818 passed, 0 failed, 36 skipped, 47 s
dotnet test src/Quartz.Tests.AspNetCore  Passed! 549 passed, 0 failed, 6 s
dotnet fallout VerifyDocsSnippets        up to date (102 markdown files checked)
dotnet fallout BenchmarkSmoke            Succeeded, 282 benchmarks executed
```

Baselines regenerated after rebasing onto `f3ec1497f`; no `*.received.*` remains after either test run.
The `PublicApiTest_Quartz` delta is the two `ScheduleJob` collapses, `AddQuartzHealthChecks` and
`OfType(string)`; `PublicApiTest_Quartz.HttpClient` is `HttpScheduler`'s matching `ScheduleJob` pair;
`PublicApiTest_Quartz.Aspire` is the one settings property.

Integration tests were not run — the suite needs the container matrix — but the ADO and XML call sites
the sweep touched are type-checked by the solution build, and `XmlSchedulingDataProcessorTest`'s
FakeItEasy expectations were updated for the new arity in both the unit and integration projects.

`docs/documentation/quartz-3.x/` untouched.


lahma:
Reviewed in full. Rulings: (1) the three cuts and the Aspire enum change are exactly the spec; (2) the WithCronSchedule(CronScheduleBuilder) REVERSAL is ratified with approval — cutting one family's member while four siblings teach the shape would have manufactured the very inconsistency this issue exists to remove; recording the ruling instead is the mature call; (3) every keep's reasoning read and accepted, and the AGENTS.md audited-and-frozen section (with the byte-cap check) is precisely how this campaign's decisions stay decided; (4) the positional-token compile trap is documented where it belongs. Retroactive ratification for #3631's sonar.yml paragraph replacement: replacing the opposite-argument prose rather than contradicting it beside itself was right, and naming what surfaced the change keeps the file honest. One addition requested before CI finishes (message sent): the Examples.HttpClient exclusion. Merging on green after that push.

lahma:
Amended with the coverage-exclusion completion. One finding while doing it changed the scope, and it is worth a look before merge.

`src/Quartz.Examples.HttpClient/**` went in as asked. But checking which example projects these two PRs actually touched turned up that **this PR already edits an uncovered example**: `src/Quartz.Examples.AspNetCore/Startup.cs`, one line, the `services.AddQuartzHealthChecks()` to `services.AddHealthChecks().AddQuartz()` rewrite. That is the #3597 trap firing again on #3598 — a new line in a project no test references — so `AspNetCore` was not hypothetical, it was live on this branch and would have had the gate charge this PR for it.

So the exclusion list now names every `src/Quartz.Examples*` project: `AspNetCore`, `HttpClient` and `Worker` join the four already there. `grep -rn "Quartz.Examples" --include=*.csproj src/Quartz.Tests.Unit src/Quartz.Tests.AspNetCore` returns nothing, so none of the seven carries coverage and none ever could without a test nobody wants. Adding the two beyond `HttpClient` is a deviation from "add HttpClient"; the alternative was shipping a known-red gate and a third round trip for `Worker`.

The paragraph is rewritten to match. It no longer claims `src/Quartz.Examples` is "the last member of that family" — that stopped being true the moment `HttpClient` was added — and it says why all of them are listed rather than only the ones that have tripped it.

`.github/workflows/sonar.yml` is the only file in the amendment: 15 insertions, 10 deletions, and it still parses as YAML. No product or test code moved, so the verification in the description stands unchanged.

lahma:
The all-seven amendment is RATIFIED — checking which examples this branch actually touches before obeying the one-project instruction is exactly the right discipline: AspNetCore was live on this PR, not hypothetical, and the grep proof (no test project references any example) closes the whole trap class rather than the last-tripped member. The paragraph rewrite over a bolted clause follows the same no-self-contradiction rule as the PR-1 edit. Merging on green.

PR #3634, opened 2026-08-31
Reads have two altitudes on purpose; everything else has one

### LINKED ISSUE #3599: Reads have two altitudes on purpose; everything else has one
The read-API and data-map consistency batch. The headline is a ruling, not a removal: `IScheduler`'s seven `Query*` members (paged, filtered, 4.x-native) and `SchedulerQueryExtensions`' `Get*` conveniences are **two altitudes of one API, kept deliberately** — query objects for paging and filters, one-line conveniences implemented over them. The ruling goes into AGENTS.md's settled list so the apparent duplication is never re-audited; the wrapper tail gets trimmed to real presets (`QueryTriggersInError` earns its place; pure synonyms do not).

The fixes around it, all breaking-or-never:

- **Pause/resume return types**: `PauseJobs`/`ResumeJobs`/`PauseTriggers`/`ResumeTriggers` return `List<string>` (group names) from the matcher overload but `List<JobKey>`/`List<TriggerKey>` from the key-collection overload — the only family with this asymmetry (`DeleteJobs`/`UnscheduleJobs` are consistent). Align on keys.
- **`ResetTriggersFromErrorState(matcher)`** is a mutation living in a class named for queries while its key-collection twin sits on `IScheduler` — move it.
- **`OneOffJobOptions`** is the only options type carrying loose `Name`/`Group` strings and a bare `bool Replace` with no `Replacing` preset; align with its siblings, and give `AddJobOptions` the preset its second flag never got.
- **Query filter shapes**: six query types spell the `Name` filter three ways (`NameMatcher` ×3, `string` ×2, `CalendarNameMatcher` ×1) and `FireInstanceQuery` prefixes its trigger filters while `TriggerQuery` does not — one shape.
- **Matcher spellings**: "match everything" has five spellings; `CalendarNameMatcher` re-implements the matcher family outside `IMatcher<T>` — fold it in or ratify why not; the `Matchers.` statics vs per-type factories split gets a ruling.
- **`DataMapExtensions`** carries 74 members; `Get<T>`/`TryGet<T>` subsume every typed accessor. Keep the mainstream scalar set twenty years of tutorials teach, delete the exotic tail (~74 → ~40), and ratify the `SchedulerContext`/`JobDataMap` twin divergence (`PutAsString` and equality participate in job-data change tracking) as deliberate.
- **Small placement fixes, now-or-never**: `ScheduledJobKey<TJob>` is a non-extension static in an extension class; `GetTriggerBuilder` is an interface member while `GetJobBuilder` is an extension; `TriggerDetailsUpdate` spells misfire six ways.

lahma:
The read-API and data-map consistency batch. The headline is a ruling, not a removal, and the rest of
the batch is names and shapes that meant two things.

Closes #3599

## The ruling: two altitudes, kept

`IScheduler`'s `Query*` members (paged, filtered, header-returning) and `SchedulerQueryExtensions`'
`Get*` conveniences (one line, bare keys and names) are two deliberate altitudes of one read API. The
rationale is written into the migration guide as *Reading has two altitudes on purpose*, so the
apparent duplication is not re-audited. (The AGENTS.md settled-list row is #3603's, per the brief; I
did not touch AGENTS.md.)

### The wrapper tail, audited one by one

| Wrapper | Verdict | Why |
|---|---|---|
| `QueryJobs()` | **cut** | `QueryJobs(new JobQuery())` — saves the `new`, knows nothing |
| `QueryTriggers()` | **cut** | same |
| `QueryFireInstances()` | **cut** | same |
| `QueryFireInstancesOfJob(jobKey)` | **cut** | moves a caller-supplied argument into a property; no preset value of its own |
| `QueryTriggersInError()` | **kept** | a preset: it knows `State = TriggerState.Error`, and "which triggers are broken?" is a question an operator asks |

## Pause/resume: aligned by name, not by return type — please read this one

The issue says "align on keys". I did not, and this is the one thing here a reviewer should decide.
What I found looking at what the group-name return is actually for:

* **It is the state the call writes.** `PauseTriggerGroups` writes rows in `QRTZ_PAUSED_TRIGGER_GRPS`
  / `QRTZ_PAUSED_JOB_GRPS`. Those rows survive a restart, reach every node, are what
  `IsTriggerGroupPaused` reports, and — for trigger groups — are imposed on triggers stored into the
  group *afterwards* (`RAMJobStore`'s store path and `AdoJobStoreBase.StoreTrigger` both check). The
  group names are exactly the rows written.
* **An equality matcher pauses an empty group**, deliberately, so that what is added next is born
  paused. There is no key to return for that, and a keys answer would report "nothing happened".
* **Keys would cost a query the ADO path had been careful to avoid.** `PauseTriggerGroups` is one
  set-based `UPDATE`; answering with keys means selecting every affected trigger key first, unbounded.

So instead of changing the return, I changed the name, which fixes the actual complaint — one name,
two return shapes:

| Before | After |
|---|---|
| `PauseJobs(GroupMatcher<JobKey>)` → `List<string>` | `PauseJobGroups(GroupMatcher<JobKey>)` → `List<string>` |
| `ResumeJobs(GroupMatcher<JobKey>)` | `ResumeJobGroups(…)` |
| `PauseTriggers(GroupMatcher<TriggerKey>)` | `PauseTriggerGroups(…)` |
| `ResumeTriggers(GroupMatcher<TriggerKey>)` | `ResumeTriggerGroups(…)` |

The key-set overloads keep their names and their key answers, so every name has one shape, and the
`Delete`/`Unschedule` precedent still reads as the contrast it is: a delete records nothing per group,
so it answers with keys. On `IScheduler`, `IJobStore` and every implementation; bodies untouched.
`TracingJobStore` gives the group forms four span names of their own, where the group and key-set
forms used to share one.

**If you would rather have the plain "align on keys",** say so — it is a mechanical follow-up, but it
loses the empty-group case and adds a key SELECT to the ADO pause path.

## The rest

* **`ResetTriggersFromErrorState(matcher)`** moves off `SchedulerQueryExtensions` (a class named for
  queries) onto `IScheduler`, where the other matcher-shaped mutations live, as a **default interface
  member** over `QueryTriggers` + the key-set reset. No forwarding in four schedulers, and an
  implementation whose store can do it in one statement can override. One cost, called out in the
  guide: a variable typed as a *concrete* scheduler needs
  `((IScheduler) x).ResetTriggersFromErrorState(matcher)`.
* **Query filter shapes.** `CalendarNameMatcher` → `NameMatcher` (arity-free twin of
  `NameMatcher<TKey>`), and `JobGroupQuery.Name` / `TriggerGroupQuery.Name` become that matcher
  instead of an exact `string` — so every listing's name filter is one family, and a group listing can
  be filtered by prefix (a dashboard listing one tenant's groups). ADO gets four LIKE predicates and
  reuses the existing escaping; the HTTP group listings take the four `name*` parameters like every
  other listing. `FireInstanceQuery`'s `TriggerGroup`/`TriggerName` prefixes are **ratified**: a query
  names a filter for what it selects on, and a firing has no key of its own, so its trigger is a
  reference like `Job` and `CalendarName` are.
* **Matcher spellings.** Five "match everything" → three with distinct value: `Matchers.AllJobs()`,
  `Matchers.AllTriggers()` and `GroupMatcher<TKey>.AnyGroup()` (which the members taking a
  non-nullable `GroupMatcher` need). `EverythingMatcher<TKey>.All()` is internal, and
  `NameMatcher<TKey>.AnyName()` — zero callers, added in a 4.0 preview — is gone, because every filter
  taking one is nullable. `CalendarNameMatcher` stays outside `IMatcher<T>` (verified: the interface
  and the combinators are constrained to `Key<T>`), it simply gets the family's name. The
  `Matchers`-vs-per-type-factory split is ratified with a reason: a factory on the type names its
  comparison; a root on `Matchers` takes the comparison as a value, which is what the HTTP API holds.
* **`DataMapExtensions`, 66 public members → 34.** Kept per receiver: `GetInt`, `GetLong`, `GetFloat`,
  `GetDouble`, `GetBoolean`, `GetString`, `GetDateTimeOffset` (+ `TryGet…` twins) and
  `Get<T>`/`TryGet<T>`/`GetValueOrDefault<T>`. Cut: `GetChar`, `GetDecimal`, `GetDateTime`,
  `GetDateOnly`, `GetTimeOnly`, `GetTimeSpan`, `GetGuid`, `GetEnum<TEnum>` and their `TryGet…` twins.

  **Correction to the issue's premise:** `Get<T>`/`TryGet<T>` did *not* subsume the typed accessors —
  they were a pure type test, so under `StoreJobDataAsStrings` (or after any `PutAsString`)
  `Get<Guid>` failed where `GetGuid` parsed. Deleting accessors on that premise would have deleted
  reading paths. So the generic accessors **now coerce**, with exactly the named accessors' rule, and
  the trim is safe: `Get<Guid>` reads what `GetGuid` read. That is a behaviour change for
  `TryGet<T>`/`GetValueOrDefault<T>` on a string entry, and it has a "Changed in 4.x" box.
  The `SchedulerContext`/`JobDataMap` divergence is ratified in the class doc and the guide:
  `PutAsString` participates in the map's change tracking, which an extension cannot do and a context
  has nothing to do.
* **Options.** `OneOffJobOptions.Replacing(name)` — a *method*, because `Replace` without a name does
  nothing (a generated name has nothing to replace), which is argued in the XML docs and the guide.
  `AddJobOptions.ReplacingAndStoringNonDurable` gives the second flag the preset
  `AddCalendarOptions.ReplacingAndUpdatingTriggers` always had. `OneOffJobOptions`' loose
  `Name`/`Group` strings are **ratified, not changed**: the two halves have different defaults (a
  generated name, a group named after the job type) and each is worth setting without the other,
  which a `TriggerKey` cannot express.
* **Placement.** `ScheduledJobKey<TJob>()` — a non-extension static in an extension class, so every
  call site named the class anyway — moves to `SchedulerConstants`, beside `ScheduledJobGroup`.
  `ITrigger.GetTriggerBuilder` **stays an interface member** where `GetJobBuilder` is an extension:
  the builder must be created against the trigger's own `TimeProvider`, which is not on `ITrigger`
  (verified — an extension cannot recover the clock). `TriggerDetailsUpdate` **keeps all six** misfire
  spellings: the five typed ones are what let the store reject a cross-family update, and
  `WithMisfireInstructionCode(int)` is the only path for a trigger outside the five built-in families
  (`TriggerFamilyExtensions.Family()` answers `null` for a custom `ITrigger`, so every typed overload
  would be rejected for it).

## Verification

* `dotnet build Quartz.slnx` — succeeded.
* `dotnet test src/Quartz.Tests.Unit` — **4831 passed, 0 failed, 36 skipped**.
* `dotnet test src/Quartz.Tests.AspNetCore` — **550 passed, 0 failed**.
* `dotnet fallout VerifyDocsSnippets` — up to date (102 markdown files).
* `dotnet fallout BenchmarkSmoke` — succeeded (282 benchmarks).
* Integration tests were **not** run: no Docker daemon in this environment. `JobStoreContractTest`,
  `AdoJobStorePagingTest` and `SmokeTestPerformer` compile against the renames and carry new
  group-pattern assertions, but they have not been executed here.

Baselines moved deliberately: `PublicApiTest_Quartz`, `PublicApiTest_Quartz.HttpClient` and
`TracingJobStoreTest_SpanNames`, regenerated and reviewed line by line; no `.received.txt` remains.
Migration-guide rows exist for every removal with its replacement spelled, including appendix rows.


lahma:
Reviewed in full. RULING on the flagged decision: the rename is RATIFIED over the issue's align-on-keys — the group-name return is load-bearing (it is the set of paused-group rows the call writes; an equality matcher pauses a group holding nothing, so keys do not even exist to return, and fetching them would add an unbounded SELECT to a set-based UPDATE). PauseJobGroups saying groups is the honest fix; the complaint was name-vs-shape incoherence, not the shape. Further rulings, all ACCEPTED: (2) coercion-first-then-trim — the subsumption premise was wrong and you proved it before cutting reading paths, with the behavior change boxed; (3) the arity-free NameMatcher and the widened group filters (pre-freeze wire change, correctly taken now); (4) every ratify-not-change verified in code rather than asserted; (5) Replacing(name) as a method — a parameterless preset for doing nothing would be decoration; (8) the arity pair sharing a file is the clearer convention deviation. Rebase over #3633 when I send the notice; merging on green after.

Issue #3636, opened 2026-08-31
The freeze audit's two blocking findings are fixed before the tag


lahma:
The final freeze read-through (alpha5-freeze-audit.md) found two BLOCKING items and one RECORD item promoted to a fix; this issue carries them to one PR.

**B1 — ExternalTransactionJobStore is documented as a choice and cannot be chosen from code.** The ADO internalization removed the only typed selector: UsePersistentStore<T> needs a nameable T, the type is internal, and only the legacy quartz.jobStore.type string still reaches it — while AdoJobStoreOptions.OpenConnection stays public with a doc saying it is read only by that store. Fix: a typed selector in the UseClustering idiom — IPersistentStoreBuilder.UseAmbientTransactions() (name final unless implementation shows better; it selects the internal ExternalTransactionJobStore) — plus the migration-guide prescription updated from the no-longer-compiling UsePersistentStore<ExternalTransactionJobStore> call, and a test that the ambient store is what gets built.

**R1 — IQuartzApiClient drops the applied flag on five members** (Interrupt, InterruptFireInstance, DeleteJob, UnscheduleJob, DeleteCalendar) while its siblings return it, IScheduler returns it for all ten, the HTTP endpoints expose all ten as OperationAppliedResponse, and the in-process client discards with a literal '_ ='. Widening a return type later is breaking, so the five gain the flag now; the two members with no XML doc gain theirs.

B2 (the guide's H-in-five-field claim) is docs-precision folded into #3602's sweep — the combination works via CronScheduleBuilder.Create(s, CronFormat.Unix), which normalizes before hash deferral; only direct ParseWithHash lacks a format overload, which is filed additively.

PR #3640, opened 2026-08-31
The documentation catches up with the freeze

### LINKED ISSUE #3602: The documentation catches up with the freeze
The documentation debt the alpha.5 breaking waves create, paid in one sweep after they merge.

- **`cron-expressions.md` is rewritten** for the new semantics: the restricted-field day rule (with the `?`-is-a-synonym story), gap-end DST behavior beside the existing fall-back documentation, the death of `everyNthWeek` (with the `RecurrenceScheduleBuilder` migration), the rejected forms, and — once #1253's dialect lands — the 5-field Unix format with its 0–7 day-of-week numbering and macros, clearly separated from the 6/7-field Quartz format. `best-practices.md:439` stops advertising fortnight steps.
- **The migration guide gets a consolidation pass.** Every alpha.5 PR carries its own rows; this pass orders them, dedups, and verifies the appendix still answers "the name you would have typed" for every removed or renamed member (the campaign removes more members than any previous alpha — the appendix is the contract).
- **Package readmes and doc snippets regenerate** (`dotnet fallout DocsSnippets`) wherever samples referenced trimmed surface; `VerifyDocsSnippets` green proves it.
- **Release-notes draft** for the tag assembles the freeze story: what was removed, what changed meaning (cron semantics get their own loud section), and the sentence every integrator needs: *this is the last release that breaks you*.

Runs in the docs wave, strictly after the semantic and trim PRs merge; blocks the tag.

lahma:
The documentation debt the alpha.5 waves created, paid in one sweep. Twenty-two PRs each reconciled the
documentation locally and nobody read the whole of it afterwards, so this reads it end to end and fixes
what only a whole-document read can see.

**No product code, no baselines.** `docs/` and one package readme.

Rebased onto `46919393a`. #3641 landed the two product fixes this sweep found and the docs edits that
went with them; the overlapping hunks take its facts, and what is left here is what it did not reach.

## Two documentation bugs, both about daylight saving

`faq.md` said a spring-forward cron trigger "will be skipped, since 2:15 am never occurs that day", and
`cron-expressions.md` that "the time shift can cause a skip or a repeat". Neither is true on either
branch — a cron trigger is never skipped, and 4.x fires at the gap's end where 3.x shifts by the
transition delta. Both are rewritten around what the code does, with the FAQ's sentences version-split
because that page serves both branches. `best-practices.md` attached its Australia/Lord_Howe readings to
`0 30 2 * * ?`, which is not in the gap in a half-hour-delta zone; the expression is named now
(`0 15 2 * * ?`, which is what `CronExpressionDstTests` pins).

## The ADO chapter reads in one order

`## The ADO.NET store is a store, not a base class` was sitting between the serialization sections. It
moves to the head of the ADO run, where the four sections describing the pre-internalization seam already
pointed at it, and the road map's pass 4 now says to read it first. Those four say so themselves rather
than presenting a `protected` member of an internal type as an override point.

The appendix carried three contradicting pairs: `JobStoreSupport` "renamed `AdoJobStoreBase`, a derived
store updates its base list" three rows below `AdoJobStoreBase` "Internal", and the same for
`JobStoreTX` / `JobStoreCMT`. They agree now.

## Where this sits relative to #3641

#3641 fixed the two product defects this sweep turned up — the `CronExpressionBuilder` message that
claimed cron forbids naming both day fields, and `quartz.jobStore.openConnection` being accepted and
silently ignored — and corrected the pages that named them. The overlap resolved in favour of its facts:
the `OpenConnection` reference row says the flat key is read, and the migration guide's `OpenConnection`
sample keeps its chained `UseAmbientTransactions().ConfigureStore(…)` form.

Three places it did not reach still prescribed a type argument for a store that has no public name, and
this fixes them: the removed-constant table's `quartz.jobStore.type` row, the sentence saying
`UseInMemoryStore` and `UsePersistentStore<T>` are how a shipped store is selected, and
`how-tos/custom-job-store.md`. The guide also now says what `UsePersistentStore<T>(configure)` is *for*
now that neither shipped store can be named — a persistent store of your own.

`H` is the other audit item folded in. It does not compose with the five-field Unix form through every
door, and the guide and the cron reference both said it did. Verified against the built assembly with
`dotnet fsi`: `CronScheduleBuilder.Create("H 4 * * 1", CronFormat.Unix)` on a trigger identified as
`nightly` gives `0 13 4 ? * MON`, because it rewrites to the Quartz form before deferring the hash; while
`CronExpression.Parse("H 4 * * 1", CronFormat.Unix)` throws and `ParseWithHash` rejects five fields. Both
pages name the builder as the one route and say a format-taking `ParseWithHash` is filed for 4.1.

The event-id range table listed 6400–6499 and 6500–6599 for two retired plugins, and a bullet described a
message that no longer ships. The ranges stay reserved and the table says so.

## Counts, checked against the baselines

Each was verified by diffing `PublicApiTest_*.verified.txt` rather than by reading:

| Claim | Was | Is |
|---|---|---|
| Types 4.0 added to `Quartz` | "Ninety-four names" | **Ninety-nine** — the block was missing `CronFormat`, `JobTimeoutAttribute`, `ScheduledOneOffJob`, `SchedulerFactoryExtensions`, `SchedulerNotFoundException` |
| `Quartz.JobStore.*` span names | 29, in three places | **33** — and `opentelemetry-integration.md` said "twenty-nine" in its table and "thirty-three" in the paragraph under it |
| Types keeping `[Serializable]` | "Nineteen … the other 30" | **Seventeen**, which is what the guide's own keep table lists and what the 4.x baselines carry |
| `IJobStore` members | "around fifty" | **sixty-odd** (61) |
| Removed-type appendix | "all 100" | a completeness claim that is true, plus which families have one entry rather than a row per name |

Counts I checked and left alone because they are right: `AdoJobStoreBase`'s "162 members and roughly a
hundred `protected` hooks" (162 and 100 exactly), the 47 `StdSchedulerFactory` constants,
`ISchedulerListener`'s twenty-three members, and the 56 `protected virtual` members 3.x's
`JobStoreSupport` had.

## Members a wave removed and a sentence kept

* `JobBuilder<TJob>.OfType(string)` is gone — the trimming section, the runtime-check list and the
  `OfType(JobType)` row all still named it. They say `OfType((JobType) typeName)`.
* `GetDecimal` / `TryGetDecimal` became `Get<decimal>` / `TryGet<decimal>`, which one section said and
  two others denied.

## Gaps the waves left with no documentation at all

Nine campaign commits touched product code and no docs. Six were internal; three were not:

* **`50f99fd4d`** made a duration longer than `Task.Delay` accepts a startup failure on six settings
  (`MisfireHandlerFrequency`, `MisfireThreshold`, `DbRetryInterval`, `TransientRetryInterval`,
  `ClusteringOptions.CheckinInterval`, `QuartzHostedServiceOptions.StartDelay`), plus a lock handler's
  `RetryPeriod` and `StartDelayed`'s argument. It reached no page; the validation section names them.
* **`956191c33`** made a retry whose instant would overflow a `DateTimeOffset` decline rather than throw.
  Added to the guide's retry rules and to `how-tos/retrying-failed-jobs.md`.
* **`2691e6a55`** made `SchemaResourceName` walk the base chain, so a delegate deriving from a shipped
  dialect inherits its script. The guide and the job-stores tutorial both said an override was required.

## The cron reference

Verification pass against merged reality, per the scope adjustment. The restricted-field rule, `?` / `*`
synonymy, `CronFormat.Unix`, the macros and `ParseWithHash` were all already there and correct. What was
missing:

* **The rejected forms had no home on the reference page.** A `## Forms the parser refuses` section lists
  all seven — `1-5W`, `? * L-3` / `? * LW`, `MON,FRI#3`, `5C`, `*/0`, `0-10/120`, `MON/2` — with the
  `RecurrenceScheduleBuilder.Create("FREQ=WEEKLY;INTERVAL=2;BYDAY=MO")` migration for the fortnight and a
  pointer to the pre-upgrade audit query. Every row is pinned by
  `CronExpressionTest.ExpressionsThatSaidOneThingAndDidAnotherAreRejected`.
* **The page had two headings called Examples**, so the deduplicating slugger gave the hash table
  `#examples` and the expression table `#examples-1` — a bare `#examples` resolves, to the one nobody
  means. The hash one is `### Hash examples` now. Nothing in the tree linked either, and the two anchors
  the migration guide records as externally linked (`#h-hash-for-load-distribution`,
  `#building-cron-expressions-programmatically`) are untouched.
* `best-practices.md` already had the fortnight and `MON/2` material and needed only a link to the new
  section, plus the Lord Howe fix above.

The builder's one-day-field rule keeps #3641's wording, which states the union example inline instead of
linking the ambiguous fragment.

## Package readmes

All thirteen checked against the merged surface. One stale reference: `src/Quartz/README.md` advertised
`Quartz.Plugins` as carrying "the interrupt monitor", which is retired. `Quartz.Plugins`,
`Quartz.Plugins.TimeZoneConverter` and `Quartz.Serialization.Newtonsoft` were already current — the
Newtonsoft one correctly describes the shared refused-value set after the revert.

## Left alone

Audit R1 and R3 are API and guard questions rather than documentation. `AGENTS.md`'s job-store paragraph
was outside this PR's scope and #3641 fixed it.

## Verification, after the rebase

```text
dotnet build Quartz.slnx                          0 Warning(s), 0 Error(s)
dotnet test src/Quartz.Tests.Unit                 Passed: 4839, Failed: 0, Skipped: 36
dotnet test src/Quartz.Tests.AspNetCore           Passed: 559, Failed: 0
dotnet fallout VerifyDocsSnippets                 Succeeded
npm run docs:check-links                          220 pages, 1130 internal links, 702 fragments
                                                  no dead links or anchors
npm run docs:check-links-test                     7 pass, 0 fail
```

Before the rebase, `npm run docs:build` also rendered all 221 pages and
`dotnet build src/Quartz.Documentation.Samples` was clean. `node_modules` and the VuePress `dist` were
removed after the last run. Every changed file was checked to be UTF-8 with no BOM and CRLF-only.

Integration tests were not run: this PR touches no product code, and no Docker daemon was available.

Closes #3602

### DIFF
--- a/docs/documentation/best-practices.md
+++ b/docs/documentation/best-practices.md
@@ -450,7 +450,10 @@ for a cron expression with a workaround in it, is usually the sign.
 Monday there. 4.x rejects it, because the fortnight's phase was anchored to whatever last asked the
 expression a question: a misfire, a restart or a failover recomputed it from a different day and
 moved it. `FREQ=WEEKLY;INTERVAL=2;BYDAY=MO` anchors on the trigger's start time instead, so the
-fortnight is a property of the trigger rather than of the caller.
+fortnight is a property of the trigger rather than of the caller. It is one of seven shapes 4.x
+refuses that 3.x accepted and then quietly reinterpreted;
+[Forms the parser refuses](/documentation/quartz-4.x/cron-expressions#forms-the-parser-refuses) lists
+them, and the ones to audit a database for before upgrading.
 
 ### The two daylight saving rules
 
@@ -462,7 +465,8 @@ is not what this implementation does, and the difference matters enough to state
   never skipped. *Where* it fires differs by version. On 4.x it fires at the **end of the gap**, the
   instant the clocks moved: a daily 02:30 over a 02:00–03:00 gap fires at 03:00. On 3.x the fire is
   shifted forward by the transition delta instead, to 03:30. In a zone whose delta is not a whole
-  hour — Australia/Lord_Howe — the two rules read 02:30 and 02:45. Only 4.x's answer is an instant
+  hour — Australia/Lord_Howe, where 02:00 becomes 02:30 — a daily `0 15 2 * * ?` reads 02:30 on 4.x
+  and 02:45 on 3.x. Only 4.x's answer is an instant
   the expression itself matches: ask `IsSatisfiedBy` about the fire time and 4.x says yes where 3.x
   says no.
 - A wall-clock time that **occurs twice** on a fall-back day fires **once**, at the first of the two
--- a/docs/documentation/faq.md
+++ b/docs/documentation/faq.md
@@ -292,28 +292,58 @@ since its firing on Saturday morning at 10:00 am will not be 24 hours,
 but will instead be 23 or 25 hours respectively.
 
 There is one additional point users must understand about CronTrigger with
-respect to daylight savings. This is that you should take careful thought
-about creating schedules that fire between midnight and 3:00 am (the critical
-window of time depends on your trigger's locale, as explained above).
-The reason is that depending on your trigger's schedule, and the particular
-daylight event, the trigger may be skipped or may appear to not fire for an
-hour or two. As examples, say you are in the United States, where daylight
-savings events occur at 2:00 am. If you have a CronTrigger that fires every
-day at 2:15 am, then on the day of the beginning of daylight savings time
-the trigger will be skipped, since, 2:15 am never occurs that day. If you
-have a CronTrigger that fires every 15 minutes of every hour of every day,
-then on the day daylight savings time ends you will have an hour of time
-for which no triggers will occur, because when 2:00 am arrives, it will become
-1:00 am again, however all of the firings during the one o'clock hour have
-already occurred, and the trigger's next fire time was set to 2:00 am
+respect to daylight savings, and it is worth reading carefully because Java
+Quartz's own page says something Quartz.NET does not do. A CronTrigger is
+**never skipped** by a daylight saving transition, and never fires twice for
+one scheduled occurrence. What a transition decides is *which instant* a
+wall-clock time that is missing or repeated resolves to, and the answer
+depends on whether the expression names a fixed time of day or an interval.
 
-* hence for the next hour no triggering will occur.
+A **fixed-time** expression is one whose second, minute and hour fields are
+plain values or comma lists of plain values - `0 15 2 * * ?`, or
+`0 0,30 2 * * ?`. Say you are in the United States, where daylight saving
+events occur at 2:00 am, and you have a CronTrigger that fires every day at
+2:15 am:
+
+* On the day daylight saving time **begins**, 2:15 am does not exist. The
+  trigger fires once anyway. On **4.x** it fires at the *end of the gap* -
+  3:00 am, the instant the clocks moved - which is the one in-gap instant the
+  expression itself matches, so `IsSatisfiedBy` agrees with the fire time. On
+  **3.x** the fire is shifted forward by the transition's delta instead, to
+  3:15 am. A zone whose delta is not a whole hour - Australia/Lord_Howe - splits
+  the same two ways, on its own numbers. An expression matching several of the
+  swallowed wall clocks, such as `0 0,15,30,45 2 * * ?`, still fires once: they
+  all name the same instant.
+* On the day daylight saving time **ends**, 2:15 am occurs twice. The trigger
+  fires once, at the first of the two occurrences. This is the same on both
+  versions.
+
+An **interval** expression - one with a wildcard, a step or a range in the
+second, minute or hour field, such as a trigger that fires every 15 minutes of
+every hour of every day - is where the versions differ most. On **4.x** it
+keeps firing through the repeated hour, so both passes of it run. On **3.x**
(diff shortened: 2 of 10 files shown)


lahma:
Reviewed in full. This is the read nobody had done — three cross-wave contradiction pairs, five counts corrected against the baselines themselves, complete removed-member coverage proven from the diff side, three undocumented behaviors given their paragraphs, and the Lord Howe example corrected against the pinned test rather than by eye. Deviations accepted: the section move (anchors heading-derived, checker-proven), and the three unlisted files carrying the same gaps. Your three product-code finds are routed to the #3636 fix PR (its agent owns those files): the builder's false reason string, the accepted-and-ignored openConnection key, and the AGENTS.md job-store sentence. Merge order stands as you flagged: #3636 lands first, then this. Merging on green after it.

lahma:
Rebase reviewed. Catching the paragraph that would have merged silently wrong outside any conflict marker — the no-flat-key claim #3641's bridge falsified — is exactly the vigilance a generated-prose merge needs; taking #3641's better sentences where they were better, keeping yours where yours were, and renaming the second Examples heading so the slugger trap dies for the next writer are all right. Merging on green as the campaign's final PR.

PR #3641, opened 2026-08-31
The freeze audit's two blocking findings are fixed

### LINKED ISSUE #3636: The freeze audit's two blocking findings are fixed before the tag

lahma:
The freeze audit's two blocking items, plus the three addenda the docs sweep (#3640) turned up in the same files.

Closes #3636

## B1 — the ambient-transaction store can be chosen from code

`ExternalTransactionJobStore` was documented as one of the two shipped stores and reachable only through the `quartz.jobStore.type` string: internalizing the ADO hierarchy left `UsePersistentStore<T>` needing a nameable `T`, while `AdoJobStoreOptions.OpenConnection` stayed public with a doc naming a store nothing typed could select.

`IPersistentStoreBuilder.UseAmbientTransactions()` selects it. It is parameterless — both stores are constructed from the same `AdoJobStoreDependencies`, so *which type is built* is the whole of the difference, and the one setting that distinguishes them (`OpenConnection`) already lives on `AdoJobStoreOptions` and is reachable through `ConfigureStore`.

Because registration is first-wins and the choice is made inside the callback, `UsePersistentStore(configure)` now registers the store **after** running the callback. `UsePersistentStore<T>(configure)` still registers `T` first, and reports the contradiction if the callback also calls the selector rather than silently keeping `T` — a scheduler that commits transactions its caller believed somebody else owned is not something to resolve quietly.

- The legacy string path is untouched and is now *measured against* the new member: `ConfiguredTypeNamesResolveTest.TheAmbientTransactionStoreIsOneStoreWhicheverWayItIsChosen` builds the store from `quartz.jobStore.type` and from `UseAmbientTransactions()` and asserts both arrive at the same type.
- `AdoJobStoreOptions.OpenConnection`'s doc now names the typed route, as do `configuration/reference.md` and `tutorial/job-stores.md`.
- Migration guide: the `UsePersistentStore<ExternalTransactionJobStore>(…)` sample (which did not compile) becomes the new member, the "code that names the type has to be updated" line says what to update it *to*, and the appendix row for the type names the selector.
- Baseline: `PublicApiTest_Quartz` moves by exactly one added member, reviewed and accepted.

**Reviewer note — overlap with #3640.** The surrounding section (guide ~:6502-6558) is being rewritten by the docs sweep; my edit is confined to the prescription lines (the diff sample, the "updated to what" sentence, the appendix row) so that the two can be sequenced with this one first.

## R1 — the dashboard client reports the applied flag for all ten mutations

`IScheduler` answers `bool` for each of its ten "did this find anything to act on" mutations and the HTTP API puts all ten on the wire as `OperationAppliedResponse`, but `IQuartzApiClient` returned it for five and discarded it for five with a literal `_ =`. `Interrupt` was renamed to the scheduler's verb in this delta, which is what turns the asymmetry into two members that read as the same operation and are not.

`Interrupt`, `InterruptFireInstance`, `DeleteJob`, `UnscheduleJob` and `DeleteCalendar` return `ValueTask<bool>`, in the shape the five siblings already use, and gain the XML doc the siblings have. Widening a public interface's return type after the tag would be a binary and source break for any implementer, which is why it happens now.

The pages use the answer rather than discarding it one level up:

- deleting a job, unscheduling a trigger or deleting a calendar that a cluster peer already removed reports "was not deleted — it no longer exists" instead of claiming success, and the two detail pages navigate away only when something was actually removed;
- **Currently Executing** now injects `ToastService` and says when the firing behind the button had already finished — the case its one-second refresh makes routine.

**No wire change.** The endpoints call `IScheduler` directly and already answered with `OperationAppliedResponse`; only the dashboard's own client interface moved. Baseline `PublicApiTest_Quartz.Dashboard` moves by the five signatures, reviewed and accepted, with a migration row and a paragraph in "The client speaks the scheduler's verbs".

## Addenda from the #3640 sweep

1. **`CronExpressionBuilder` blamed cron for its own rule.** The refusal said "Cron expressions do not support specifying both day-of-month and day-of-week", which is false since the union rule was ratified: `0 0 0 13 * FRI` parses and fires on every Friday *and* the 13th. The builder's one-day-field constraint stays; the message, the class remark and `cron-expressions.md` now say whose rule it is and what to write instead. The test pins the new reason and asserts the expression the builder declines to compose is a valid one.
2. **`quartz.jobStore.openConnection` had no reader.** 3.x set it by writing the property onto the store object; here the key sits under a supported prefix, so a migrated file was accepted and silently ignored. **Deviation from the brief:** the brief asked for a targeted *rejection* on the grounds that "OpenConnection is a delegate no string can express". It is a `bool` (`AdoJobStoreOptions.OpenConnection`), 3.x's `JobStoreCMT.OpenConnection` was a settable `bool` property, and rejecting the key would break a 3.x file that meant something by it — so it is bridged onto the option instead, with two pin tests and a row in the reference's flat-key table. (`AdoJobStoreOptions.IsTransient` is the setting that genuinely cannot come from a string.)
3. **AGENTS.md** — the job-store paragraph now says the three ADO types are internal and names the two ways to select them. One sentence; the file is 31,072 bytes against the 32,768 cap.

## Verification

```
dotnet build Quartz.slnx                              0 warnings, 0 errors
dotnet test src/Quartz.Tests.Unit                     4839 passed, 36 skipped, 0 failed
dotnet test src/Quartz.Tests.AspNetCore                559 passed, 0 failed
dotnet fallout VerifyDocsSnippets                     up to date (102 markdown files)
dotnet test src/Quartz.Tests.Integration
  --filter ExternalTransactionJobStoreContractTest      93 passed (SQLite file, no container)
```

The rest of the integration suite was not run: it wants a container per dialect and this worktree is disk-constrained. Nothing here touches store internals only an integration test reaches — the ADO change is which type the builder registers, and the contract test above covers that store's behaviour. `Quartz.Benchmark` names none of the types touched, so `BenchmarkSmoke` is not implicated.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -270,7 +270,7 @@ Quartz.NET is an enterprise job scheduling library. The core domain model:
 ### Job Stores
 
 - **`RAMJobStore`** (`Quartz.Impl`) — in-memory, volatile. Default.
-- **`AdoJobStoreBase`** → `LocalTransactionJobStore` / `ExternalTransactionJobStore` (`Quartz.Impl.AdoJobStore`) — ADO.NET-based persistence with database-specific delegates (`SqlServerDelegate`, `PostgreSQLDelegate`, `OracleDelegate`, `MySQLDelegate`, `SQLiteDelegate`, `FirebirdDelegate`). On 3.x these are `JobStoreSupport` → `JobStoreTX` / `JobStoreCMT`.
+- **`AdoJobStoreBase`** → `LocalTransactionJobStore` / `ExternalTransactionJobStore` (`Quartz.Impl.AdoJobStore`) — ADO.NET-based persistence with database-specific delegates (`SqlServerDelegate`, `PostgreSQLDelegate`, `OracleDelegate`, `MySQLDelegate`, `SQLiteDelegate`, `FirebirdDelegate`). On 3.x these are `JobStoreSupport` → `JobStoreTX` / `JobStoreCMT`. All three are **internal**: `UsePersistentStore` builds the local one and `UsePersistentStore(s => s.UseAmbientTransactions())` the other, the driver delegate is the seam, and `quartz.jobStore.type` still names either by string.
 
 ### Database scripts
 
--- a/docs/documentation/quartz-4.x/configuration/reference.md
+++ b/docs/documentation/quartz-4.x/configuration/reference.md
@@ -762,6 +762,7 @@ Two differences are worth knowing:
 | `quartz.jobStore.makeThreadsDaemons` | `JobStore:UseBackgroundThreads` |
 | `quartz.jobStore.clustered` | `JobStore:Clustering:Enabled`, or `UseClustering()` |
 | `quartz.jobStore.acceptEnlistedTransactions` | `JobStore:AcceptEnlistedTransactions` |
+| `quartz.jobStore.openConnection` | `JobStore:OpenConnection`, read only by the ambient-transaction store |
 | `quartz.jobStore.clusterCheckinInterval` | `JobStore:Clustering:CheckinInterval` |
 | `quartz.jobStore.clusterCheckinMisfireThreshold` | `JobStore:Clustering:CheckinMisfireThreshold` |
 | `quartz.jobStore.dataSource` | set for you by the database methods |
--- a/docs/documentation/quartz-4.x/configuration/reference.md
+++ b/docs/documentation/quartz-4.x/configuration/reference.md
@@ -177,7 +177,7 @@ services.AddQuartz(q => q.UsePersistentStore(store =>
 | `UseBackgroundThreads` | bool | `false` | Runs the misfire handler and cluster manager on background threads, which do not keep the process alive. These two are the only real threads Quartz creates. |
 | `SchemaProvisioning` | `SchemaProvisioning` | `Validate` | What the store does about its schema at startup: `None` does nothing, `Validate` verifies the expected tables exist, `CreateIfMissing` creates whatever is missing and then verifies. |
 | `SelectWithLockSql` | string? | none | Overrides the row-lock statement, defaulted to SQL Server's `WITH (UPDLOCK,ROWLOCK)` form when that is the database. Read only when the store builds a database-locking handler for itself — see [Locking](#locking). |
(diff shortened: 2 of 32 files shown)


lahma:
Reviewed in full. Deviation 1 is the PR's best moment: verifying the 3.x source before obeying the reject instruction — OpenConnection is a bool JobStoreCMT bound from that very key, so bridging preserves what a 3.x file meant while rejection would have broken it; my relay of the 'delegate' premise was the error and your check caught it. The UsePersistentStore ordering fix (callback-then-register, with the generic overload throwing on a conflicting selector instead of silently winning) is the right first-wins hygiene. The applied-flag alignment with the interface-level ratifying paragraph, the ToastService addition it forced, and the two one-line docs edits are all accepted; the undocumented-members remainder is correctly out of scope. Merging on green; #3640 follows and will reconcile the reference.md overlap on its rebase.

PR #3657, opened 2026-09-02
The 4.0 upgrade is rehearsed on data a released 3.20 wrote

### LINKED ISSUE #3645: The 4.0 upgrade is rehearsed on data a released 3.20 wrote
**Milestone:** 4.0.0-beta.1 · **Labels:** tests, infra, docs, 4.x · **Baseline:** none
**Closes:** —

### Why
N4-1 is the one FIX-L of the whole campaign: **no test ever applies the 4.0 migration to a schema with 3.x
rows in it.** All six dialects migrate an *empty* 3.16 schema and then write new 4.0 rows on top
(`src/Quartz.Tests.Integration/Impl/AdoJobStore/MigrationScriptTest.cs` — `BuildMigratedSchemaAsync` inserts
nothing, and `MigratedSchemaWorkload` reads back only rows 4.0 just wrote). So the single claim beta.1 most
needs to make — that a 3.20 production database survives the upgrade — is untested on every dialect. The
same rehearsal produces the fixtures N4-5 wants, because the "verbatim 3.x" JSON literals in the unit suite
are hand-transcribed rather than captured from a 3.20 run.

### Evidence
- **N4-1.** `MigrationScriptTest.cs:116-253` runs two routes per dialect over a frozen 3.16 DDL copy under
  `src/Quartz.Tests.Integration/SchemaBaselines/3.16/tables_<dialect>.sql`; the end-state assertion is
  structural (tables, columns, indexes) and the only `INSERT` in the file serves
  `FreshInstallHonoursDropSwitch`. Untested on every dialect, therefore: a 3.x `JOB_DATA` blob read after
  migration, a 3.x `QRTZ_CALENDARS` row, a `BLOB_TRIGGERS` row, a `QRTZ_PAUSED_TRIGGER_GRPS` row governing a
  trigger stored afterwards, a 3.x `QRTZ_FIRED_TRIGGERS` row surviving into recovery, a 3.x
  `QRTZ_SIMPROP_TRIGGERS` row read by 4.0's persistence delegate, and `RETRY_POLICY` / `RETRY_ATTEMPT`
  reading as null on a migrated row.
- **N4-5.** `src/Quartz.Tests.Unit/Simpl/LegacyJsonPayloadTest.cs:33-35` describes its twelve literals as
  "verbatim output from 3.x" — a comment, not a fact the suite can re-derive; there is no capture harness and
  no fixture directory (`git ls-tree -r --name-only origin/main -- src/Quartz.Tests.Unit | grep -i "serial\|json"`
  lists only `.cs` and `Verify/*.verified.txt`, which are 4.x output). The Newtonsoft **default** shape — the
  plain object graph 3.x wrote with `RegisterTriggerConverters = false` — has no 3.x-sourced sample at all;
  the fixture that pins it captured its literals at a 4.x commit.
- **N4-2 / N4-3, guide misses.** `CronExpressionBuilder.OnDaysOfWeek`, `.OnDayOfWeekRange` and
  `.OnDayOfWeekIncrements` were renamed to `With*` and have **no** migration-guide row:
  `grep -c "\bOnDaysOfWeek\b" docs/documentation/quartz-4.x/migration-guide.md` → 0, while the type itself
  appears eleven times. `Get/SetPreviousFireTimeUtc` on all five `*TriggerImpl` are covered only in prose
  that spells three sibling names, and `SchedulerConstants.FailedJobOriginalTriggerScheduledFiretime` is
  written with an ellipsis at `migration-guide.md:8571`, so neither is searchable. With those rows the
  release-notes claim becomes exactly true: **606 removed or renamed public names checked mechanically, 605
  with a row, a family entry or a stated exemption.**
- **N4-14.** The one uncovered name is `Quartz.Dashboard.Components._Imports` (and its `Pages` sibling), a
  Razor compiler artefact present in 3.x's `PublicApiTest_Quartz.Dashboard.verified.txt` and absent from
  main's. It is not a contract; it belongs out of the baseline, not in the guide.
- **N4-4, binary blobs.** `docs/documentation/quartz-4.x/packages/json-serialization.md:129-155` tells a
  deployment to migrate off `BinaryFormatter` while still on 3.x and never says which blobs rewrite
  themselves: `QRTZ_JOB_DETAILS.JOB_DATA` does (on re-store, or after a firing with
  `[PersistJobDataAfterExecution]`), `QRTZ_TRIGGERS.JOB_DATA` only when the trigger is stored again,
  `QRTZ_CALENDARS.CALENDAR` only on `AddCalendar` — so a calendar added once at deployment is **never**
  rewritten — `BLOB_TRIGGERS.BLOB_DATA` never (and cannot be migrated from 4.x at all), and anything on a
  paused or far-future trigger never until it resumes. `grep -rn "BinaryFormatter" docs/documentation/quartz-4.x/`
  → 13 hits, none enumerating the columns and none offering a detection query.
- **N4-6, mixed window.** `docs/documentation/quartz-4.x/operations.md:79-177` is a good and honest account
  of a mixed 3.x/4.0 cluster, and `grep -n "RETRY\|retry"` over it returns only the command-timeout section
  at `:595-619`. Two facts are missing: a 3.x node never honours a retry policy, and a 3.x node that
  *reschedules* a trigger drops it, because 3.x's `ReplaceTrigger` is delete + insert
  (`origin/3.x:src/Quartz/Impl/AdoJobStore/JobStoreSupport.cs:2318-2349`) and its `INSERT` names no
  `RETRY_POLICY`/`RETRY_ATTEMPT`; an in-place update is safe. The same paragraph calls the index realignment
  "section 5" where the generated scripts number it 6, and calls sections 1-4 "the required ones" where the
  header marks 1-5 required.
- **N4-8, wording.** `migration-guide.md:260` says "4.x validates the whole schema at startup".
  `ValidateSchema` (`src/Quartz/Impl/AdoJobStore/StdAdoDelegate.cs:688`) issues `SELECT 1 FROM <table>` for
  each of the twelve names in `AdoConstants.AllTableNames` and checks no column, so a half-applied script or
  a mis-typed prefix on one table starts cleanly and fails later with a raw provider error.

### Done means
- **A new console project `src/Quartz.Tests.Integration.Seeder/`** — `net10.0`, in `Quartz.slnx` so CI
  compiles it and it rots loudly, `IsPackable=false`, coverage-excluded like the examples — referencing the
  **released NuGet package `Quartz` 3.20.0** (plus `Quartz.Serialization.Json` 3.20.0 for the Newtonsoft
  run), never the project reference. Central package management applies, so the pin is a
  `VersionOverride="3.20.0"` on those `PackageReference` items (or an explicit `Directory.Packages.props`
  entry) and never floats: the point of the fixture is a *named* released version.
- The seeder takes `--dialect`, `--connection-string`, `--table-prefix` and `--serializer json|stj`, applies
  the vendored 3.20 fresh-install DDL (`src/Quartz.Tests.Integration/SchemaBaselines/3.20/tables_<dialect>.sql`,
  copied from `origin/3.x:database/tables/` at `v3.20.0`), builds a 3.20 scheduler on the ADO store, and
  writes the Part D seed set: one trigger of each of the five persisted families; a blob-persisted custom
  trigger, so a `BLOB_TRIGGERS` row exists; calendars of all six kinds including a chained pair; a job data
  map holding every value type the 4.0 write gate admits plus a `Dictionary<string,string>` (the #3582 shape)
  and a `JobKey`; a paused trigger group **and** a paused job group with a trigger and a job stored *after*
  the pause; a deliberately orphaned `QRTZ_FIRED_TRIGGERS` row; a `QRTZ_SCHEDULER_STATE` row; and a
  `PREFERRED_NODE` pin.
- It finishes by writing `seed.json` — the manifest the assertions name their expectations from — and dumping
  the blob columns to `src/Quartz.Tests.Unit/TestData/Legacy/3.20/`, which `LegacyJsonPayloadTest` then reads
  instead of its inline literals (N4-5). Both serializer runs are captured.
- **`UpgradeRehearsalTest`, beside `MigrationScriptTest`.** Per dialect: run the seeder out of process
  against the existing Testcontainers environment under table prefix `QRTZU_`; apply
  `database/migrations/4.0/schema_30_to_40_upgrade_<dialect>.sql` through the existing `ExecuteScriptAsync` /
  `MigrationScript` helpers; assert the structural equivalence `AssertSchemaMatchesAsync` already asserts;
  **then** start a 4.0 scheduler on the migrated prefix and assert against the manifest — every trigger reads
  back with its schedule, misfire instruction, calendar name, priority and pin; every calendar answers
  `IsTimeIncluded` as the manifest says; the job data map reads back value-for-value; the paused groups and
  the born-paused items are still paused; `RETRY_POLICY`/`RETRY_ATTEMPT` read as no policy and 0; the
  orphaned fired row is recovered rather than stranded; every trigger fires within a deadline.
- Each test carries `[Category("db-<dialect>")]` so the existing `pr-integration-<dialect>` workflows pick it
  up unchanged.
- **Commit order:** seeder + fixtures + the SQLite leg first (S, no container — that leg alone retires most
  of the risk and produces the fixtures), then SQL Server, PostgreSQL, MySQL, then Oracle, then Firebird (L:
  the 3.x DDL's unprefixed triggers need the prefix surgery `MigrationScriptTest.cs:766-769,809` already
  does for 3.16).
- Docs, in the same PR: the N4-2/N4-3 guide rows with the names spelled out; `_Imports` excluded from the
  Dashboard baseline generator (N4-14); a table on `packages/json-serialization.md` saying which binary blob
  columns rewrite themselves and which never do, with per-dialect detection SQL over the four columns (a
  binary payload starts `0x00 0x01 0x00 0x00 0x00`, a JSON one `0x7B`); the `RETRY_*` paragraph and the
  section-numbering correction on `operations.md`; and `migration-guide.md:260` reworded to "validates that
  every table it will read exists".
- Report which legs actually ran. Integration legs need Docker; if it is unavailable for a dialect, say so
  explicitly rather than skipping silently.

### Not in this PR
- The ordered upgrade runbook (N4-7 / D18) — P11, with the rest of the new pages.
- The 3.x-side companions (N4-9 … N4-13) — P13, on the `3.x` branch.
- Any change to `ValidateSchema` itself: beta.1 corrects the sentence; a column check is a 4.x candidate.
- The store-contract gaps (N2b) — P3.

lahma:
No test ever applied the 4.0 migration to a schema with 3.x rows in it. All six dialects migrated an
*empty* 3.16 schema and then wrote new 4.0 rows on top, so the single claim beta.1 most needs to make —
that a 3.20 production database survives the upgrade — was untested everywhere. This makes it tested,
on all six dialects, against rows a **released `Quartz` 3.20.0** wrote.

## What changed

**`src/Quartz.Tests.Integration.Seeder/`** — a `net10.0` console project on the released `Quartz`
3.20.0, `Quartz.Serialization.Json` 3.20.0 and `Quartz.Serialization.SystemTextJson` 3.20.0 packages,
pinned with `VersionOverride` so the fixture is a *named* released version. It is a separate process
because it has to be: those packages' assembly identity is this repository's own, so nothing can
reference both. It is in `Quartz.slnx` (so `Compile` builds it and it rots loudly), `IsPackable=false`,
and coverage-excluded in `sonar.yml` beside the canary.

It seeds one trigger of each of the five persisted families, the same families again blob-stored, a
trigger carrying `EXECUTION_GROUP` and `PREFERRED_NODE`, all six calendar kinds plus a chained pair, a
job data map holding every value type 4.0's write gate admits (and, on a job of its own, a `JobKey`,
which the gate would refuse), a paused trigger group and a paused job group each with a member stored
*after* the pause, and a `QRTZ_FIRED_TRIGGERS` row abandoned by killing the process mid-firing. It
writes `seed.json` — read from the tables, not from its own intentions — and dumps the blob columns
byte for byte.

**`UpgradeRehearsalTest`**, beside `MigrationScriptTest`, one `[Test]` per dialect carrying
`[Category("migrations")]` and `[Category("db-<dialect>")]` so the existing `pr-integration-*` workflows
pick it up unchanged. Per dialect: build the 3.20 schema under `QRTZU_`, seed it twice (once per
serializer, under a scheduler name each), run `schema_30_to_40_upgrade_<dialect>.sql`, assert the
structural equivalence `MigrationScriptTest` already asserts, then start a 4.0 scheduler and check every
row against the manifest — schedule, misfire instruction, calendar name, priority, pin, trigger state,
the job data map value for value, every calendar answering `IsTimeIncluded` as 3.20 answered it,
`RETRY_POLICY`/`RETRY_ATTEMPT` empty on every migrated row, the abandoned firing recovered rather than
stranded, and every runnable trigger firing.

**The fixtures** land in `src/Quartz.Tests.Unit/TestData/Legacy/3.20/`, and `LegacyJsonPayloadTest`
reads them. Its own literals stay where they are an *older* 3.x shape — timestamps with no zone marker,
a map holding a null, triggers from before pinning and before #3294 — and the three that were
shape-identical to the capture have gone.

**Documentation**: the missing `CronExpressionBuilder` `On*` → `With*` rows, the `Get`/`Set` fire-time
family spelled out, `SchedulerConstants.FailedJobOriginalTriggerScheduledFiretime` de-ellipsised, the
`ValidateSchema` sentence corrected, the binary-blob rewrite table with per-dialect detection SQL, and
the mixed-window `RETRY_*` paragraph with the section-numbering correction.

## Verification

Every leg was run locally, one dialect at a time. Docker was available throughout.

| Leg | Result |
|---|---|
| `dotnet build Quartz.slnx` | succeeded, 0 warnings, 0 errors |
| `dotnet test src/Quartz.Tests.Unit` | Failed: 0, Passed: 4893, Skipped: 36, Total: 4929 |
| `dotnet test src/Quartz.Tests.AspNetCore` | Failed: 0, Passed: 559, Total: 559 |
| `UpgradeRehearsalTest` sqlite | Passed 1 (3 s), no container |
| `UpgradeRehearsalTest` sqlServer | Passed 1 (5 s) |
| `UpgradeRehearsalTest` postgres | Passed 1 (2 s) |
| `UpgradeRehearsalTest` mysql_innodb | Passed 1 (23 s) |
| `UpgradeRehearsalTest` oracle | Passed 1 (13 s) |
| `UpgradeRehearsalTest` firebird | Passed 1 (17 s) |
| `MigrationScriptTest` + `UpgradeRehearsalTest`, sqlite filter | Passed 4, Skipped 20 (the container legs, ignored without a connection string) |
| `npm run docs:check-links` | 220 pages, 1131 internal links, 703 fragments, no dead links or anchors |
| `dotnet fallout VerifyDocsSnippets` | up to date, 103 markdown files checked |

The legs are fast because the fixtures fire on one-second intervals and the seeder is warm; container
start-up sits in `OneTimeSetUp`, outside the reported duration. Three gates make a vacuous pass
impossible: `AssertManifestIsWorthAsserting` refuses a manifest that arrived empty, the trigger row
count is tied to `manifest.Triggers.Count + 1`, and `AssertEveryTriggerFiresAsync` waits on real firings.

No product code changed, so no public API baseline moves.

## What a reviewer has to decide

1. **A paused job group does not survive the upgrade.** 3.x's ADO store records nothing when it pauses
   one — `IsJobGroupPaused` returns a hard-coded `false` and `PauseJobs` only pauses the individual
   triggers — so `QRTZ_PAUSED_JOB_GRPS` is empty afterwards and the group reads as running. The
   individual triggers stay paused, so nothing starts running that was not; what is lost is the group's
   pause, so a trigger added to it after the upgrade is born running. The rehearsal asserts the true
   behaviour and the migration guide now has a paragraph. Worth deciding whether it also deserves a row
   in the appendix's table.

2. **The blob-trigger case is a stock trigger type, not an application's own.** `QRTZ_BLOB_TRIGGERS`
   normally holds a type the application declares, and a blob naming an assembly the reader does not
   have proves only that it cannot be loaded. So `BlobStorageOverride` declines the persistence delegate
   for one trigger group, which puts a *shipped* trigger type through the blob path — real 3.20 bytes,
   readable by 4.0. What is not covered, and cannot be from here, is a blob naming a type the reader
   does not ship. Both files say so.

3. **The seeder is not a `ProjectReference` of the test project**, so `dotnet test` on a tree that has
   not had the solution built fails the six rehearsal tests with a message naming the build command,
   rather than skipping. A `ReferenceOutputAssembly="false"` reference would build it automatically but
   risks putting two `Quartz` assemblies in one output directory; I judged the explicit failure safer.
   CI builds the solution, so CI is unaffected.

## Findings, recorded rather than fixed

- **3.20 writes a `BLOB_TRIGGERS` payload for a daily-time-interval trigger that 3.20 itself cannot read
  back.** With the settings a 3.x deployment gets by default (`JsonObjectSerializer.RegisterTriggerConverters`
  is `false`), the trigger is written as a plain object graph whose `StartTimeOfDay` and `EndTimeOfDay`
  are `Quartz.TimeOfDay` objects, and `TimeOfDay` has neither a parameterless constructor nor a
  `[JsonConstructor]`. `SelectTrigger` throws *"Unable to find a constructor to use for type
  Quartz.TimeOfDay"* against a row it just wrote. This is a 3.x defect with no 4.0 half — no such row
  ever worked — so the seeder skips that one combination and `BlobStorageOverride.Families` records why.
  It is a candidate for the 3.x companion PR.

- **`Quartz.Serialization.Json` 3.20.0 ships no `net10.0` asset** (net462/net472/netstandard2.0 only).
  It resolves through netstandard2.0 and works; noting it because the issue assumed otherwise.

## Deviations from the issue

- **The 3.20 fresh-install DDL is applied by the test, not by the seeder**, for every dialect. Each
  dialect's script is written for that database's own command-line client — `GO`, `/`, `SET TERM` — and
  the rehearsal runs it through exactly that client inside the container, which is what
  `MigrationScriptTest` already does and the only way to be sure the script a user is told to run is the
  script that ran. The seeder keeps a `--schema` option for SQLite, which is what the documented
  fixture-capture command uses, so the path is exercised. `SchemaScript.cs` explains it.

- **N4-14 needed no change.** `src/Quartz.Tests.AspNetCore/PublicApiTest.cs` already excludes
  `_Imports` (`type.Name == "_Imports"`), and `PublicApiTest_Quartz.Dashboard.verified.txt` on `main`
  contains zero occurrences. The name only survives in the **3.x** baseline, which is P13's ground.

- **The `JobKey` job data value is on a job of its own**, not on the job every trigger points at. It is
  outside 4.0's write gate, and one unreadable entry on the worker's map would take the whole rehearsal
  down instead of failing one focused assertion.

- **All six legs are in one file** rather than added dialect by dialect, but they were *verified* in the
  issue's order: SQLite, then SQL Server, PostgreSQL, MySQL, then Oracle, then Firebird. Oracle and
  Firebird needed no special handling beyond the SQLite trigger-prefix surgery `MigrationScriptTest`
  already had — the 3.20 Firebird script names its constraints `*_QRTZ_*`, which the existing prefix
  substitution retargets on its own.

Closes #3645


lahma:
Coordinator commit: the Firebird leg failed on a metadata deadlock (SQLSTATE 40001) because the assembly runs fixtures in parallel and this fixture's DDL overlapped `MigrationScriptTest`'s on the one Firebird database — both fixtures failed. `[NonParallelizable]` on `UpgradeRehearsalTest` serialises it; CI re-running.

Issue #3624, opened 2026-08-31
A recurrence schedule cannot be written in a scheduling file


lahma:
Found while writing #3601's XML-freeze capability table (PR #3623): JsonSchedulingDataProcessor.BuildSchedule handles Simple, Cron, CalendarInterval and DailyTimeInterval — the RecurrenceTrigger (RFC 5545 rules, a 4.0 flagship) cannot be declared in quartz_jobs.json at all, and never could in XML. The freeze documentation now states this plainly for both formats. The fix is additive (a new trigger kind in the JSON format plus its round-trip tests — the file format is extensible without breaking readers), so it does not need the alpha.5 window; it does need doing before anyone writes the 'file-based scheduling covers every trigger type' sentence the docs currently avoid. The rule expression is already a string (FREQ=WEEKLY;INTERVAL=2;BYDAY=MO), so the JSON shape is one field plus start-anchor semantics.

PR #3658, opened 2026-09-02
Every documented core behaviour has a test, and a JSON schedule keeps the scheduler's clock

### LINKED ISSUE #3643: Every documented core behaviour has a test, and a JSON schedule keeps the scheduler's clock
**Milestone:** 4.0.0-beta.1 · **Labels:** bug, tests, docs, 4.x · **Baseline:** none
**Closes:** —

### Why
N2a mapped 218 scheduler-core guarantees to the test that pins them: 194 are pinned, 18 are not, and six
documented statements contradict the code. Beta.1's rule is that a documented behaviour either gets a pin
or gets named as a caveat, with no third option. One of the eighteen is a real behavioural defect —
JSON-declared triggers read the wall clock where XML-declared ones read the scheduler's `TimeProvider`
(G-3) — and four of the unpinned ones are the migration guide's own "silent behaviour changes" in job-data
reading, which is exactly the set a 3.x upgrader is told to check (G-8).

### Evidence
- **G-3, behaviour.** `src/Quartz/Configuration/JsonSchedulingHelper.cs:183` is `var startTime =
  DateTimeOffset.UtcNow;` and `:212` is `TriggerBuilder.Create()` with no clock, so a JSON-declared trigger
  is anchored to real time *and* carries `TimeProvider.System` for its own misfire and retry arithmetic. The
  XML processor does the opposite — `src/Quartz/Xml/XmlSchedulingDataProcessor.cs:373`
  (`timeProvider.GetUtcNow()`) and `:439` (`TriggerBuilder.Create(timeProvider)`) — and says so in a test,
  `src/Quartz.Tests.Unit/Xml/XmlSchedulingDataProcessorTest.cs:452`. This is the `Quartz:Schedule` section of
  `appsettings.json`, reached from `AddQuartz` and from `QuartzSchedulerBuilder`.
- **G-1, docs.** `src/Quartz/Diagnostics/HealthChecks/QuartzHealthCheck.cs:74` returns *degraded* for a
  `Created` scheduler the application starts and `:77` *unhealthy* otherwise; both arms are pinned
  (`src/Quartz.Tests.Unit/HealthChecks/QuartzHealthCheckTests.cs:72`, `:88`). Two pages say it correctly
  (`packages/hosted-services-integration.md:137`, `how-tos/external-leader.md:154`) and two say a `Created`
  scheduler is simply unhealthy: `docs/documentation/quartz-4.x/operations.md:671` and
  `docs/documentation/quartz-4.x/migration-guide.md:1487`. `AutoStart = false` is the shape three how-tos teach.
- **G-4, docs.** `docs/documentation/quartz-4.x/tutorial/trigger-and-job-listeners.md:14-17` still carries
  3.x's danger box — "Jobs can get stuck … when listener notification failed" — which
  `src/Quartz.Tests.Unit/Core/ThrowingJobListenerTest.cs:58`, `:147` and
  `src/Quartz.Tests.Unit/Core/JobListenerFailureTest.cs:97,173,229,277` disprove: a throwing listener costs
  the firing and frees the trigger and its blocked siblings.
- **G-5, docs.** `grep -c "PropagateTraceContext" docs/documentation/quartz-4.x/configuration/reference.md`
  → **0**. It is the one `QuartzSchedulerOptions` property missing from the reference, and the only one whose
  default (`true`, `src/Quartz/Configuration/QuartzSchedulerOptions.cs:115`) writes two entries into the data
  map of every trigger scheduled inside an `Activity`. It has no flat `quartz.*` key, so no legacy-table row.
- **G-7, pin (MED).** `tutorial/more-about-triggers.md:142` states the chaining semantic; all six shipped
  calendars honour it, `CronCalendar` and `DailyCalendar` by reaching `CalendarBase` directly rather than
  `base.IsTimeIncluded`. Nothing asserts it: `git grep -rn "IsTimeIncluded"` over the three test projects
  shows every assertion is on a calendar with no base, and every other `CalendarBase` occurrence is a fake, a
  clone comparison, a detach test asserting something else, or a serializer round trip.
- **G-8, pins (MED).** Four of the guide's seven "calls that still compile and now do something different"
  (`migration-guide.md:366-393`) have no test: `GetString` returning `null` for both a missing key and a
  non-string value (`src/Quartz/DataMapExtensions.cs:354` over `:744-756`); typed accessors throwing
  `InvalidCastException` rather than `KeyNotFoundException` for an absent key (`:304-310`); `GetBoolean`
  reading anything that is not `"true"` as `false` while `TryGetBoolean` reports success (`:703`);
  invariant-culture parsing (`:578`, `:610`). `git grep -rn "GetBoolean"` over all three test projects
  returns five non-baseline hits, every one storing `true` and asserting `true`.
- **G-9 / G-10 / G-11 / G-12, pins.** `[JobTimeout]` inheritance from a base class or an interface is
  documented (`tutorial/job-execution-middleware.md:303-305`) and covered nowhere — `JobTypeInformationTest`
  covers the two boolean attributes only. `RefireImmediately` beating the unschedule flags
  (`src/Quartz/Impl/Triggers/TriggerBase.cs:610-617`) is never asserted with both set. Per-fire-instance
  timeout isolation has no two-firing test. Listener notification order is registration order by
  construction — `src/Quartz/Core/ListenerManagerImpl.cs:17,20,23` are `OrderedDictionary`, which can only be
  for that — and is promised on no page and asserted in no test.
- **G-13 = N2b-14, coverage.** `MisfireMatrixTest` and `RetryMatrixTest` live in `Quartz.Tests.Integration`,
  which never feeds `dotnet fallout Compile UnitTest --coverage`, yet both need nothing but a SQLite file
  (`src/Quartz.Tests.Integration/Impl/MisfireThroughAStoreTestBase.cs:84-91`, `:166-186`). Product code whose
  only pin is one of them reads as uncovered on a Sonar PR gate.

### Done means
- `JsonSchedulingHelper.ReadTriggers` takes the resolved per-scheduler `TimeProvider`
  (`IServiceProvider.GetSchedulerTimeProvider`, `src/Quartz/Configuration/QuartzServiceRegistration.cs:416`)
  and uses it for the default start time, for `StartTimeSecondsInFuture`, and as
  `TriggerBuilder.Create(timeProvider)`. A test mirroring `XmlSchedulingDataProcessorTest.cs:452` — a
  `FakeTimeProvider`, a JSON trigger with `StartTimeSecondsInFuture`, asserting fake-now + n — fails without
  the change, plus a second assertion that the built trigger's own clock is the scheduler's.
- `CalendarChainingTest`: a `[TestCaseSource]` over the six shipped calendars, each with a `HolidayCalendar`
  underneath, asserting (a) an instant the calendar includes and the base excludes is excluded, (b) an
  instant both include is included, (c) `GetNextIncludedTimeUtc` walks past the base's exclusion.
- `JobDataMapCoercionContractTest`: one case per G-8 bullet, run against both `JobDataMap` and
  `SchedulerContext`; the culture case sets `CultureInfo.CurrentCulture` to `de-DE` and asserts `"3.14"`
  still reads as `3.14`.
- Cases for `[JobTimeout]` inheritance on `JobTypeInformationTest`; one `RetryEngineTest` case setting
  `RefireImmediately` **and** an unschedule flag and asserting `ReExecuteJob`; a two-concurrent-firings
  timeout test asserting only the overrunning fire instance is interrupted.
- **D12 is ratified as a promise:** one sentence in `tutorial/trigger-and-job-listeners.md` saying listeners
  are notified in registration order, and one test registering three listeners and asserting the call log.
- Docs: the `AutoStart` split added to `operations.md:671` and to the `migration-guide.md:1487` row; the
  listener danger box rewritten to what the fixtures prove (the firing is lost, the failure is reported
  through `ISchedulerListener.SchedulerError` wrapped in `JobExecutionProcessException`, the trigger is always
  released), keeping the "do not throw" advice with its real reason; one `PropagateTraceContext` row in
  `configuration/reference.md` linking the OpenTelemetry page.
- The SQLite halves of `MisfireMatrixTest` and `RetryMatrixTest` run from `Quartz.Tests.Unit`, so two more
  suites feed the Sonar gate.
- Every new test is shown failing without its change. `dotnet fallout VerifyDocsSnippets` and
  `npm run docs:check-links` clean.

### Not in this PR
- `IJobExecutionContext.JobRunTime`'s XML doc (G-2) — P7 owns the core doc-truth batch.
- The `SmartPolicy` per-family table (G-6): the plan lists it both under V2.3 and in P11's bundle; **P11 owns
  it**, because the rest of V2.3 is behaviour corrections and G-6 is a tutorial gap.
- Every store-side guarantee (N2b) — P3.
- `IScheduler`'s exception contract and its null-argument behaviour — P7 (Q2 item 1).

lahma:
N2a mapped 218 scheduler-core guarantees to the test that pins them and found 18 unpinned and six
documented statements that contradict the code. This is the core half of that list: one behavioural
defect, six pin batches, three docs corrections, and a move that lets the coverage gate see seventy
existing assertions.

## The behaviour change

**A trigger declared in `appsettings.json` was anchored to the wall clock.**
`JsonSchedulingHelper.ReadTriggers` read `DateTimeOffset.UtcNow` for the default start time and for
`StartTimeSecondsInFuture`, and built with `TriggerBuilder.Create()` and no clock — so a scheduler on a
test or replayed clock got its XML-declared triggers on that clock and its JSON-declared ones on real
time, and the JSON ones additionally carried `TimeProvider.System` for their own misfire and retry
arithmetic. The XML processor has done the opposite since #3212, with a test saying so.

The clock was already reachable: the trigger reading happens inside the `SchedulerContent` factory,
which is handed the scheduler's own view of the container, so `GetSchedulerTimeProvider` resolves the
same clock every other part of that scheduler is built with. Three tests mirror
`XmlSchedulingDataProcessorTest.StartTimeSecondsInFutureIsRelativeToTheTimeProvider`, including one
asserting the built trigger keeps the scheduler's clock rather than only being anchored to it.

This is **not** #3624, which is about recurrence support in the JSON scheduling file. The four schedule
kinds are untouched.

## The pins

Each was written against a documented promise that nothing asserted, and each was shown failing by
breaking the thing it pins.

* **Calendar chaining**, over all seven public calendars with a `HolidayCalendar` underneath: an instant
  only the base excludes is excluded, an instant neither excludes is included, and `GetNextIncludedTimeUtc`
  walks past the base's exclusion. `CronCalendar` and `DailyCalendar` reach `CalendarBase` directly
  instead of calling `base.IsTimeIncluded`, so a refactor centralising the chain would have left them
  behind silently. The case list is held to the assembly's public calendars, so a new calendar has to say
  what it does with its base.
* **The job-data coercion contract**, one case per bullet of the migration guide's "calls that still
  compile and now do something different" list, run against `JobDataMap` and `SchedulerContext` both.
* **`[JobTimeout]` inheritance** from a base class and from an interface, and a type's own budget beating
  both — including `"00:00:00"`, which is how a long-running job exempts itself.
* **`RefireImmediately` beats the unschedule flags.** The two existing pins each set one flag alone, so
  neither would have noticed the branch order changing.
* **Per-fire-instance timeout isolation.** Two firings of one job, staggered so their budgets differ, with
  the companion still in flight when the other is interrupted. Interrupting by job key instead of fire
  instance id fails it.
* **Listener notification order**, now a promise (plan decision D12): one sentence in the tutorial and
  three tests. A plain `Dictionary` enumerates in insertion order too, right up until an entry is removed
  and the next addition fills the hole — that case is what tells the promise apart from an accident, and
  swapping `OrderedDictionary` for `Dictionary` fails it.

## The docs

* `operations.md` and the `migration-guide.md` table said a `Created` scheduler is simply unhealthy. It is
  *degraded* when `AutoStart` is `false` — the shape three how-tos teach — and both arms have been tested
  since the check was written.
* The listener danger box was inherited verbatim from 3.x and said jobs can get stuck when a listener
  notification fails. Six fixtures say otherwise; it now says what they prove, keeping the "do not throw"
  advice with its real reason.
* `PropagateTraceContext` was the one `QuartzSchedulerOptions` property missing from the reference table,
  and the worst to omit: on by default, and it writes two entries into the data map of every trigger
  scheduled inside an `Activity`.

## Extra, and worth a reviewer's eye

Writing the culture case for the coercion contract disproved the migration guide's own sentence. **`"3,14"`
does not throw.** The floating-point styles allow a group separator, so `GetDouble` and `GetFloat` read the
comma as one and answer **314**; only `GetInt` and `GetLong` throw. The guide documented the loud half and
missed the silent one, which is the wrong way round for an upgrader auditing job data written under a
comma-decimal culture. The sentence now says both, and a test pins it. This is outside the issue's docs list,
so it is called out here rather than folded in quietly.

## Coverage

`MisfireMatrixTest` and `RetryMatrixTest` need nothing but a SQLite file, but lived in
`Quartz.Tests.Integration`, which the coverage run does not execute — so product code whose only pin is one
of them reads as uncovered on a pull request that touches it. **Moved outright rather than mirrored**:
running seventy assertions twice to protect a two-second suite buys nothing. The scaffolding they share with
the two misfire fixtures that stayed is compiled into both projects the way `TestJobStores` and
`TestTimeZones` already are, under the neutral `Quartz.Tests.Impl` namespace the shared files use.

## Verification

* `dotnet test src/Quartz.Tests.Unit` — `Failed: 0, Passed: 4983, Skipped: 36, Total: 5019`, 53 s.
* `dotnet test src/Quartz.Tests.Integration --filter MisfireCalendarExclusionTest|MisfireWhilePausedOrBlockedTest`
  — `Failed: 0, Passed: 10`, the two fixtures that still live there. No Docker was needed and none of the
  containerised legs were run.
* `dotnet fallout VerifyDocsSnippets` — up to date, 102 markdown files checked.
* `npm run docs:check-links` — 220 pages, 1135 internal links, 704 fragments, no dead links or anchors.
* No public API baseline moves; nothing in `src/**/Verify/` is touched.

Closes #3643


lahma:
Coordinator commit on top: the "3,14" finding is fixed in code rather than documented — float/double/decimal now parse with `NumberStyles.Float` (no group separator), so a comma-decimal string throws `InvalidCastException` from every numeric accessor instead of reading as 314 from the floating-point ones; the coercion contract test asserts the rejection and the guide sentence says so. Full unit suite green locally.

Issue #3646, opened 2026-09-01
The side packages keep their promises


lahma:
**Milestone:** 4.0.0-beta.1 · **Labels:** docs, tests, 4.x · **Baseline:** none
**Closes:** —

### Why
N1b found the nine side packages sound in shape — all 62 `IScheduler` members reach an endpoint with a
round-trip test, and both JSON serializers are verified against one shared snapshot so they must emit
byte-identical blobs — and found ten things that are promises the packages do not yet keep: one member that
escapes the client's error mapping, six of eight typed-exception mappings untested, a flag lost in both
directions, four trigger fields that only ever appear at their defaults in every stored and wire snapshot,
two pages that contradict each other about errors, and a public surface (`HttpScheduler`) that documents
seven of about sixty-one members. CS1591 is off repo-wide (`Directory.Build.props:34` puts `1591` in
`NoWarn`), so nothing objects.

### Evidence
- **N1b-1.** `src/Quartz.HttpClient/HttpScheduler.cs:360-366`: `SetExecutionLimits(null)` calls
  `httpClient.DeleteAsync(...)` + `EnsureSuccessStatusCode()` directly, bypassing `CheckResponseStatusCode`.
  The checked equivalent `HttpClientExtensions.Delete` exists at
  `src/Quartz.HttpClient/HttpClientExtensions.cs:133-141` and has **no callers**. Against an unbound
  scheduler name the caller gets `HttpRequestException` where every other member gives
  `HttpClientException("Scheduler not found…")`.
- **N1b-2.** `HttpClientExtensions.cs:214-225` maps eight exception names back to typed exceptions —
  `ObjectAlreadyExistsException` is at `:223` — and only `SchedulerException` and `JobExecutionException` are
  covered end to end (`src/Quartz.Tests.AspNetCore/HttpApi/CommonEndpointTest.cs:23-38`). The mapping is by
  *type-name string*, so a rename silently downgrades a caller's `catch`.
- **N1b-3.** `ExecutionLimits.UsesTriggerGroupWhenUnset` is dropped in both directions when no group limit is
  set: the server keeps `limits` null unless `request.Limits is { Count: > 0 }`
  (`src/Quartz.AspNetCore/AspNetCore/HttpApi/Endpoints/SchedulerEndpoints.cs:272-303`) and the client returns
  `null` when `response.Limits` is empty (`HttpScheduler.cs:385-388`), discarding a flag the server did send.
  No test in either project sets it.
[...]

PR #3660, opened 2026-09-02
The core surface says what it does

### LINKED ISSUE #3648: The core surface says what it does
**Milestone:** 4.0.0-beta.1 · **Labels:** docs, bug, tests, 4.x · **Baseline:** moves `PublicApiTest_Quartz` (the two nullability corrections); **third in the chain P1 → P6 → P7 → P8 → P9**
**Closes:** —

### Why
N1a audited the frozen `Quartz` surface and found **nothing that needs a breaking change**: the shapes are
sound. What is not yet a promise is what the shapes *say*. Twenty-three public types and roughly 190
public and protected members carry no documentation at all — including the whole trigger- and
calendar-serializer seam and all four `QuartzHostedService` hooks the class remarks call "a supported
extension point"; `IScheduler` has two `<exception>` tags over 65 members and raises `SchedulerException`
where a caller would write `catch (ArgumentNullException)`; about twenty public `cref`s point at types a
consumer cannot see; and eight `<throws>` tags carry promises that reach nobody because `<throws>` is not
an XML documentation element. Three of the five ratified freeze exceptions are here.

### Evidence
- **A5.** `grep -rn "<throws>" --include=*.cs src/` → **8**: `Extensibility/IJobFactory.cs:67`,
  `Extensibility/IJobStore.cs:112,144,222,376`, `Impl/AdoJobStore/AdoJobStoreBase.Execution.cs:421`,
  `Impl/PropertySettingJobFactory.cs:92`, `Impl/SimpleJobFactory.cs:73`. The compiler emits them into
  `Quartz.xml` as an unknown element, so no renderer and no IDE shows them — and `IJobStore.ScheduleJob` and
  `AddCalendar` have no other statement of what a duplicate key does.
- **A1 / A2 / A3.** `grep -rn "///.*%[A-Za-z][A-Za-z.]*(" --include=*.cs src/` → three NDoc `%…%` markers,
  all on `src/Quartz/IScheduler.cs:51,52,74`, two of which name signatures that no longer exist.
  `src/Quartz/IScheduler.cs:1138` and `:1154` name `InterruptableJob`, a 3.x type; the sentence is copied to
  `Impl/StdScheduler.cs:648` and `Core/QuartzScheduler.cs:2739,2770,2774`. `IScheduler.cs:69-70` says
  `Key{T}.DefaultGroup` is "defined on this interface" while the `cref` in the same sentence points at
  `Key<T>`.
- **A4 / A6 / A7 / A9 / A11.** `IJobStore.GetCalendar`'s summary reads "Retrieve the given `ITrigger`"
  (`src/Quartz/Extensibility/IJobStore.cs:396-405`); `StringOperator.Equals` ends both `<returns>` blocks in
  `true` where it means `false` (`src/Quartz/Matchers/StringOperator.cs:123`, `:138`);
  `CronScheduleBuilder.Create(string)` promises "only a **RuntimeException**"
  (`src/Quartz/CronScheduleBuilder.cs:149-153`) where a `FormatException` or an `ArgumentException` is what
  escapes; `SchedulingOptions` has no type doc and describes "the file" although it governs `AddQuartz(…)` and
  the `Quartz:Scheduling` section; and `Clone()` on `ITrigger`, `IJobDetail` and `ICalendar` has **no doc
  comment at all**, although `TriggerBase.Clone()` (`src/Quartz/Impl/Triggers/TriggerBase.cs:936-949`) is a
  `MemberwiseClone` plus a shallow-cloned data map — its own internal comment says so and the public doc does
  not.
- **A8.** `src/Quartz/ICalendar.cs:38-40` tells implementors to be "Serializable", the 3.x requirement. The
  4.x one is a `CalendarSerializer<TCalendar>` registered through `AddCalendarSerializer<TCalendar>`; without
  it the first store write fails in `CalendarConverter`. `ICalendar.cs` (73 lines, read in full) names
  neither, and `grep -rn "AddCalendarSerializer" --include=*.cs src/Quartz/` reaches only the registry.
- **A10.** About twenty public `cref`s resolve to twelve non-public types — all four public misfire enums
  point at the internal `MisfireInstruction` (`src/Quartz/CronTriggerMisfireInstruction.cs:26` and its three
  siblings), `AdoJobStoreOptions.cs:185,187,243` at the two internal stores on the settings whose whole
  meaning is that distinction, and the extensibility interfaces at `QuartzScheduler` /
  `QuartzSchedulerThread`. CS1574 is satisfied because the type resolves inside the assembly.
- **A12.** `Directory.Build.props:34` puts `1591` in `NoWarn`. Undocumented: 23 public types (including
  `ITriggerSerializer`, `ICalendarSerializer`, `CalendarSerializer<TCalendar>` and all five built-in trigger
  serializers), ~166 public/protected members and 25 interface and enum members —
  among them `ITrigger.Key`/`JobKey`, `IMatcher.IsMatch`, all eight `IntervalUnit` values, and
  `QuartzHostedService`'s constructor and four lifecycle hooks.
- **A14, Q2 item 1.** `grep -c "/// <exception" src/Quartz/IScheduler.cs` → **2**, over 65 members. Null
  arguments raise `SchedulerException` (`src/Quartz/Core/QuartzScheduler.cs:651-669`) while the extension
  methods next door use `ArgumentNullException.ThrowIfNull` (`src/Quartz/SchedulerQueryExtensions.cs:67,87-88,…`)
  — two families on one type with opposite contracts, neither written down. No test pins any of it:
  `grep -rn "ScheduleJob(null" src/Quartz.Tests.Unit src/Quartz.Tests.Integration` → no hits.
- **A13, Q2 item 3.** `CronExpression.TryParse(s, CronFormat, out)` (`src/Quartz/CronExpression.cs:533-551`)
  catches only `FormatException`, so `ToQuartzForm`'s `ArgumentOutOfRangeException` for an undefined
  `CronFormat` escapes a `Try*`. `new CronExpression(null!)` throws `ArgumentException` (`:288`) where
  `Parse(null!)` throws `ArgumentNullException` (`:468`).
- **A15 / A16, Q2 item 2.** `src/Quartz/Impl/Triggers/TriggerBase.cs:438` is
  `public virtual string FireInstanceId { get; set; } = null!;` — assigned only when a store fires the
  trigger, so a trigger from `TriggerBuilder` answers `null` from a non-nullable `string`.
  `src/Quartz/Impl/AdoJobStore/Common/DbMetadata.cs:159` is the same shape, and forgetting it fails as
  `Type.GetProperty(null)` two lines from a deliberate, useful message.
- **A17 / A18 / A19.** A custom `StringOperator` works on `RAMJobStore` and throws on every ADO store
  (`StdAdoDelegate.cs:447-478` tests the five built-ins by equality); `grep -rn "SQL\|LIKE" src/Quartz/Matchers/`
  → no hits, so nothing on the type says so. `SchedulingOptions`' documented booleans interact —
  `OverwriteExistingData = true` makes `IgnoreDuplicates` inert by the doc's own account — and no validator
  reports the combination. `SchedulerException`'s `ToString()` doc carries a .NET Framework 2.0
  `<PermissionSet>` block and "The MSG.".

### Done means
- All eight `<throws>` become `<exception cref="…">`, and a one-line assertion in the unit project that no
  `.cs` under `src/` contains `<throws>` keeps the shape from coming back.
- A1–A9 and A11 fixed as text, with `Clone()`'s shallow semantics — including the shared data-map values —
  stated on all three interfaces, and `ICalendar` naming `AddCalendarSerializer<TCalendar>` as the 4.x
  requirement. A unit test stores a custom `ICalendar` on a **file SQLite** store (`UseSQLite(...)` +
  `ProvisionSchema()`) with and without the serializer registered and asserts the message names the calendar
  type — which also covers that ADO path from the unit project.
- A `cref`-to-internal guard: read the generated `Quartz.xml` (or re-derive from source) and assert no `cref`
  on a public member resolves to a non-public type; the ~20 sites fixed to say what a consumer can reach.
- **CS1591 for `Quartz`:** every public and protected member documented, `/// <inheritdoc />` on interface
  implementations and overrides, one-sentence summaries in the repository's voice, `<exception cref>` where a
  documented exception is thrown; then `1591` drops out of `NoWarn` for the package and is re-added only in
  the non-packable projects. Only once the package is clean.
- **Freeze exception 1 (Q2.1):** `IScheduler`'s mutation members raise `ArgumentNullException` for a null
  argument, named by parameter. One theory over the mutation members pins it, one test pins the
  non-`IOperableTrigger` refusal, and the `<exception>` tags the tests prove land on all 65 members.
- **Freeze exception 2 (Q2.2):** `FireInstanceId` and `ParameterDbTypePropertyName` say what they are — the
  annotation corrected, and the `DbMetadata` failure becomes the deliberate message its neighbours use.
- **Freeze exception 3 (Q2.3):** `TryParse(s, format, out)` validates `format` before the `try` and answers
  `false`; the constructor and `Parse` agree on the null exception; both document it.
- A17's ADO caveat stated on `StringOperator`; a `SchedulingOptions` validator that reports the inert
  combination; `SchedulerException`'s doc debris removed.
- All three freeze exceptions are called out as **"changed since alpha.5"** and each gets its own
  migration-guide row. The baseline moves for Q2.2; accept through Verify. Opens once P6 is open; rebase and
  re-accept on its merge notice.

### Not in this PR
- `HttpScheduler`'s `<inheritdoc />` pass: the plan lists it under V1.1, but **P5 owns it** by file locality,
  since P5 is already doing that package's CS1591 sweep.
- CS1591 for the side packages — P5.
- `IJobExecutionContext.JobRunTime`'s doc is here (G-2), but the JSON-clock behaviour it sits beside is P2.
- The two remaining freeze exceptions (the 500 detail, `MaxPageSize`) — P9.

lahma:
N1a audited the frozen `Quartz` surface and found nothing that needs a breaking change — the shapes
are sound. What was not yet a promise is what the shapes *say*. This makes them say it, and puts
three compiler-enforced guards where prose alone would drift again.

Eight commits, in the order they are easiest to read.

### 1. `<throws>` → `<exception cref>`, and a guard

`<throws>` is Javadoc's tag. The C# compiler copies it into `Quartz.xml` as an unknown element and
every renderer and every IDE drops it, so the promise looked written down and reached nobody. Two of
the eight were the only statement `IJobStore.ScheduleJob` and `IJobStore.AddCalendar` made about a
duplicate key. A walk over `src/` fails the build if the shape comes back.

### 2. The doc-truth batch (A1–A9, A11, G-2)

`IScheduler`'s type documentation carried three NDoc `%…%` cross-references, two of them naming
signatures that no longer exist, and said `Key<T>.DefaultGroup` is "defined on this interface" while
the `cref` beside it pointed at `Key<T>`. `Interrupt` and `InterruptFireInstance` described
interrupting an `InterruptableJob`, a 3.x type. `IJobStore.GetCalendar`'s summary was `GetTrigger`'s.
`StringOperator.Equals` promised `true` where it means `false`, twice.
`CronScheduleBuilder.Create(string)` promised "only a **RuntimeException**", which is `java.lang`.
`SchedulingOptions` had no type documentation and described "the file".

`Clone()` on `ITrigger`, `IJobDetail` and `ICalendar` had no documentation of any kind. The first two
are shallow *and* shallow-copy the data map, so a caller who clones a trigger to edit it and then
mutates a value in the map has edited both — which `TriggerBase`'s own internal comment has always
said and the public one never did. `ICalendar`'s is not shallow, so it states the obligation instead.

`ICalendar` told implementors to be "Serializable", which is 3.x's requirement and stores nothing
here. A **file-SQLite** test writes a custom calendar with a `CalendarSerializer<TCalendar>`
registered and without one — the round trip, and the failure that names the type — which also covers
that ADO write path from the unit project.

`IJobExecutionContext.JobRunTime` promised `TimeSpan.MinValue` until the job finished, and has never
returned it. A listener written against that sentence never took its branch.

### 3. A `cref`-to-internal guard, and thirty-three fixes

CS1574 is satisfied the moment a target resolves *inside* the assembly, so all four public misfire
enums pointed a reader at the internal `MisfireInstruction` they exist to replace, and
`AcceptEnlistedTransactions` / `OpenConnection` — whose meaning is *which of the two internal ADO
stores is built* — named both by type. The guard reads the generated `Quartz.xml` and fails a `cref`
on a member a consumer can see whose target a consumer cannot. `<inheritdoc cref>` is exempt: it
copies the text in rather than linking, so the reader never learns the name.

### 4. Freeze exception 1 — `IScheduler` nulls raise `ArgumentNullException` (Q2.1)

`IScheduler` had **two** `<exception>` tags over sixty-five members, and answered a null argument
with `SchedulerException("JobDetail cannot be null")` — Java parity, undocumented, and the opposite
of what the `SchedulerQueryExtensions` methods on the same type already did. Thirty-seven refusals
across the mutation members now name the parameter, pinned by one `[TestCaseSource]` theory on
`RAMJobStore`; beside it, the non-`IOperableTrigger` refusal, which stays a `SchedulerException`
because it is a scheduling refusal rather than a missing argument. Then the tags the tests prove:
**ninety-seven** of them.

### 5. Freeze exception 3 — `TryParse(s, format, out)` (Q2.3)

`ToQuartzForm` ends in `ArgumentOutOfRangeException` for a `CronFormat` it has no arm for, and the
`try` caught only `FormatException` — so the one method whose whole promise is that it will not
throw, threw. Beside it, `new CronExpression(null)` raised `ArgumentException` where `Parse(null)` —
a null check followed by that very constructor — raised `ArgumentNullException`.

### 6. Freeze exception 2 — two annotations that were not true (Q2.2) — **moves the baseline**

`IOperableTrigger.FireInstanceId` and `TriggerBase.FireInstanceId` are `string?`;
`DbMetadata.ParameterDbTypePropertyName` is `string?` and raises the deliberate message its
neighbours do. Three lines in `PublicApiTest_Quartz.verified.txt`, accepted through Verify.

### 7. Three limits the code has and the surface never mentioned (A17, A18, A19)

A custom `StringOperator` works on `RAMJobStore` and is refused by every persistent store at query
time; the type says so now. `SchedulingOptions`' `IgnoreDuplicates` was inert whenever
`OverwriteExistingData` was left at its default — an application asking to *pass over* duplicates got
*replace* — and that pair is now a startup failure. `SchedulerException` carried a .NET Framework 2.0
`<PermissionSet>` block and "The MSG.".

### 8. CS1591 for `Quartz`

`Directory.Build.props` had `1591` in `NoWarn`, so **521** public and protected members of the
package a consumer reads first were undocumented and nothing objected — including the whole
trigger- and calendar-serializer seam, all four `QuartzHostedService` lifecycle hooks,
`ITrigger.Key`, `IMatcher.IsMatch` and every `IntervalUnit` value. The suppression is gone from
`Directory.Build.props` and re-added, with a note naming #3646, to the nine side packages that are
not clean yet, so `Quartz` alone is enforced and each of the others has one line to delete.

## What a reviewer has to decide

- **The `DbMetadata` shape.** I chose `string?` over `required`, because the property is required
  *exactly when* `DbBinaryTypeName` is set: a provider whose parameters need no explicit database
  type sets neither, and `required` would demand it of them. Say so if you would rather it were
  `required` and the two settings moved together.
- **`SchedulingOptions` validation is a startup refusal**, not a warning. A deployment that set
  `IgnoreDuplicates = true` and left `OverwriteExistingData` alone starts today and silently does the
  opposite of what it asked; after this it does not start. That is the point, and it is also the one
  change here that can stop an existing application. It is not a freeze exception — nothing about the
  shape changed — but it gets a migration-guide row of its own beside the three that are, so the guide
  carries four "changed since alpha.5" entries rather than three.
- **The scheduler's `ArgumentNullException` may change an HTTP status.** `ExceptionHandler` maps
  `SchedulerException` to 400 and everything else to 500. No test in `Quartz.Tests.AspNetCore`
  moved (561 pass), because the endpoints validate their own input before the scheduler sees it —
  but P9 owns the 500 body and is the place to confirm nothing reaches the scheduler with a null.
  `HttpScheduler` and the HTTP API are untouched here, as briefed.
- **Read members were left alone.** Q2.1 is scoped to the mutation members, which are the family
  that raised `SchedulerException`. `GetJobDetail(null)` and friends still fail however they fail.
  Widening it is a second behaviour change and wants its own decision.

## Deviations from the brief

- The brief's A18 says "the inert combination". I first also refused
  `ScheduleTriggerRelativeToReplacedTrigger` with `OverwriteExistingData = false`, which is inert by
  the same reading — and it broke `AddQuartz_SchedulingSectionInJson_PopulatesSchedulingOptions`,
  which sets all three deliberately. I dropped that second rule rather than invent a refusal nobody
  ratified.
- `CronScheduleBuilder.Create(string)` and `Create(string, CronFormat)` were folded into Q2.3 along
  with `CronExpression`'s constructor, since they had the same split against their own private
  constructor. `ArgumentNullException` is an `ArgumentException`, so this only widens what a `catch`
  sees.
- A2 sketched a test for `Interrupt` with a token cancelled between two matching firings. It is
  inherently racy and the issue lists A1–A9 as text fixes; the sentence is corrected, the test is not
  written.

## Verification

```
dotnet build src/Quartz/Quartz.csproj -t:Rebuild      Build succeeded. 0 Warning(s) 0 Error(s)   ← CS1591 enforced
dotnet test src/Quartz.Tests.Unit                     Passed! 5126 passed, 36 skipped, 0 failed
dotnet test src/Quartz.Tests.AspNetCore               Passed!  561 passed, 0 failed
dotnet fallout VerifyDocsSnippets                     Succeeded (104 markdown files checked)
npm run docs:check-links                              221 pages, 1154 links, 708 fragments; no dead links
dotnet fallout BenchmarkSmoke                         Succeeded (282 benchmarks)
```

Integration tests were not run: they need a Docker daemon, and the ADO path this touches is covered
from the unit project on a file SQLite database instead.

Closes #3648


lahma:
Coordinator commit: Sonar's reliability gate flagged three `if (matcher is null)` fallbacks left under the new `ThrowIfNull` guards in `QuartzScheduler` (always false) — removed; the guide already states a null matcher throws rather than meaning the default group. Also the redundant pattern null-check in the scheduling validator and `SchedulerException` overload adjacency. CI re-running.

Issue #3066, opened 2026-04-30
Adding dashboard policy/role auth applies the policy to the entire site, not just /quartz endpoint


daver77:
Hi, I have applied a policy as shown in the [documentation here](https://www.quartz-scheduler.net/documentation/quartz-3.x/packages/dashboard.html#policy-and-role-based-authorization). It is setup on an existing Blazor Server application which already has its own policies. When I apply the new policy for Quartz it applies it to the entire website, not just /quartz, so I get a 404 when I access my site when I am not in the quartz specific role.

v3.18.1 on .NET 10.

```
public static class Policies
{
    ...
    public const string ScheduledJobAdmin = "ScheduledJobAdmin";

   ...

    public static AuthorizationPolicy IsScheduledJobAdminPolicy()
    {
        return new AuthorizationPolicyBuilder().RequireAuthenticatedUser()
                                               .RequireRole(ScheduledJobAdmin)
                                               .Build();
    }
}
```

```
services.AddAuthorizationCore(options =>
{
    ...
    options.AddPolicy(Policies.ScheduledJobAdmin, Policies.IsScheduledJobAdminPolicy());
});
```

```
services.AddQuartzDashboard(options =>
{
    options.AuthorizationPolicy = Policies.ScheduledJobAdmin;
});
```

I would expect the new ScheduledJobAdmin policy to only apply to the /quartz endpoint.

PR #3664, opened 2026-09-02
Neither HTTP surface starts unless something says who may reach it

### LINKED ISSUE #3650: An unauthenticated Quartz endpoint fails at startup instead of at the incident
**Milestone:** 4.0.0-beta.1 · **Labels:** bug, docs, tests, 4.x · **Baseline:** moves `PublicApiTest_Quartz.AspNetCore` (`MaxPageSize`); **fifth in the chain P1 → P6 → P7 → P8 → P9**
**Closes:** —

### Why
Neither HTTP surface is safe by default. `app.MapQuartzHttpApi()` with no options applies authorization only
when `SchedulerAuthorizationPolicy` is set, so all sixty routes — including `shutdown` and `clear` — answer
anonymously; the dashboard is the same and its `ReadOnly` defaults to `false`. Both surfaces schedule a job
whose type is a **string from the request**, stored unresolved and resolved later with `Type.GetType`, with
no allow-list. With `Quartz.Jobs` on the probing path that reaches `NativeJob`, which starts a process named
in job data: an unauthenticated endpoint here is remote code execution, and no page says so. Q1 ratified
failing closed, and Q2 ratified the two remaining freeze exceptions, both of which live on this surface.

### Evidence
- **OPS-F8 / N1b-10.** `src/Quartz.AspNetCore/QuartzAspNetCoreConfigurationExtensions.cs:151-161` applies
  `RequireSchedulerAuthorization` **only** when `options.SchedulerAuthorizationPolicy` is non-blank, and
  nothing else adds authorization; it returns a `QuartzApiConventionBuilder`
  (`src/Quartz.AspNetCore/AspNetCore/HttpApi/QuartzApiConventionBuilder.cs:5`) holding every mapped endpoint.
  `src/Quartz.Dashboard/QuartzDashboardEndpointRouteBuilderExtensions.cs:245-312` is the mirror: with a
  policy it authorizes the hub, the assets and the pages; without one the asset endpoints are explicitly
  `AllowAnonymous()` and **the pages and the SignalR hub are left with no authorization metadata at all**.
  `grep -n "anonymous\|no policy\|nothing configured\|by default\|unauthenticated" docs/documentation/quartz-4.x/packages/http-api.md`
  returns one line, about the per-scheduler policy, so the API page never states its own default;
  `packages/dashboard.md:435-438` does state the dashboard's, accurately.
- **The consequence, undocumented.** `JobDetailDto.cs:80-89` builds the detail with the *string* cast, so the
  type name is stored unresolved; `src/Quartz/JobType.cs:30` resolves it later with `Type.GetType`, and
  `JobDetailDto.cs:109-132` validates shape only (length ≤ 1024, no leading comma, no control characters).
  `src/Quartz.Jobs/NativeJob.cs:188-200` builds a `Process` whose `FileName` and `Arguments` come from job
  data. `grep -rn -i "arbitrary\|remote code\|untrusted\|any type\|execute any" docs/documentation/quartz-4.x/`
  → 10 hits, none about this; `packages/http-api.md`'s Production hardening list says "require
  authentication/authorization" without naming what it prevents.
- **OPS-F1, Q2 item 5.** `src/Quartz.AspNetCore/AspNetCore/HttpApi/Util/EndpointHelper.cs:159-170`
  (`AssertPaging`) rejects only negatives, and the six call sites write `take` straight onto the query record.
  The bulk key fetch **is** capped, at `MaxKeysToFetch = 1000` (`:157`), so one paged request can materialize
  every trigger in the store while the neighbouring endpoint refuses 1001 keys.
- **OPS-F2, Q2 item 4.** `src/Quartz.AspNetCore/AspNetCore/HttpApi/Util/ExceptionHandler.cs:66` returns
  `Problem(exception, exception.Message, 500, nameTheExceptionType: false)`, and the XML doc four lines above
  states the rule it breaks — a 500 is a fault the caller cannot act on, and naming what produced it buys
  nothing it does not also leak. The type is withheld; the message is not, and a driver message routinely
  names the server, database, login or constraint.
  `src/Quartz.Tests.AspNetCore/Verify/WireFormatSnapshotTest_ServerFaultProblemDetailsBody.verified.json`
  carries it today.
- **N6 D19.** `packages/dashboard.md:67-78`'s "Basic setup" maps
  `app.MapQuartzHttpApi().RequireAuthorization()` and then a bare `app.MapQuartzDashboard()`; the
  `src/Quartz.Dashboard/README.md` snippet has no `RequireAuthorization()` at all, where the
  `Quartz.AspNetCore` readme does. Two nuget front pages, opposite defaults.

### Done means
- **`AddQuartzHttpApi` and `AddQuartzDashboard` each register a startup guard** — an `IHostedLifecycleService` whose `StartingAsync` runs before
  any hosted service's `StartAsync` (so before Kestrel listens), enumerating `EndpointDataSource.Endpoints` — over the endpoints Quartz
  mapped, identified by a marker metadata applied through the existing `QuartzApiConventionBuilder` and the
  dashboard's mapping. It throws `InvalidOperationException` unless **every** Quartz endpoint carries
  `IAuthorizeData` or `IAllowAnonymous` metadata, or the host's `AuthorizationOptions.FallbackPolicy` is
  non-null. Metadata applied on a parent `MapGroup` flows into endpoint metadata, so a grouped map passes.
- The message names the three fixes: `.RequireAuthorization()` on the returned builder,
  `QuartzHttpApiOptions.SchedulerAuthorizationPolicy` / `QuartzDashboardOptions.AuthorizationPolicy`, or an
  explicit `.AllowAnonymous()` opt-out.
- The dashboard's deliberately anonymous asset endpoints already carry `AllowAnonymous` and pass unchanged;
  its **page and hub** endpoints are the ones the guard covers.
- Tests in `Quartz.Tests.AspNetCore` over `WebApplicationFactory`, for both surfaces: a bare map throws at
  start with that message; `RequireAuthorization` starts; an explicit `AllowAnonymous` starts; a non-null
  `FallbackPolicy` starts; `SchedulerAuthorizationPolicy` starts.
- **Freeze exception 4:** `QuartzHttpApiOptions.MaxPageSize` (default 1000, `0` = unbounded), clamped in
  `EndpointHelper.AssertPaging` → `400` above it, with a test that `take=2000` is 400 at the default and 200
  once the option is raised.
- **Freeze exception 5:** a 500's problem-details `detail` becomes a fixed sentence unless
  `IncludeStackTraceInProblemDetails` is on; the real message keeps being logged.
  `WireFormatSnapshotTest_ServerFaultProblemDetailsBody` is updated and a negative case asserts the exception
  message does **not** appear with the option off.
- Docs: the paragraph drafted in N5 §2 ("The one docs sentence a deployment needs") on
  `packages/http-api.md`, `packages/dashboard.md` and `packages/quartz-jobs.md` — the surfaces add no
  authentication of their own, both can schedule a job named by a request string, and with `Quartz.Jobs` on
  the path that is remote code execution rather than an information leak. `dashboard.md`'s "Basic setup" and
  `src/Quartz.Dashboard/README.md` authorize the dashboard as well as the API. `http-api.md` states the old
  default plainly and the new startup failure.
- Migration-guide rows for all three changes — the startup throw, `MaxPageSize` and the 500 detail — each
  labelled **"changed since alpha.5"**, since the throw is a break for an alpha user who mapped a bare API
  and the opt-out is one explicit call.
- The baseline moves for `MaxPageSize`; accept through Verify. Opens once P8 is open; rebase and re-accept on
  its merge notice.

### Not in this PR
- Rate limiting on either surface — none exists anywhere and none is added; caveat.
- An allow-list for job type names — a design of its own, not a beta.1 change.
- `SendMailJob` reading SMTP credentials out of persisted job data — documented on the type; caveat.
- The dashboard's `ReadOnly` default — deliberate, and the guard makes the surface reachable only with
  authorization anyway.

lahma:
Neither HTTP surface was safe by default. `app.MapQuartzHttpApi()` applied authorization only when
`SchedulerAuthorizationPolicy` was set, so all sixty routes — `shutdown` and `clear` among them —
answered anonymously; `MapQuartzDashboard()` left its pages and its live-events hub with no
authorization metadata at all, with `ReadOnly` defaulting to `false`. Both schedule a job whose type is
a **string the request supplies**, stored unresolved and resolved later with `Type.GetType` against
whatever is on the host's probing path, with no allow-list. With `Quartz.Jobs` on that path the string
reaches `NativeJob`, which starts a process named in job data — so an unauthenticated endpoint here is
remote code execution rather than an information leak, and no page said so.

Q1 ratified failing closed. Q2's two remaining freeze exceptions both live on this surface and ship
here too.

## What changed

**1. A mapping that says nothing about authorization refuses to start.** `AddQuartzHttpApi` and
`AddQuartzDashboard` register one hosted lifecycle service whose `StartingAsync` — which every hosted
service completes before any of them is started, so before the server binds its listener — enumerates
the endpoints Quartz mapped, identified by a marker applied through the existing
`QuartzApiConventionBuilder` and the dashboard's mapping, and throws `InvalidOperationException` unless
every one carries `IAuthorizeData` or `IAllowAnonymous`, or the host has a non-null
`AuthorizationOptions.FallbackPolicy`. The message names the three fixes. A per-scheduler policy
counts (it is a filter over the route rather than metadata, so the metadata check cannot see it);
`AddQuartz*Api` without a map is left alone; authorization on a `MapGroup` above the mapping counts.

**2. The dashboard's hub is covered by the same statement as its pages.** `MapQuartzDashboard` now
returns an `IEndpointConventionBuilder` over the pages **and** the hub instead of the
`RazorComponentsEndpointConventionBuilder`. This is the reviewer decision below.

**3. Freeze exception 4 — `QuartzHttpApiOptions.MaxPageSize`** (default `1000`, `0` unbounded).

**4. Freeze exception 5 — a `500`'s `detail`** is one fixed sentence unless
`IncludeStackTraceInProblemDetails` is on; the real message keeps being logged.

**5. Docs.** The security paragraph on `packages/http-api.md`, `packages/dashboard.md` and
`packages/quartz-jobs.md`; `dashboard.md`'s Basic setup and `src/Quartz.Dashboard/README.md` authorize
the dashboard as well as the API; `http-api.md` states the old default and the new failure, the
`MaxPageSize` row and paging rule, the fixed 500 detail, and a production-hardening list that admits
there is no rate limiting; `packages/http-client.md` gains the 500 detail and the paging interaction;
`src/Quartz.AspNetCore/README.md` says what authorizing prevents. Migration-guide rows for the startup
throw, `MaxPageSize` and the 500 detail, each labelled **changed since alpha.5**.

### `MapQuartzDashboard`'s return type

All three overloads now return `IEndpointConventionBuilder` rather than
`RazorComponentsEndpointConventionBuilder`. This is beyond the issue's stated baseline delta and is
deliberate:

- the hub is mapped separately from the pages because SignalR needs its own route, so
  `MapQuartzDashboard().RequireAuthorization()` used to reach the pages and **not** the half that
  streams the same scheduler data over `/quartz/hub`. Without this, "the dashboard, authorized" could
  not be said in one call and the spec's "`RequireAuthorization` starts" test could only have passed by
  the guard ignoring a hole it exists to close;
- in the overload that takes an existing components builder, the old return value was the *host's*
  builder, so a convention on it reached the host application's own pages — the thing #3066 had to work
  around. What comes back now is the dashboard's pages and its hub, and nothing else.

Cost: `.AddAdditionalAssemblies(...)` and the other Razor-specific members are no longer callable on
the result. Nothing in this repository called them, and in standalone hosting the dashboard has already
made those calls itself.

### `?take=all` is bounded by `MaxPageSize` rather than refused by it

The brief said `all` should be a `400` when a cap is set. Implemented literally, that breaks every
`SchedulerQueryExtensions` listing through `HttpScheduler`: `GetJobKeys`, `GetTriggerKeys`,
`GetCalendarNames`, `GetJobGroupNames`, `GetTriggerGroupNames` and `GetPausedTriggerGroups` ask for
`PagedQuery.All` **regardless of how many items exist**, so a 1000-item cap would `400` a three-job
listing and the two shipped packages would not work together out of the box.

So the two spellings are answered differently:

- a **number** above the cap is a `400` naming the cap and the setting that would raise it —
  `?take=2147483647` included, which is a behaviour change from alpha.5's "the number behind the
  sentinel is still accepted";
- **`all`** names no number — it says "as many as you will give me" — so it is bounded by the cap and
  `hasMore` reports the truth. A listing that fits under the cap answers exactly as it would with no cap
  at all.

To keep that from becoming a silent truncation, `Quartz.HttpClient` now sends `take=all` for
`PagedQuery.All` (one line in `QueryStringBuilder`) and raises `HttpClientException` naming
`MaxPageSize` when a request for everything comes back bounded, rather than handing a short list to the
3.x-compatible members that return a bare `List<T>`. Pinned by
`MaxPageSizeTest.TheClientRefusesATruncatedAnswerToARequestForEverything`.

The alternative designs and why they were rejected are in the report accompanying this PR; if you would
rather have `all` refused and the client left to fail, say so and I will flip it.

### `InternalsVisibleTo("Quartz.Dashboard")` on `Quartz.AspNetCore`

Added so both surfaces share one guard, one marker and one message shape instead of carrying a copy
each. `Quartz.Dashboard` already has a project reference to `Quartz.AspNetCore`; this is a new grant
between two shipped packages.

### `ArgumentException` → `400` in `ExceptionHandler`: not done

P7 noted that a null reaching the scheduler through the API is now a `500` because the core throws
`ArgumentNullException`. Left alone deliberately: the endpoints validate first, so no request is known
to reach the handler that way, the mapping would be untestable, and the `400` path is the one that
still returns `exception.Message` — mapping a broad exception type onto it would re-open, for any path
nobody enumerated, exactly the leak this PR closes for `500`.

## Baselines

- `PublicApiTest_Quartz.AspNetCore` — `MaxPageSize`.
- `PublicApiTest_Quartz.Dashboard` — the three `MapQuartzDashboard` return types.
- `WireFormatSnapshotTest_ServerFaultProblemDetailsBody` — the fixed detail.

All three accepted through Verify; the diffs are exactly those three things and nothing else.

## Not in this PR

Rate limiting on either surface (none exists anywhere; now a stated caveat on `http-api.md`), an
allow-list for job type names, `SendMailJob`'s SMTP credential in job data (documented on the type and
now cross-referenced), and the dashboard's `ReadOnly` default.

## Verification

```
dotnet build Quartz.slnx                      Build succeeded. 0 Warning(s) 0 Error(s)
dotnet test src/Quartz.Tests.AspNetCore       Passed! Failed: 0, Passed: 588, Skipped: 0, Total: 588
dotnet test src/Quartz.Tests.Unit             Passed! Failed: 0, Passed: 5145, Skipped: 36, Total: 5181
dotnet fallout VerifyDocsSnippets             Documentation snippets are up to date (104 markdown files checked)
npm run docs:check-links                      221 pages, 1166 internal links, 714 fragments; no dead links or anchors
```

`Quartz.Tests.Integration` was not run: it needs Docker, and nothing here touches a path it covers.
`BenchmarkSmoke` was not run: `Quartz.Benchmark` references only `Quartz.Jobs`, and no type it uses
moved.

Closes #3650

### DIFF
--- a/docs/documentation/quartz-4.x/migration-guide.md
+++ b/docs/documentation/quartz-4.x/migration-guide.md
@@ -541,6 +541,15 @@ shape changed — only what a mapping that stated nothing, and a contradictory p
   or `AllowAnonymous()`. Authorization on a `MapGroup` above the mapping counts, and registering the
   services without mapping anything is left alone. **An alpha user who mapped a bare API or dashboard
   has to add one call**, and `AllowAnonymous()` is a supported answer.
+* **`MapQuartzDashboard()` returns an `IEndpointConventionBuilder`, not a
+  `RazorComponentsEndpointConventionBuilder`.** **Changed since alpha.5:** the dashboard is two sets of
+  endpoints — the Blazor pages and the SignalR hub that streams the same data — and the old return type
+  was the pages' builder alone, so `MapQuartzDashboard().RequireAuthorization()` protected the pages and
+  left the hub open, which the startup guard above would now refuse. The builder handed back covers
+  both, so one `RequireAuthorization()` or `AllowAnonymous()` is the whole answer. Code that used
+  members specific to `RazorComponentsEndpointConventionBuilder` on the result (a render-mode call, for
+  instance) configures those on its own `MapRazorComponents` and passes that builder to the
+  `existingComponents` overload instead. The dashboard baseline moved for all three overloads.
 
   `MapQuartzDashboard` returns an `IEndpointConventionBuilder` rather than the
   `RazorComponentsEndpointConventionBuilder` for the same reason: the hub is mapped separately from the
--- a/docs/documentation/quartz-4.x/migration-guide.md
+++ b/docs/documentation/quartz-4.x/migration-guide.md
@@ -473,7 +473,7 @@ exactly on it is listed.
 
 ### Between the alphas and beta.1
 
-The 4.0 surface froze at `4.0.0-alpha.5`. Three exceptions were taken deliberately, because each was a
+The 4.0 surface froze at `4.0.0-alpha.5`. Five exceptions were taken deliberately, because each was a
 statement the API made that was not true; if you are coming from 3.x rather than from an alpha, read
 these as part of 4.0 and nothing more.
 
@@ -504,9 +504,49 @@ these as part of 4.0 and nothing more.
   raises too. A store or a driver description of your own may need a `!` or a null check where the
   compiler now says so; that is the point. `IJobExecutionContext.FireInstanceId` is unchanged and
   non-nullable — a context exists only for a firing.
+* **A `500` from the HTTP API no longer carries the exception's message.** **Changed since alpha.5:**
+  `ExceptionHandler` withheld the exception's *type* on that path — a fault the caller cannot act on
+  says nothing about the server's internals — and returned its *message* verbatim, which for a driver
+  fault names the server, the database, the login or the constraint. The `detail` is now one fixed
+  sentence, *"The scheduler failed to handle the request. The failure is recorded in the server's
+  log."*, and the real message keeps being logged.
+  `QuartzHttpApiOptions.IncludeStackTraceInProblemDetails` — the switch that already says "I am
+  debugging this" — puts it back. A client matching on a `500`'s `detail` was matching on whichever
+  driver failed; it now matches on a constant.
+* **A paged HTTP API request is bounded by `QuartzHttpApiOptions.MaxPageSize`, which is 1000.**
+  **Changed since alpha.5:** `take` rejected only negatives, so one request could ask the store for
+  every trigger in it while the bulk key fetch next door refused a thousand and one keys. A `take`
+  naming a **number** above the cap is now a `400` naming it — `?take=2147483647` included — while
+  `?take=all`, which names no number, is *bounded* by the cap rather than refused by it and reports
+  `hasMore`. That asymmetry is what keeps the 3.x-compatible listings working: `GetJobKeys` and its
+  neighbours ask for everything whether the answer is three rows or three million, and
+  `HttpScheduler` turns a truncated answer to one of those into an exception rather than a short list.
+  Set `MaxPageSize` to `0` for the old behaviour.
 
-One more startup refusal arrived with them, and is not a freeze exception because nothing about the
-shape changed — only what a contradictory pair of values does:
+Two more startup refusals arrived with them, and are not freeze exceptions because nothing about any
+shape changed — only what a mapping that stated nothing, and a contradictory pair of values, now do:
+
+* **`MapQuartzHttpApi()` and `MapQuartzDashboard()` refuse to start when nothing authorizes them.**
+  **Changed since alpha.5:** neither surface was safe by default. `app.MapQuartzHttpApi()` applied
+  authorization only when `SchedulerAuthorizationPolicy` was set, so all sixty routes — `shutdown` and
+  `clear` among them — answered anonymously; the dashboard left its pages and its live-events hub with
+  no authorization metadata at all, with `ReadOnly` defaulting to `false`. Both schedule a job whose
+  type is a string the request supplies, resolved later with `Type.GetType`, so with `Quartz.Jobs` on
+  the probing path an open endpoint is remote code execution rather than an information leak. A hosted
+  lifecycle service now enumerates the endpoints Quartz mapped before the server binds its listener and
+  throws unless every one of them carries authorization metadata or an explicit `AllowAnonymous()`, or
+  the host has a non-null `AuthorizationOptions.FallbackPolicy`. The message names the three fixes:
+  `RequireAuthorization()` on what the map call returns, the surface's policy option
+  (`QuartzHttpApiOptions.SchedulerAuthorizationPolicy` / `QuartzDashboardOptions.AuthorizationPolicy`),
+  or `AllowAnonymous()`. Authorization on a `MapGroup` above the mapping counts, and registering the
+  services without mapping anything is left alone. **An alpha user who mapped a bare API or dashboard
+  has to add one call**, and `AllowAnonymous()` is a supported answer.
+
+  `MapQuartzDashboard` returns an `IEndpointConventionBuilder` rather than the
+  `RazorComponentsEndpointConventionBuilder` for the same reason: the hub is mapped separately from the
+  pages, so the old return value reached the pages and not the hub — and in the overload that takes an
+  existing components builder it reached the host application's own pages too. What comes back now is
+  the dashboard's pages and its hub, and nothing else.
 
 * **`Quartz:Scheduling` refuses `IgnoreDuplicates` while `OverwriteExistingData` is on.**
   **Changed since alpha.5:** the two are answers to the same question and `OverwriteExistingData`
--- a/docs/documentation/quartz-4.x/packages/dashboard.md
+++ b/docs/documentation/quartz-4.x/packages/dashboard.md
@@ -73,7 +73,7 @@ app.UseAuthorization();
 app.UseAntiforgery();
 
 app.MapQuartzHttpApi().RequireAuthorization();
-app.MapQuartzDashboard();
+app.MapQuartzDashboard().RequireAuthorization();
 ```
 <!-- endSnippet -->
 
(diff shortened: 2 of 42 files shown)


lahma:
Coordinator: the `MapQuartzDashboard` return-type change is accepted as a sixth freeze exception (forced by fail-closed: the returned builder must cover the hub). Added its migration-guide row beside the three the PR wrote. CI re-running.

PR #3670, opened 2026-09-02
What 4.0 costs per firing, and what it does over a night, are numbers we can take

### LINKED ISSUE #3653: What 4.0 costs per firing, and what it does over a night, are numbers we have measured
**Milestone:** 4.0.0-beta.1 · **Labels:** tests, docs, infra, 4.x · **Baseline:** none
**Closes:** —

### Why
Beta.1 asks people to run 4.0 in production, and two of the questions a production reader asks have no
answer. **No fire-throughput or per-fire-allocation figure exists for 4.0 at all** (OPS-F11): the migration
guide claims the batched fire path replaced "six to nine round trips" with one and gives no number, and the
cron and `ScheduleJob` numbers that do exist live in `src/Quartz.Benchmark/README.md`, off the docs site
(OPS-F10). And nothing has ever run two clustered nodes through a full workload for an extended period: the
clustered fixtures are minutes long and single-purpose, `[DisallowConcurrentExecution]` non-overlap across
two live nodes is settled by mechanism rather than end to end (N2b §4(b)), and `Quartz.Extensions.Redis`,
the Aspire package and the HTTP surfaces over a persistent store are the least-exercised things we ship.
Q3 item 2 ratified both as **release gates, not CI legs**.

### Evidence
- **OPS-F10.** `grep -rn -i "benchmark" docs/documentation/quartz-4.x/` → three hits, all pointing at the
  acquisition-index analysis in `db/index.md:170-196` and the execution-ceiling figures in
  `tutorial/execution-groups.md:326-338`. `src/Quartz.Benchmark/README.md:30-241` has the substantial numbers
  — cron parse and next-fire against 3.14 and NCrontab, job and trigger build, `ScheduleJob` into
  `RAMJobStore`, the 14-shape cron suite — and nothing on the docs site links them.
- **OPS-F11.** `grep -rn -i "3\.x.*faster\|per second\|throughput" docs/documentation/quartz-4.x/` → three
  hits, none a comparison. `BenchmarkSmoke` (`build/Build.cs:430-443`) runs every benchmark once, excluding
  `RequiresDatabase` and `LongRunning`, and measures nothing — it is a liveness check, and `Build.cs:416-421`
  records that its first run found every `SchedulerBenchmark` case silently failing.
- **N2b §4(b) and the caveats.** No end-to-end test runs a `[DisallowConcurrentExecution]` job on two real
  nodes and asserts non-overlap; `ClusteredExactlyOnceTestBase.cs:145-148` says why its own job is concurrent
  by design. `ClusteredHardening*`, `PreferredNode*` and `SharedDatabaseTenancy*` each assert one property
  over a short run. Nothing has ever watched heap or handle count over time.

### Done means
- **The benchmark (OPS-F11).** Fires per second and allocated bytes per fire over a `NoOpJob`, on
  `RAMJobStore` and on PostgreSQL, at `MaxConcurrency` 10 and 50, with `[MemoryDiagnoser]`;
  `TriggerFirePathBenchmark` and `SchedulerBenchmark` are the harnesses to build on and
  `AcquisitionIndexBenchmark`'s `RequiresDatabase` plumbing is the pattern for the ADO half. **The baseline is
  the same run on `origin/3.x`'s own `Quartz.Benchmark`**, so the round-trip claim finally has a figure.
  The result is a `##` section appended to `src/Quartz.Benchmark/README.md` in the existing house style
  (machine, runtime, the Error-column caveat), plus a short **"Performance"** section in
  `docs/documentation/quartz-4.x/operations.md` that links the README and carries the headline table
  including the existing cron and `ScheduleJob` numbers (OPS-F10).
- **The soak (V6.9).** An opt-in `[Category("LongRunning")]` NUnit test in `Quartz.Tests.Integration`,
  parameterised by `QUARTZ_SOAK_MINUTES` (default 30): two clustered scheduler nodes over one PostgreSQL
  container sharing a `SCHED_NAME` — and the same run against SQL Server as a second configuration — with a
  workload covering every trigger family including recurrence, `[DisallowConcurrentExecution]` jobs behind an
  overlap detector, a retry policy over a job that fails on a schedule, `AddJobTimeout` over a job that
  overruns, induced misfires (a node put in standby past the threshold), and one node killed mid-run —
  shut down without check-in, the hardening tests' technique — and recovered by the survivor.
- At the end the soak asserts: no `ACQUIRED` or `BLOCKED` rows left, `QRTZ_FIRED_TRIGGERS` empty, fire counts
  within tolerance of expectation, **zero overlaps**, no unobserved exceptions, and GC heap and handle count
  sampled every minute and flat.
- Neither is wired into a CI leg. Both are run before the tag, on a quiet machine, and their reports go into
  the PR body and into a sentence on `docs/documentation/quartz-4.x/operations.md` saying what was run and
  for how long — so the release notes' maturity ledger cites a measurement rather than an intention.
- Because the harness touches trigger and scheduler types, `dotnet fallout BenchmarkSmoke` runs locally as
  well: CI runs it inside the unit-test leg, where a BenchmarkDotNet child failure is masked behind a
  `LogIssues` `NullReferenceException` and a green unit suite proves nothing about it.
- The soak needs Docker. If it is unavailable when the PR is prepared, say so and land the harness with the
  runs reported separately before the tag.

### Not in this PR
- Making either a CI leg — they are release gates; `BenchmarkSmoke` stays the CI-side liveness check.
- Fixing anything the soak finds: a finding becomes its own issue and its own PR, so the harness lands green
  or lands with its finding named.
- The acquisition-index and execution-ceiling measurements already published — unchanged.
- Any Redis, Aspire or HTTP-over-persistent-store coverage beyond what the soak happens to exercise; those
  stay on the caveats list.

### LINKED ISSUE #3673: A sub-millisecond repeat interval wedges a simple trigger in ACQUIRED, silently
### What happens

Schedule a `SimpleTrigger` whose repeat interval is shorter than a millisecond into any ADO job
store. It fires once, and then never again. Nothing is reported to the caller, nothing shows in the
scheduler's state, and the trigger's row sits in `ACQUIRED` for the life of the deployment.

Found while writing the fire-throughput benchmark for #3653, which used a one-tick interval to keep
its triggers permanently overdue. On `RAMJobStore` that works. On PostgreSQL the harness fired zero
times and every trigger it acquired stayed `ACQUIRED`:

```
 sched_name                  | trigger_state | count
-----------------------------+---------------+-------
 PostgresThroughputBenchmark | WAITING       |   150
 PostgresThroughputBenchmark | ACQUIRED      |    50
```

With a logger attached, the scheduler thread is looping on this, once per acquisition round, forever:

```
[Error] Quartz.Impl.AdoJobStore.LocalTransactionJobStore: Caught exception: Attempted to divide by zero.
System.DivideByZeroException: Attempted to divide by zero.
   at Quartz.Impl.Triggers.SimpleTriggerImpl.GetFireTimeAfter(Nullable`1 afterTimeUtc) in SimpleTriggerImpl.cs:line 575
   at Quartz.Impl.Triggers.SimpleTriggerImpl.Triggered(ICalendar calendar) in SimpleTriggerImpl.cs:line 408
   at Quartz.Impl.AdoJobStore.AdoJobStoreBase.TriggerFired(...) in AdoJobStoreBase.Acquisition.cs:line 587
   at Quartz.Impl.AdoJobStore.AdoJobStoreBase.<>c__DisplayClass6_0.<<TriggersFired>b__0>d.MoveNext() in AdoJobStoreBase.Acquisition.cs:line 393
```

### Why

Two things, each harmless on its own.

**The store rounds the interval to zero.** `StdAdoDelegate.GetDbTimeSpanValue` casts
`TotalMilliseconds` to `long`:

```csharp
public object? GetDbTimeSpanValue(TimeSpan? timeSpanValue)
{
    return timeSpanValue is not null ? (long?) timeSpanValue.Value.TotalMilliseconds : null;
}
```

One tick is 0.0001 ms, so `REPEAT_INTERVAL` is written as `0`. Reading it back,
`GetTimeSpanFromDbValue` only builds a `TimeSpan` when `millis > 0`, so
`SimpleTriggerPersistenceDelegate` falls to its `?? TimeSpan.Zero`.

**`GetFireTimeAfter` divides by it.** `SimpleTriggerImpl.GetFireTimeAfter` reaches

```csharp
long numberOfTimesExecuted = (timeBetweenAfterAndStart.Ticks / repeatInterval.Ticks) + 1;
```

with `repeatInterval.Ticks == 0`.

**And the failure is invisible.** `AdoJobStoreBase` catches it, logs it and returns; the scheduler
thread's own catch-all logs and continues. The trigger is never released, so it stays `ACQUIRED` —
which no other node will acquire either. A deployment with no log pipeline sees a job that simply
stopped running.

### Scope

- **Both branches.** `main` and `3.x` have the same cast (`StdAdoDelegate.cs:265` on 3.x) and the
  same division (`SimpleTriggerImpl.cs:650` on 3.x).
- **Every ADO dialect**, since the cast is in `StdAdoDelegate`.
- **`RAMJobStore` is unaffected** — it keeps the `TimeSpan` it was given — so the two stores disagree
  about the same trigger, which is the sort of difference the store contract suite exists to catch.
- Any `TimeSpan` the store persists goes through `GetDbTimeSpanValue`, so a sub-millisecond
  `DailyTimeIntervalTrigger` interval is truncated the same way; only `SimpleTriggerImpl` divides by
  it, so only that one throws.

### What to do about it

Three candidates, not exclusive:

1. **Reject it at the door.** `SimpleScheduleBuilder.WithInterval` (and the other builders) could
   refuse an interval that a persistent store cannot represent, the way `RetryPolicy.Explicit`
   refuses a policy longer than its column. That turns a silent stall into an
   `ArgumentOutOfRangeException` at schedule time, which is where the mistake was made.
2. **Do not divide by zero.** `GetFireTimeAfter` returning `null` (or treating a zero interval as
   one-shot) would at least stop the trigger wedging, though it would still be silently doing
   something other than what was asked.
3. **Store ticks rather than milliseconds.** That is a schema and data-format change, so it is a
   different conversation; noted only because it is the fix that loses nothing.

(1) plus (2) looks right: refuse what cannot be stored, and never wedge a trigger on arithmetic.

### Not part of this

There is no evidence anyone schedules a sub-millisecond repeat in production — the benchmark did it
to keep triggers permanently overdue, and now uses one millisecond, which is the smallest interval
both stores agree on. This is filed because the failure mode is *silent*, not because the interval is
a common one.

lahma:
Two of the questions a production reader asks about 4.0 have never had an answer, and both are
measurements rather than code. This lands the two harnesses that produce them, the numbers they
produced, the fix for the bug the first of them found, and the one build change that keeps neither
harness in CI.

## The benchmark (OPS-F11)

`FireThroughputBenchmark` and `FireThroughputPostgresBenchmark` measure a firing end to end — acquire,
fire, run a job that does nothing, complete — against `RAMJobStore` and against a real PostgreSQL
store, at `MaxConcurrency` 10 and 50, with `[MemoryDiagnoser]`. `baseline-3x/` carries the same
workload written against 3.x's API, so both sides came off one machine in one sitting.

The scheduler is built once in `[GlobalSetup]` and left running, so an invocation is *wait for the
next N fires*. That is what makes `Mean` the time one firing took, and it is why the worker threads
are inside the measured window: BenchmarkDotNet reads `GC.GetTotalAllocatedBytes`, which is
process-wide, so `Allocated` is what a firing costs the process rather than what one thread of it did.

### The numbers

AMD Ryzen 9 5950X, .NET 10.0.11, PostgreSQL 15.1 in Docker over loopback at its shipped durability
settings. 3.20 rows taken on `origin/3.x` (`b33c70487`) against **3.x's own** schema. This is a
working machine with other things on it; these runs came out tight anyway — every Error is under
1.5 % of its Mean, and the allocation column is exact whatever the load.

| Store | MaxConcurrency | 3.20 | 4.0 | Fires/second (4.0) |
|---|---|---|---|---|
| PostgreSQL | 10 | 9.865 ms / 136.09 KB | **6.177 ms** / **56.47 KB** | 162 |
| PostgreSQL | 50 | 9.212 ms / 132.10 KB | **6.129 ms** / **52.78 KB** | 163 |
| `RAMJobStore` | 10 | 2.314 µs / 3.25 KB | 3.436 µs / 3.62 KB | 291,000 |
| `RAMJobStore` | 50 | 2.228 µs / 3.25 KB | 3.024 µs / 3.63 KB | 330,700 |

**On a persistent store 4.0 is 1.5× faster per firing than 3.20 and allocates 2.4× less.** That is
the batched fire path, and it is the figure the migration guide's "six to nine round trips replaced
with one" has never had. Counted at the database rather than in the client, a firing now costs
**1.27 commits** — 42,337 transactions over 33,404 firings in one run.

**On `RAMJobStore` 4.0 is about 1.4× slower and allocates 11 % more.** Recorded, not acted on: against
six milliseconds of database, three microseconds of scheduler is invisible in any deployment that has
one. Filed as #3674 and left there.

**Pool size does not start firings faster.** Five times the threads bought 14 % on `RAMJobStore` and
nothing measurable on PostgreSQL, because a node's store operations serialise on `TRIGGER_ACCESS`.
That is `operations.md`'s "adding nodes does not make a single trigger fire faster", measured.

**Two settings could not be left at their defaults**, and both are stated beside the numbers.
`MaxBatchSize` tracks `MaxConcurrency`, because `QuartzOptionsValidators` refuses a batch larger than
the pool that would run it. `BatchTriggerAcquisitionFireAheadTimeWindow` is one second rather than
the shipped zero: the store ends a batch at the first acquired trigger's fire time plus the window,
so at the default a batch is one trigger however large `MaxBatchSize` is.

## The soak (V6.9), and what both runs said

`ClusteredSoakTestBase` runs two clustered nodes over one database for `QUARTZ_SOAK_MINUTES` (30 by
default): every trigger family including recurrence, a `[DisallowConcurrentExecution]` job behind an
overlap detector, a retry policy over a job that always fails, a job that overruns the budget
`AddJobTimeout` enforces, and a job only a recovery trigger can reach — with the failures induced
along the way. Both nodes run in one process, which is what makes the overlap detector possible: a
static counter sees a firing on either of them.

**PostgreSQL, 30 minutes — passed.**

```
  08:42:22 workload scheduled; running for 00:30:00
  08:49:52 both nodes in standby for 00:00:30 to induce misfires
  08:50:22 both nodes resumed
  08:58:52 'soak-node-b' killed mid-flight, leaving an EXECUTING row behind
  08:58:52 survivor recovered the killed node's work
  09:03:22 'soak-node-b' replaced and rejoined the cluster
  09:12:22 run complete, shutting the cluster down
  family-calendarInterval 446 · family-cron 357 · family-dailyTimeInterval 594
  family-recurrence 297 · family-simple 890
  serial 2662 firings, max concurrent 1 · failing 534 attempts, 356 of them retries
  timed out 178 · recovered 1
  heap 4-5 MB and 624-645 handles across all 31 samples
```

**SQL Server, 30 minutes — passed.**

```
  09:20:16 workload scheduled; running for 00:30:00
  09:27:46 both nodes in standby … 09:28:16 resumed
  09:36:46 'soak-node-b' killed … 09:36:46 survivor recovered its work
  09:41:16 'soak-node-b' replaced and rejoined
  09:50:16 run complete
  family-calendarInterval 440 · family-cron 351 · family-dailyTimeInterval 587
  family-recurrence 294 · family-simple 867
  serial 2367 firings, max concurrent 1 · failing 521 attempts, 346 of them retries
  timed out 175 · recovered 1
  heap 4-5 MB and 725-746 handles across all 31 samples
```

End-state assertions, both runs: **no `ACQUIRED` or `BLOCKED` trigger rows**, **`QRTZ_FIRED_TRIGGERS`
empty**, the killed node's `ENTRY_ID` gone, every family's fire count inside the band its schedule
implies, **peak observed concurrency of the `[DisallowConcurrentExecution]` job = 1**, at least one
retry re-fire, at least one timeout, exactly one recovered firing and it was the survivor's, no
unexpected scheduler error, no unobserved task exception, and heap and handle count no larger at the
end than once the run had settled.

**N2b §4(b)'s open question — can a `[DisallowConcurrentExecution]` job double-fire across two nodes —
is now answered by observation on two engines rather than by mechanism.**

## The bug the benchmark found, fixed here (#3673)

`REPEAT_INTERVAL` holds whole milliseconds — `StdAdoDelegate.GetDbTimeSpanValue` casts
`TotalMilliseconds` to `long` — so a `SimpleTrigger` with a sub-millisecond repeat interval was
stored as `0`. Not merely lossy: the trigger read back has a zero repeat interval,
`SimpleTriggerImpl.GetFireTimeAfter` divides by it, and `AdoJobStoreBase` catches the
`DivideByZeroException`, logs it and returns — leaving the row in `ACQUIRED` for ever. A job that
stops running and says nothing. **The benchmark's first PostgreSQL run hit exactly this**: zero
firings, nothing reported anywhere the caller could see.

`AdoJobStoreUtil.RequireStorableDuration` now refuses any `TimeSpan` carrying ticks below a whole
millisecond, naming the trigger, the column and what to do about it, and the three places that write
a simple trigger's interval call it. `GetDbTimeSpanValue` carries the same rule as a backstop, so no
other path can truncate a duration silently either. **`AdoJobStoreBase`'s catch is untouched** — it is
there for the failures nobody has thought of, and the refusal is what makes this one unreachable.
`RAMJobStore` goes on accepting the interval, because it has no column to round it to.

`SubMillisecondIntervalSqliteTest` pins all three claims on a file SQLite database — refusal, a whole
millisecond accepted, and the in-memory store still taking it — so the rule is held where the
coverage gate can see it. It failed before the fix and passes after. One line added to the migration
guide under *Between the alphas and beta.1*.

## The build change

`GetTestFilter` mapped a database to its own `TestCategory` and nothing else, so a fixture carrying
both `db-postgres` and `LongRunning` would have run in the postgres leg — half an hour of wall time
in a pull request, which reads as a hung job rather than as thoroughness. Every leg now excludes
`LongRunning`, whether or not a database was named, because an unnamed one applies no filter at all.

## Docs

`src/Quartz.Benchmark/README.md` gains a `##` section in the house style. `operations.md` gains a
`## Performance` section linking it, carrying the headline table plus the cron and `ScheduleJob`
numbers that existed and were unreachable from the docs site (OPS-F10), and saying what was soaked
and for how long — so the maturity ledger cites a measurement rather than an intention.

## Verification

- `dotnet build Quartz.slnx` — succeeded, 0 warnings.
- `dotnet test src/Quartz.Tests.Unit` — **5205 passed**, 0 failed, 36 skipped.
- `dotnet test src/Quartz.Tests.AspNetCore` — **605 passed**, 0 failed.
- `Quartz.Tests.Integration`, `db-sqlite` leg — 15 passed, 0 failed.
- Both 30-minute soaks — passed, above.
- `dotnet fallout BenchmarkSmoke` — succeeded, 284 benchmarks; both `FireThroughputBenchmark` cases
  in it, `FireThroughputPostgresBenchmark` correctly excluded.
- `dotnet fallout VerifyDocsSnippets`, `VerifyDocsLogEvents`, `npm run docs:check-links` (223 pages,
  1270 links, 778 fragments, no dead anchors), `docs:check-links-test` (7/7), `npm run docs:build`
  (224 pages) — all clean.
- No public API baseline moved.

## For a reviewer

- **The issue says "a node put in standby"; the soak puts *both* nodes in standby.** With two nodes,
  one in standby leaves the other acquiring and nothing misfires. Ruled fine, and now said so on the
  fixture.
- **The PostgreSQL benchmark runs one node, not clustered.** Clustering adds a check-in loop, a
  cluster-wide lock per acquisition cycle and a second node competing for the same rows; none of that
  is in the number. The soak is where two nodes are exercised.
- **The PostgreSQL figure is a commit-latency figure as much as a Quartz one.** With
  `synchronous_commit = off` the 4.0 arm runs at 3.605 ms / 4.140 ms — 1.7× faster — so roughly 40 %
  of the 6.1 ms is this container's fsync. The 3.20-to-4.0 ratio is the durable part.
- **`ClusteredJobStoreTestBase.CreateScheduler` gains an optional `Action<IQuartzBuilder>`**, because
  `AddJobTimeout` registers middleware and middleware has no property key. Additive.
- **The soak's check-in misfire threshold is 15 s against a 2 s interval**, where the minute-long
  fixtures use 2 s against 1 s: thirty times the run length is thirty times the chances for a
  collection and a slow round trip to make a live node look dead, and a spurious failover would
  replay work the soak then counts.

Closes #3653
Closes #3673

### DIFF
--- a/build/Build.cs
+++ b/build/Build.cs
@@ -496,18 +496,43 @@ partial class Build : FalloutBuild, ICompile, IPack
     static readonly string[] DatabaseCategories =
         ["db-postgres", "db-sqlserver", "db-mysql", "db-oracle", "db-firebird", "db-sqlite", "db-redis"];
 
-    string GetTestFilter(string database) => database switch
+    /// <summary>
+    /// The category a release gate carries, and the one thing every integration leg leaves out.
+    /// </summary>
+    /// <remarks>
+    /// The clustered soak (<c>ClusteredSoakTestBase</c>) runs for half an hour by design, and a leg
+    /// that ran it would read as a hung job rather than as thoroughness. It is run by hand before a
+    /// tag, exactly as the benchmarks are; <c>BenchmarkSmoke</c> excludes the same name on the
+    /// benchmark side, through <c>BenchmarkCategories</c>.
+    /// </remarks>
+    const string LongRunningCategory = "LongRunning";
+
+    /// <summary>
+    /// What one integration leg runs: the fixtures for its database, and never a release gate.
+    /// </summary>
+    /// <remarks>
+    /// The <c>LongRunning</c> exclusion is applied whether or not a database was named, because an
+    /// unnamed one applies no filter at all — which is what a local <c>dotnet fallout IntegrationTest</c>
+    /// does, and it would otherwise pick up an hour of soak.
+    /// </remarks>
+    string GetTestFilter(string database)
     {
-        "postgres" => "TestCategory=db-postgres",
-        "sqlserver" => "TestCategory=db-sqlserver",
-        "mysql" => "TestCategory=db-mysql",
-        "oracle" => "TestCategory=db-oracle",
-        "firebird" => "TestCategory=db-firebird",
-        "sqlite" => "TestCategory=db-sqlite",
-        "redis" => "TestCategory=db-redis",
-        "basic" => string.Join("&", DatabaseCategories.Select(c => $"TestCategory!={c}")),
-        _ => null
-    };
+        string databaseFilter = database switch
+        {
+            "postgres" => "TestCategory=db-postgres",
+            "sqlserver" => "TestCategory=db-sqlserver",
+            "mysql" => "TestCategory=db-mysql",
+            "oracle" => "TestCategory=db-oracle",
+            "firebird" => "TestCategory=db-firebird",
+            "sqlite" => "TestCategory=db-sqlite",
+            "redis" => "TestCategory=db-redis",
+            "basic" => string.Join("&", DatabaseCategories.Select(c => $"TestCategory!={c}")),
+            _ => null
+        };
+
+        var excludeLongRunning = $"TestCategory!={LongRunningCategory}";
+        return string.IsNullOrEmpty(databaseFilter) ? excludeLongRunning : $"{databaseFilter}&{excludeLongRunning}";
+    }
 
     Target IntegrationTest => _ => _
         .DependsOn<ICompile>()
--- a/docs/documentation/quartz-4.x/migration-guide.md
+++ b/docs/documentation/quartz-4.x/migration-guide.md
@@ -630,6 +630,16 @@ shape changed — only what a mapping that stated nothing, and a contradictory p
   settings and the default that made one of them inert. `OverwriteExistingData` defaults to `true`, so
   a configuration that only sets `IgnoreDuplicates` is the one this catches; set
   `OverwriteExistingData` to `false` beside it.
+* **A persistent store refuses a repeat interval it cannot hold exactly.**
+  **Changed since alpha.5:** duration columns keep whole milliseconds, and
+  `StdAdoDelegate.GetDbTimeSpanValue` cast `TotalMilliseconds` to `long`, so a `SimpleTrigger` whose
+  repeat interval was shorter than a millisecond was stored as `0` - and a trigger read back with a
+  zero repeat interval throws `DivideByZeroException` out of `GetFireTimeAfter` on its next firing,
+  which the store logs and swallows, leaving the row in `ACQUIRED` for ever. A job that stopped
+  running and said nothing. Scheduling such a trigger into an ADO store now raises
+  `ArgumentException` naming the trigger, the column and the rule; `RAMJobStore` keeps taking it,
+  because it has no column to round the value to. Round the interval to a whole number of
+  milliseconds. See [#3673](https://github.com/quartznet/quartznet/issues/3673).
 
 ### What is on nobody's list because it did not change
 
--- a/docs/documentation/quartz-4.x/operations.md
+++ b/docs/documentation/quartz-4.x/operations.md
@@ -677,6 +677,81 @@ Adding nodes does not make a single trigger fire faster, and it does not shorten
 capacity for concurrent firings and it adds a node to fail over to. If the schedule is dominated by one
 long job, a second node changes nothing about it.
 
+## Performance
+
+The numbers below are the only end-to-end firing figures published for 4.0. They come from
+[`src/Quartz.Benchmark/README.md`](https://github.com/quartznet/quartznet/blob/main/src/Quartz.Benchmark/README.md),
+which carries the full method, the machine, the settings and the caveats; this is the summary, and it
+is here because a page about running a cluster is where somebody asks the question.
+
+**Read them as ratios, not as a capacity plan.** One machine, one PostgreSQL container over loopback,
+one node, no clustering. Your storage, your network and your jobs decide the absolute; what carries
+across is how 4.0 compares with 3.20 on the same box on the same day.
+
+### A firing, end to end
+
+One firing means acquire, fire, run a job that does nothing, complete. `MaxBatchSize` tracks
+`MaxConcurrency` in these runs, because the scheduler refuses a batch larger than the pool that would
+have to run it, and `BatchTriggerAcquisitionFireAheadTimeWindow` is one second rather than the shipped
+zero — without a window a batch is one trigger however large `MaxBatchSize` is, which is the shape a
+deployment left at the defaults gets.
+
+| Store         | MaxConcurrency | 3.20               | 4.0                | Per firing  |
+|-------------- |--------------- |------------------- |------------------- |------------ |
+| PostgreSQL    | 10             | 9.87 ms / 136 KB   | 6.18 ms / 56 KB    | 1.6x faster, 2.4x less garbage |
+| PostgreSQL    | 50             | 9.21 ms / 132 KB   | 6.13 ms / 53 KB    | 1.5x faster, 2.5x less garbage |
+| `RAMJobStore` | 10             | 2.31 µs / 3.25 KB  | 3.44 µs / 3.62 KB  | 1.5x slower |
+| `RAMJobStore` | 50             | 2.23 µs / 3.25 KB  | 3.02 µs / 3.63 KB  | 1.4x slower |
+
+That is about 162 firings a second per node on PostgreSQL and about 300,000 on `RAMJobStore`, on this
+machine. The persistent-store gain is the batched fire path: a firing now costs **1.27 database
+commits**, where it used to cost one round trip per statement. The in-memory loss is 4.0's per-firing
+machinery — a DI scope, the middleware pipeline, the execution-group ledger — and against six
+milliseconds of database it is invisible in any deployment that has one
+([#3674](https://github.com/quartznet/quartznet/issues/3674)).
+
+**A bigger pool does not start firings faster.** Five times the threads bought 14 % on `RAMJobStore`
+and nothing measurable on PostgreSQL, because a node's store operations serialise on the
+`TRIGGER_ACCESS` lock. What `MaxConcurrency` buys is more *jobs* running at once, which is the point
+of it — see [Sizing a cluster](#sizing-a-cluster) above.
+
+### Scheduling and cron
+
+Also measured against 3.14, and also in the benchmark README:
+
+| Operation                                       | 3.14                | 4.0                 |
+|------------------------------------------------ |-------------------- |-------------------- |
+| Parse a `CronExpression`                        | 3.0–4.6 µs / 8.7–12.9 KB | 236–338 ns / 576–688 B |
+| `GetNextValidTimeAfter`                         | ~1.29 µs / ~3.2 KB  | 315–373 ns / **0 B** |
+| `ScheduleJob`, cron trigger, into `RAMJobStore` | 31 µs / 38.7 KB     | 10.5 µs / 5 KB      |
+
+Two other measurements live closer to what they are about:
+[the acquisition index](db/index.md#indexes-and-the-acquisition-index-in-particular), over a hundred
+thousand triggers on four engines, and
+[the cluster-wide execution ceiling](tutorial/execution-groups.md#cluster-scoped-limits), which costs
+one aggregate per acquisition attempt.
+
+### What has been run against a cluster
+
+Before the 4.0.0-beta.1 tag, two clustered nodes sharing one scheduler name were run for **30 minutes
+against PostgreSQL and 30 minutes against SQL Server**, carrying every trigger family including
+recurrence, a `[DisallowConcurrentExecution]` job behind an overlap detector, a retry policy over a
+job that always fails, a job that overruns the budget `AddJobTimeout` enforces, and induced failures
+along the way: both nodes into standby past the misfire threshold, then one node killed with an
+`EXECUTING` row left behind and its check-in aged, then replaced once the survivor had recovered it.
+
+Both runs passed. Each ended with no `ACQUIRED` or `BLOCKED` triggers and an empty
+`QRTZ_FIRED_TRIGGERS`, every trigger family still firing at its schedule, **no overlapping firing of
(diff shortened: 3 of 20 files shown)


lahma:
Coordinator commit: Sonar's new-code coverage gate (68 % of 13 lines) missed the `GetDbTimeSpanValue` backstop branch; `DurationStorageRuleTest` pins both doors of the rule directly. CI re-running.

PR #3716, opened 2026-09-07
Fix body margin

NEXXDev:
## Summary

Added body { margin: 0 } to the dashboard css file.

## Details

The dashboard renders inside its own Blazor document and doesn't reset the host document's <body>, so the browser's user-agent default margin (8px) leaks through. This produces an unappealing gap around the edges of every dashboard page.

image


lahma:
Thank you!

Issue #3690, opened 2026-09-02
CronCalendar.GetNextIncludedTimeUtc can walk a century one second at a time


lahma:
`CronCalendar.GetNextIncludedTimeUtc` (`src/Quartz/Impl/Calendar/CronCalendar.cs:172-203`) calls `CronExpression.GetNextInvalidTimeAfter` (`CronExpression.cs:395-425`, whose own comment says "Performance will be very bad here"). For an expression that excludes everything — `* * * * * ?` — the loop advances one second per iteration until `GetTimeAfter` gives up at `TriggerConstants.YearToGiveUpSchedulingAt` (`now.Year + 100`): roughly 3×10⁹ full cron computations. Not reachable from any HTTP request or scheduler path (only the `ICalendar` declaration names it outside `Impl/Calendar/`; the trigger paths use `IsTimeIncluded`), so a public-API hazard for an application that calls it, found by the rc.1 security gate. Fix: compute the next included instant from the expression's structure (the complement of a cron set is itself enumerable), or bound the walk and throw a clear exception.

Issue #3338, opened 2026-08-22
Runtime tenant lifecycle: add and restart schedulers in a running container


lahma:
Two related gaps from the multi-tenancy audit, both deliberately deferred at the alpha:

1. **Onboarding**: AddQuartz mutates IServiceCollection, so a tenant scheduler cannot be added after BuildServiceProvider; QuartzSchedulerBuilder builds a separate container whose scheduler is invisible to the host's repository/HTTP API/dashboard (manual Bind works and the repository now self-heals dead entries, #3297).
2. **Restart**: a shut-down scheduler cannot be rebuilt in the same container — its parts are keyed singletons — and DefaultSchedulerFactory now throws an honest SchedulerException (#3297) instead of resurrecting a dead pool. Real restart support means re-creatable component graphs.

Both are architectural; they should be designed together.

lahma:
## Design spike: runtime tenant lifecycle

Read-only spike against `f04f5d0`. No code written, nothing changed. Everything below marked **verified**
was read in the tree or measured; everything marked **inferred** was not.

**Recommendation in short:** build one additive seam — an `ISchedulerManager` whose runtime-created
schedulers live in their own small child containers, owned by the manager and made visible through the
container's `ISchedulerRepository` — and express restart as a second phase on the same machinery rather
than as a separate feature. Onboarding is purely additive and should land first; restart touches
`DeferredScheduler` and the registration path and should land second, in the same milestone.

---

### 1. The root cause, stated precisely

Both gaps are the same sentence: **a scheduler's parts are singletons whose descriptors were fixed when
the container was built, and neither a new tenant nor a second generation of an old one can get a new
set.**

Three measurements pin it down (verified empirically, `Microsoft.Extensions.Hosting` 10.0.0, .NET 10):

| | result |
|---|---|
| descriptor added to `IServiceCollection` after `BuildServiceProvider()` | invisible to the already-built provider |
| `hostBuilder.Services.Add(...)` after `Build()` | `InvalidOperationException: The service collection cannot be modified because it is read-only.` |
| keyed singleton disposal | disposed with the provider, not before |

So under the generic host — the common case — a runtime `AddQuartz` throws; under a bare
`ServiceCollection` it silently does nothing. Both roads end at "no", which is why this is architectural
rather than a missing overload.

And the parts genuinely cannot be reused. Verified in the tree:

- `TaskSchedulingThreadPool.shutdownCancellation` is a `readonly` source created at construction and
[...]

PR #3733, opened 2026-09-08
A textual day-of-week takes a step, and it is the numeric one

### LINKED ISSUE #3732: A textual day-of-week takes a step, and it means what the numeric one means
`MON/2` is rejected on 4.0. The numeric `2/2` beside it parses and means a step through the week — Monday, Wednesday and Friday — so the field has one construct with two spellings, and only one of them is accepted.

4.0 refused the token deliberately (#3591, ratified in the alpha.5 audit, #3603): Quartz.NET 3.x read a textual day-of-week step as *every N weeks*, so accepting it as the step would have turned a 3.x fortnight into a thrice-weekly schedule in silence. That decision was recorded as "`MON/2` stays rejected; 4.1 may accept it additively", and this is 4.1 taking it up.

What it should mean:

- `MON/2` is `2/2` — MON, WED, FRI. The step runs to Saturday and does not wrap, as it does after a number.
- `MON-FRI/2` is `2-6/2`, by the same rule; the two spellings stay in lockstep, whatever either can express.
- The value after `/` is a whole number from 1 to 7. `MON/X` and `SUN/9` stay rejected, and `MON/0` says what `2/0` says.
- 3.x's fortnight is `RecurrenceScheduleBuilder.Create("FREQ=WEEKLY;INTERVAL=2;BYDAY=MO")`, which anchors on the trigger's start time and so keeps a phase the cron reading never had.

The hazard this creates is the reason 4.0 said no, and it does not go away by accepting the token: an expression carried from 3.x to 4.1 changes meaning and nothing can report it, because the expression is valid. The documentation has to lead with that — `cron-expressions.md`, the migration guide, and `best-practices.md`, which is where an operator looks before an upgrade — and the migration guide's audit has to say that this is the one shape `TryParse` cannot find for you.

lahma:
`MON/2` is `2/2`: Monday, Wednesday and Friday, the same step through the week its numeric twin has always named. `MON-FRI/2` is `2-6/2` by the same rule. The value after the `/` still has to be a whole number from 1 to 7, so `MON/X` and `SUN/9` are rejected — with a message that now says what a step has to be as well as where the fortnight lives — and `MON/0` says what `2/0` says.

Stacked on #3731 (the #3690 fix, milestone 4.0.1), which is where this branches from. **Rebase onto `main` once that merges**; the two touch `CronExpression.cs` in different places.

## The reviewer decision

Quartz.NET 3.x read `MON/2` as *every second Monday*. 4.0 refused the token rather than re-read it as a step in silence, and #3603 recorded the refusal with "4.1 may accept it additively". 4.1 accepting it means **an expression carried across both majors turns a fortnightly job into a thrice-weekly one — 26 fires a year become 156 — and nothing logs it, because on 4.1 the expression is valid.** That is precisely what 4.0's rationale said it would not do, and the mitigation is documentation rather than code: nothing can warn on a valid expression.

So the question a reviewer has to settle is whether the documentation carries it. What is written:

- `cron-expressions.md` gains a section of its own, [`MON/2` is a step through the week], with a `::: danger` box for the 3.x reading, and the `/` bullet points at it.
- The migration guide gains a **From 4.0 to 4.1** section (two rows) linked from *Start here*; the "parser refuses" table row becomes "`FormatException` on 4.0; parses on 4.1"; and *Before you upgrade* now says that `TryParse` against the new assembly is **not** the authoritative check for this one shape, because it parses — a name in front of a `/` has to be read by eye.
- `best-practices.md` says the expression means three different things on three versions, which is what an operator planning an upgrade needs in one sentence.
- `AGENTS.md`'s ruling becomes "`MON/2` is 4.1's `2/2`", so the next audit does not reopen it.

The `From 4.0 to 4.1` section is deliberately one small table; it will merge with whatever #3338 lands.

## A second behaviour change, found on the way

**`MON-FRI/2` parsed on 4.0 and meant `MON-FRI`.** The textual range read its two day names, stopped, and dropped whatever followed — so an expression asking for three days fired on five, silently. That is exactly the shape #3591 removed everywhere else (`1-5W`, `1-5X`, `? * L-3`), and it survived because nobody looked past the `/` in the *textual* range. Reading the step fixes it; `AStepAfterATextualRangeIsNotDropped` keeps it fixed, and the 4.0 → 4.1 table carries the row.

## What the parser does

`StoreExpressionGeneralValue`'s day-of-week `/` case no longer throws: it validates the step and hands off to `HandleSlashOption`, the same method the numeric path uses, so the two produce the same field by construction rather than by parallel code. The `-` case reads a following step through the same method.

A step of `0` is left to `CheckIncrementRange`, so `MON/0` and `2/0` give the same message. A step that is not a number, or is above 7, keeps the textual message, because a caller who writes one is usually reaching for 3.x's fortnight and the message is where that is explained. `2/9` still says `Increment > 7 : 9` — the two messages differ, which is deliberate: only the textual spelling has a fortnight to explain.

The Unix rewriter keeps emitting `2/2` for crontab's `1/2`. It could emit `MON/2` now, but an expression already stored must not change spelling under an upgrade.

## Public API

No change. No baseline line moves; the parser accepts a string it used to reject, which is behaviour, not surface.

## Tests

- `ATextualDayOfWeekStepIsItsNumericTwin` — seven forms, each compared with its numeric twin both as a resolved field and as a week of fire times, so neither spelling can drift from the other.
- `AStepAfterATextualRangeIsNotDropped` — `MON-FRI/2` is three days, not five.
- `ATextualDayOfWeekStepThatIsNotAStepIsRejected` — `MON/X`, `SUN/9`, `MON/0`. These were rows in `ExpressionsThatSaidOneThingAndDidAnotherAreRejected`, standing beside `MON/2`; that fixture is about shapes 3.x reinterpreted, and what is left here is a different claim.
- `ExpressionsTheNewRejectionsMustNotCatchStillParse` gains `MON/2` and `MON-FRI/2`.
- `UnixCronFormatTest` and `CronExpressionBuilderTest` keep their assertions; their reasons said "Quartz rejects `MON/2`" and now say why the numeric form is still what is emitted.

## Verification

```
dotnet build Quartz.slnx                  Build succeeded, 0 warnings
dotnet test src/Quartz.Tests.Unit         Passed: 6116, Failed: 0, Skipped: 36 (50 s)
dotnet test src/Quartz.Tests.AspNetCore   Passed:  630, Failed: 0
dotnet fallout VerifyDocsSnippets         Succeeded, 115 markdown files
npm run docs:check-links                  224 pages, 1412 internal links, 869 fragments, no dead links or anchors
npm run docs:lint                         92 files, 0 errors
```

Documentation sites touched: `docs/documentation/quartz-4.x/cron-expressions.md`, `docs/documentation/quartz-4.x/migration-guide.md`, `docs/documentation/best-practices.md`, and `AGENTS.md`. `docs/documentation/quartz-3.x/` is untouched, as it must be — on 3.x the token really does mean a fortnight.

Integration tests were not run: no Docker daemon on this machine.

Closes #3732

https://claude.ai/code/session_013XXTwomkp4QLt6vbamKxEK

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -208,7 +208,9 @@ Extension policy, not taste. `how-tos/extending-quartz.md` is the reader-facing 
   runs the default on the forwarder; `DelegatingForwardingTest` sweeps both delegating types for it.
 - **Read-replica routing, if ever, is a DIM `IDbProvider.CreateReadConnection()`** — never a `readOnly`
   parameter on `CreateConnection`, which would break both public `IDbProvider` implementations.
-- **`MON/2` stays rejected** (ratified 2026-09-01); 4.1 may accept it additively.
+- **`MON/2` is 4.1's `2/2`**, and `MON-FRI/2` is `2-6/2`: a textual day-of-week takes a step, in lockstep
+  with the numeric form. 4.0 refused it because 3.x read it as a fortnight; that silent change of meaning
+  is a documented hazard, not a reason to reopen this.
 
 ## Build & Test
 
--- a/docs/documentation/best-practices.md
+++ b/docs/documentation/best-practices.md
@@ -446,14 +446,20 @@ for every other Monday and `FREQ=DAILY;INTERVAL=3` for a three-day cadence that 
 [3.x](/documentation/quartz-3.x/tutorial/recurrencetrigger). Reaching for several cron triggers, or
 for a cron expression with a workaround in it, is usually the sign.
 
-**Quartz 3.x only:** `0 0 0 ? * MON/2` — a textual day-of-week with a step — means every other
-Monday there. 4.x rejects it, because the fortnight's phase was anchored to whatever last asked the
-expression a question: a misfire, a restart or a failover recomputed it from a different day and
-moved it. `FREQ=WEEKLY;INTERVAL=2;BYDAY=MO` anchors on the trigger's start time instead, so the
-fortnight is a property of the trigger rather than of the caller. It is one of seven shapes 4.x
-refuses that 3.x accepted and then quietly reinterpreted;
+**The one expression that means three different things:** `0 0 0 ? * MON/2` — a textual day-of-week
+with a step. On **3.x** it is every other Monday. On **4.0** it is a `FormatException`. From **4.1**
+it parses and means what `0 0 0 ? * 2/2` means: Monday, Wednesday and Friday, 156 fires a year rather
+than 26, with nothing logged either way. So audit for it before upgrading from 3.x, whichever 4.x you
+are going to:
+[`MON/2` is a step through the week](/documentation/quartz-4.x/cron-expressions#mon-2-is-a-step-through-the-week)
+is the whole story, and
 [Forms the parser refuses](/documentation/quartz-4.x/cron-expressions#forms-the-parser-refuses) lists
-them, and the ones to audit a database for before upgrading.
+the six shapes 4.x rejects that 3.x accepted and then quietly reinterpreted.
+
+The fortnight went because its phase was anchored to whatever last asked the expression a question: a
+misfire, a restart or a failover recomputed it from a different day and moved it.
+`FREQ=WEEKLY;INTERVAL=2;BYDAY=MO` anchors on the trigger's start time instead, so the fortnight is a
+property of the trigger rather than of the caller.
 
 ### The two daylight saving rules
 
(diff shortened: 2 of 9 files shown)


lahma:
**No checks will appear on this PR while its base is a branch.** Every workflow here is `on: pull_request: branches: [main, 3.x]`, so a PR whose base is `401/cron-calendar-walk` triggers nothing — build, unit tests, docs and Sonar all wait until the base flips to `main`, which GitHub does automatically when #3731 merges. Rebase then, and CI runs on the real thing.

Run locally in the meantime, on `b2b9305a4f`:

```
dotnet build Quartz.slnx                  Build succeeded, 0 warnings
dotnet test src/Quartz.Tests.Unit         Passed: 6116, Failed: 0, Skipped: 36
dotnet test src/Quartz.Tests.AspNetCore   Passed:  630, Failed: 0
dotnet fallout VerifyDocsSnippets         Succeeded, 115 markdown files
npm run docs:check-links                  no dead links or anchors
npm run docs:lint                         92 files, 0 errors
```

PR #3792, opened 2026-09-14
The tag build drafts its own release and attaches the zip

lahma:
`PackZip` has always built `Quartz.NET-<version>.zip` on every `Pack`, and on a tag build it went into the workflow artifact and no further. Attaching it to the GitHub release was a manual step, and it was quietly dropped during the 4.0 campaign — of the releases cut since `v4.0.0`, only that one carried a zip, though every release back to 2.x had one.

Those are already backfilled by hand, each zip taken from that tag's own publish run rather than rebuilt: `v4.1.0`, `v3.22.0`, `v3.21.0`, `v4.0.1`, `v3.20.1`, `v3.20.0`. `v4.0.0`'s existing asset was verified sha256-identical to its run's artifact. The 4.0.0 prereleases were left bare — superseded, and prereleases have no zip precedent.

## What this adds

`DraftRelease` (`build/Build.Release.cs`) runs after `Publish` has pushed the packages, opens the release for the tag **as a draft** with the notes GitHub generates from `.github/release.yml`, and attaches the archive. Nothing becomes public: a human edits the notes and presses Publish, as before.

It is written to be re-runnable, because re-running the publish workflow is the recovery path for a release that broke halfway:

- an existing draft with notes already written is left exactly as it is — accumulating notes ahead of a tag is how this repository works;
- an already published release has its notes left alone too, and only gains the asset;
- an asset of the same name is deleted first, rather than hitting the 422 GitHub answers a duplicate with.

Two API details are load-bearing and are commented as such. `releases/tags/{tag}` answers with **published** releases only and 404s on a draft, so a miss is followed by a walk of the release list — the only place a draft shows up. And a `PATCH` of a draft resends `tag_name`, because a body-only patch clears it.

`Fallout.Components`' own `ICreateGitHubRelease` was considered and rejected: it takes its body from `IHasChangelog`/`ChangelogFile`, and there is deliberately no changelog file on either branch, and it cannot ask GitHub to generate notes.

## Permissions

`contents: write` now sits next to `id-token: write` in the one workflow that can mint a nuget.org API key. That was weighed rather than assumed, and the reasoning is on the attribute: the archive is already on disk in that job, so a separate workflow would have to fetch it back across runs through `workflow_run` (which resolves against the default branch's workflow file, not the tag's); the job is reachable only by pushing a v-tag; it is gated by the `nuget` environment; and it already runs the tagged commit's own build code.

## Verification

- Workflow regeneration is exactly the three expected lines — permissions, the target list, the `GITHUB_TOKEN` env entry — and no other workflow moved.
- `POST /releases/generate-notes` was confirmed against this repository for both `v4.1.0` and a not-yet-existing `v4.2.0`; both come back categorised through `.github/release.yml`.
- `DraftRelease` on an untagged build skips on `OnlyWhen: IsTaggedBuild` and drags nothing in behind it — it is `TriggeredBy(Publish)`, not `DependsOn`, so running it by hand cannot push packages.
- **Still to run before merge:** the end-to-end rehearsal against a throwaway draft (create → idempotent re-run → hand-written body survives → cleanup). Script is ready; it needs a token, so it has not been run from this branch yet.

## Note

`AGENTS.md` is now at 32,767 of its 32,768-byte cap on a CRLF checkout, which is what `AgentInstructionsTest` measures. It passes, but the file is full — the release-notes bullet had to be tightened to fit this in at all. A trim pass is worth doing separately.

Companion 3.x pull request follows; `build/Build.Release.cs` is byte-identical there.

https://claude.ai/code/session_01K7MFZ3oDEHdwRnxjLjQT6n


lahma:
Holding this until after the 4.1.1 tag, deliberately: 4.1.1 is a two-fix patch and a poor thing to make the first customer of new release automation. It will be cut on the machinery that shipped 4.0.1, 4.1.0 and 3.22.0; this one can prove itself on 4.2.0, where there is a schema migration to watch anyway.

Nothing to change here — it is green and mergeable, it just goes second.

lahma:
Reviewed and rehearsed before merging. One defect found and fixed in the second commit; the rest holds.

**The ordering.** A trigger only says "after Pack", not "before the next invoked target". The v4.1.1 tag build's own summary ran `PackZip` last of all, after `Publish`, and the planner schedules an invoked target ahead of a merely triggered one, so as written `DraftRelease` ran before `PackZip` and failed on the missing archive — after `Publish` had pushed the packages. Rehearsed with `Pack Publish DraftRelease` in the plan, the order was `Pack → Publish → PackZip → DraftRelease` only once `.After(PackZip)` was added. `After`, not `DependsOn`, so a by-hand recovery run does not rebuild the archive it is about to attach. `build/Build.Release.cs` stays byte-identical on both branches.

**The rehearsal** (against a throwaway draft in this repository, deleted afterwards), run as the tag build runs it — `GITHUB_ACTIONS=true GITHUB_REF_TYPE=tag GITHUB_REF_NAME=v0.0.0-rehearsal GITHUB_REPOSITORY=quartznet/quartznet GITHUB_TOKEN=$(gh auth token) dotnet fallout Pack DraftRelease`:

| Run | Result |
|---|---|
| 1. create | `Drafted v0.0.0-rehearsal as a prerelease with generated release notes` (2,549 chars), `Attached Quartz.NET-0.0.0-rehearsal.zip (46,412,007 bytes)` |
| 2. re-run | `already has a draft with notes written — attaching the archive and leaving them alone`, asset removed and re-attached |
| 3. body patched by hand, re-run | body `hand-written notes - must survive` intact, `tag_name` intact, one asset |

Two things the rehearsal taught, now in AGENTS.md on both branches: a local `v*` tag does not rehearse this (`IsLocalBuild` restamps the suffix to `-dev` after the tag is read, so every run would name a different release), and `--skip` skips a target's dependencies rather than the named target, so `Publish` cannot be held out of a plan it is invoked in.

**Against the real v4.2.0 draft** (already written, targeting the candidate): the list walk finds it by `tag_name`, the body is non-empty, so the tag build will leave the notes alone and attach the archive. Publishing and "latest" stay a human's.

PR #3793, opened 2026-09-14
The tag build drafts its own release and attaches the zip

lahma:
Companion to #3792, which carries the reasoning in full.

`PackZip` has always built `Quartz.NET-<version>.zip`, and on a tag build it went into the workflow artifact and no further. `v3.20.0`, `v3.20.1`, `v3.21.0` and `v3.22.0` all shipped without the zip that every 3.x release before them carried. Those are already backfilled by hand, each taken from that tag's own publish run.

`DraftRelease` runs after `Publish` has pushed the packages, opens the release for the tag **as a draft** with the notes GitHub generates from `.github/release.yml`, and attaches the archive. Nothing becomes public: a human edits the notes and presses Publish — and leaves **"Set as the latest release" unticked**, because 4.x is the current line. The target asks for `make_latest: "legacy"`, which settles it by semantic version and creation date, but the publish dialog's checkbox is what finally decides it.

## Ported, and the one thing that is not

`build/Build.Release.cs` is **byte-identical** to main's copy. It leans only on `ArtifactsDirectory`, `FullVersion`, `VersionSuffix`, `IsTaggedBuild`, `GitRepository` and `Publish`, which mean the same thing on both branches, and the archive glob covers `artifacts/` here and `artifacts/packages/` there without a difference.

`.github/release.yml` is the one file that is not a port: this branch had none, so the generated notes would have come out as one flat "What's Changed" list. The categories match main's, and `breaking` is listed first precisely because it should never appear on a maintenance branch — if it ever does, it is at the top where somebody will see it. Every label it names already exists on the repository.

## Verification

- The build project compiles unchanged with the file added, and `DraftRelease` on an untagged build skips on `OnlyWhen: IsTaggedBuild` without dragging `Publish` in.
- The regenerated `publish.yml` diff is the same three lines as on main.
- **Still to run before merge:** the same end-to-end rehearsal as #3792.

https://claude.ai/code/session_01K7MFZ3oDEHdwRnxjLjQT6n


lahma:
Reviewed and rehearsed before merging. One defect found and fixed in the second commit; the rest holds.

**The ordering.** A trigger only says "after Pack", not "before the next invoked target". The v4.1.1 tag build's own summary ran `PackZip` last of all, after `Publish`, and the planner schedules an invoked target ahead of a merely triggered one, so as written `DraftRelease` ran before `PackZip` and failed on the missing archive — after `Publish` had pushed the packages. Rehearsed with `Pack Publish DraftRelease` in the plan, the order was `Pack → Publish → PackZip → DraftRelease` only once `.After(PackZip)` was added. `After`, not `DependsOn`, so a by-hand recovery run does not rebuild the archive it is about to attach. `build/Build.Release.cs` stays byte-identical on both branches.

**The rehearsal** (against a throwaway draft in this repository, deleted afterwards), run as the tag build runs it — `GITHUB_ACTIONS=true GITHUB_REF_TYPE=tag GITHUB_REF_NAME=v0.0.0-rehearsal GITHUB_REPOSITORY=quartznet/quartznet GITHUB_TOKEN=$(gh auth token) dotnet fallout Pack DraftRelease`:

| Run | Result |
|---|---|
| 1. create | `Drafted v0.0.0-rehearsal as a prerelease with generated release notes` (2,549 chars), `Attached Quartz.NET-0.0.0-rehearsal.zip (46,412,007 bytes)` |
| 2. re-run | `already has a draft with notes written — attaching the archive and leaving them alone`, asset removed and re-attached |
| 3. body patched by hand, re-run | body `hand-written notes - must survive` intact, `tag_name` intact, one asset |

Two things the rehearsal taught, now in AGENTS.md on both branches: a local `v*` tag does not rehearse this (`IsLocalBuild` restamps the suffix to `-dev` after the tag is read, so every run would name a different release), and `--skip` skips a target's dependencies rather than the named target, so `Publish` cannot be held out of a plan it is invoked in.

**Against the real v4.2.0 draft** (already written, targeting the candidate): the list walk finds it by `tag_name`, the body is non-empty, so the tag build will leave the notes alone and attach the archive. Publishing and "latest" stay a human's.

lahma:
The first push of the ordering fix did not compile here: 3.x wrote the archive at the end of `Pack` itself, so there was no `PackZip` to order after. The third commit moves that zip step out of `Pack` verbatim into a `PackZip` target `Pack` triggers, the shape `main` has, so `build/Build.Release.cs` stays byte-identical on both branches. `Pack` still cleans the artifacts folder and packs the same projects; the archive still lands in `artifacts/`; the build schema gains the target.

Rehearsed on this branch the way 3.x reads a tag (from git, not the Actions environment): a local `v0.0.0-rehearsal` tag, then `Compile Pack DraftRelease` (Compile because the archive copies the examples' binaries):

| Run | Result |
|---|---|
| 1. create | `Drafted v0.0.0-rehearsal as a prerelease with generated release notes`, `Attached Quartz.NET-0.0.0-rehearsal.zip (51,716,667 bytes)` |
| 2. re-run | notes kept, asset removed and re-attached |
| 3. body patched by hand, re-run | body intact, `tag_name` intact, one asset |

Draft and tag deleted afterwards. AGENTS.md here carries the 3.x recipe.

Issue #3811, opened 2026-09-19
Quartz.NET 4.2


lahma:
**Theme: "Quartz.NET, declared."** 4.0 made the API a promise, 4.1 made schedulers runtime-managed and the dashboard remote-capable. 4.2 closes the two gaps every comparison names — declaring a job in one attribute with the cron checked by the compiler, and telling one job to run after another — and publishes performance numbers a competitor cannot wave away. Everything is additive (default interface members, context objects, nullable columns); the one schema change is a generated `database/migrations/4.2/` that is safe under a mixed 4.1/4.2 cluster.

Decided by @lahma on 2026-09-14 from the landscape analysis (TickerQ 10.4.0, Hangfire 1.8.25, our 4.1 inventory and benchmark suite): scope A–G below; continuations store-owned with the schema change; the analyzer inside `Quartz.nupkg`; no outward reaction to TickerQ's comparison table or benchmark PR — profile where we lag instead.

| Order | Item | Issue | After |
|---|---|---|---|
| 0 | AGENTS.md trimmed so 4.2's conventions fit | #3810 | — |
| 0 | Comparison page, coming-from guides, stale claims | #3808 | — (docs only, runs beside anything) |
| 1 | **Profile** every surface where we lag (D1) | #3802 | — |
| 2 | Cron next-occurrence fast path | #3801 | D1's ranking |
| 3 | Head-to-head harness + published numbers (D2), RAMJobStore fire-path cuts (D3) | #3802 | #3801 |
| 4 | `Quartz.Analyzers`: compile-time cron/timeout checks | #3803 | — |
| 5 | `[QuartzJob]`/`[CronTrigger]` + source generator | #3804 | #3803 |
| 6 | Continuations: store + API + the 4.2 migration | #3805 | — |
| 7 | Continuations: dashboard, HTTP API, files, docs | #3806 | #3805 |
| 8 | Retries exhausted + jitter | #3807 | #3805 |
| 9 | ADO-backed execution history (same migration folder) | #3771 | #3805 |
| 10 | Store-attached dashboard targets | #3772 | #3771 |
| filler | Health check: stalled firing | #3809 | — |

Left for 4.3, unassigned: #3770 (operation catalogue), #3773 (agent target), #3774 (fleet pane), #3775 (Aspire resource), #3687 (document DB), #3720 (`Quartz.Cron` — interesting only after #3801 makes the parser fast).

[...]

PR #3841, opened 2026-09-19
A scheduler built without a container gets what ConfigureAllQuartzSchedulers said

lahma:
Found by the 4.2 release gate's consumer run (G4/G5 on #3811): the database-backed execution history was silently inert on a scheduler built by `QuartzSchedulerBuilder`.

## What was wrong

`AddQuartz` carries the delegates `ConfigureAllQuartzSchedulers` recorded to a scheduler in the pass that runs straight after the scheduler's own callback (`registry.ApplyConfigureAll`, `QuartzServiceCollectionExtensions.cs`). `QuartzSchedulerBuilder.Build()` re-implements the phases and never made that pass. A delegate recorded while the builder's callback ran was therefore recorded against a scheduler that did not exist yet and applied to nothing:

| Registration | Standalone builder, before |
|---|---|
| `UsePersistentStore(s => s.UseExecutionHistory())` (4.2) | store registered, recorder never installed: 0 rows for every firing |
| `quartz.jobStore.executionHistory = true` (4.2) | same |
| `AddQuartzSchedulerEvents()` (since 4.0.0) | plugin never installed |
| `Services.ConfigureAllQuartzSchedulers(q => q.AddPlugin<T>())` | dropped |

The legacy key had a second layer: the bridge records its delegate while the property-derived options are built in a seed `ServiceCollection` of their own, so `SchedulerNameRegistry.For(seed)` created a *second* registry that sat in front of the builder's once the seed was inserted.

## The fix

- `Build()` makes the same pass `AddQuartz` makes, in the same place: after the callback, before the property-named registrations, so what code registered still beats what a string named.
- The seed shares the builder's registry, and the extra descriptor is dropped before insertion.
- `ApplyProperties()` returns the merged bag and `Build()` applies the registrations half, so the phase order reads in one place.

No public API change (baselines untouched); no schema change; no 3.x counterpart (3.x has no `QuartzSchedulerBuilder`).

## Tests

`QuartzSchedulerBuilderTest`: `ContainerWideConfigurationReachesTheStandaloneScheduler` (code route) and `ContainerWideConfigurationFromTheLegacyKeyReachesTheStandaloneScheduler` (property route), through a fake `IExecutionHistoryStore` that completes a task on the first `AddExecution`. Both fail on `main` (20 s timeout each) and pass here.

```
without the fix:  Failed:  2, Passed:    0
with the fix:     Failed:  0, Passed:    2
unit suite:       Failed:  0, Passed: 6646, Skipped: 36
```

The release-gate probe that found it (`scratchpad/g5/probe`, cases C, D and F) will be re-run against the re-packed candidate after merge.

Milestone 4.2; part of #3811.

https://claude.ai/code/session_0121rLvZfqcUZeZDmJAshq2S


lahma:
The release-gate probe (`scratchpad/g5/probe`: six SQLite-backed scenarios, the packed 4.2.0 resolved from this branch, assembly hash checked against the pack) before and after:

| Scenario | `main` | this branch |
|---|---|---|
| A. container + `UsePersistentStore(UseExecutionHistory)` | 6 firings, 6 rows | 6 / 6 |
| B. container + `quartz.jobStore.executionHistory=true` | 6 / 6 | 6 / 6 |
| C. `QuartzSchedulerBuilder` + `UsePersistentStore(UseExecutionHistory)` | 5 firings, **0 rows**, no listener | **6 / 6**, `quartzExecutionHistory` listener |
| D. `QuartzSchedulerBuilder` + `quartz.jobStore.executionHistory=true` | 5 firings, **0 rows**, no listener | **6 / 6**, `quartzExecutionHistory` listener |
| E. `QuartzSchedulerBuilder` + `AddPlugin<T>()` directly | plugin arrives | plugin arrives |
| F. `QuartzSchedulerBuilder` + `Services.ConfigureAllQuartzSchedulers(AddPlugin<T>)` | **dropped** | **arrives** |

